PR #658, opened 2013-09-14
Rebased version of JoinNode PR 

satra:
@FredLoney this replaces #643 

i have rebased this on the current master.


mwaskom:
> I agree that the documentation might need some work, but we can do it later in a separate PR 

Sadly, this never happened!

PR #868, opened 2014-06-27
0.9.2 cpac tested branch with tapattern nipype/interfaces/afni/preprocess.py fix

pintohutch:


dclark87:
Yep: https://github.com/nipy/nipype/pull/868

PR #938, opened 2014-10-07
ENH: Allow chained name_source

oesteban:
An input trait with ```name_source``` defined and pointing to another trait that defines ```name_source``` pointing to a third trait now is sequentially resolved, adding the appropriate suffixes.
  - [x] Improve former behavior
  - [x] Add regression tests for name_sources:
    - [x] With a 3 elements chain of name_sources
    - [x] With a 3 elements ring of name_sources (raises NipypeInterfaceError)
    - [x] With a 3 elements ring of name_sources, one is set and therefore the ring can be solved. Also checks that NipypInterfaceError is not raised.
  - [x] Update documentation accordingly


oesteban:
Hi @satra, I think this PR is ready to be considered. I've updated the description at the top, and the WIP flag replaced.

PR #978, opened 2014-10-24
fugue not listing outputs

bpinsard:


oesteban:
I was using the interface intensively without problems... not sure if this PR doesn't break it.

bpinsard:
you might have been using the interface without problem if it was not chained to any other node in the pipeline. I hope it won't break anything, but regarding the changes it should be ok.

oesteban:
@bpinsard it was actually connected in the middle of the pipeline, and I should check it out, but I think using all the possible outcomes (sometimes the shifts, sometimes the fieldmap, sometimes the (un)warped files).

PR #999, opened 2014-11-19
FIX: Remove nonfunctional ANTs registration flag

JDWarner:
As of 25 days ago, ANTs now has a "command line robustness" feature ([introduced in their PR no. 126](https://github.com/stnava/ANTs/pull/126)) which forces all provided options to actually mean/do something. This particular option, `--collapse-linear-transforms-to-fixed-image-header`, has been [nonfunctional since December 2012](https://github.com/stnava/ANTs/blame/3b906d86a1b455714983083989417003425cd43c/Examples/antsRegistration.cxx#L76) but is still passed by the nipype interface. Until the robustness feature was introduced, it was just silently ignored.

However, now with a fresh-from-source ANTs build, all nipype.interface.ants.Registration operations fail with the message:
```
Standard output:
ERROR:  Invalid flag provided collapse-linear-transforms-to-fixed-image-header
Standard error:
ERROR:  Invalid command line flags found! Aborting execution.
Return code: 1
```

There is no provision to turn this flag off completely; the interface always passes _something_ so Ants always throws a fit. Until this flag actually works again on ANTs' end, we should disable the flag in the registration interface.

Since it may be reintroduced I opted to comment it out rather than delete the lines.


satra:
could this be merged or rebased on the current master and the auto tests updated with `make specs`?

also add a line to the CHANGES file.

JDWarner:
Merged current master, ran `make spec`, added minor changes to tests. After `make spec` two new test files were created; I included all the changes from `make spec`, even though not all seem directly related to my modifications.

Suite passes on my machine now. I'll make a new PR for the other changed flags in #1014.

JDWarner:
I didn't include the `transformRigid` change noted in #1014 because I'm not familiar enough with the change to know if backwards compatibility should be a factor. If so, it would belong with the other changes noted there.

nipype/interfaces/ants/registration.py
```diff
@@ -320,8 +320,8 @@ class RegistrationInputSpec(ANTSCommandInputSpec):
         low=0.0, high=1.0, value=1.0, argstr='%s', usedefault=True, desc="The Upper quantile to clip image ranges")
     winsorize_lower_quantile = traits.Range(
         low=0.0, high=1.0, value=0.0, argstr='%s', usedefault=True, desc="The Lower quantile to clip image ranges")
-    collapse_linear_transforms_to_fixed_image_header = traits.Bool(
-        argstr='%s', default=False, usedefault=True, desc='')
+    # collapse_linear_transforms_to_fixed_image_header = traits.Bool(
+    #     argstr='%s', default=False, usedefault=True, desc='')
```
satra: these two lines should be deleted.

JDWarner:
Comment-out lines removed thanks to @nicholsn. Apologies, @satra, for some reason I never saw your comment or would have cleaned this up long past.

PR #1002, opened 2014-11-21
ENH: Add support for reduced SGEgraph on re-run

hjmjohnson:
During the re-running of a workflow with SGEgraph, we can greatly
reduce the dependancies by determining which nodes have completed
prior to generating the submit-graph.

Job numbers need to be based on job ID's not on job names.  Job names are not unique.

BUG: Job names must be valid bash variables.

### DIFF
--- a/nipype/pipeline/plugins/sgegraph.py
+++ b/nipype/pipeline/plugins/sgegraph.py
@@ -9,6 +9,25 @@ from .base import (GraphPluginBase, logger)
 from ...interfaces.base import CommandLine
 
 
+def node_completed_status( checknode):
+    """
+    A function to determine if a node has previously completed it's work
+    :param checknode: The node to check the run status
+    :return: boolean value True indicates that the node does not need to be run.
+    """
+    """ TODO: place this in the base.py file and refactor """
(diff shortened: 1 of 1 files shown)


nipype/pipeline/plugins/sgegraph.py
```diff
@@ -45,56 +64,91 @@ def make_job_name(jobnumber, nodeslist):
...
-            return 'j{0}_{1}'.format(jobnumber, nodeslist[jobnumber]._id)
+            job_name='j{0}_{1}'.format(jobnumber, nodeslist[jobnumber]._id)
+            # Condition job_name to be a valid bash identifier (i.e. - is invalid)
+            job_name=job_name.replace('-','_').replace('.','_').replace(':','_')
+            return job_name
         batch_dir, _ = os.path.split(pyfiles[0])
         submitjobsfile = os.path.join(batch_dir, 'submit_jobs.sh')
+
+        cache_doneness_per_node = dict()
+        if True: ## A future parameter for controlling this behavior could be added here
```
chrisfilo: This is where we can add the config option.

PR #1003, opened 2014-11-21
BUG: matplotlib is supposed to be optional

hjmjohnson:
This code was failing to find matplotlib and
the first import was silently continuing, but
then the second call that requires matplotlib
began failing.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 Next release
 ============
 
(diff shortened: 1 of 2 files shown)


satra:
@hjmjohnson - looks good - could you please update the CHANGES file? and then this is ready for merge.

hjmjohnson:
@satra Comment added, and local testing done.

Will merge.

PR #1004, opened 2014-11-21
ENH: Adding new ants command line feature.

hjmjohnson:
A newer version of ANTs has a command line feature
added that allows for more optimized processing in
multi-stage registrations.

ENH: Added new state saving interface to ANTs

In order to improve registration efficiency, it is important that
we can save and restore the state of a registration so taht
continuation may occur from a previous state.

ENH: Make transform writing mutually exclusive

Writing composite or non-composite transforms needs
to be mutually exclusive in the new version of
ants.


chrisfilo:
wouldn't this only influence people who use both composite and individual transforms? Kind of a niche usecase

hjmjohnson:
@chrisfilo  Reverted to old behavior so that previous  composite and individual transforms syntax behavior with respect to nipype usage.

hjmjohnson:
@satra Added new saveState and restoreState optional command line options.

This will hopefully be ready for merging now.

PR #1009, opened 2014-12-08
ENH: antsIntroduction handles RA and RI transformations

lsqshr:
when generating the _list_outputs if -t RA or -t RI is set, *_warp_field
outputs will not be required

### DIFF
--- a/nipype/interfaces/ants/legacy.py
+++ b/nipype/interfaces/ants/legacy.py
@@ -97,7 +97,7 @@ class antsIntroduction(ANTSCommand):
         transmodel = self.inputs.transformation_model
 
         # When transform is set as 'RI'/'RA', no wrap fields will be outputed
-        if transmodel not in ['RI', 'RA']:
+        if isdefined(transmodel) and transmodel not in ['RI', 'RA']:
             outputs['warp_field'] = os.path.join(os.getcwd(),
                                  self.inputs.out_prefix +
                                  'Warp.nii.gz')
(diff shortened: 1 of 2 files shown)


nipype/interfaces/ants/legacy.py
```diff
@@ -94,16 +94,20 @@ class antsIntroduction(ANTSCommand):
 
     def _list_outputs(self):
         outputs = self._outputs().get()
+        transmodel = self.inputs.transformation_model
+
+        # When transform is set as 'RI'/'RA', no wrap fields will be outputed
+        if isdefined(transmodel) and transmodel not in ['RI', 'RA']:
```
satra: if someone unsets/undefines this, what will antsIntroduction do? do you know what the ants default behavior is?

lsqshr: Hi @satra, thanks for reviewing. I just double-checked, if transmodel is undefined, antsIntroduction will set '-t  GR' which is 'greedy SyN' by default.

satra: @lsqshr -  in that case the if statement should perhaps be:

`if not isdefined(transmodel) or (isdefined(transmodel) and transmodel not in ['RI', 'RA']):`

lsqshr: @satra Thx for pointing out. You are right about it. I just made the according change and locally tested it with real nii images.

PR #1016, opened 2015-01-13
FIX+ENH: refactor area weighted surface distance + fields operations

oesteban:
- [x] Fix area-weighted path
  - [x] Refactor former code to be more robust (now using numpy when possible)
  - [x] Add zero-distance regression test
  - [x] Add translation-distance regression tests
  - [x] Add area-weighting regression tests


oesteban:
This PR is ready for code-review :)

nipype/algorithms/mesh.py
```diff
@@ -13,55 +13,78 @@
...
-    """Calculates a point-to-point (p2p) distance between two corresponding
-    VTK-readable meshes or contours.
+    """
+    Calculates a the vertex-wise warping to get surface2 from surface1.
+    It also reports the average distance of vertices, using the norm specified
+    as input.
+
+    .. warning:
+
+      A point-to-point correspondence between surfaces is required
```
satra: what if these two files are spheres? is there a way to weight them differently, for example with curvature information or some such?

satra:
also as a side note, i wanted to write compy (a python computational geometry library) a while back that would allow us to do such operations easily (perhaps with vtk this is not longer necessary, or something already exists that i haven't followed), 

i implemented a set of computational geometry algorithms (on triangular meshes) a long while back in matlab. they should be relatively straightforward to convert to python + cython.

https://github.com/satra/SLT/tree/master/SurfTools

if you are interested we can start compy (if no such other library exists) and chip away at this.

oesteban:
Hi @satra, I definitely agree on that computational geometry algorithms would be much better developed and maintained as a package different from nipype. The algorithms you had implemented in matlab look a good starting point, but also we could benefit a lot from vtk. From this side, everythings looks good to me in favor of compy.

However, I am currently struggling to finish up my PhD. For now, I can only contribute with stuff I really need for the results of my thesis... Thus, compy would start with a very particular and limited choice of algorithms from my side. What do you think?

nipype/algorithms/mesh.py
oesteban: You're right, I did not observe that detail, but I could add a new weighting keyword that reads the vertex-wise properties of the vtk file, and definitely curvature may be a potential weight. I'll look at this ASAP.

PR #1019, opened 2015-01-15
ENH: improve sdc_fmb (fieldmap-based susceptibility distortion correction)

oesteban:
- [x] Replace flirt by antsRegistration
  - [x] Execute workflow with HCP data
  - [x] Write parameters in docstrings
  - [x] Read parameters from JSON file (?), instead of getting them as arguments (requires #1020)
  - [x] Set default parameters (requires #1047)
  - [x] Update workflows depending on ```sdc_fmb```


oesteban:
I've submitted PR #1020 . If this was accepted, I will include a JSONFileGrabber interface to read parameters from a file in execution time, instead of hard-setting them as arguments of the workflow's creation function. I also edit the original tasks list.

oesteban:
This PR is ready for review

nipype/workflows/dmri/fsl/artifacts.py
```diff
@@ -545,66 +550,93 @@ def sdc_fmb(name='fmb_correction',
...
+    fmm2b0.inputs.sigma_units = ['vox'] * 2
+    fmm2b0.inputs.shrink_factors = [[6], [1]]  # ,[1] ]
+    fmm2b0.inputs.use_estimate_learning_rate_once = [True] * 2
+    fmm2b0.inputs.use_histogram_matching = [True] * 2
+    fmm2b0.inputs.initial_moving_transform_com = 0
+    fmm2b0.inputs.collapse_output_transforms = True
+    fmm2b0.inputs.winsorize_upper_quantile = 0.995
+
+    applyxfm = pe.Node(ants.ApplyTransforms(
+        dimension=3, interpolation='BSpline'), name='FMp_to_B0')
```
satra: `BSpline` gave me a whole lot of trouble on non-negative variance images (and appears to be an ITK issue). so make sure that this works here. otherwise you may want to go with `Linear`. it will make the field smoother, but may not be that big an issue here.

oesteban: Totally agree, let's go with ```Linear``` by default.

oesteban:
@satra, may I merge this? I will need it in the future, I still have the idea of revise all the distortion correction framework we are building.

satra:
@oesteban - i would request a change of name before the merge, but feel free to merge after the name change. as implemented this looks specific to diffusion rather than any epi images, so you may want to change the name. 

for general application to epi images: perhaps using an index to focus on a set of images for getting the mean image. the index would be b0 images for diffusion and could be all images for other functional paradigms.

oesteban:
you are right, that's the only difference between dmri and fmri. I will see how to cover all the cases. If the workflow worked for both families of epi images, would the name be appropriate?

satra:
yes

oesteban:
Now there is no ```in_bval``` input, being replaced by ```in_ref``` a list of indices of the images to be averaged as baseline.

PR #1020, opened 2015-01-16
ENH: IO interfaces to JSON files

oesteban:
A grabber/sink pair to read/write parameters from json files.


satra:
this looks good. could you please update the changes file?

oesteban:
@satra sorry, I forgot about that. Fixed.

nipype/interfaces/io.py
```diff
@@ -1801,3 +1801,104 @@ def _get_ssh_client(self):
...
+            raise RuntimeError('JSON input has no dictionary structure')
+
+        outputs = {}
+        for key, value in data.iteritems():
+            outputs[key] = value
+
+        return outputs
+
+
+class JSONFileSinkInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
```
satra: how about adding an input that accepts a dictionary?

oesteban: Sorry @satra, I don't understand what you mean :S.

satra: something like `in_dict = Dict(desc='input JSON dictionary')`

so that you can pass a dictionary output of another node directly into the jsonsink node? and it will populate the relevant fields.

oesteban: Ok I see :+1:

satra:
should we consider:
a. adding the elements using `on_trait_change` for `in_dict`, i.e., whenever this is changed, instead of only a single input? 
or 
b. making it behave more like datasink with the `a[.[@]b[.[@]c]` formulation to create arbitrary dictionaries?

this would allow connecting multiple nodes to the same JSONsink.

oesteban:
Hi @satra, I've been thinking this around, but I can't see what would be a good end to this PR.

From your last comment, I don't see why it is not possible to connect multiple nodes to the same JSONSink, provided with you specify the input in input_names.

Regarding your considerations, I'd prefer not to go for b), as I'm not fond of the datasink formulation. I find it particularly cumbersome (of course, this is just one opinion).

Regarding a), you mean that every time ```in_dict``` changed, a new json file would be written/updated?.

satra:
@oesteban 

> Regarding your considerations, I'd prefer not to go for b), as I'm not fond of the datasink formulation. I find it particularly cumbersome (of course, this is just one opinion).

happy to entertain other thoughts on that. but we will leave it for a separate issue.

> Regarding a), you mean that every time in_dict changed, a new json file would be written/updated?.

i meant that this input acts as an accumulator, but i can now see how that wouldn't be great. let's merge this and then consider the datasink formulation (also what an alternative to becoming less cumbersome would be).

PR #1023, opened 2015-01-20
Fix bug in Normalize with modulation

spikefairway:
I fixed bug reported in http://neurostars.org/p/2768/, wrong prefix in case to execute 'Normalize' with modulation.
I'd appreciate it if you could reflect this change.

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -497,20 +497,24 @@ class Normalize(SPMCommand):
                 outputs['normalized_files'] = self.inputs.apply_to_files
             outputs['normalized_source'] = self.inputs.source
         elif 'write' in self.inputs.jobtype:
+            if isdefined(self.inputs.write_preserve) and self.inputs.write_preserve:
+                prefixNorm = ''.join(['m', self.inputs.out_prefix])
+            else:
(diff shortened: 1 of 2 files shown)


satra:
@spikefairway - thanks. could you please add an updated CHANGES file?

nipype/interfaces/spm/preprocess.py
```diff
@@ -497,20 +497,24 @@ def _list_outputs(self):
                 outputs['normalized_files'] = self.inputs.apply_to_files
             outputs['normalized_source'] = self.inputs.source
         elif 'write' in self.inputs.jobtype:
+            if isdefined(self.inputs.write_preserve) and self.inputs.write_preserve:
+                prefixNorm = ''.join(['m', self.inputs.out_prefix])
```
satra: one quick question, just to ensure i understand this correctly. so if write_preserve is set, then the output prefix doesn't matter?

spikefairway: No. Output prefix is fixed such that it corresponds to actual output from SPM (its filename starts 'm') in case that 'write_preserve' is set to True.

satra: thx.

spikefairway:
@satra Sure. I added an updated CHANGES file. Please refer it.

PR #1027, opened 2015-01-21
FIX: Interfaces redirecting X crashed if $DISPLAY not defined

oesteban:


nipype/interfaces/base.py
```diff
@@ -987,7 +987,8 @@ def run(self, **inputs):
                     time.sleep(0.2)  # give Xvfb time to start
                     if xvfb_proc.poll() is not None:
                         raise Exception('Error: Xvfb did not start')
-                    old_displaynum = os.environ['DISPLAY']
+
+                    # old_displaynum = os.getenv('DISPLAY')
```
satra: delete line instead of comment.

satra:
@oesteban  - needs to be remerged with master

PR #1028, opened 2015-01-22
FIX: AddCSVRow failed when using infields

oesteban:
This PR contains two fixes:

  - [x] Import ```Undefined``` that was missing and made the interface fail when using ```infields```
  - [x] Fix behavior when using both ```infields``` and dynamically added traits.


nipype/algorithms/misc.py
```diff
@@ -850,22 +853,26 @@ def _run_interface(self, runtime):
...
-            warnings.warn(('Python module lockfile was not found: AddCSVRow will not be thread-safe '
-                          'in multi-processor execution'))
+            from warnings import warn
+            warn(('Python module lockfile was not found: AddCSVRow will not be'
+                  ' thread-safe in multi-processor execution'))
 
         input_dict = {}
         for key, val in self.inputs._outputs.items():
             # expand lists to several columns
+            if key == 'trait_added' and val in self.inputs._outputs.keys():
```
oesteban: Hi @satra, I'm not very happy with this way of checking that ```'trait_added'``` is not interpreted as one more output. Is there a way to check this more cleanly?. If not, this PR should be merged as it fixes the bad behavior in case there are Undefined inputs.

oesteban: This situation also happens now in #1047

nipype/algorithms/misc.py
```diff
@@ -850,22 +853,26 @@ def _run_interface(self, runtime):
         try:
             import pandas as pd
         except ImportError:
-            raise ImportError('This interface requires pandas (http://pandas.pydata.org/) to run.')
+            raise ImportError(('This interface requires pandas '
+                               '(http://pandas.pydata.org/) to run.'))
 
         try:
             import lockfile as pl
```
satra: i know this wasn't the change here, but in nipype.externals there is a lock file code that might work across multiprocessor and possibly in nfs settings. in nfs, the default cache time could be 30s or higher, so the lock may not propagate across nodes without timeout settings in place.

nipype/algorithms/misc.py
satra: i haven't looked into that in a while. it's possible to use some combination of np.setdiff1d with copyable_trait_names.

oesteban: I've written the same code using  copyable_trait_names, but it is still necessary to check that the key is not ```'trait_added'```

nipype/algorithms/misc.py
oesteban: @satra, I've been looking at ```portalocker``` (http://portalocker.readthedocs.org/en/latest/usage.html#examples) included in nipype.externals, and before integrating it in this interface I have a minor concern:
  - I think that the script included in nipype.externals is a bit outdated. The new version includes the class Lock that can be used with contexts. Therefore, I would update the external code to this new version.

I would proceed with this PR as it fixes a bug on this interface. Then update portalocker and finally integrate it in all io classes.

PR #1029, opened 2015-01-23
Camino tracking interfaces update

mick-d:
Camino tracking interfaces has been updated so that to include:
- new deterministic tracking from bedpostx output: `TrackBedpostxDeter`
- new probabilistic tracking from bedpostx output: `TrackBedpostxProba`
- add tracking options previously not implemented in NyPipe: `tracker`, `interpolator`, `stepsize`, `curveinterval`

Important note: `input_file` is not mandatory anymore for tracking, as for bedpostx input only `bedpostxdir` is required. As a result, the output name to be generated was set to `bedpostx_tracked`, instead of based on `input_file`, when bedpostx data are used as input (since `input_file` is undefined in this case).


satra:
thanks @mick-d - could you please merge with master and redo `make check-before-commit`.

also could you please update the CHANGES file.

mick-d:
Hi Satra, I did again a merge with master and a `make check-before-commit` but it did not result in any changes. I updated the CHANGES file.

PR #1043, opened 2015-02-05
Moved the log format and the log date format to the config file. 

robbert-harms:
This would allow an user to alter the log format if they want to do so. This
does not change the default logging format, it only changes where the
logging format is defined.

In order to do so I had to:
    - add two fields to the config file 'log_format' and
      'log_dateformat'
    - the percentage (%) needs to be escaped with an additional percentage: '%%' in order to please the config parser
    - the config parser is changed to the SafeConfigParser to handle
      this escaping
    - since the default config is in a string, the newline and tab had
      to be escaped with an additional \, that is \n and \t. In a
      actual file one would not need that.
    - the formatting of the config file was initially done with using
      the string modulo operator %. This had to be changed to the new
      string formatting option .format() in order not to trip over all
      the percentages in the new entries
    - logger.py now reads the log formats from the config file. The
      decode() function is necessary since the strings read from the
      config file are raw python strings r'', which need to be converted
      to normal strings in order for the logging to understand it needs
      to print a real newline instead of a \n character.

Having done all that, it is now possible to change the logging format in
the config file.

The changes should be fully backwards compatible, given that nowhere in
the application the class variable Logging.fmt and Logging.datefmt are
used.

As a sidenote. All this escaping business makes one wonder if it would
not be better to convert from ConfigParser to something like Yaml.

### DIFF
--- a/nipype/pipeline/engine.py
+++ b/nipype/pipeline/engine.py
@@ -1914,7 +1914,9 @@ class JoinNode(Node):
         # the new field name
         name = self._join_item_field_name(field, index)
         # make a copy of the join trait
-        trait = self._inputs.trait(field, False, True)
+        # This always returns NoneType for some reason. I replaced it by the traits Any type
+        #trait = self._inputs.trait(field, False, True)
+        trait = traits.Any
         # add the join item trait to the override traits
         self._inputs.add_trait(name, trait)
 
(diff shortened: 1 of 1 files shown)


satra:
@robbert-harms thanks for this.

in the early days of nipype yaml wasn't part of python. but perhaps we can use json? for something like this config file, i don't have a strong preference. since the internal representation is a dictionary, json would fit pretty simply, and thus having a python dict instead of a string, would also be nice.

chrisfilo:
Seems like some tests are failing - could you rebase and fix the tests?

djarecka:
@robbert-harms are you planning to work more on this PR? do you need help?

djarecka:
@robbert-harms - please let us know if you're planning to work on this PR, otherwise we will either close or finish it.

djarecka:
@oesteban I believe some of the suggested changes are already included in`config.py` or `logger.py`. Can you please let me know if you would like to include something more in the current master?

oesteban:
Yes, this should be even easier now :+1:

djarecka:
@oesteban ok, I understand your comment that I shouldn't close it.

PR #1046, opened 2015-02-09
fix: udpated indexing to ensure edge cases are considered

satra:


mwaskom:
Looks like there are some pep8 violations in there that could be spruced up while you're touching that code :)

satra:
@mwaskom - cleaned up as much as i could!

mwaskom:
:+1:

PR #1047, opened 2015-02-10
EHN: Improve JSON interfaces

oesteban:
Following #1020, the idea of this PR is improving these new interfaces.
  - [x] Add a ```defaults``` input to ```JSONFileGrabber``` so that one can set default settings to workflows in case the JSON file with settings does not exist.
  - [x] Solve the ```JSONFileSink``` outputs, as @satra suggested in #1020 or maybe with a novel paradigm.
  - [x] Add a regression test for the new JSONFileSink feature


nipype/interfaces/io.py
```diff
@@ -1885,34 +1911,52 @@ class JSONFileSink(IOBase):
...
+        for key in infields:
+            self.inputs.add_trait(key, traits.Any)
+            self.inputs._outputs[key] = Undefined
+            undefined_traits[key] = Undefined
+        self.inputs.trait_set(trait_change_notify=False, **undefined_traits)
+
+        if force_run:
+            self._always_run = True
+
+    def _process_name(self, name, val):
```
oesteban: hi @satra, thinking about the ```[string[.[@]string]]``` format, I guess there's no use for the ```@``` here right? I've implemented only the dot element to create nested dictionaries

satra: the `@` in datasink allows people to dump multiple outputs to the same location and that nipype won't allow connecting two different outputs to the same input. 

i'm not sure yet of all the use cases for this interface, but perhaps we can implement it later if needed.

oesteban: +1 on implementing it later, let's see what happens when people use this

satra:
@oesteban - sync with master and update changes for merge. feel free to merge this.

PR #1052, opened 2015-02-11
add missing auto test files removed after some PR

oesteban:


satra:
@oesteban - on my machine it always deletes the test_auto_overlap because there is an explicit test_overlap. is there a particular reason, why you don't get this? perhaps something to do with case?

satra:
basically any file that has an explicit test doesn't have an auto test. we encourage testers to put the contents of the auto test into their explicit test when they write one. or name their test differently from test_`interfaceName`.py

oesteban:
@satra ok, I think I get this. So, either I set a different name for test_overlap.py or I copy the contents of test_auto_overlap into that file, right?.

Anyways, when I do ```make specs``` they are automatically generated on my setup.

PR #1056, opened 2015-02-17
Add sfDM to featured software

chrisfilo:


hjmjohnson:
@chrisfilo This seems like a stale patch set.  Is there a reason why this has not been merged yet?

PR #1057, opened 2015-02-18
FIX Constant column in motion regressor causes singular matrix

nathanntg:
The constant column in the motion regressor results in the FSL GLM reporting a SingularMatrix exception. I can not find reference to a constant (intercept) term in the ``fsl_glm`` documentation. It could potentially be a result of the ``--demean`` argument when calling the ``fsl_glm`` command. Do you have any insights into what may be causing this?

For now, I am opening the pull request to start discussion. If this seems like a reasonable fix, I will go through the other steps listed in the "contributing" document.

### DIFF
--- a/examples/rsfmri_vol_surface_preprocessing.py
+++ b/examples/rsfmri_vol_surface_preprocessing.py
@@ -542,7 +542,7 @@ def create_reg_workflow(name='registration'):
 
     warpmean = Node(ants.ApplyTransforms(), name='warpmean')
     warpmean.inputs.input_image_type = 3
-    warpmean.inputs.interpolation = 'BSpline'
+    warpmean.inputs.interpolation = 'Linear'
     warpmean.inputs.invert_transform_flags = [False, False]
     warpmean.inputs.terminal_output = 'file'
     warpmean.inputs.args = '--float'
@@ -771,7 +771,7 @@ def create_workflow(files,
     warpall = MapNode(ants.ApplyTransforms(), iterfield=['input_image'],
                       name='warpall')
     warpall.inputs.input_image_type = 3
-    warpall.inputs.interpolation = 'BSpline'
+    warpall.inputs.interpolation = 'Linear'
     warpall.inputs.invert_transform_flags = [False, False]
     warpall.inputs.terminal_output = 'file'
     warpall.inputs.reference_image = target_file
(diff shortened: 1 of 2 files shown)


satra:
@nathanntg - thanks - this is indeed a bug when used with fsl_glm (especially with the demean flag set to True)

satra:
and while you are at it, would you mind replacing the `BSpline` interpolation in applytransforms with `Linear`

nathanntg:
No problem. I'll also update ``rsfmri_vol_surface_preprocessing_nipy.py`` with the corresponding change as well. I will let you know once those changes are complete. Thanks!

nathanntg:
@satra - I made similar changes to the other resting state example script, and changed the transform interpolation to linear for both scripts. I tested the revised script up on my data set and it is working correctly. Thanks!

satra:
thanks @nathanntg

satra:
@nathanntg - i'm trying to regenerate the error on some old data and couldn't get it to do so on fsl 5.0.6/7/8 (these are centos distributions). i still think this is a good fix, but it would be good to know what system you are getting the message on.

nathanntg:
Hi @satra - I am using CentOS 6.6 with fsl_glm build 507. I reran the script this morning on one of my data sets and the error message returned is:

```
Exception: Subnodes of node: filtermotion failed:
Subnode 0 failed
Error:
Command:
fsl_glm -i spacetime_realign/corr_rest_01.nii.gz -d makemotionbasedfilter/filter_regressor00.txt -o corr_rest_01_glm.nii.gz --demean --out_f=F_mcart.nii.gz --out_pf=pF_mcart.nii.gz --out_res=corr_rest_01_filtermotart.nii.gz
Standard output:

Standard error:
terminate called after throwing an instance of 'NEWMAT::SingularException'

Return code: -6
Interface GLM failed to run.
```

Let me know if you want me to try any other configurations or update FSL.

satra:
@nathanntg : would it be possible to put those two files (`corr_rest_01.nii.gz` and `filter_regressor00.txt`) somewhere (e.g., dropbox, bittorrent sync, etc.,.)

nathanntg:
@satra Sorry for the delay, I don't think that I can share the `corr_rest_01.nii.gz` file (I need to check with my lab director). Is there a good open source data set that I can use to test and see if it produces the same issue?

satra:
plenty of open data sets - search for fcp-indi, nki, abide, etc.,.

nathanntg:
Hi @satra . Unfortunately, I tried a couple of the files from those open data sets and it did not produce the same output. I am not sure what specifically is accounting for the difference. When I looked at the design matrix, nothing jumped out at me. But the "singular matrix" exception was consistent across subjects in our work. Any other ideas I should try to potentially uncover what is going on?

PR #1059, opened 2015-02-19
Updates to JoinNode

robbert-harms:
My use case scenario was this:

I had a class with an inputspec as follows:
```Python
class JoiningIS(DynamicTraitedSpec):
    subject_id_list = traits.List(trait=traits.Str, imandatory=True, desc="The list of subject ids")

class Joining(BaseInterface):
    input_spec = JoiningIS
    ...
```

This was wrapped in an JoinNode:
```Python
node = pe.JoinNode(Joining(), joinsource='InfoSource', name='Joining', joinfield=['subject_id_list'])
```

Whenevery I tried to run this I got the following error message:

traits.trait_errors.TraitError: The 'subject_id_listJ1' trait of a DynamicTraitedSpec instance must be an implementor of, or can be adapted to implement, NoneType or None, but a value of '000000642263' <type 'str'> was specified.

I traced this in the code to 
```Python
trait = self._inputs.trait(field, False, True)
```
in the engine.py on line 1917. This always returned a NoneType. Replacing this with a single call to:
```Python
trait = traits.Any
```
circumvented this error.

Next, I ran into the problem that my lists were not concatenated. The loop on line 1968 had:
```Python
for field in self.inputs.copyable_trait_names():
```
This did not iterate over the fields specified by the joinfields. It therefore never concatenated the results. I split the loop into two loops, to make sure it always iterates over the join fields.

The end result is that the JoinNode now works for me as expected. I now get in my interface a list of concatenated inputs.

Please let me know if you need more information about this patch.

### DIFF
--- a/nipype/pipeline/engine.py
+++ b/nipype/pipeline/engine.py
@@ -1914,7 +1914,9 @@ class JoinNode(Node):
         # the new field name
         name = self._join_item_field_name(field, index)
         # make a copy of the join trait
-        trait = self._inputs.trait(field, False, True)
+        # This always returns NoneType for some reason. I replaced it by the traits Any type
+        #trait = self._inputs.trait(field, False, True)
+        trait = traits.Any
         # add the join item trait to the override traits
         self._inputs.add_trait(name, trait)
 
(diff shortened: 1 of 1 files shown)


chrisfilo:
Thanks for your contribution. I managed to fix the first issue you reported (with proper type checking instead of forcing everything to Any) in #1151.

I don't quite understand your second concern (about concatenating inputs). Could you elaborate on how do you expect this to work and what did you get?

PR #1069, opened 2015-03-18
ENH/FIX?: Drop support for Python 2.6

chrisfilo:
Python 2.7 brings argparse and some useful testing functions
(http://stackoverflow.com/questions/17585207/python-unittest-assetraises).
Both debian stable and ubuntu LST come with Python 2.7.

### DIFF
--- a/doc/users/install.rst
+++ b/doc/users/install.rst
@@ -107,7 +107,7 @@ recommendations.
 Must Have
 ~~~~~~~~~
 
-Python_ 2.6 - 2.7
+Python_ 2.7
 
 Nibabel_ 1.0 - 1.4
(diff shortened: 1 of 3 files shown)


doc/users/install.rst
```diff
@@ -107,7 +107,7 @@ recommendations.
 Must Have
 ~~~~~~~~~
 
-Python_ 2.6 - 2.7
+Python_ 2.7
 
 Nibabel_ 1.0 - 1.4
```
satra: should we update the ranges for the rest of the libraries?

chrisfilo: yes - but let's do this in a separate PR

PR #1070, opened 2015-03-18
Make terminal_output not mandatory (since it has a silent default).

chrisfilo:
This will improve the clarity of documentation.


chrisfilo:
The decreased coveraged is due to this:

	$ make specs
	python tools/checkspecs.py
	150318-11:57:23,977 workflow DEBUG:
		 networkx 1.4 dev or higher detected
	ErrorMap has nonautotest
	Overlap has nonautotest

satra:
ah now i remember why the default was essential - we need to ensure that any interface that relies on terminal output (e.g., fslstats) always sets this to stream or file. 

it also makes me think that it might be nice for the command line class to have a terminal_output output spec as well (we don't have to make the change here, but would be nice to add during the sprint). that way people can sink those as well. we do capture it in the report (in some cases) and in the prov. so it may not be complete necessary.

otherwise this looks good. can you update the CHANGES?

chrisfilo:
On Wed, Mar 18, 2015 at 12:15 PM, Satrajit Ghosh <notifications@github.com>
wrote:

> ah now i remember why the default was essential - we need to ensure that
> any interface that relies on terminal output (e.g., fslstats) always sets
> this to stream or file.
>
That could be easily done by overriding this in the InputSpec. Like here:
https://github.com/nipy/nipype/blob/732cef3f0259fec92f9dd15205c7c209e2606499/nipype/interfaces/afni/preprocess.py#L1494


>  it also makes me think that it might be nice for the command line class
> to have a terminal_output output spec as well (we don't have to make the
> change here, but would be nice to add during the sprint). that way people
> can sink those as well. we do capture it in the report (in some cases) and
> in the prov. so it may not be complete necessary.
>
> otherwise this looks good. can you update the CHANGES?
>
Happy to do this, but isn't it too small?

>

satra:
no - this is a pseudo API change, so i don't think it hurts to be included. i know practically it doesn't affect anyone really.

PR #1073, opened 2015-04-05
fixed fsl.utils.Slicer outfile file argstr

DeSistoNV:
wrapper for Slicer didn't have the correct prefix.


mwaskom:
The `-a` argument triggers one specific way of creating an image with `slicer`, but the nipype interface is general and can handle other options. To get the behavior you would see with `-a` in nipype, you need to set the `middle_slices` input to True.

PR #1078, opened 2015-04-09
Fix/vol2surf

soligschlager:
surf_reg option was broken.
We ran the tests, but weren't sure whether that went ok


nipype/interfaces/freesurfer/utils.py
```diff
@@ -89,8 +89,8 @@ class SampleToSurfaceInputSpec(FSTraitedSpec):
     subject_id = traits.String(desc="subject id")
     target_subject = traits.String(argstr="--trgsubject %s",
                      desc="sample to surface of different subject than source")
-    surf_reg = traits.Bool(argstr="--surfreg", requires=["target_subject"],
-                           desc="use surface registration to target subject")
+    surf_reg = traits.String(argstr="--surfreg %s", requires=["target_subject"],
+                           desc="registration surface to target subject (default sphere.reg)")
```
satra: in such a case, you may want to explicitly set the default value of the traits to `sphere.reg`

`surf_reg = traits.String('sphere.reg', ..., usedefault=True)`

SaOligg88:
we considered that, too, but thought rather to mimic freesurfer's original behavior:
If --surfreg is not specified, it defaults to using sphere.reg.
What do you think?

satra:
@SaOligg88 -

for this to work we need to ensure versioning is in place (and it's only somewhat in place). say for example, freesurfer changes the default from sphere.reg to fancynew.reg. then from a nipype perspective it won't rerun a node because it doesn't know that some default has changed, and from a reproducibility perspective running an older script may not generate the same output because it's actually being called with different parameters.

so i would suggest that we start moving towards encoding the defaults where we can. we will over time start moving the specification to the underlying software themselves (e.g., like slicer), such that these things are automatically picked up together with versioning.

djarecka:
@soligschlager are you still working on this PR? Do you need help?

djarecka:
@soligschlager - please let us know if you're planning to work on this PR, otherwise I'll try to review and finish.

soligschlager:
Hi there, no, unfortunately I'm not. Thanks for checking.

djarecka:
@soligschlager - thank you for answering. I will review and check what is still missing.

PR #1083, opened 2015-04-11
fixed typo op.abspath to os.abspath

fliem:


mwaskom:
You need to make this `os.path.abspath`, there isn't an `os.abspath` function.

I'm actually a little surprised the tests passed?

satra:
this is inside list_outputs and is dependent on an input. the tests don't catch that (yet!).

@fliem - once you make the change @mwaskom suggested we can merge this.

fliem:
Thanks @mwaskom. I changed it.
Best,
Franz

PR #1085, opened 2015-04-17
MultiTensor simulator interface

oesteban:
An interface to simulate diffusion weighted images (DWIs) using the multi-tensor model included in dipy in a ball-and-stick -like fashion:
  * Up to 3 fibers are simulated with the multi-tensor model
  * Signal is simulated for 3 isotropic compartments (CSF, GM, WM).

Diffusivity of the isotropic compartments and the single-fiber tensor are taken from:
> Alexander et al., Detection and modeling of non-Gaussian apparent diffusion coefficient profiles in human brain data, MRM  48(2):331-340, 2002, doi: 10.1002/mrm.10209


oesteban:
It'd be nice if @Garyfallidis or @arokem could give this PR a quick look before merging :)

mwaskom:
It seems strange that the main code here is going to live in nipype and not dipy?

arokem:
Well - depending on what you mean by 'main', it is. That's the `multi_tensor` function, the `add_noise` function and all the functions being used in `generate_gradients`. The latter could arguably find a home in `dipy.core.gradients`, or in `dipy.sims`.

nipype/interfaces/dipy/simulate.py
```diff
@@ -0,0 +1,274 @@
...
+        outputs = self._outputs().get()
+        outputs['out_file'] = op.abspath(self.inputs.out_file)
+        outputs['out_mask'] = op.abspath(self.inputs.out_mask)
+        outputs['out_bvec'] = op.abspath(self.inputs.out_bvec)
+        outputs['out_bval'] = op.abspath(self.inputs.out_bval)
+
+        return outputs
+
+
+def _compute_voxel(args):
```
arokem: I don't really know how this usually works in nipype, but it would be good to have clear documentation for what goes into `args`. There's lots of them!

nipype/interfaces/dipy/simulate.py
```diff
@@ -0,0 +1,274 @@
...
+
+    .. [Alexander2002] Alexander et al., Detection and modeling of non-Gaussian
+      apparent diffusion coefficient profiles in human brain data, MRM
+      48(2):331-340, 2002, doi: `10.1002/mrm.10209
+      <http://dx.doi.org/10.1002/mrm.10209>`_.
+    .. [Pierpaoli1996] Pierpaoli et al., Diffusion tensor MR imaging
+      of the human brain, Radiology 201:637-648. 1996.
+    """
+    D_ball = [3000e-6, 960e-6, 680e-6]
+    sf_evals = [1700e-6, 200e-6, 200e-6]
```
arokem: So the tensor evals are always this value? Seems limiting.

nipype/interfaces/dipy/simulate.py
```diff
@@ -0,0 +1,274 @@
...
+    mevals = [sf_evals] * nf
+    sf_vf = np.sum(ffs)
+    ffs = ((np.array(ffs) / sf_vf) * 100)
+
+    # Simulate sticks
+    signal, _ = multi_tensor(gtab, np.array(mevals), S0=1,
+                             angles=sticks, fractions=ffs, snr=None)
+    signal *= sf_vf
+
+    # Simulate balls
```
arokem: Interesting. Should we also enable this functionality in the simulations in dipy (adding isotropic components)?

nipype/interfaces/dipy/simulate.py
```diff
@@ -0,0 +1,274 @@
...
+      48(2):331-340, 2002, doi: `10.1002/mrm.10209
+      <http://dx.doi.org/10.1002/mrm.10209>`_.
+    .. [Pierpaoli1996] Pierpaoli et al., Diffusion tensor MR imaging
+      of the human brain, Radiology 201:637-648. 1996.
+    """
+    D_ball = [3000e-6, 960e-6, 680e-6]
+    sf_evals = [1700e-6, 200e-6, 200e-6]
+
+    vfs = [args[0], args[1], args[2]]
+    ffs = [args[3], args[4], args[5]]  # single fiber fractions
```
arokem: Always exactly three? Why not take lists as arguments and allow users to determine how many fibers they want?

nipype/interfaces/dipy/simulate.py
Garyfallidis: Hi @oesteban and @arokem. Having isotropic components is already supported (maybe it needs more documentation in the dipy docs).
For example here I am adding one anisotropic component and then an isotropic
mevals = np.array(([0.0015, 0.0003, 0.0003], [0.0020, 0.0020, 0.0020])
S = multi_tensor(gtab, mevals)

So, I think the lines below are not necessary. The fractions should also add to 1 (or 100%) which is crucial.

nipype/interfaces/dipy/simulate.py
oesteban: We need to use args to make the method 'pickleable' by multiprocessing

nipype/interfaces/dipy/simulate.py
oesteban: Of course, we could easily add them in args. I wanted to start with a proof of concept.

nipype/interfaces/dipy/simulate.py
oesteban: @Garyfallidis I started out doing that, but I got faster results by simulated the isotropic compartments outside ```multi_tensor```

nipype/interfaces/dipy/simulate.py
oesteban: As before, we could easily extend this, but for a first implementation I think three is a good tradeoff

oesteban:
@arokem @mwaskom @Garyfallidis I think there are only a couple of features that may be more appropriate in a PR to dipy:
  * Processing full images: parallel processing of many pixels and the simple mask computation.
  * Simulation of isotropic compartments. This is actually doable already as @Garyfallidis pointed out.

If there's an insterest in dipy of having these two features, I'll be happy to contribute.

arokem:
Those both sound interesting. I am curious to hear how much of a gain you got from simulating the isotropic component separately. As pointed out by @Garyfallidis, this is already doable, by passing a set of identical evals, but if there's a substantial gain by doing it separately, it might be worth doing. Concerning your second point, we're planning to have a hangout to discuss parallel processing in dipy, and we'd be happy to have your input on that, see link to scheduling on https://github.com/nipy/dipy/pull/631

oesteban:
@arokem I'll see to participate in the parallel processing chat. However I'm far from an expert. I just know how to use ```multiprocessing``` to map a list to a function, and get the results. I don't have any numbers about the performance gain simulating compartments outside ```multi_tensor```. With ```multi_tensor``` the b-vectors and the sticks are projected and scaled by the tensor. However, outside, it is just about matrix operations optimized by numpy.

@chrisfilo @satra: what do you think about merging this as is now, and updating this interface when the features under discussion are added into dipy?

oesteban:
@arokem @Garyfallidis I finally integrated the simulation of isotropic compartments using ```multi_tensor```. However, resulting MD in these regions is lower than that of single-fiber regions. FA doesn't look bad, though. What could I be doing wrongly?

satra:
pinging to see where this is at.

oesteban:
I think that with the last merge with master, everything is ready from my side and all the questions were answered :)

PR #1090, opened 2015-04-21
[ENH] Add new dipy interfaces

oesteban:
Added interfaces to RESTORE, response estimator and CSD in reconstructions. Added StreamlineTractography in tracks.py.


oesteban:
This is another PR to be reviewed by dipy's crew (@arokem, @Garyfallidis, etc)

oesteban:
@satra, @chrisfilo How could I check that dipy has certain version, and make the interfaces unavailable if too old?

chrisfilo:
It would be great to add this, but there are still some tests failing.

oesteban:
hey @chrisfilo , yep, the problem is on checking the dipy version. My problem here is two-fold: 1) I don't know exactly what is the policy we want to set on dipy versions, since it is developing very (very) fast; and 2) I would need to check how you do this in other interfaces where version is important.

oesteban:
@satra @chrisfilo , tests are passing (circle hook didn't update, but passed), how do you see this?

oesteban:
Cool!

PR #1103, opened 2015-05-15
MRG: fixes to get Windows tests to run

matthew-brett:
The config module used `os.environ['HOME']` that is often undefined in Windows.

Sevaral places use the `pwd` package to get the login name of the user, but
`pwd` is specific to Unix.

`getpass.getuser()` seems to be a simpler and more general alternative:

http://stackoverflow.com/questions/842059/is-there-a-portable-way-to-get-the-current-username-in-python


satra:
thanks matthew.

PR #1110, opened 2015-05-25
Revert #978 adding tests (see also #977)

### LINKED ISSUE #977: fugue broken by the refactor
No _list_outputs is implemented in FUGUE interface, so no outputs is set by default.
this seems to be broken by this PR: https://github.com/nipy/nipype/pull/857
it seems that no tests are checking for outputs to be listed if file created.

oesteban:
PR #978 broke FUGUE (see #977). This PR reverts back that one, and adds a test to ensure that outputs of interface are actually filled in.

Close #977. 

@bpinsard, could it be that in your pipelines you are trying to get an output that is actually undefined by design of the FUGUE interface?, something like you want to apply the forward warping but look into unwarped_file at the output?

### DIFF
--- a/nipype/interfaces/fsl/preprocess.py
+++ b/nipype/interfaces/fsl/preprocess.py
@@ -1388,26 +1388,6 @@ class FUGUE(FSLCommand):
 
         return super(FUGUE, self)._parse_inputs(skip=skip)
 
-    def _list_outputs(self):
-        outputs = self._outputs().get()
-        if self.inputs.forward_warping:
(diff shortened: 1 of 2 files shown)


oesteban:
@chrisfilo @satra please, check this PR & merge if OK

oesteban:
@chrisfilo, @satra Any progress on this? 

@bpinsard: could you check that this PR works for you? Should we investigate the reason why ```name_source``` does not work on your settings?

PR #1124, opened 2015-06-25
Fix an input and corresponding output in FIRST

### LINKED ISSUE #1123: FIRST fails generating segmentation_file output path
A FileNotFoundError is thrown when executed with default parameters, it looks for segmented_all_fast_firstseg.nii.gz but segmented_all_none_firstseg.nii.gz was generated.

Conversely, if ```method``` is set, FIRST fails when formatting the argument

oesteban:
This PR fixes #1123

  - [X] Fix template for input method
  - [x] Fix output

Still necessary to check if output name changes with the input method.


oesteban:
I've written the theoretical names of outputs, however ```run_first_all``` doesn't work with threshold method in my settings, so I can't check it out.

oesteban:
@satra @chrisfilo , I think this PR is ready to go. Even though I complain about the threshold method in my last update, I think it just doesn't work in the original script shipped with FSL. Documentation of the script doesn't help at all. Additionally, the alternative enum works well. With this PR, at least the interface can be used.

PR #1128, opened 2015-07-06
Add -newgrid option to Warp interface

blakedewey:
Newgrid option was missing from the Warp interface. Added.


blakedewey:
@satra This is another one from my list that came up when I rebuilt. Feel free to check it out.

Issue #1127, opened 2015-06-30
Parallel execution of nipype failing


dclark87:
Hello,

When submitting a nipype workflow execution script in parallel with a job scheduler (in this case we're using SGE), nipype crashes when more than one job tries to create the ~/.nipype directory (as one of the jobs creates it before the others). Specifically:

  File "/usr/local/lib/python2.7/dist-packages/CPAC/anat_preproc/anat_preproc.py", line 1, in <module>
    from nipype.interfaces.afni import preprocess
  File "/usr/local/lib/python2.7/dist-packages/nipype/__init__.py", line 10, in <module>
    config = NipypeConfig()
  File "/usr/local/lib/python2.7/dist-packages/nipype/utils/config.py", line 64, in __init__
    os.makedirs(config_dir)
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/root/.nipype'

I think what is happening is that multiple jobs actually make it through the if statement on https://github.com/nipy/nipype/blob/master/nipype/utils/config.py#L65, then one of the jobs creates the directory before the others. Then when the other jobs attempt to create that directory, they get the OSError exception.

This can be fixed by adding a try block around the https://github.com/nipy/nipype/blob/master/nipype/utils/config.py#L66 that handles if it raises an OSError in the case that folder exists.

blakedewey:
I'm not sure of cross-OS compatibility, but I added a function for this into filemanip.

PR #1130, opened 2015-07-07
Enh/add mkdir p

blakedewey:
Add function to mimic mkdir -p in Linux. Used to correct race condition brought up in parallel execution error from #1127.

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -13,12 +13,12 @@ import json
 import os
 import re
 import shutil
-import errno
 
 import numpy as np
 
 from ..interfaces.traits_extension import isdefined
 from .misc import is_container
+from .config import mkdir_p
 
 from .. import logging, config
 fmlogger = logging.getLogger("filemanip")
@@ -431,13 +431,3 @@ def write_rst_dict(info, prefix=''):
     for key, value in sorted(info.items()):
         out.append(prefix + '* ' + key + ' : ' + str(value))
     return '\n'.join(out)+'\n\n'
-
-
-def mkdir_p(path):
-    try:
-        os.makedirs(path)
-    except OSError as exc:
-        if exc.errno == errno.EEXIST and os.path.isdir(path):
-            pass
-        else:
-            raise
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -13,6 +13,7 @@ import json
 import os
 import re
 import shutil
+import errno
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


blakedewey:
I'm getting a strange error with `make specs`. Any ideas?

```
python tools/checkspecs.py
  File "tools/checkspecs.py", line 15, in <module>
    from nipype.external import six
  File "/Users/deweybe/Documents/PythonProjects/nipype/nipype/__init__.py", line 9, in <module>
    from utils.config import NipypeConfig
  File "/Users/deweybe/Documents/PythonProjects/nipype/nipype/utils/config.py", line 20, in <module>
    import filemanip
  File "/Users/deweybe/Documents/PythonProjects/nipype/nipype/utils/filemanip.py", line 20, in <module>
    from ..interfaces.traits_extension import isdefined
  File "/Users/deweybe/Documents/PythonProjects/nipype/nipype/interfaces/__init__.py", line 10, in <module>
    from io import DataGrabber, DataSink, SelectFiles
  File "/Users/deweybe/Documents/PythonProjects/nipype/nipype/interfaces/io.py", line 45, in <module>
    from nipype.interfaces.base import (TraitedSpec, traits, File, Directory,
  File "/Users/deweybe/Documents/PythonProjects/nipype/nipype/interfaces/base.py", line 34, in <module>
    from ..utils.filemanip import (md5, hash_infile, FileNotFoundError,
ImportError: cannot import name hash_infile
make: *** [specs] Error 1
```

nipype/utils/filemanip.py
```diff
@@ -430,3 +431,13 @@ def write_rst_dict(info, prefix=''):
     for key, value in sorted(info.items()):
         out.append(prefix + '* ' + key + ' : ' + str(value))
     return '\n'.join(out)+'\n\n'
+
+
+def mkdir_p(path):
+    try:
+        os.makedirs(path)
```
satra: i don't think this is an atomic operation, and hence likely to run into the same issue. this is an instance where a lock may be required.

blakedewey: even with the try is on the creation aspect only? because the only error is EEXISTS according to the initial issue.

satra:
@blakedewey - i think the mkdir bit is fine, but there is a recursive loop because filemanip imports config and config imports filemanip. for this you may want to put the mkdir function in config itself.

blakedewey:
Made the move to config, but I still imported into filemanip, because after all it is a file operation. @satra let me know what you think.

PR #1136, opened 2015-07-10
SLURMGraph plugin

chrisfilo:


blakedewey:
I think the documentation on the "Graph" plugins is a little weak and I always have trouble distinguishing the plugin functions. Maybe this would be a good time to update the docs on these, since we are adding more graph options.

chrisfilo:
Good point!

chrisfilo:
What about now?

nipype/pipeline/plugins/slurmgraph.py
```diff
@@ -0,0 +1,158 @@
...
+        submitjobsfile = os.path.join(batch_dir, 'submit_jobs.sh')
+
+        cache_doneness_per_node = dict()
+        if self._dont_resubmit_completed_jobs: ## A future parameter for controlling this behavior could be added here
+            for idx, pyscript in enumerate(pyfiles):
+                node = nodes[idx]
+                node_status_done = node_completed_status(node)
+                ## If a node has no dependencies, and it is requested to run_without_submitting
+                ## then run this node in place
+                if (not node_status_done) and (len(dependencies[idx]) == 0 ) and (node.run_without_submitting == True):
```
satra: run_without_submitting shouldn't apply to a graph plugin - that option is to ensure nodes can run on the monitoring/submitting node. in the graph plugin case, since the scheduler takes over, this check would be unnecessary. i know this was taken from the SGEgraph code (and i didn't review it closely enough :( ).

chrisfilo:
I cleaned it up and added a small fix to SGEGraph

satra:
LGTM

PR #1142, opened 2015-07-16
MRG:Allow extra wait time for Xvfb if needed

blakedewey:
Change to _redirect_x block to allow more wait time for Xvfb if needed. Max wait time could potentially be set in config. Hard coded for now. @chrisfilo, thoughts?


satra:
sounds reasonable. agree that a config option for this would be good.

oesteban:
+1 to any improvement on headless operations :)

blakedewey:
Added config option "xvfb_max_wait"

satra:
@blakedewey - could we also please update the CHANGES file and add the new option to config.rst?

blakedewey:
@satra should be good now. I changed the docs in config_file.rst. I hope that is the right one.

PR #1148, opened 2015-07-21
Updates in nipype2boutiques and nipype_cmd.

glatard:


nipype/utils/nipype_cmd.py
```diff
@@ -42,7 +42,11 @@ def run_instance(interface, options):
                     value = float(value)
                 except:
                     pass
-
+                # try to cast string input to boolean
```
satra: there is a utility function to do this:

https://github.com/nipy/nipype/blob/master/nipype/utils/misc.py#L199

glatard:
Fixed, thanks for the hint.

PR #1152, opened 2015-07-22
New interface for fslcpgeom

mick-d:


mick-d:
@satra I had to spend quite some time cleaning everything up as I coulnd't get rid of all the test files. Everything is good now.

To ensure that no input files as overwritten (more preciesly `dest_file`) I had to copy the file locally. I am not sure this the most elegant implementation so don't hesitate to let me know if there is a better solution. In any case I tested it and it works fine.

nipype/interfaces/fsl/utils.py
```diff
@@ -26,11 +26,56 @@
...
-                                fname_presuffix)
+                                fname_presuffix, copyfile)
 
 warn = warnings.warn
 warnings.filterwarnings('always', category=UserWarning)
 
+class CopyGeomInputSpec(FSLCommandInputSpec):
+    in_file = File(exists=True, mandatory=True, argstr="%s", position=0,
+                   desc="source image")
+    dest_file = File(exists=True, mandatory=True, argstr="%s", position=1,
```
satra: you could add another two pieces of metadata to this:
 `output_name='out_file', name_template='%s_newhd'`

and then remove all three functions (run_interface, gen_filename, list_outputs) from interface

satra: oops. this may not work  for this scenario where dest_file is an existing file.

mick-d: I see. So the current implementation is ok?

nipype/interfaces/fsl/utils.py
```diff
@@ -26,11 +26,56 @@
...
+    from one image to another. Note that only copies from Analyze to Analyze
+    or Nifti to Nifti will work properly. Copying from different files will result
+    in loss of information or potentially incorrect settings.
+
+    """
+    _cmd = "fslcpgeom"
+    input_spec = CopyGeomInputSpec
+    output_spec = CopyGeomOutputSpec
+
+    def _run_interface(self, runtime):
```
satra: you can replace this code with something like this - we don't like the input state getting altered in a function.

```
        # Copy destination file to new local file to prevent overwriting                                        
        # and update destination file                                                                           
        runtime = super(CopyGeom, self)._run_interface(runtime)                                                 
        out_file = self._gen_filename('out_file')                                                               
        copyfile(self.inputs.dest_file, out_file, copy=True)                                                    
        return runtime                                                                                          
                                                                                                                
        if name == 'out_file':                                                                                  
            return self._gen_fname(self.inputs.dest_file, suffix="_newhd")                                      
        return None                                                                                             
                                                                                                                
    def _list_outputs(self):                                                                                    
        # Use the new destination file updated by _run_interface                                                
        outputs['out_file'] = self._gen_filename('out_file')                                                    
        return outputs                                                                                          
```

satra: also the destination file will be unchanged only if this is used in a workflow. as an interface it will do exactly what fslcpgeom does, which is to alter the destination file.

mick-d: In the code i wrote, a new file is created in the `cwd`, even when run as an interface, as preserving the original file prevents from making an irreversible change otherwise. Also it may be likely that this interface is used on the first `SelectFiles` node in a workflow (as in my case) and so the raw data would be overwritten if the file is not copied before running `fslcpgeom`.

satra: in a workflow there is no issue, as the `copyfile` directive will make a local copy. however, as an interface the python interface should have similar functionality as the original commandline one which is destructive - and we should note it in the docstring as well. the only issue is that since an interface doesn't know if it's running in a workflow, it will still create a copy of the dest file.

mick-d: But in the code you wrote `copyfile` is called after `runtime = super(CopyGeom, self)._run_interface(runtime)` so wouldn't the original file be overwritten and only after that, copied inside the workflow `cwd`?

satra: different `copyfile` reference. the `copyfile=True` in the traits for `dest_file` is what copies things into the local working directory during workflow operation. then the `fslcpgeom` is applied to this, copied to the new file and sent further downstream. practically what i would like to figure out is if there is a more generic approach for such situations where an input file is modified. but here is an even simpler version of the code, i think (see if this works in a node):

```
class CopyGeomInputSpec(FSLCommandInputSpec):
    in_file = File(exists=True, mandatory=True, argstr="%s", position=0,
                   desc="source image")
    dest_file = File(exists=True, mandatory=True, argstr="%s", position=1,
                     desc="destination image", copyfile=True, output_name='out_file')
    ignore_dims = traits.Bool(desc=('Do not copy image dimensions'),
                              argstr='-d', position="-1")

class CopyGeomOutputSpec(TraitedSpec):
    out_file = File(exists=True, desc="image with new geometry header")

class CopyGeom(FSLCommand):
    """Use fslcpgeom to copy the header geometry information to another image.
    Copy certain parts of the header information (image dimensions, voxel dimensions,
    voxel dimensions units string, image orientation/origin or qform/sform info)
    from one image to another. Note that only copies from Analyze to Analyze
    or Nifti to Nifti will work properly. Copying from different files will result
    in loss of information or potentially incorrect settings.
    """
    _cmd = "fslcpgeom"
    input_spec = CopyGeomInputSpec
    output_spec = CopyGeomOutputSpec
```

mick-d: Got you. I'll test it as an interface and Node.

mick-d: @satra What is the new argument `output_name='out_file'` supposed to do? 

When i define `CopyGeom` as written above and run it as an interface, `outputs.out_file` is undefined

satra: that's weird. here is what output_name is supposed to do:

https://github.com/nipy/nipype/blob/master/nipype/interfaces/base.py#L1592

after you initialize the node or interface, could you do:

`obj._filename_from_source('dest_file')`

mick-d: I see. So I don't know why it is not working either since the command you wrote return the correct path to the file. Here is the full code and output.
```
from nipype.interfaces.fsl.utils import CopyGeom
cpg = CopyGeom(in_file='../test_fslcpgeom_satra/SUBJ01_FAfsl.nii.gz',
               dest_file='../test_fslcpgeom_satra/SUBJ01_T1_brainseed.nii')
cpg_res = cpg.run()
```
```
> cpg_res.outputs.out_file
<undefined>
```
```
> cpg._filename_from_source('dest_file')
 '../test_fslcpgeom_satra/SUBJ01_T1_brainseed.nii'
```
Does the following have to do anything with it?
```
metadata = dict(name_source=lambda t: t is not None)
traits = cpg.inputs.traits(**metadata)
```
```
> traits
{}
```

satra: aha - that's it. add the following to the dest_file attributes.

name_source='dest_file', name_template='%s'

mick-d: Yes, now it works!

mick-d: I'll test it in a Node now

mick-d: All good, i'll update the PR

mick-d:
@satra Done

PR #1156, opened 2015-07-29
[FIX] Improve redirections of X

### LINKED ISSUE #1155: [BUG] BaseInterface with _redirect_x: object has no attribute '_exists_in_path'
When ```_redirect_x = True``` interfaces fail because ```_exists_in_path()``` is defined only for ```CommandLineInterface```

Additionally, ```CommandLineInterface``` should better use ```xvfb-run```, instead of manually create the process

oesteban:
This PR fixes #1155. Additionally, ```CommandLineInterface``` has been also modified to be run with the ```xvfb-run``` script.

  - Two tests checking 1) BaseInterface redirection, and 2) CommandLineInterface after this change.
  - The code may  fail when running many BaseInterfaces at the same time, but at least the workflow doesn't fail.

Question: maybe the approach should be opening just one virtual display and redirecting all the nodes with redirect_x in the host to that display. What do you think? Currently, a lot of new virtual displays need to be open, but only one per host is enough.

### DIFF
--- a/nipype/interfaces/base.py
+++ b/nipype/interfaces/base.py
@@ -983,32 +983,12 @@ class BaseInterface(Interface):
                         version=self.version)
         try:
             if self._redirect_x:
-                exist_val, _ = self._exists_in_path('Xvfb',
-                                                    runtime.environ)
-                if not exist_val:
-                    raise IOError("Xvfb could not be found on host %s" %
-                                  (runtime.hostname))
-                else:
-                    vdisplay_num = _search_for_free_display()
-                    xvfb_cmd = ['Xvfb', ':%d' % vdisplay_num]
-                    xvfb_proc = subprocess.Popen(xvfb_cmd,
-                                                 stdout=open(os.devnull),
-                                                 stderr=open(os.devnull))
-                    wait_step = 0.2
-                    wait_time = 0
-                    while xvfb_proc.poll() is not None:
-                        if wait_time > config.get('execution', 'xvfb_max_wait'):
-                            raise Exception('Error: Xvfb did not start')
-                        time.sleep(wait_step)  # give Xvfb time to start
-                        wait_time += wait_step
-
-                    runtime.environ['DISPLAY'] = ':%s' % vdisplay_num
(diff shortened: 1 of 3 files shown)


satra:
yes, i think each host should open a single display. we could check if there is a host display running under the users uid. would probably need some local locking mechanism to ensure two jobs don't start creating displays concurrently.

oesteban:
Should we merge this as a temporary patch? Is there a place to list the different feature requests apart from the issue tracker?

nipype/interfaces/base.py
```diff
@@ -1117,6 +1131,25 @@ def version(self):
                                  self.__class__.__name__)
         return self._version
 
+    def _exists_in_path(self, cmd, environ):
```
satra: since we are moving this, i think this could be a standalone function.

nipype/interfaces/base.py
```diff
@@ -1169,32 +1202,39 @@ def _read(self, drain):
...
     """Run a command, read stdout and stderr, prefix with timestamp.
 
     The returned runtime contains a merged stdout+stderr log with timestamps
     """
     PIPE = subprocess.PIPE
 
+    cmdline = runtime.cmdline
+    if redirect_x:
+        exist_xvfb, _ = self._exists_in_path('xvfb-run', runtime.environ)
+        if not exist_val:
```
satra: shouldn't this be `exist_xvfb`?

requirements.txt
```diff
@@ -5,3 +5,4 @@ traits>=4.0
 python-dateutil>=1.5
 nibabel>=1.0
 nose>=1.0
+xvfbwrapper>=0.2
```
satra: i'm not sure this should be a requirement

nipype/interfaces/base.py
oesteban: ugh yes!

requirements.txt
oesteban: ok, that's true, if it is not installed it just fails.

PR #1159, opened 2015-08-04
20150804 an ts float test

hjmjohnson:
This patch supersedes: https://github.com/nipy/nipype/pull/1024 

In this patch, improved testing and backwards compatibility are preserved for the command line.


hjmjohnson:
@satra @chrisfilo 

This patch incorporates ideas and discussion from : https://github.com/nipy/nipype/pull/1024

More complete testing of command line was also performed.  Backwards compatibility is maintained such that if no "float" is specified, it does not show up on the command line.  If it is specified, then True maps to 1 and False maps to 0 as required by ANTs.

It would be great to finally get this commit incorporated.  Now that ITK bugs related to floating point transforms are fixed, using float = True is rapidly becoming the preferred method (not default) for running deformation based registrations.

nipype/interfaces/ants/registration.py
```diff
@@ -291,6 +291,11 @@ class RegistrationInputSpec(ANTSCommandInputSpec):
           'each stage directly updates the estimated linear transform from the previous '
           'stage. (e.g. Translation -> Rigid -> Affine). '
           ))
+    ## NOTE: Even though only 0=False and 1=True are allowed, ants uses integer values instead of booleans
+    float = traits.Bool(
+        argstr='--float %d', default=False,
+        usedefault=False, ## If --float is not specified, do not add it to the command line
```
satra: i would omit `usedefault`

hjmjohnson:
@satra All requested changes made and tests pass, so merging.

PR #1161, opened 2015-08-04
ENH: Attempt to use hard links for data sink.

hjmjohnson:
In many cases on Unix, the data sink is on the same
physical drive as the internal nipype cache.  In
that case, we can use a hard link to save both
time necessary to duplicate data, and space
necesssary to hold the same data at two different
inodes.  This allows removal of the cache
directory without modifying the results
directory.

In large analysis, this optimization can save
several terabytes of storage consumption.

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -32,6 +32,9 @@ import sqlite3
 from nipype.utils.misc import human_order_sorted
 from nipype.external import six
 
+from ..utils.misc import str2bool
+from .. import config
+
 try:
     import pyxnat
 except:
@@ -53,7 +56,7 @@ from .. import logging
 iflogger = logging.getLogger('interface')
 
 
-def copytree(src, dst):
+def copytree(src, dst, use_hardlink=False):
     """Recursively copy a directory tree using
     nipype.utils.filemanip.copyfile()
 
@@ -75,9 +78,10 @@ def copytree(src, dst):
         dstname = os.path.join(dst, name)
         try:
             if os.path.isdir(srcname):
-                copytree(srcname, dstname)
+                copytree(srcname, dstname, use_hardlink)
             else:
-                copyfile(srcname, dstname, True, hashmethod='content')
+                copyfile(srcname, dstname, True, hashmethod='content',
+                         use_hardlink=use_hardlink)
         except (IOError, os.error), why:
             errors.append((srcname, dstname, str(why)))
         # catch the Error from the recursive copytree so that we can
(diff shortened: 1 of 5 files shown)


nipype/interfaces/io.py
```diff
@@ -32,6 +32,8 @@
 from nipype.utils.misc import human_order_sorted
 from nipype.external import six
 
+from nipype.utils.config import NipypeConfig
```
satra: use relative path here:

`from ... import config` 

and use config directly. no need to initialize `NipypeConfig`

nipype/interfaces/io.py
```diff
@@ -75,9 +77,9 @@ def copytree(src, dst):
         dstname = os.path.join(dst, name)
         try:
             if os.path.isdir(srcname):
-                copytree(srcname, dstname)
+                copytree(srcname, dstname,use_hardlink)
             else:
-                copyfile(srcname, dstname, True, hashmethod='content')
+                copyfile(srcname, dstname, True, hashmethod='content',use_hardlink=use_hardlink)
```
satra: pep8 space after `,`

nipype/interfaces/io.py
```diff
@@ -322,6 +324,11 @@ def _list_outputs(self):
                 if isinstance(files[0], list):
                     files = [item for sublist in files for item in sublist]
 
+            config = NipypeConfig()
```
satra: no need to initialize config if you use the relative import

nipype/interfaces/io.py
```diff
@@ -322,6 +324,11 @@ def _list_outputs(self):
...
+            try_hard_link_datasink = config.get('execution',
```
satra: you can also use str2bool function from nipype utils:

`str2bool(config['execution']['try_hard_link_datasink'])`

nipype/interfaces/io.py
```diff
@@ -338,7 +345,7 @@ def _list_outputs(self):
                             else:
                                 raise(inst)
                     iflogger.debug("copyfile: %s %s" % (src, dst))
-                    copyfile(src, dst, copy=True, hashmethod='content')
+                    copyfile(src, dst, copy=True, hashmethod='content',use_hardlink=use_hardlink)
```
satra: pep8 - space after `,`

nipype/interfaces/io.py
hjmjohnson: done

nipype/interfaces/io.py
hjmjohnson: done

nipype/interfaces/io.py
hjmjohnson: done

nipype/interfaces/io.py
hjmjohnson: done

mwaskom:
Nice!

hjmjohnson:
@mwaskom @satra 

I tested this in a real pipeline analysis last night, and it seems to be working as expected. (i.e. saving a TON of disk space :)

I think this is ready to merge.

satra:
@hjmjohnson - could we please add this to the CHANGES file?

satra:
also add config option to config.rst

hjmjohnson:
@satra Latest requested changes to config_file.rst and CHANGELOG committed.

PR #1162, opened 2015-08-05
[ENH] MultiProc plugin with multithreaded nodes

oesteban:
This PR adds an attribute to the node: ```_singleworker=True```. This means that the node only uses one worker. 

When set to ```False``` (the node is multithreaded, versus single-threaded), the plugin will wait for all the nodes currently executing, close the pool, run this node, and start a new pool when finished.

Question: should we maintain the old MultiProc plugin as deprecated and integrate this one under a new name?


chrisfilo:
I'm a bit confused about this new approach. Is the main difference that you use Manager instead of Pool? How one can set number of workers in this new version?

BTW @ccraddock was working on something similar.

oesteban:
@chrisfilo nope, actually it still uses Pool. I have to clean up the code because at some point I had to decide between Manager and Pool. I finally left Pool as it was. It's been long since then and I don't remember the reasons. With no expectation to submit it as PR by that time, I just went forward.

One main difference is the use of ```pool.map()``` to send the jobs. Jobs are sent in batches of runable nodes. When a bottleneck comes (meaning, one of these nodes that will require all cpu cores), then the node is run isolated (no other jobs will be around). When finished, a new Pool is open. 

Moreover, with pool.map it is much easier to keep jobs (and results) under control, as compares to the async call of the current MultiProc plugin.

One second improvement would be the use of maxtasksperchild. It seems it is a good practice for long pipelines to wait the pool for finishing and create a new one.

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -97,9 +97,10 @@ def test_callback_multiproc_exception():
                plugin_args={'status_callback': so.callback})
     except:
         pass
-    assert_equal(len(so.statuses), 2)
+    assert_equal(len(so.statuses), 3)
     for (n, s) in so.statuses:
         yield assert_equal, n.name, 'f_node'
     yield assert_equal, so.statuses[0][1], 'start'
-    yield assert_equal, so.statuses[1][1], 'exception'
+    yield assert_equal, so.statuses[1][1], 'end'
```
oesteban: The new MultiProc plugin raises 3 statuses instead of 2. Is it relevant that the *end* status is also generated?

chrisfilo:
How one can set number of workers in this new version?

oesteban:
Same way, in plugin_args one should set 'n_procs': https://github.com/oesteban/nipype/blob/enh/MultiProcAllWorkersNodes/nipype/pipeline/plugins/multiproc.py#L118

nipype/pipeline/plugins/multiproc.py
```diff
@@ -49,41 +75,93 @@ class MultiProcPlugin(DistributedPluginBase):
 
     """
 
+    def _init_worker(self):
+        """
+        Ignore interrupt signal, handle in main thread
+        https://noswap.com/blog/python-multiprocessing-keyboardinterrupt
+        """
+        signal.signal(signal.SIGINT, signal.SIG_IGN)
```
oesteban: The interruption signal handling is a WIP:
https://noswap.com/blog/python-multiprocessing-keyboardinterrupt
http://jtushman.github.io/blog/2014/01/14/python-%7C-multiprocessing-and-interrupts/
http://jessenoller.com/blog/2009/01/08/multiprocessingpool-and-keyboardinterrupt

chrisfilo:
could you describe the new option in documentation and/or example workflow?

oesteban:
Sure! Some more questions:
- What about calling this workflow 'MultiProc2', so that we can still offer the former one? Or call the former something like 'MultiProcDeprecated', but keeping it?
- Is it important if the plugin raises 3 status instead of 2 (see above)
- I also think we should not merge this before I solve the signal handling.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -53,50 +76,207 @@ class MultiProcPlugin(DistributedPluginBase):
...
+
+                if continue_with_submission:
+                    self.pending_tasks.insert(0, jobid)
+                    sworker = getattr(self.procs[jobid]._interface,
+                                      '_singleworker', True)
+                    nosubmit = getattr(self.procs[jobid],
+                                       'run_without_submitting', False)
+                    if sworker and not nosubmit:
+                        forkjids.append(jobid)
+                    else:
```
chrisfilo: Codacy Issue found: [Too many positional arguments for function call](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/1939263659/issues/source?bid=2767207&fileBranchId=2790691#l234)

nipype/pipeline/plugins/multiproc.py
```diff
@@ -53,50 +87,223 @@ class MultiProcPlugin(DistributedPluginBase):
...
```
chrisfilo: Codacy Issue found: [Too many positional arguments for function call](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/1940156842/issues/source?bid=2767207&fileBranchId=2790691#l261)

chrisfilo: Codacy Issue found: [Too many positional arguments for function call](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/1941803899/issues/source?bid=2767207&fileBranchId=2790691#l261)

chrisfilo: Codacy Issue found: [Too many positional arguments for function call](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/1947759271/issues/source?bid=2767207&fileBranchId=2790691#l261)

oesteban:
We probably want to split this as a new plugin, what do you think?

PR #1039, opened 2015-02-02
ENH: add average error to ErrorMap

lsqshr:
Add an average distance output to the `nipype.algorithms.metrics.ErrorMap`, which simpy get the square root average/the sum average of the errormap vector.

PR #1163, opened 2015-08-05
ENH: Remove temporary test files.

hjmjohnson:
As new patches are created, testing should be done
before a PR is requested.

The recommended testing is
make specs; python setup.py install; make check-before-commit

In this instance, many temporary test files are created
that obscure the view of the code that was actually changed.

The temporary test files (which may change during test runs)
should NOT be included in the repository.

Step 1: Remove temporary files
  find . -name "test_auto*.py" -exec git rm {} \;
  git rm nipype/testing/data/von_errmap.nii.gz -f
  git rm nipype/testing/data/von-ray_errmap.nii.gz -f

Step 2: Add file patterns to .gitignore


satra:
@hjmjohnson - these are not temporary files. we have an automatic test generator for interfaces. if for some reason your branch is not uptodate with master, or someone forgot to generate the tests, then you will suddenly get many of them. these should be part of the repo. however any data file that's generated during a test, should not be. so `test_auto*.py` should stay, but `von_*.nii.gz` should not.

oesteban:
For some reason ```von_*.nii.gz``` have been around since https://github.com/nipy/nipype/pull/1039/files#diff-2b03173337e8ea0ce182da39dd4ad5ddR68

At least, this test should clean up the temp folder, shouldn't it?

satra:
@oesteban - i'm cleaning up a bunch of things. the problem with the test is that it runs in the testing folder. it never switches to the testing folder. so the output is in the testing/data folder.

oesteban:
:+1:

PR #1174, opened 2015-08-10
fix: change how hash values are computed

satra:
@chrisfilo, @hjmjohnson, @oesteban - please review this PR. 

while working on the py3 branch, i realized that the way we were calculating the hash values left it up to a random number generator whenever a dict was used. this fixes that. what this does mean is that every current working directory will be rendered invalid.


hjmjohnson:
@satra Is there a way to use a configuration variable to control which of the two methods is used?
      1) Defaults to current method until the next major nipy release is made, then changes to the new method.  In 1 year, the current method will be planned to be removed completely.
       2) Does this affect every pipeline, or only pipelines that use dict?

Thanks,
Hans

satra:
@hjmjohnson - this is technically and actually a bug (unless i missed something in my crazy two days of looking over 800+ files). so please do verify that i'm doing the correct thing.

being a bug, i would rather not have a config option. essentially, every input spec is converted to a dict and is hashed. the problem was that we were storing the hashed values back in a dict and then hashing the whole thing by converting the dict to a string representation. python doesn't guarantee order of elements in a dict.

strictly speaking this is one thing we should even backport to older versions.

that being said, you can always run a workflow with `updatehash=True` if you are confident that no inputs have changed.

hjmjohnson:
@satra :  I forgot about "updatehash=True "  That addresses all my concerns, and provides a way to continue using an existing internal cache while updating nipype.

I think that "updatehash=True" provides all the functionality needed to prevent years of data re-processing.

Hans

hjmjohnson:
@satra :  This looks good to me.  If I understand correctly, you convert the dict to a list where the elements of the list are sorted versions of the dictionary keys.

This approach should give consistent orderings in all cases.

Issue #1175, opened 2015-08-10
BUG: Registration interface failed multi-modal


hjmjohnson:
Registration interface was failing to support multimodal input when they are given.
    
    ex)
    nipype:
      antsWF.inputs.fixed_image=['fixed_t1.nii.gz',
                                 'fixed_t2.nii.gz']
      antsWF.inputs.moving_image=['moving_t1.nii.gz',
                                  'moving_t2.nii.gz']
    
    corresponding correct commandline:
      --metric MI[\
          fixed_t1.nii.gz,\
          moving_t1.nii.gz,\ ...
      --metric MI[\
        fixed_t2.nii.gz,\
        moving_t2.nii.gz,\ ...

oesteban:
Fixed in #1176

PR #1176, opened 2015-08-10
BUG: Registration interface failed multi-modal

hjmjohnson:
Registration interface was failing to support
multimodal input when they are given.

ex)
nipype:
  antsWF.inputs.fixed_image=['fixed_t1.nii.gz',
                             'fixed_t2.nii.gz']
  antsWF.inputs.moving_image=['moving_t1.nii.gz',
                              'moving_t2.nii.gz']

corresponding correct commandline:
  --metric MI[\
      fixed_t1.nii.gz,\
      moving_t1.nii.gz,\ ...
  --metric MI[\
    fixed_t2.nii.gz,\
    moving_t2.nii.gz,\ ...

Reported in issue: https://github.com/nipy/nipype/issues/1175

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 Next release
 ============
 
+* BUG: ANTs Registration interface failed with multi-modal inputs
+    (https://github.com/nipy/nipype/pull/1176) (https://github.com/nipy/nipype/issues/1175)
 * FIX: Bug in XFibres5 (https://github.com/nipy/nipype/pull/1168)
 * ENH: Attempt to use hard links for data sink.
     (https://github.com/nipy/nipype/pull/1161)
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@satra Someone in my lab found a bug in using ANTs in a multi-modal registration.  Is it OK to merge this bug fix?

oesteban:
This looks OK to me. I was wondering, is it possible to use ```antsRegistration``` with multichannel images? Because, this is multimodal in the sense that you can create a composite metric that works in different modalities, right?

satra:
@hjmjohnson - i'm not sure this is doing what you want it to do? it would also be good clarify how one uses it. i think the specs need to be clarified in the docstring and with examples.

for example, the old example1 makes no reference to fixed2 and moving2.

oesteban:
@satra because it only uses the first modality, isn't it?

satra:
correct - but that's why we need clear examples of how to do the multimodal stuff. also it seems example 5 should have been:
CC[fixed1, moving1] CC[fixed2.moving2] Mattes[fixed2. moving2]

oesteban:
Well, @hjmjohnson will correct me, but that example would be monomodal operation... with a level-wise change of the images you try to register. I agree that this behavior is not clearly explained with ```antsRegistration``` and we have an opportunity here to make it clearer.

PR #1181, opened 2015-08-12
Revert "BUG: Registration interface failed multi-modal"

satra:
Reverts nipy/nipype#1176

@hjmjohnson, @chrisfilo - reverting this change - breaks nipype.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,8 +1,6 @@
 Next release
 ============
 
-* BUG: ANTs Registration interface failed with multi-modal inputs
-    (https://github.com/nipy/nipype/pull/1176) (https://github.com/nipy/nipype/issues/1175)
 * FIX: Bug in XFibres5 (https://github.com/nipy/nipype/pull/1168)
 * ENH: Attempt to use hard links for data sink.
     (https://github.com/nipy/nipype/pull/1161)
--- a/nipype/interfaces/ants/registration.py
+++ b/nipype/interfaces/ants/registration.py
@@ -446,7 +446,7 @@ class Registration(ANTSCommand):
     >>> reg5.inputs.sampling_strategy = ['Random', None] # use default strategy in second stage
     >>> reg5.inputs.sampling_percentage = [0.05, [0.05, 0.10]]
     >>> reg5.cmdline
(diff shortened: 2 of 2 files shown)


satra:
@hjmjohnson - please test both the single modality and multimodal cases (and command lines) when you resubmit. also the doctests should be updated for the single modality case.

satra:
also i just got this error - should also be fixed in an update.

```
restore-state option is mutually exclusive with initial-moving-transform & initial-fixed-transform options.
```

hjmjohnson:
@satra @chrisfilo 

reg1 - reg4 were all single modal tests.  When a list is specified in a single modal test, only the first element of the list is used (as driven by the metric).

reg5 was the multi-modal test case.  NOTE:  Even in reg5 test, the first stage is single modal (and only uses the first image) and the second stage is where multi-modal information is used.


NOTE: The multi-modal imaging test was originally written by @chrisfilo back in 2012
https://github.com/nipy/nipype/commit/18cb53389319dfdb4dd7ae61b7b09005c5647734#diff-6e6dd1764769fff7edcdb5a00c00b1f8R91

We were trying to make the minimum changes necessary to fix the identified bug.  

I'll make another patch set that uses the simpler notation in the base class, and the for reg5 demonstrates the multi-modal fixed and moving images.

hjmjohnson:
@satra  RE: "restore-state option is mutually exclusive with initial-moving-transform & initial-fixed-transform options."

This is true.  You can not provide both restore-state and "initial-moving-transform or inital-fixed-transform"

Did you get that error in the doc-tests or in an external pipeline usage?

chrisfilo:
@hjmjohnson <https://github.com/hjmjohnson> I pledge guilty as charged! I
have corrected the tests in a recent PR so now they all are unimodal (since
this is the only way ANTS works correctly right now). When you fix the
multimodal support please add some multimodal tests. Tanhks!

hjmjohnson:
@satra @chrisfilo I'm struggling to figure out how the multi-modal case is wrong.  We are using the codebase before the reversion, and it seems to be working correctly for us for the 100's of cases that we have run.

Can you please elaborate on how it is broken?

Thanks,
Hans

chrisfilo:
It breaks the unimodal cases (I expect you did not test those). Try running
doctests from current master (the ones I made for unimodal cases) on the
codebase from before reversal. You should get an error somewhere here:
https://github.com/nipy/nipype/commit/07ec0cc09c908bda7703572c043767a2b7518538#diff-aad5256804f940ebdb8e591a59e282e6L509

I hope this helps!

hjmjohnson:
Before submitting PR  I ran the following:

python setup.py install; make specs; make check-before-commit

There were no errors.  Are these the doctest you are referring to?

satra:
@hjmjohnson - the doctests pass because the single modality case was not considered. if you remove the second input from the fixed and moving images, the doctests start failing. i.e a user with the single case would suddenly have run into errors with the change. that's why chris, refined the tests to focus on the single input scenario.

hjmjohnson:
@satra @chrisfilo 
OHH!    I see now!   The failure is when the fixed and moving images specified are NOT provided as a list.    

(I was confused because we run uni-modal registrations all the time successfully, but we alway pass them in as lists with single elements)

============
I now know what to fix! :)  I'll work on it right away.  It should be an easy fix.

chrisfilo:
On Fri, Aug 14, 2015 at 1:30 PM, Hans Johnson <notifications@github.com>
wrote:

> @satra <https://github.com/satra> @chrisfilo
> <https://github.com/chrisfilo>
> OHH! I see now! The failure is when the fixed and moving images specified
> are NOT provided as a list.
>
> (I was confused because we run uni-modal registrations all the time
> successfully, but we alway pass them in as lists with single elements)
>
This is weird - InputMultiPath should've turned the a string into a single
element list automatically.

PR #1182, opened 2015-08-12
Current ANTS doctests are only valid for unimodal registrations

chrisfilo:
@hjmjohnson please add (not replace) multimodal doctests along with the corrected multimodal PR


hjmjohnson:
@chrisfilo Sorry, I don't understand what I need to do here.

satra:
@hjmjohnson - we removed fixed2 and moving2 from these doctests. when the multimodal PR is submitted, @chrisfilo suggested adding new doctests for those conditions, that re-using these.

PR #1183, opened 2015-08-13
ants composite transform is a single file not a list

chrisfilo:


satra:
LGTM. @hjmjohnson - any particular reason why composite was a list? if it was a list to support multimodal, this should be refactored for the multimodal PR.

hjmjohnson:
@satra @chrisfilo 

The list was used in order to maintain parity between the various ways that ANTs produces output transforms.  If composite transforms are not used, then then the resulting transforms will be a list of file names of the different components.

hjmjohnson:
@satra @chrisfilo I don't have strong opinions on this.  If it is changed, a large commit message should indicate that backwards compatibility will be broken.  

I know that I'll need to change (i.e. simplify) some of my pipelines if this is accepted.

Hans

satra:
@hjmjohnson - this PR is specifically for these two outputs:

`composite_transform` and `inverse_composite_transform`

any reason why these would be lists?

hjmjohnson:
@satra There is only one "ill-conceived, bad" reason to keep them as lists.  We should merge this patch set and fix downstream code as necessary to be conformant.

satra:
> @satra There is only one "ill-conceived, bad" reason to keep them as lists

:)

PR #1185, opened 2015-08-14
BROCCOLI

wanderine:
Added wrappers for 3 BROCCOLI functions

Here is an example of how to use the functions

https://github.com/wanderine/BROCCOLI/blob/master/examples/nipypetest.py

### DIFF
--- a/nipype/interfaces/broccoli/preprocess.py
+++ b/nipype/interfaces/broccoli/preprocess.py
@@ -45,25 +45,18 @@ class RegisterTwoVolumesInputSpec(BROCCOLICommandInputSpec):
                    exists=True,
                    copyfile=False)
 
-
-    platform = traits.Int(argstr='-platform %d', desc='OpenCL platform to use')
-
-    device = traits.Int(argstr='-device %d', desc='OpenCL device to use')
-
+   
     iterationslinear = traits.Int(argstr='-iterationslinear %d', desc='Number of iterations for linear registration')
 
     iterationsnonlinear = traits.Int(argstr='-iterationsnonlinear %d', desc='Number of iterations for non-linear registration')
 
-    sigma = traits.Float(argstr='-sigma %s', desc='Amount of Gaussian smoothing for regularization of displacement field (default 5.0)')
+    sigma = traits.Float(argstr='-sigma %f', desc='Amount of Gaussian smoothing for regularization of displacement field (default 5.0)')
 
-    quiet = traits.Bool(argstr='-quiet', desc='Dont print anything')
-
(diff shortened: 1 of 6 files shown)


satra:
@wanderine - could you please merge with current master? there are a few things in the setup that have changed. otherwise, this will be a fun set of additions to play with.

also you could add a few tests to run_examples and use circle ci to test the broccoli interfaces as well.

once you are done writing the interfaces you should do a `make check-before-commit` to ensure that the auto tests and the docs are still building. the make command may add new files to your directory, so merging with upstream master is important before doing the make.

chrisfilo:
I'm not sure CircleCI provides any GPU we can test on...

wanderine:
BROCCOLI runs on CPUs as well.

chrisfilo:
Oh - I think they have some of those ;)

wanderine:
I don't understand this error

File "/home/andek/Research_projects/nipype/nipype/interfaces/broccoli/preprocess.py", line 71, in nipype.interfaces.broccoli.preprocess.RegisterTwoVolumes
Failed example:
    reg.cmdline
Expected:
    'RegisterTwoVolumes structural.nii mni.nii -platform 1 -device 2'
Got:
    'RegisterTwoVolumes structural.nii mni.nii -device 2 -platform 1'

nipype/interfaces/broccoli/preprocess.py
```diff
@@ -0,0 +1,168 @@
...
+
+    >>> from nipype.interfaces import broccoli
+    >>> from nipype.testing import example_data
+    >>> reg = broccoli.RegisterTwoVolumes()
+    >>> reg.inputs.in_file = 'structural.nii'
+    >>> reg.inputs.reference = 'mni.nii'
+    >>> reg.inputs.platform = 1
+    >>> reg.inputs.device = 2
+    >>> reg.cmdline
+    'RegisterTwoVolumes structural.nii mni.nii -platform 1 -device 2'
```
oesteban: This doctest expects ```'RegisterTwoVolumes structural.nii mni.nii -device 2 -platform 1'```

oesteban:
If ordering of -platform and -device is not important, then the change I suggest should fix the problem.

wanderine:
OK, thanks.

CircleCI needs an OpenCL driver if I'm going to test BROCCOLI on it.

chrisfilo:
You can add installation of required dependencies in the circle.yml file.

wanderine:
What is the  operating system on CircleCI?

Does it have Intel or AMD CPU(s) ?

chrisfilo:
Circle CI runs Ubuntu AFAIK. It's using EC2 so probably Intel but no
guarantees.

PR #1188, opened 2015-08-15
STYLE: Remove end of line trailing white space

hjmjohnson:
Trailing whitespace is often automatically
removed by several tools (editors, ide's, autopep8, & scm tools).
In order to minimize code differences that are intended from
those that are side effects of auto-cleanups, this patch set
removes end of line spaces in one patch set.


hjmjohnson:
@chrisfilo @satra 

The failures are due to "OSError: [Errno 28] No space left on device:"  They are not due to this patch set.

Would you please consider these changes?  They make editing and contributing to nipype easier.

Without them, I need to reconfigure my editing environment everytime I try to develop for nipype.

Other projects I work on refuse to accept commits with end-of-line spaces.

Thanks,
Hans

satra:
@hjmjohnson - this looks fine. any reason why `make check-before-commit` didn't capture these?

https://github.com/nipy/nipype/blob/master/Makefile#L28

hjmjohnson:
@satra 

I am not well versed in perl, but it seems that the perl command is not performing the action properly.

1) grep -l "  *$" $(find . -name "*.py")
2) find . -name "*.py" | xargs perl -pi -e 's/[ \t]*$$//'
3) grep -l "  *$" $(find . -name "*.py")

=====
step 1 provides a list of 17 files.
step 2 ---->  I would have expected it to fix the files
step 3 should provide a list of 0 files if step 2 succeeds

hjmjohnson:
@satra 

This seems to work:

find . -name "*.py" | xargs perl -pi -e 's/\s+$/\n/'

hjmjohnson:
@satra 
So does this:
find . -name "*.py" | xargs perl -pi -e 's/[ \t]*$//'

I'm wondering if the $$ is needed for windows, and $ is needed for linux/mac.

Hans

hjmjohnson:
@satra  Regarding "make check-before-commit didn't capture these?"

The problem was that on my computer, it was catching these for ALL .py files, even many that I did not edit.    Then I needed filter out all the noise and only commit the changes related to my patch set.

Perhaps we need to also add .rst to the list of files to be cleaned during the "make check-before-commit".

satra:
@hjmjohnson - let's try to ensure that PRs run a `make check before commit` and generate an ERROR, that travis reports (i'll send a PR later).

hjmjohnson:
@satra  That would be great!  

PS:  I really need to go back to funded work this week.  I've had fun playing with nipype, but I need to *STOP* working on nipype for a while.  :(  

Hans

PR #1189, opened 2015-08-17
Trailing whitespace romoval

hjmjohnson:
@satra 

These patches allow me to run:
python setup.py install; make specs; make check-before-commit

cleanly.  This allows for repeated runs do not report files that need to be committed to git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@
 .idea/
 /documentation.zip
 .DS_Store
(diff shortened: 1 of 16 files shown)


hjmjohnson:
@satra When you make the PR run make check-before-commit, you might want to build upon this patch set to get initial clean versions.

I'll hand this over to you to do with it what you want.

Hans

hjmjohnson:
@satra Did you have any more concerns?

PR #1196, opened 2015-08-27
fix: address multimodal and unimodal inputs

satra:
@hjmjohnson - please check this.

should fix the multimodal and unimodal case.


hjmjohnson:
@satra  This looks good.  I can not test this week.  I'll be happy to test next week.

hjmjohnson:
@satra To be clear. Happy if you merge.

satra:
@hjmjohnson - i'll merge, but please do test on your work when you get a chance. 

we should also augment the metric description to support something like this.

metric specs:

#### Single input
```
fix = 'f_in1'
moving = 'm_in1'

'CC' - apply only to 1
['CC', 'Mattes'] - apply both to 1
```

#### Multiple inputs
```
fix = ['f_in1', 'f_in2', 'f_in3']
moving = ['m_in1', 'm_in2', 'm_in3']

'CC' - apply only to 1
['CC', 'CC'] - apply to 1, 2
['CC', None, 'CC'] - apply to 1 (CC), 3(CC)
[['CC', 'Mattes'], None, 'CC'] - apply to 1, 3, but apply CC, Mattes to 1, and CC to 3
```

for multiple inputs, we could also add an input arg `apply_metric_all` to handle the following scenario:

```
multiple inputs specified and apply_metric_all, 
'CC' - would get applied to 1, 2, and 3 
```

Issue #1195, opened 2015-08-25
Lock when importing nipype.interfaces.spm in version 0.10.0


monade:
I just encountered a deadlock (or alike) after updating to 0.10.0. This lock (by which I simply mean that the code doesn't proceed) happens already at the import of `nipype.interfaces.spm` when running the following code:

    from nipype.interfaces.matlab import MatlabCommand
    MatlabCommand.set_default_paths("/opt/matlab/64bit/R2015a/toolbox/spm12")
    MatlabCommand.set_default_matlab_cmd("/opt/matlab/64bit/R2015a/bin/matlab -nodesktop -nosplash")
    import nipype.interfaces.spm as spm

This issue does not occur for me on 0.9.2 (here the import of nipype.interfaces.spm is executed in less than a second). I have switched back and forth a couple of times now between 0.10.0 and 0.9.2, and the pattern is consistent. Upon pressing Ctrl+C, I get the following output in 0.10.0: http://pastebin.com/9S1BJjsw

Of note, a similar lock was previously reported for 0.9.1 in these two links:

http://stackoverflow.com/questions/24544726/subprocess-get-stucks-at-poller-poll
https://groups.google.com/forum/#!topic/nipy-user/ICJzyP6sLek

Versions: Ubuntu 15.04 / Python 2.7.8 / Matlab R2015a / SPM12

satra:
@monade - could you please update to the latest master and try?

in the past nipype used to create a lock file in the home directory. could also try deleting the lock file if there is one? (~/.nipype/nipype.json)

monade:
@satra: 

Calling `pip install https://github.com/nipy/nipype/zipball/master` (correct?) installed 0.11.0.dev0. The problem still persists: http://pastebin.com/1QM0JDEs

(nb: there was no such file in the .nipype directory)

satra:
@monade - i see that you are using matlab. are you on a debian/ubuntu system?

if so could you try setting the matlab command to this (replacing the variables):

$pathtomatlabdir/bin/$platform/MATLAB

monade:
@satra - thanks for your suggestions. I tried it and it did not change the outcome.

monade:
Now it gets a bit more complicated. I can reproduce the above behavior on two Ubuntu machines running Matlab R2015a and R2015b. However, on the computer with Matlab R2015b, which has a pretty identical setup otherwise, I now end up in a similar lock when executing the following minimal working example (using 0.9.2):

[...]

PR #1199, opened 2015-08-31
fix: let spm use base class for version

satra:
temporary solution for #1195


nipype/interfaces/spm/preprocess.py
```diff
@@ -766,7 +766,8 @@ class Segment(SPMCommand):
 
     """
 
-    if Info.version() and Info.version()['name'] == "SPM12":
+    _local_version = SPMCommand().version
+    if _local_version and _local_version['name'] == "SPM12":
```
chrisfilo: not 

	if _local_version and _local_version == "SPM12"
	
?

Issue #1198, opened 2015-08-29
model.SpecifySPMModel weird behavior


MartinPerez:
When there is only one event for a given condition for a given run. SpecifySPMModel will create the session_info file with the correct total number of onsets across runs for a condition, but it will ignore the durations when only one event is reported, leading to less total number of durations. Which will of course crash spm.Level1Design due to non matching total number of onsets and durations.

satra:
@MartinPerez - are you specifying durations as shown in this example (note double brackets).

http://nipy.org/nipype/interfaces/generated/nipype.algorithms.modelgen.html#id1

satra:
it also looks like this may be the source of the bug:

https://github.com/nipy/nipype/blob/3444b7588a1dea15675dc8a4f5be5f8b753adbd8/nipype/algorithms/modelgen.py#L478

satra:
actually it appears to be fine:

```
    info = [Bunch(conditions=['cond1'], onsets=[[2]], durations=[[1]]),
            Bunch(conditions=['cond1'], onsets=[[3]], durations=[[1]])]
```

generates:

```
{'regress': [], 'cond': [{'onset': [2.0], 'duration': [1.0], 'name': 'cond1'}], 'hpf': 128.0, 'scans': '/tmp/test1.nii'}, {'regress': [], 'cond': [{'onset': [3.0], 'duration': [1.0], 'name': 'cond1'}], 'hpf': 128.0, 'scans': '/tmp/test2.nii'}
```

MartinPerez:
I didnt pass the double brackets.... it works well with [2.0, 2.0] for example

satra:
this is what i am getting with your inputs, which seems right - i'm using the current nipype master:

```
{'regress': [], 'cond': [{'onset': [2.0, 3.0], 'duration': [1.0, 1.0], 'name': 'cond1'},
{'onset': [2.0], 'duration': [1.0], 'name': 'cond2'}], 'hpf': 128.0, 'scans': '/test1.nii'},
{'regress': [], 'cond': [{'onset': [2.0, 3.0], 'duration': [1.0, 1.0], 'name': 'cond1'},
{'onset': [2.0, 4.0], 'duration': [1.0, 1.0], 'name': 'cond2'}], 'hpf': 128.0, 'scans': 'test2.nii'}
```

MartinPerez:
i will retest and reinstall the master...

MartinPerez:
[...]

PR #1200, opened 2015-09-01
fix for SpecifySPMModel

MartinPerez:
@satra you were right, I tested simply correcting the line you pointed out solved the problem. Fix for #1198.

### DIFF
--- a/nipype/algorithms/modelgen.py
+++ b/nipype/algorithms/modelgen.py
@@ -476,9 +476,7 @@ class SpecifySPMModel(SpecifyModel):
                                  sum(nscans[0:(i + 1)])
                         infoout.onsets[j].extend(onsets.tolist())
                 for j, val in enumerate(info.durations):
-                    if len(val) > 1:
-                        print 'new script installed'
-                        import pdb; pdb.set_trace()  # breakpoint aaf99b04 //
+                    if len(val) > 0:
                         infoout.durations[j].extend(info.durations[j])
                 if hasattr(info, 'amplitudes') and info.amplitudes:
                     for j, val in enumerate(info.amplitudes):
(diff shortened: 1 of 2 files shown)


nipype/algorithms/modelgen.py
```diff
@@ -476,7 +476,7 @@ def _concatenate_info(self, infolist):
                                  sum(nscans[0:(i + 1)])
                         infoout.onsets[j].extend(onsets.tolist())
                 for j, val in enumerate(info.durations):
-                    if len(val) > 1:
+                    if len(val) > 0:
```
satra: i think this can simply be: `if len(val):`

also could you please add your test to test_modelgen? it should fail without this change, and pass with it? there is already a section that checks concatenation.

nipype/algorithms/modelgen.py
```diff
@@ -497,7 +497,7 @@ def _concatenate_info(self, infolist):
             onelist = np.zeros((1, sum(nscans)))
             onelist[0, sum(nscans[0:i]):sum(nscans[0:(i + 1)])] = 1
             infoout.regressors.insert(len(infoout.regressors),
-                                      onelist.tolist()[0])
+                                      onelist.tolist()[0])
```
satra: extra white space added.

MartinPerez:
@satra I implemented the test and corrected based on your comments. In the test I added some comments to separate the intention of the different tests and added mine at the end. Also modified the last test to contrast it with mine when there are three sessions and they are not concatenated the durations are detected just fine for all runs.

satra:
thanks @MartinPerez

MartinPerez:
thanks @satra

chrisfilo:
CircleCI tests are not passing on this PR. We should consider reverting the merge until this is resolved.

satra:
@chrisfilo - is there an error message? this should only affect the spm test.

MartinPerez:
Effectively this should be reverted. The key problem is the same that I had (non matching number of onsets and events / Number of event onsets (16) does not match the number of durations (4)) but i guess now its about specifying all durations with only one number?

satra:
i think we should fix it to address both cases. 

```
if len(val) == 1:
    # then replicate to number of onsets
else:
    if len(val) != len(number_onsets):
        raise ValueError('Mismatch in number of onsets and durations for session %d, condition %s')
    # append all durations
```

MartinPerez:
I would say that we have to check the number of onsets, like:

if len(ons) > 1 and len(val) == 1:
# fill ons with val
elif len(ons) ==  len(val):
# add the durations
else:
    raise ValueError('Mismatch in number of onsets and durations for session %d, condition %s')

you agree?

PR #1201, opened 2015-09-01
Added support for grabbing and putting data on S3.

shariqiqbal2810:
* Added S3DataGrabber to nipype/interfaces/io.py
* Added S3DataSink to nipype/interfaces/io.py
* Added tests to nipype/interfaces/tests/test_io.py
    * Currently the S3 tests are using a bucket on our lab's account.
    * We would like to have a conversation about a more long-term testing solution.

### DIFF
--- a/nipype/interfaces/tests/test_io.py
+++ b/nipype/interfaces/tests/test_io.py
@@ -84,31 +84,31 @@ def test_selectfiles_valueerror():
     yield assert_raises, ValueError, sf.run
 
 def test_s3datagrabber_communication():
-    dg = nio.S3DataGrabber(infields=['subj_id','run_num'], outfields=['func','struct'])
-    dg.inputs.bucket = 'nipype-test'
-    dg.inputs.bucket_path = 'data/'
+    dg = nio.S3DataGrabber(infields=['subj_id', 'run_num'], outfields=['func', 'struct'])
+    dg.inputs.bucket = 'openfmri'
+    dg.inputs.bucket_path = 'ds001/'
     tempdir = mkdtemp()
     dg.inputs.local_directory = tempdir
     dg.inputs.sort_filelist = True
     dg.inputs.template = '*'
-    dg.inputs.field_template = dict(func='%s/%s.txt',
-                                    struct='%s/struct.txt')
-    dg.inputs.subj_id = ['subj1','subj2']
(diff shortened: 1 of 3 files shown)


satra:
for testing the grabber you can use the openfmri bucket (which is a public bucket). for the sink, i need to think of a better solution.

shariqiqbal2810:
I changed the s3datagrabber test to use the openfmri data.  Unfortunately, these files are fairly large, so the tests are slowed (even though I'm only downloading 2).  Additionally, these modules require the use of boto3 (AWS Python API), and I was wondering what your solution would be to keep this optional. Currently I have this with my imports
```
try:
    import boto3
except:
    pass
```

shariqiqbal2810:
FYI the only tests that are failing are due to a lack of boto3 on the testing system.  Tests all pass on my system.

chrisfilo:
We should skip those tests on systems without boto3 (since not all users
will use it), but also add it to circle.yml config file so it would be
installed on the test server.

shariqiqbal2810:
Ok done!  Any more thoughts on how to implement the S3DataSink test?

nipype/interfaces/tests/test_io.py
```diff
@@ -8,10 +8,17 @@
...
+from nipype.testing import assert_equal, assert_true, assert_false, skipif
 import nipype.interfaces.io as nio
 from nipype.interfaces.base import Undefined
 
+noboto = False
+try:
+    import boto3
+except:
+    noboto = True
+    pass
```
chrisfilo: Codacy Issue found: [Unnecessary pass statement](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/935437085/issues/source?bid=2767207&fileBranchId=2841000#l20)

shariqiqbal2810:
So I think the only remaining things to sort out are:
* Figuring out a more permanent testing solution for S3DataSink
* Getting AWS credentials for the testing server

chrisfilo:
What about using a fake S3 server? https://github.com/jubos/fake-s3

shariqiqbal2810:
I modified the code to use a fake s3 server on the S3DataSink tests.  I added a 'gem install fakes3' line to the circle.yml file. Hopefully that will be enough to get everything working.

PR #1202, opened 2015-09-01
refix conditions and duration matching issue

MartinPerez:
to fix bad merge in #1200, to solve issue #1198


MartinPerez:
@satra @chrisfilo I hope this will make it.

PR #1207, opened 2015-09-08
SPM realign used to omit the first volume/run

chrisfilo:
See: https://neurostars.org/p/3047/


satra:
hey chris, have you tested this under both write and estwrite? 

the only reason i ask is that if statement looks very specific about the first file not having a prefix under some condition.

chrisfilo:
it works correctly - I have added some tests on real data. this is ready to merge

satra:
should i wait for circle ci or take you word (you can merge instead of replying :) ).

PR #1208, opened 2015-09-08
fix FLIRT output specs with view  to applyxfm

dmordom:
Regarding the FLIRT interface, when using the applyxfm option or the Applyxfm interface no output matrix will be created (as is natural), so if in output specs the output matrix has exists=True as it stands now, the interface will fail in execution as the output matrix file will not be found

### DIFF
--- a/nipype/interfaces/fsl/preprocess.py
+++ b/nipype/interfaces/fsl/preprocess.py
@@ -510,8 +510,7 @@ class FLIRTInputSpec(FSLCommandInputSpec):
 class FLIRTOutputSpec(TraitedSpec):
     out_file = File(exists=True,
                     desc='path/name of registered file (if generated)')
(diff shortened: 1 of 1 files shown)


chrisfilo:
Instead of removing the exists=True flag you should make filling the outputs in _list_outputs conditional on the applyxfm input.

dmordom:
Yes that sounds better. My python is a bit rusty though... wouldnt want to mess that up. Just wanted to raise awareness on the issue :). Maybe simply reject this request and make the fix when u have time (no rush on my part at all).

djarecka:
Hi @dmordom, are you still working on this? Do you need some help?

djarecka:
@dmordom - please let me know if you still want to work it, otherwise I'll close it and review the issue you described

PR #1221, opened 2015-09-17
Apply python-future for Python 3.4 support

alexsavio:


alexsavio:
For Python 3 I need to futurize `prov` as well, which is in https://github.com/trungdong/prov/archive/rdf.zip

satra:
@alexsavio - don't worry about prov for now. in fact if you want you can strip it out for now or use dummy functions in provenance.py

i'm working on updating nipype's prov support to be compatible with current master which already supports py3.

alexsavio:
FAIL in both versions: nipype.algorithms.tests.test_splitmerge.test_split_and_merge

alexsavio:
Python 3 is struggling with the `sorted` here: https://github.com/alexsavio/nipype/commit/d8aa26efe95e3ee63703c5d6e4298e31454c5ec5#diff-f7320b522ddfa2bd56e0ca3c5ea19cc1R782
The test is nipype.pipeline.tests.test_utils:test_identitynode_removal.
Could you please explain how do you want to order a `list` of: `tuples of tuples` mixed with `tuples of str` ? So I can reprogram it. Thanks!

alexsavio:
I fixed the link, sorry.

alexsavio:
If I remove `dumps` and `loads` in `getsource` and `create_function_from_source` in https://github.com/nipy/nipype/blob/master/nipype/utils/misc.py#L60, it fixes many tests in `pipeline.test.test_utils`, this turns to be using `str` instead of `bytes` in Python 3. In addition, assigning `bytes` to a traits.Str raises an error.
Is there any reason that I'm missing for you to be using "dumped" strings in `function_str` in [Function](https://github.com/nipy/nipype/blob/master/nipype/interfaces/utility.py#L399?

satra:
@alexsavio - the thought process behind using pickle is to ensure that the contents of function_str could be hashed properly and also turned into a python function that can be executed. if there is a better way to achieve the same effect, we can definitely try it out.

alexsavio:
I have been busy with life. I'm currently stuck with unicodes, strings vs. Traits.
For example: `nosetests -s nipype.interfaces.tests.test_base:test_TraitedSpec`
In Python3 it works, now that everything is a unicode string, even what Traits returns, e.g., from `get_hashval()`.
On Python2 the string literals are now `unicode` (thanks to `from __future__ import unicode_literals`) but `TraitedSpec.get_hashval()` is returning `str`. There are a few options to fix this, but I can't decide yet which one is better.

bcipolli:
@alexsavio can you rebase on current master?

bcipolli:
>If I remove dumps and loads in getsource and create_function_from_source, it fixes many tests

This is a place where using `six` would make things easy; they take care of the import location change for `cPickle`...

alexsavio:
Hi @bcipolli, do you mean merging the current master to this PR?

bcipolli:
`git rebase upstream/master`. It replays your commits on top of the current master, rather than merging master on top of your changes.

alexsavio:
Like this?

bcipolli:
`This branch has conflicts that must be resolved` - have to fix this. Can you make me a collaborator on your repo?

alexsavio:
done

alexsavio:
I know, I haven't fixed everything yet

bcipolli:
@alexsavio I will try to rebase. I added some questions from reviewing some commits. Will let you know how it goes!

bcipolli:
I believe the Trait error is because `environ = traits.DictStrStr`, which likely requires a `str` (and not unicode). Will try to read a bit about how others have worked around things, if this is the issue.

alexsavio:
Yes, definitely that is the case. Thanks @bcipolli!

bcipolli:
Temporary fix is to relax from `DictStrStr` to `Dict`

bcipolli:
@alexsavio I force-pushed the rebase to your branch; you will need to rename/remove your current local branch to work on the force-pushed one.

I will do a bit more investigation on other errors, now that the `TraitError` workaround is in place. We can revisit that later; looks like there are plenty of errors to address in Py2 and Py3 in the meantime.

bcipolli:
@alexsavio I think the strategy of applying changes globally (e.g. `traits.Str => traits.Unicode, `from __future__ import unicode_literals`) is going to be problematic.

* I think we need a `Py23Str` that is `traits.Str` for py2, `traits.Unicode` for py3
* I think literals should not be unicode, as that means the calling code will have to be using unicode exclusively, and we can't guarantee that--users using py2 will be passing values to compare that are bytes.

I made the first change. The second one, however, I am starting to think should be removed everywhere. Otherwise, I don't think we have a hope for Py2 compatibility. What do you think?

alexsavio:
@bcipolli Thanks!

Maybe you are right. I did the `from future import unicode_literals` because it was the recommendation in `python-future`. 

Of course I realised it could be problematic once I stomped on the Traits failures. That is why I was asking. I suppose we have also the option to use `six` for strings (so, remove all the `unicode_literals` from the code) and then implement the `Traits.Py6Str` or `Traits.Py23Str` which was something in my mind, but I needed a hand to decide.

bcipolli:
See the "Drawbacks" here, particularly the "user code" example:
http://python-future.org/imports.html#should-i-import-unicode-literals

My intuition says to try and migrate without `unicode-literals`. Is there a specific commit we can remove to try it out (rather than adding a commit to delete the code insertion)?

alexsavio:
I guess you are right.
Sadly I used the command futurize, which makes many different changes not only the `unicode_literals` import.
This was done in the first commits that say 'Futurize ...'.
I can only see that the best thing is to add a commit to delete these imports. :/

bcipolli:
@alexsavio I made a commit to remove `unicode_laterals`. Now, checks such as this fail: `isinstance\(.*, str\)`. This should migrate to `six` (`isinstance\(.*, string_types\)`). Do you have time for that? I have to move on to other work today.

bcipolli:
@alexsavio I need a break from my work, so I will try out the above replacement.

alexsavio:
Hi @bcipolli I was out. How is it going can I help?

bcipolli:
@alexsavio I just pushed the changes; please take a look at the testing results and try to debug remaining failures. If you have any questions let me know! I will go back to my own work for now.

alexsavio:
ok, I will. Thanks!

satra:
i can also lend a little bit of a hand this weekend. we should try to make a push and get this operational soon. i fear that as each of the other PRs get merged we will just be rebasing for a while.

bcipolli:
:+1: agreed!

bcipolli:
@alexsavio I am working on relative imports ATM. Let's meet in the gitter room to chat, when we're working! https://gitter.im/nipy/nipype

bcipolli:
Making progress on this. I am seeing many doctest failures; I don't understand why they worked before:
* Many doctests use filenames that don't exist; this fails the `traits` checks. They must be skipped?
* Many doctests output dictionaries. Dicts are unordered, so string compare often fails. Using pprint to fix.

@satra or others.. any thoughts or recommendations? This isn't Python 3-specific; not sure why we're hitting this and master branch is not...

satra:
@bcipolli - the beginning of each file has a docstring that runs a piece of code to situate the execution in the testing directory. perhaps the top level docstring also needs the `unicode` annotation?

alexsavio:
@satra the culprit for that was `from builtins import str` that made all strings be unicode, which made the docstrings fail in py2k. @bcipolli removed these imports and now it is fixed.

bcipolli:
@alexsavio @satra Things seem to be working! I am cleaning up the commit history (fixing messages, squashing break/fix cycles, deleting unnecessary changes). 

In the meantime, here's a summary; not all changes are without controversy:

**Things that changed that need to be noted:**
* *Everything* uses relative imports now, except tests and examples.

**Broken / disabled tests:**
* test_multifield_join_node - fails sometimes, only on Py3.4. Ordering of objects seems not guaranteed. Is this a real failure or no?
* fsl/maths.py: doctests were disabled; tmp directories were being prepended to filenames.

**Unexpected changes that should be reviewed:**
* synchronize_iterables was broken. I reimplemented, but it's ugly. I need to go back and understand this more deeply. (BF: fix tests, fix broken function.)
* Remove filterwarnings('always', UserWarning) from modules (this is the default)
* Added __init__.py to all interface/XXX/tests directories (allows nosetests nipype.interface.XXXX.tests.test_blah:test_me)
* Use simplejson, not json (byte/unicode support mirrors Python 2/3; json does not)
    * added to requirements.txt, but without version currently.
* Use requirements.txt in .travis.yml (so dependencies are installed)
* Now some doctests use pprint.pprint(dict) (in standard library) to compare output.
* added paramiko to requirements.txt (it's used in the code...)
* suppress all but first "FSL" warnings from within .travis.yml (makes test output more readable)

**Things to do:**
* Whatever generates code for interfaces/semtools needs to be updated to use relative imports.

satra:
@chrisfilo - thanks to the heroic efforts of @bcipolli and @alexsavio, we now have a python 3 compatible version of nipype. it seems that circleci is choking on installing the `future` package, although the package is available on pypi.

we should get this merged asap before the other PRs are merged. when you get a chance can you take a look at the circleci issue? also we should now enable circleci to test on both python 2.7+ and 3.4+

satra:
@bcipolli - i've updated your summary.

bcipolli:
Thanks @satra, looks great!

I tried adding `pip install -r requirements.txt` to `circle.yml`, so that dependencies are installed before `setup.py` is called. An alternative would be to remove `future` dependencies from `setup.py`, which wouldn't be hard.

bcipolli:
Looks like the tweak to `circle.yml` worked. I looked at the failures, and there seem to be two issues:

* In interfaces, `results.stdout`/`stderr` is always unicode. Need to check that out... and preferably create a unit test that runs in `Travis.ci`.
* In the examples, the `future` imports are declared at the top of the file. Apparently that's not cool. In addition, I suspect that the examples may be uglier now; we should manually look them over and make them as pretty as possible.

Finally, we may want to review the use of `old_div`. I think the migration tool inserted it across the code, but it's likely not necessary in most places, and it definitely makes the code less readable.

satra:
@bcipolli - the latest commits should take care of all the issues - let's see what circle says.

also if you have a chance to scroll through this commit that would be great.
https://github.com/alexsavio/nipype/commit/df8fbaf2074db5df5be101fa69945d466ff7b37f

circle.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
     - "~/examples/fsl_course_data"
   override:
     - pip install --upgrade pip
+    - pip install -r requirements.txt
```
chrisfilo: Why do we need this? All dependencies should be in the setup.py file and picked up by

	pip install -e .
	
I recommend removing requirements.txt from circle.yml. Keeping it may lead to tests passing, but pip-installed packages failing to run because some dependency was in in requirements.txt, but was left out of setup.py.

satra:
@chrisfilo - it's hard to read the circleci output, but am i understanding that the only error is in doc generation?

chrisfilo:
both nosetest and doc generation are currently failing on this branch

nosetest:

	======================================================================
	FAIL: nipype.interfaces.spm.tests.test_base.test_spm_path(<type 'unicode'>, <type 'str'>)
	----------------------------------------------------------------------
	  File "/home/ubuntu/virtualenvs/venv-system/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
	    self.test(*self.arg)
	  File "/home/ubuntu/virtualenvs/venv-system/local/lib/python2.7/site-packages/numpy/testing/utils.py", line 354, in assert_equal
	    raise AssertionError(msg)
	AssertionError: 
	Items are not equal:
	 ACTUAL: <type 'unicode'>
	 DESIRED: <type 'str'>
	 
documentation:

	/home/ubuntu/nipype/doc/users/examples/fmri_ants_openfmri.rst:35: ERROR: Unexpected indentation.
	/home/ubuntu/nipype/doc/users/examples/fmri_ants_openfmri.rst:49: ERROR: Unexpected indentation.
	
	and many similar

bcipolli:
Thanks @chrisfilo . I was able to reproduce the doc build issues locally; they are due to a missing blank line between the file docstring and first import. I can fix those.

The other issue is due to output streams coming out unicode on python 2/3. Haven't figured out why this is yet.

bcipolli:
@satra @alexsavio tests are passing!! :) One last issue I thought of last night; the use of `loads`/`dumps`. I believe some were removed (per the discussion above) without replacement; please take a look for example at: 
https://github.com/alexsavio/nipype/commit/6ceb87eaa00a0c82b113dcabbee4515371906e12
https://github.com/alexsavio/nipype/commit/f531ed6e869032f80d91a0b264f26992b97cd9e1

to confirm this won't cause any problem.

satra:
@bcipolli - i'm going to open an issue for loads/dumps - i have no recollection why i wrote it with pickling. so we will monitor it in case we run into trouble, but otherwise i think this is ready for merging.

@chrisfilo - could i leave it to you to take any last look and hit the merge button?

chrisfilo:
Did you guys try to switch Circle to python 3 for one commit and see if the workflows still work? That would convince me that we are indeed P3 compatible.

bcipolli:
@chrisfilo no, we didn't. Any suggestion how to do that?

chrisfilo:
According to this https://circleci.com/docs/language-python you need to add the following lines to circle.yml:

	machine:
	  python:
	    version: 3.5.0
	    
There are more versions available: https://circleci.com/docs/environment#python

bcipolli:
@chrisfilo could you help look over the failures? There were four. I looked over the text, and it wasn't clear to me what the problem might be, nor how to reproduce them locally.

The easiest one to reproduce locally seemed to be: `nipype.interfaces.tests.test_io:test_s3datagrabber_communication`, which I guess just required installing `boto`, but when I run the following command, it hangs for 10 minutes:

python3 `which nosetests` nipype.interfaces.tests.test_io:test_s3datagrabber_communication

chrisfilo:
The errors are indeed strange. I'll try rebuilding without cache. Hold tight it will take a while.

bcipolli:
Please see #1253

satra:
woohoo!

alexsavio:
:)

PR #1225, opened 2015-09-21
Fix FreeSurfer RobustRegister gen out_reg_file

afloren:
Current implementation uses the source file's path for the automatically generated file instead of the cwd. The result is that if multiple RobustRegister nodes are connected to a single source file, the resulting autogenerated files will overwrite each other.


chrisfilo:
please rebase/merge master - we have fixed the CircleCI tests

PR #1234, opened 2015-10-05
[ENH] Remove warnings in tSNR calculation

oesteban:
Also simplified the interface using default inputs.


nipype/algorithms/misc.py
```diff
@@ -262,6 +262,11 @@ class TSNRInputSpec(BaseInterfaceInputSpec):
     in_file = InputMultiPath(File(exists=True), mandatory=True,
                              desc='realigned 4D file or a list of 3D files')
     regress_poly = traits.Range(low=1, desc='Remove polynomials')
+    tsnr_file = File('tsnr.nii.gz', usedefault=True, desc='output tSNR file')
```
satra: since these are all output_files, we should add `hash_files=False`

satra: actually could we use `name_source` and `name_template` here? i'm not sure though whether they work for all types of interfaces. hopefully then do.

oesteban: Unfortunately, ```name_source``` & co. are only defined for ```CommandLine```

satra:
by making things default and not being able to use `name_source` we loose the ability to name things based on the input. which means if multiple files are passed in, nipype will create the same named file for each one. i think we need to allow for different names here. so why don't we hold off on this till we move `name_source` and `name_template` one level up.

chrisfilo:
please rebase/merge master - we have fixed the CircleCI tests

oesteban:
Ok, I'll remove the inputs/outputs stuff, so that the default parameters using name_source will be done on #1240 .

oesteban:
Ok, I have started a list on #1240 with those interfaces that should be revised when ```name_source``` works also in ```BaseInterface```s. I have added ```TSNR``` to the list, and I will be updating the list with the candidate interfaces.

I think we should review and merge this for now, and come back to the ```name_source``` problem when the implementation is available.

oesteban:
Good to go?

PR #1249, opened 2015-10-13
Fix FreeSurfer SurfaceTransform gen out_file

afloren:
SurfaceTransform may specify either a source_file or a source_annot_file (mutually exclusive). However, if out_file is not specified and source_annot_file is specified, _list_outputs will try to generate the out_file name using the source_file (which won't be defined in this case). Added an extra check to see if self.input.source_file is defined and if it's not then use self.inputs.source_annot_file instead.


blakedewey:
Looks good. Merging.

PR #1255, opened 2015-10-17
FIX: PEP8 correction for all error E100-E399

miykael:
This PR covers part of the PEP8 violations mentioned in #597. Almost all of them were done with autopep8, by using the following code:

    pep8ID=E303

    for f in `find . -name "*.py"`
    do
        if [[ "$f" != *"test_auto"* ]]
        then

            count=`pep8 --select $pep8ID $f | wc -l`

            if [ "$count" -ne "0" ]
            then
                autopep8 -i --select $pep8ID $f
                pep8 --select $pep8ID $f
            fi
        fi
    done

I've only adjusted a few spaces and indentations, where autopep8 couldn't solve the problem. And any files with "test_auto" were ignored during the correction.


miykael:
I'm not sure how to tackle the circleci and travis test fails. Are there any problems with my commits? And what can I do to facilitate the merging?

miykael:
Ah, I see. Thanks @chrisfilo. I think this is because it compares it to the auto generated code from [nipype/utils/nipype_cmd.py](https://github.com/nipy/nipype/blob/master/nipype/utils/nipype_cmd.py#L16). Should I know change my code or the file ``nipype/utils/nipype_cmd.py``?

I'm still struggling with github modifications. How can I easly change a file in a specific commit?

chrisfilo:
I would just revert the problematic lines of the test code in a new commit.

oesteban:
I'd propose to add your script to the make target ```check-before-commit```. This way we can have PEP8 violations under control in the long term. What do you think?. How this PR coordinates with the new Python 3 integration, @satra ?

satra:
@oesteban - i think this PR is being done on top of the merged py3 refactor.

miykael:
Yes, the PR is on top of the merged py3 refactor. I've tried to be as fast as possible with the PEP8 corrections to make sure that it doesn't interfere with to many other PR.

@oesteban, I'm not sure if it's the best idea to put the ``pep8`` check or ``autopep8`` into ``check-before-commit``. There might be some PEP8 errors or warnings that perhaps shouldn't be corrected. At least not for now. It might be an idea to look at it again, once most PEP8 issues are taken care of?

chrisfilo:
I'm happy merging it the way it is (also worried we will have to wait a long time for everyone to rebase their PRs).

miykael:
What would be the best way to tackle the remaining PEP issues? Having many small PRs or one big one containing all remaining corrections?
And what would be the best time point? Now or perhaps better just after a release?

blakedewey:
Have we decided to merge this? I think now would be better than later. As many files will begin conflicting.

satra:
merging this. we will open a separate PR for the remaining PEP8s.

PR #1259, opened 2015-10-26
Added support for the OAR task scheduler http://oar.imag.fr 

demianw:


chrisfilo:
This looks like a great addition! Could you add OAR to the https://github.com/nipy/nipype/blob/master/doc/users/plugins.rst as well as https://github.com/nipy/nipype/blob/master/CHANGES. This will let other users lean about this new feature.

demianw:
Thanks @chrisfilo! Hope you are doing well. Did what you said, merged committed and pushed.

PR #1260, opened 2015-10-28
ENH: Cleanup Registration in ANTs

blakedewey:
This PR is to fix a few typos and tests in `Registration`, as well as tackle some of the `TODO` items. I have also added a `verbose` input that defaults to `True`. The default to `True` is to allow for the same output as previous versions of antsRegistration, where iterations and stages were recorded to stdout.


chrisfilo:
@hjmjohnson could you have a look?

blakedewey:
I added one more tweak. The inverse composite transform may not be created if any of the initial transforms are not invertable.

PR #1264, opened 2015-11-01
STY: PEP8 style correction for all warnings, and almost all errors. Except E402, E501 and E731

miykael:
This PR covers the rest of PEP8 violations mentioned in #597, that were not already covered by #1255. Only exceptions are:
- E402 module level import not at top of file (N=402)
- E501 line too long (80 > 79 characters) (N=8631)
- E731 do not assign a lambda expression, use a def (N=40)

In my opinion, E402 cannot be corrected and E731 would be a mess to correct.
E501 can be corrected, but just needs a lot of manual correction. Therefore, I think it makes sense to correct them with another PR.

Also, there are the following PEP8 violations that I haven't corrected and I would be happy to have a feedback if they should or shouldn't be changed:

E712 comparison to False should be 'if cond is False:' or 'if not cond:'
The four errors are all in file ``nipype/pipeline/plugins/base.py``, on line [235 + 236](https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L235), [331](https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L331) and [341](https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L341). Because they compare the values of a numpy array to a boolean, we cannot replace '==' with is.

W191 indentation contains tabs and E101 indentation contains mixed spaces and tabs
The six W191 and two E101 violations exist because of ``nipype/utils/tests/test_cmd.py`` on line [84-89](https://github.com/nipy/nipype/blob/master/nipype/utils/tests/test_cmd.py#L84), because of the tab character in file ``nipype/utils/nipype_cmd.py`` on line [17](https://github.com/nipy/nipype/blob/master/nipype/utils/nipype_cmd.py#L17).

E116 unexpected indentation (comment)
The 5 E116 violations are caused by ``nipype/interfaces/fsl/preprocess.py`` on line [243-247](https://github.com/nipy/nipype/blob/master/nipype/interfaces/fsl/preprocess.py#L243) and I don't dare to delete this.

E265 block comment should start with '# '
The 28 E265 violations are caused by multiple lines in file ``doc/conf.py``, for example line [53](https://github.com/nipy/nipype/blob/master/doc/conf.py#L53), [73](https://github.com/nipy/nipype/blob/master/doc/conf.py#L73), [77](https://github.com/nipy/nipype/blob/master/doc/conf.py#L77), etc. I didn't add the space after `#` so that it's still clear what is a comment and what is commented code.


nipype/utils/tests/test_docparse.py
```diff
@@ -50,9 +50,9 @@ def test_build_doc():
     The name of the input file
 outfile : str
     The name of the output file
-outline : 
+outline :
      something about an outline
-fun : 
+fun :
```
satra: this change is affecting a test. the test should be updated.

satra:
@miykael -  can you merge this with master now?

miykael:
Yes, but how do I do this? Do I merge my [pep8StyleCorrections branch](https://github.com/miykael/nipype/tree/pep8StyleCorrections) with master and than just correct any given conflict with a new commit?

satra:
yes

PR #1278, opened 2015-11-17
Fix: ANTs cleanup

blakedewey:
Cleanup of ANTs interfaces.
-- Various PEP8 and other style corrections. Cleanup of `_format_xarray`
-- Finished the `TODO` items in ApplyTransforms. 
-- Added option for `use_random_seed` for Atropos
-- Added options for `verbose` in ApplyTransforms/Atropos
-- Added option for float computations in ApplyTransforms (`float`)


nipype/interfaces/ants/__init__.py
```diff
@@ -7,11 +7,13 @@
 from .registration import ANTS, Registration
 
 # Resampling Programs
-from .resampling import ApplyTransforms, ApplyTransformsToPoints, WarpImageMultiTransform, WarpTimeSeriesImageMultiTransform
+from .resampling import ApplyTransforms, ApplyTransformsToPoints, WarpImageMultiTransform, \
```
satra: instead of using a `\` we could use a tuple representation

```python
(ApplyTransforms, ApplyTransformsToPoints, WarpImageMultiTransform,
 WarpTimeSeriesImageMultiTransform)
```

satra:
LGTM

blakedewey:
Pushed changes to imports. Merge when ready.

nipype/interfaces/ants/segmentation.py
```diff
@@ -486,24 +491,26 @@ class antsCorticalThicknessOutputSpec(TraitedSpec):
     BrainVolumes = File(exists=True, desc='Brain volumes as text')
 
 
-class antsCorticalThickness(ANTSCommand):
+class CorticalThickness(ANTSCommand):
```
chrisfilo: Isn't this breaking backwards compatibility?

blakedewey:
So the removal of "ants" from the class names would break backwards compatibility. What would be the process to go through this? It has been done in the AFNI interfaces before. `antsCorticalThickness` was added 1 year ago and `antsBrainExtraction` was just added.

nipype/interfaces/ants/segmentation.py
satra: actually that's a good point. we can add a deprecation warning to the class:

```
class antsCorticalThickness(CorticalThickness):
    DeprecationWarning('This class has been replaced by CorticalThickness and will be removed in version 0.13')
```

chrisfilo:
why not just do a 

	import segmentation.CorticalThickness as antsCorticalthickness
	
Adding a deprecation warning would be good as well.

blakedewey:
Updated to include deprecated classes.

satra:
@blakedewey - i just ran into the issue that the verbose flag for antsRegistration is not available in version 2.1.0. i saw that you added it recently. i think we need some form of versioning for this flag.

blakedewey:
@hjmjohnson Do you know when this flag came up?

blakedewey:
Nvm. I found the commit and it was after 2.1.0 came out. I didn't realize they added a completely new option. I will remove the additions.

hjmjohnson:
commit c37982901a1b8c2c51d1b4353b289a44547d63e2
Author: Hans Johnson <hans-johnson@uiowa.edu>
Date:   Wed Jan 14 11:18:40 2015

    ENH: Add version command line option
    
    In order to identify different versions of
    the ANTs program, it is benificial to have
    command line arguments use introspection to
    report their version and git hash codes.
    
    ./bin/antsRegistration -v
    All_Command_lines_OK
    ANTs Version: 15.0.0rc3.dev1-g39ad7
    Compiled: Jan 14 2015 10:58:35
    
    ./bin/antsRegistration --version
    All_Command_lines_OK
    ANTs Version: 15.0.0rc3.dev1-g39ad7
    Compiled: Jan 14 2015 10:58:35
    
    This patch set adds the version information
    for the antsRegistration program.

satra:
@hjmjohnson - hoping there is an ants release soon :)

hjmjohnson:
@satra   What constistutes a release?  Do you just want a new version number presented?

I could make a new release version easily so that we have something to identify in nipype.

Do you want version set to 2.2.0.???

ANTs Version: 2.1.0.post407-g8a995

Hans

satra:
essentially - the last release was 2.1.0 - if the API has changed it should be 3.0.0. see http://semver.org/

PR #1289, opened 2015-12-10
FIX: S3io

shariqiqbal2810:
Fixed issue with S3DataGrabber not converting paths properly. More information in commit messages.  Also, fixed tests to catch this.


chrisfilo:
Should fix #1307

PR #1292, opened 2015-12-10
SGEGraph fix

billbrod:
SGEGraph was incorrectly formatting job names for `-hold_jid` in the
qsub calls in the submit_jobs.sh script. qsub wants `-hold_jid
j0_jobname,j1_jobname`, while SGEGraph was wrapping the names with `${}`
so they were `-hold_jid ${j0_jobname},${j1_jobname}`. This was causing
the submission of all jobs with `-hold_jid` to fail.

### DIFF
--- a/nipype/pipeline/plugins/sgegraph.py
+++ b/nipype/pipeline/plugins/sgegraph.py
@@ -122,7 +122,7 @@ class SGEGraphPlugin(GraphPluginBase):
                         for jobid in dependencies[idx]:
                             # Avoid dependancies of done jobs
                             if not self._dont_resubmit_completed_jobs or cache_doneness_per_node[jobid] == False:
-                                values += "${{{0}}},".format(make_job_name(jobid, nodes))
+                                values += "{0},".format(make_job_name(jobid, nodes))
                         if values != ' ':  # i.e. if some jobs were added to dependency list
                             values = values.rstrip(',')
                             deps = '-hold_jid%s' % values


satra:
@hjmjohnson - could you please take a look here?

hjmjohnson:
I will look at this on Sunday.   The proposed changes do not look correct to me.  

Hans

billbrod:
The proposed changes were necessary for SGEGraph to work on the cluster I'm using. If they're not correct for other clusters, does anyone have a sense why mine would be different?

hjmjohnson:
This code generates a bash script that is used to submit the Graph jobs to SGE.  In the command " -hold_jid ${j0_jobname}", j0_jobname is a bash variable that holds the job_id returned from qsub.  j0_jobname is *NOT* a python variable name.  

In order for this to work properly across many job runs submitted simultaneously, it is essential that j0_jobname be globally unique for all jobs running on the cluster. For this reason, we must use the autogenerated identifier produced by qsub at the time of submission rather than the python name for that job.    To reiterate, if two people are running the same nipype pipeline on the cluster at the same time, if you use the python job names, then the graph dependencies cause the two users jobs to depend on each others jobs.   This is only valid if you can ensure that users never run more than one version of the pipeline (or sub-pipeline) at a time.

image

billbrod:
Ah okay, I understand. It looks to me like the `awk '{print $3}'` is where the issue is: it's extracting some extra info because of the way the cluster handles email notifications. If the user doesn't set an email, it prints `Invalid EMAIL detected! tuning off notifications. update to valid email address in future. Your job jobid ("j0_jobname") has been submitted`. I'm not familiar with `awk`, do you know of a way to use it to grab the jobid if the output looks like the above?

hjmjohnson:
@satra This patch set needs to be aborted.  The solution to fix the problem is incorrect.  

The problem is related to extracting the unique job identificaiton number from SGE itself.  It appears that their installation has a non-conformant response when a job is submitted, and the code for extracting the globally unique number assigned by SGE is not working for them.


If a correct solution is found, then this can be re-opened.

Issue #878, opened 2014-07-15
Nodes for Conditional Execution


blakedewey:
Is there any room in the code base for a node for use in conditional execution? I was thinking a specialized node class that ran code and returned a boolean that could then be used to terminate a branch by using the donotrun in the linear plugin or similar in the distributed plugins. Any thoughts?

Issue #819, opened 2014-03-19
Dynamic iterables


dmwelch:
@satra @chrisfilo Having runtime defined iterable attributes would be awesome, especially for big data sets.  Issue has been raised on the mailing list before: https://groups.google.com/forum/#!msg/nipy-user/pYW70-xAeu4/4MNbiDVQvAMJ 

The current option is a MapNode of a Function that contains the workflow to iterate...

Cheers,
Dave

AlanKuurstra:
+1, This feature would be useful in combination with the experimental `remove_node_directories` for large datasets whose processing eats up disk space. 

I have an fmri dataset that must be registered to a high resolution atlas.  A number of preprocessing steps are done before the fmri file is split into separate time volumes. It wold be nice to send the split volumes into the next node's iterables list.  After the split, each volume would be processed by a chain of subsequent nodes separately.  The subsequent chain of node directories would be removed (freeing up space) before moving on to the next time volume.

effigies:
This will not be implemented in Nipype 1, but is being implemented in [Pydra](https://github.com/nipype/pydra/), which will be the engine for Nipype 2.

PR #882, opened 2014-07-24
WIP: Interfaced Workflows and GraftWorkflow

oesteban:
These workflows provide automatic input/output nodes generation, with some other utilities such as fast connection (avoiding to specify the connecting fields).

GraftWorkflow is intended to create evaluation workflows, where all the inputs are the same but several different methods are to be compared, stacking the outputs in lists.

download

Two additional interfaces have been coded: CollateInterface and MultipleSelectInterface. These interfaces multiplex/demultiplex outputs/inputs respectively.
- [x] InterfacedWorkflow class
- [x] GraftWorkflow class
- [x] CollateInterface and MultipleSelectInterface (mux/demuxer)
- [x] doctests
- [x] regression tests
- [x] ~~open new _experimental_ modules within distribution~~
- [x] documentation, on the same vein of iterables, mapnodes, etc

EDIT [Dec 31, 2015]:
After commenting with @chrisfilo,  we have added the following features/requirements for this PR:
- [ ] Making InterfacedWorkflows a true blackbox (like Nodes are):
  - [x] Do not expose (in/out)putports of internal nodes as inputs/outputs
  - [ ] Do not allow inserting sub-workflows with connections to external nodes.

Once we get these InterfacedWorkflows to work, we will then focus on #1299.

PR #1299, opened 2015-12-15
[WIP,ENH] Conditional Nodes and Workflows

oesteban:
This is a first take on conditional execution of nodes in workflows. Runtime decisions in nipype are old feature requests (i.e. #819 and #878 for nodes, and #1081 for workflows).

Edit: this PR is documented here: https://github.com/oesteban/nipype/blob/enh/ControlNodes/doc/users/runtime_decisions.rst

~~The new `ConditionalNode` adds an special input to the inner interface named `donotrun` that can be used to disable the node. It is the user's responsibility to maintain the consistency along the whole graph.~~

~~The new `ConditionalWorkflow` adds an special input node: when setting `checknode.donotrun` to `True`, the workflow is disabled. These workflows automatically cast all its nodes to `ConditionalNode`s and connected to the `checknode.donotrun` input.~~

~~Finally, a new `CachedWorkflow` is derived from `ConditionalWorkflow`, with the addition of a `cache_map` (list of tuples indicating the input name and the output to which it is mapped). If all the inputs of a `cachenode` (as listed in the first element of the `cache_map` list of tuples) are defined, the `donotrun` of the workflow is activated and all the nodes skipped. Additionally, the workflows outputs are mapped from the `cachenode`. Otherwise, they are filled with the normal execution of the workflow.~~

~~Example:~~

```
[example moved to documentation]
```

~~This will return `outputnode.out1 = 5` and `outputnode.out2 = 5`~~
~~If the line `# cwf.inputs.cachenode.c = 1` is uncommented, then `outputnode.out1 = 1` and `outputnode.out2 = 7` since all the cache inputs are set and mapped.~~

This PR is labeled as WIP because there are some missing elements:
- [x] Include this example in the documentation, improve documentation
- [x] Analyze side-effects of skipping nodes, currently this is just a hard decision. Presumably, something should be cached when `donotrun` is `True`.
- [x] Convert sub-workflows of ConditionalWorkflow and CachedWorkflow in ConditionalWorkflows and connect the `donotrun` input.

~~This PR requires #1298.~~ [Already merged]


satra:
this is great. one thing is that i have been contemplating rewriting the workflow engine code to integrate node and mapnode into the same construct. i was hoping to work on this towards the end of this month. should i outline the changes in a separate issue? i haven't completely thought through the changes, but i was hoping that it would at least implement all these patterns: http://www.workflowpatterns.com/ and be scalable and expandable completely at runtime. it would also allow additional execution semantics to run subgraphs or chains of nodes using node/storage/memory isolation.

oesteban:
Hi @satra I really agree with you on the convenience of attaching to those patterns you propose (great resource!, btw).

In this case, this could be a kind of pattern of data precondition adding signaling inputs to the nodes (the ConditionalNode special input ```donotrun```, in this case). I don't think my approach implements the pattern very well. An alternate approach I've tried is graph editing (see https://github.com/oesteban/nipype/tree/enh/ConditionalWorkflow) but I found that route a bit unreliable.

I think the redesign you propose deserves a separate issue, identifying which patterns are currently necessary (to start implementation from them) and then organize the remaining patterns in order of priority. What do you think?

satra:
@oesteban - thanks - perhaps we can have a google hangout next week or the week after to go over this. that will give me a little bit of time to even attempt a basic refactor.

nipype/pipeline/engine/workflows.py
```diff
@@ -0,0 +1,1096 @@
...
+    def _check_conditional_nodes(self):
+        from nipype.interfaces.utility import IdentityInterface
+
+        def _checkdefined(val):
+            from nipype.interfaces.base import isdefined
+            if isdefined(val):
+                return bool(val)
+            return False
+
+        if checknode is None:
```
chrisfilo: Codacy Issue found: [Using variable 'checknode' before assignment](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/1590241444/issues/source?bid=2767207&fileBranchId=3022732#l977)

oesteban: Yes, since https://github.com/oesteban/nipype/commit/6f3a3aa2cf8c95ba08af3a95b95430419c63ef91 I'm doing a bit of refactoring, now that I have the implementation clear.

oesteban:
Hi @satra, the more I advance on implementing this, the strongest is my feeling in favor of your proposed refactoring. I think there are several concepts from http://www.workflowpatterns.com/ that would really improve nipype.

First of all, having *atomic tasks* (like current nodes in the flat graph before execution) and *block tasks* (workflows) would ease implementation.

Second, the current construct of Node(IdentityInterface) is just an implementation of *data buffers* and *data channels*. Here and in #882 I'm finding a problem on this because workflows are not aware of their interface. This forces the user to apply the typical 'inputnode.inputname' construct when connecting inputs to workflows. If workflows were *block tasks*, they'd have some inputs defined and plugins would be able to run these *block tasks* as such, or dissect the internal *atomic tasks* if required.

In summary, let's go for this meeting :+1:

satra:
@oesteban - sounds good. while we are using this as a sounding board, here are some things to dwell on:

https://github.com/nipy/nipype/issues/1301

oesteban:
Cool! I like the features list.

nipype/pipeline/engine/graph.py
```diff
@@ -18,142 +18,34 @@
...
+from .utils import get_print_name
 
-from .. import logging, config
-logger = logging.getLogger('workflow')
 
+package_check('networkx', '1.3')
+import networkx as nx
 try:
     dfs_preorder = nx.dfs_preorder
+    logger.debug('detected networkx < 1.4 dev')
```
chrisfilo: Codacy Issue found: [Using variable 'logger' before assignment](https://www.codacy.com/app/krzysztof-gorgolewski/nipy_nipype/file/1605601501/issues/source?bid=2767207&fileBranchId=3022732#l44)

blakedewey:
Super excited about this features list. Lots of awesome improvements. Let me know if you need help or feature testing.

oesteban:
Hi there, I think this is ready to be reviewed. So yes, @blakedewey, I'd really appreciate some help with testing :)

satra:
@oesteban - thanks! 

i'll start taking a look in a day. is that ok?

chrisfilo:
@sgiavasis and @carlohamalainen  could you also have a look - this might be useful/relevant to C-PAC and QAP.

sgiavasis:
@chrisfilo Interesting..! I will also take a look.

oesteban:
Basically, this PR consists on several patches. I think we should definitely go for a refactoring sooner or later. But for now, this can be useful.

effigies:
I'm a little inclined to call this an orphan. Is there any interest in reviving this or splitting it into PRs that provide partial functionality, or is this wholly superceded by the 2.0 refactor?

oesteban:
Yeah, I think this is orphaned. We can cherry pick from it if we wanted something in either 1.0 or 2.0.

PR #1302, opened 2015-12-17
ENH: Add interfaces for MINC tools.

carlohamalainen:
I wrote NiPype interfaces for some MINC tools when I ported volgenmodel to NiPype (https://github.com/carlohamalainen/volgenmodel-nipype).

Recently Chris Filo Gorgolewski asked me to submit a PR of my MINC interfaces as he is working on wrapping other MINC tools.


chrisfilo:
Thanks for doing this! A few comments
* I am concerned is the size of the test data. Our usual patter for doctests to work is creating empty files in https://github.com/nipy/nipype/tree/master/nipype/testing/data and add a snippet on top of the module file https://github.com/nipy/nipype/blob/master/nipype/interfaces/fsl/preprocess.py#L7
* Could you split the base classes from actual interface classes like we do for FSL and others?

carlohamalainen:
I've split the base module.

With the tests, am I meant to generate actual test data at run time? I notice that most of the doctests in the FSL module have {{#doctest: +SKIP}} so they are never run?

carlohamalainen:
Sorry, disregard my question about doctests. I've updated things to match the FSL snippet etc.

mwaskom:
I think that ultimately you'll want to rebase and remove the commits with the data files, otherwise they'll still end up in the git history.

carlohamalainen:
For some reason github is not picking up my rebase-squash so I've opened a new PR: https://github.com/nipy/nipype/pull/1304

PR #1307, opened 2015-12-25
FIX proposed for S3io

jason-wg:
When I tried running this code, type(outputs[key]) was never == list or == str, so the method, self.s3tolocal, never got called. When I tried debugging, type(outputs[key]) turned out to be of type 'unicode'


jason-wg:
@chrisfilo  FIX: S3io #1289 fixed this. Closing this pull request.

PR #1314, opened 2016-01-03
py3 and ipython fixes

poldrack:
this PR has two fixes:
- I discovered a bug in the ipython kernel (https://github.com/ipython/ipykernel/issues/91) that caused faulthandler.enable() to fail with an IOError.  I added explicit catching of IOError to catch this
- decoding of stderr/stdout using the standard .decode() failed in py3.  I added decoding using ISO-8859-1

### DIFF
--- a/nipype/interfaces/base.py
+++ b/nipype/interfaces/base.py
@@ -1271,9 +1271,16 @@ def run_command(runtime, output=None, timeout=0.01, redirect_x=False):
     if output == 'allatonce':
         stdout, stderr = proc.communicate()
         if stdout and isinstance(stdout, bytes):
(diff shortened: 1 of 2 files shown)


nipype/interfaces/base.py
```diff
@@ -1271,9 +1271,16 @@ def _process(drain=0):
     if output == 'allatonce':
         stdout, stderr = proc.communicate()
         if stdout and isinstance(stdout, bytes):
-            stdout = stdout.decode()
+            try:
+                stdout = stdout.decode()
+            except:
```
chrisfilo: could you add an expected exception class here?

PR #1319, opened 2016-01-15
Pass on the oarsub args

srothmei:
Previously, the arguments supposed to be passed onto the oarsub command just were discarded. Now they are correctly passed to the command such that one can for example set higher wall times or restrict job submission to a certain subset of nodes etc

Tested and working on our OAR cluster


chrisfilo:
There was something wrong with the test server (CircleCI) I'm rebuilding it.

chrisfilo:
Could add a line about this option to the execution plugins documentation? https://github.com/nipy/nipype/blob/master/doc/users/plugins.rst

srothmei:
Actually i only made sure that the arguments sued to execute this plugin are really passed onto it. As i described, in the previous code they were simply discarded. 
From what i can tell, the current section about the OAR-plugin in the documentation you linked is quite comprehensive.

PR #1321, opened 2016-01-15
Ensure presence of lxml's binary deps

aharding:
Bug:
The Vagrant shell provisioner fails out when the lxml wheel fails to build.

Cause:
lxml's binary dependencies libxml2 and libxslt are missing from the image's default installation.

Proposed solution:
Use the Vagrant shell provisioner to ensure the presence of Ubuntu header packages "libxml2-dev" and "libxslt1-dev", which grab their respective libs as dependencies (see diff).
Edit: title/message should perhaps include "in Vangrant config".

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,6 +25,13 @@ chmod +x miniconda.sh
 ./miniconda.sh -b
 echo "export PATH=$HOME/miniconda/bin:\\$PATH" >> .bashrc
 
+# install nipype dependencies' dependencies:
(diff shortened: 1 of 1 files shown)


chrisfilo:
I think the failing test is unrelated to your PR. Nonetheless could you rebase/merge master?

satra:
@aharding - can you check if the current vagrant provisioner (in nipype master) is still failing? if not let's close this PR, if yes, lets' fix it.

PR #1322, opened 2016-01-15
ENH: Add new command line for BRAINSABC tool

hjmjohnson:
A new featue set has been added to improve
the quality of segmentations from the BRAINSABC
tool.

purePlugsThreshold = If this threshold value is greater than zero, only pure samples are used to compute the distributions in EM classification, and only pure samples are used for KNN training. The default value is set to 0, that means not using pure plugs. However, a value of 0.2 is suggested if you want to activate using pure plugs option.

numberOfSubSamplesInEachPlugArea -- How many samples in each plug area to create.


hjmjohnson:
Ran 14705 tests in 549.389s

OK (SKIP=57)
removed spaces
built docs
ran test
generated spec tests

hjmjohnson:
@satra A small change for a new feature we added to the BRAINSABC program.  Could this please be merged soon?

PR #1328, opened 2016-01-23
[ENH] Check AFNI version in SkullStrip to avoid xvfb

oesteban:
Added version checking in afni.SkullStrip to avoid use of virtual display with AFNI >= 16

(comes from https://github.com/poldracklab/mriqc/issues/25#issuecomment-173917499)


oesteban:
It'd be great for our team if somebody else could review this PR soon :)

( @satra, @chrisfilo ...?)

satra:
LGTM

oesteban:
Since the last change passed travis, and it was just removing one unnecessary import, I merge this before circleci passes (right now it shows degraded service and the build has been working for more than 4h, besides it already passed in the previous commit)

PR #1334, opened 2016-01-28
[FIX] Missing link to fsl-feeds from apt (rel. #1333)

oesteban:
Since ```~/examples/fsl-feeds``` was cached, this problem passed unnoticed. Amends #1333.


oesteban:
I was to merge it now, since the soft link worked in a neurodebian virtual machine, and this pr is important. Thanks.

PR #1341, opened 2016-02-01
[ENH] Improvements on fsl.Smooth

### LINKED ISSUE #1324: documentation of fsl.utils.Smooth
documentation says:

fwhm: (a float)
        flag: -kernel gauss %f -fmean, position: 1

when it should note that the value is passed after transforming from fwhm to sigma

oesteban:
- Added new input ```sigma``` which passes value as given, and made it compatible with the input ```fwhm```.
- Added 3 doctests
- Improved documentation.

Fixes #1324


nipype/interfaces/fsl/utils.py
```diff
@@ -150,33 +154,51 @@ class SmoothOutputSpec(TraitedSpec):
...
-        outputs['smoothed_file'] = os.path.abspath(outputs['smoothed_file'])
-        return outputs
-
     def _format_arg(self, name, trait_spec, value):
         if name == 'fwhm':
             sigma = float(value) / np.sqrt(8 * np.log(2))
             return super(Smooth, self)._format_arg(name, trait_spec, sigma)
         return super(Smooth, self)._format_arg(name, trait_spec, value)
 
+    def _parse_inputs(self, skip=None):
```
satra: i think you can have `mandatory=True` for the fields, which should take care of the need to do this.

mwaskom:
I still think there is the broader issue that we now print the argument string in the help but provide no information about possible internal reformatting of arguments.

This also probably isn't the only place in our fsl interfaces where we accept a `fwhm` parameter and convert it to sigma internally. Even without solving the broader problem, it might be good to extend this change to those interfaces for the sake of consistency? `fsl.SUSAN` comes to mind.

nipype/interfaces/fsl/utils.py
oesteban: Ok, I wasn't sure.

oesteban:
Yep, my idea was to give a short-term answer to the problem raised by @poldrack . I have also updated the documentation and mentioned the conversion of the argument, and showed it with an example commandline.

I agree this should be definitely extended to other interfaces and, if possible, take on the broader issue you point out. But I would do this in a separate issue/PR so that this one goes through.

nipype/interfaces/fsl/utils.py
oesteban: Ok, it does, but only refers to the first of the xor parameters specified on the error message. For now it's ok.

oesteban: Mmmm it's a bit worse than that. Sometimes it asks for the new sigma, sometimes for the old fwhm. We should print all the xor parameters in the message (or simplify it).

oesteban:
@mwaskom @satra @chrisfilo do you think this is fine to go ahead?. Then the ideas proposed by @mwaskom would be included through a different PR, what do you think?. Even a general polishing of all fsl interfaces (the issue here and any others) would be great news.

PR #1295, opened 2015-12-12
new interface for Bru2

TheChymera:
Following up on #1293

PR #1353, opened 2016-02-08
added new interface bru2nii

TheChymera:
Updated #1295

### DIFF
--- a/nipype/interfaces/bru2nii.py
+++ b/nipype/interfaces/bru2nii.py
@@ -31,7 +31,7 @@ class Bru2(CommandLine):
 
 	>>> from nipype.interfaces.bru2nii import Bru2
 	>>> converter = Bru2()
-	>>> converter.inputs.input_dir = os.getcwd()
+	>>> converter.inputs.input_dir = brukerdir
 	>>> converter.cmdline
 	'Bru2 -o brukerdir'
 	"""
--- a/nipype/interfaces/bru2nii.py
+++ b/nipype/interfaces/bru2nii.py
@@ -31,10 +31,9 @@ class Bru2(CommandLine):
 
 	>>> from nipype.interfaces.bru2nii import Bru2
 	>>> converter = Bru2()
-	>>> converter.inputs.input_dir = '/data/20151103_213035_4001_1_1/10/'
-	>>> converter.inputs.nii_file = '/data/20151103_213035_4001_1_1/10'
+	>>> converter.inputs.input_dir = '.'
 	>>> converter.cmdline
-	'Bru2 -o /data/20151103_213035_4001_1_1/10 /data/20151103_213035_4001_1_1/10/'
+	'Bru2 -o .'
 	"""
 	input_spec = Bru2InputSpec
 	output_spec = Bru2OutputSpec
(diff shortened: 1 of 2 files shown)


TheChymera:
@chrisfilo @satra can you help me figure out why my pr fails the tests?

circleci fails because

```
TraitError: The 'input_dir' trait of a Bru2InputSpec instance must be an existing directory name, but a value of '/data/20151103_213035_4001_1_1/10/' <type 'str'> was specified.
```
Am I supposed to give it a path relative to the nipype root directory so that the example can actually run? If so, there is no bruker sample data shipped with nipype. Should I add data as well? Nipype does not seem to ship with data, though.

And from the travis log I was only able to dig up this:

```
The command "python -W once:FSL:UserWarning:nipype `which nosetests` --with-doctest --with-cov --cover-package nipype --cov-config .coveragerc --logging-level=INFO" exited with 1.
```

What does it mean?

nipype/interfaces/bru2nii.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Bru2(CommandLine):
+	"""Uses bru2nii's Bru2 to convert Bruker files
+
+	Examples
+	========
+
+	>>> from nipype.interfaces.bru2nii import Bru2
+	>>> converter = Bru2()
+	>>> converter.inputs.input_dir = '/data/20151103_213035_4001_1_1/10/'
```
satra: for example purposes this can be set to `.` or `os.getcwd()`

nipype/interfaces/bru2nii.py
```diff
@@ -0,0 +1,56 @@
...
+	>>> converter.inputs.nii_file = '/data/20151103_213035_4001_1_1/10'
```
satra: there is no inputs called `nii_file`

satra:
for interface development it may be useful to read this: http://nbviewer.jupyter.org/urls/dl.dropbox.com/s/mxud29oj2mxid6o/nipype_dev_tutorial.ipynb

TheChymera:
@satra I had a read, corrected the mistakes you pointed out, and a few others I noticed while reexamining the interface.

Still, it seems the build fails. at least for travis, I am also unsure how to locate/interpret the error.

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp4hvza70o/.coverage.testing-gce-9c8ce19f-80aa-42d3-973a-aa2beb30b90e.2744.449699'
The command "python -W once:FSL:UserWarning:nipype `which nosetests` --with-doctest --with-cov --cover-package nipype --cov-config .coveragerc --logging-level=INFO" exited with 1.
```

This is the best I could dig up, but does that have anything to do with my added code? I have no idea where that temp file might be coming from...

TheChymera:
@chrisfilo @satra can you help me figure out what is going wrong with these tests?

nipype/interfaces/bru2nii.py
```diff
@@ -0,0 +1,54 @@
...
+
+class Bru2(CommandLine):
+	"""Uses bru2nii's Bru2 to convert Bruker files
+
+	Examples
+	========
+
+	>>> from nipype.interfaces.bru2nii import Bru2
+	>>> converter = Bru2()
+	>>> converter.inputs.input_dir = os.getcwd()
```
chrisfilo: Use a relative path to an existing folder in "testing/data". For example "dicomdir". If this name does not make sense in this context you can create a new folder (and add an empty dummy file in it so git would register it).

Issue #1350, opened 2016-02-05
AttributeError: 'module' object has no attribute 'write_dot'


TheChymera:
I am trying to run [this pipeline](https://github.com/TheChymera/chyMRI/blob/f4b9469189623c9f23c6338099ec2ed856333673/glm.py) and I am getting this error message:

```
/usr/lib64/python2.7/site-packages/nipy/labs/glm/glm.py:9: FutureWarning: Module nipy.labs.utils.routines deprecated, will be removed
  from ..utils import mahalanobis
/usr/lib64/python2.7/site-packages/nipype/interfaces/nipy/model.py:18: FutureWarning: Module nipy.labs.glm deprecated, will be removed. Please use nipy.modalities.fmri.glm instead.
  import nipy.labs.glm.glm as GLM
/usr/lib64/python2.7/site-packages/nipype/interfaces/base.py:419: UserWarning: Input tr requires inputs: slice_times
  warn(msg)
/usr/lib64/python2.7/site-packages/nipype/interfaces/base.py:419: UserWarning: Input slice_info requires inputs: slice_times
  warn(msg)
160205-22:03:39,570 workflow INFO:
         Creating detailed dot file: /home/chymera/NIdata/ofM.dr/FSL_GLM_work/graph_detailed.dot
Traceback (most recent call last):
  File "glm.py", line 166, in <module>
    fsl_glm(workflow_base="~/NIdata/ofM.dr/", functional_scan_type="7_EPI_CBV", structural_scan_type="T2_TurboRARE>", experiment_type="<ofM>", omit_ID=["20151027_121613_4013_1_1"])
  File "glm.py", line 162, in fsl_glm
    pipeline.write_graph(graph2use="flat")
  File "/usr/lib64/python2.7/site-packages/nipype/pipeline/engine.py", line 558, in write_graph
    format=format, simple_form=simple_form)
  File "/usr/lib64/python2.7/site-packages/nipype/pipeline/utils.py", line 905, in export_graph
    nx.write_dot(pklgraph, outfname)
AttributeError: 'module' object has no attribute 'write_dot'
```

It seems to be due to [this line](https://github.com/TheChymera/chyMRI/blob/f4b9469189623c9f23c6338099ec2ed856333673/glm.py#L162). And if I cmment it, the pipeline works. This happens both with release `0.11.0` and with the live nipype version from github. I have matplotlib. Any ideas what might be the problem?

mwaskom:
[...]

PR #1358, opened 2016-02-08
Fixed write_dot error for networkx>=1.11 in nipype/pipeline/utils.py

polosecki:
This fixes an incompatibility with networkx>=1.11 reported here:
https://github.com/nipy/nipype/issues/1350#issuecomment-181495121

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -1021,7 +1021,10 @@ def export_graph(graph_in, base_dir=None, show=False, use_execgraph=False,
                                suffix='.dot',
                                use_ext=False,
                                newpath=base_dir)
-    nx.write_dot(pklgraph, outfname)
+    if float(nx.__version__)<1.11:                           
+        nx.write_dot(pklgraph, outfname)
+    else:
+        nx.drawing.nx_pydot.write_dot(pklgraph, outfname)
     logger.info('Creating dot file: %s' % outfname)
     cmd = 'dot -T%s -O %s' % (format, outfname)
     res = CommandLine(cmd, terminal_output='allatonce').run()
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -27,7 +27,6 @@ import re
 import numpy as np
 from nipype.utils.misc import package_check
 from functools import reduce
-from distutils.version import LooseVersion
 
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/utils.py
```diff
@@ -1021,7 +1021,10 @@ def export_graph(graph_in, base_dir=None, show=False, use_execgraph=False,
                                suffix='.dot',
                                use_ext=False,
                                newpath=base_dir)
-    nx.write_dot(pklgraph, outfname)
+    if float(nx.__version__)<1.11:
```
mwaskom: This won't work because, e.g., `1.9` is numerically greater than `1.11`.

polosecki: True, how dumb

satra: you can use: `LooseVersion(nx.__version__) < LooseVersion('1.11')`

mwaskom: Is it necessary to do the version check? NX removed the top-level import, but did they move the function?

polosecki: I changed it to do the proper version check suggested by @satra

nipype/pipeline/engine/utils.py
```diff
@@ -1021,7 +1022,10 @@ def export_graph(graph_in, base_dir=None, show=False, use_execgraph=False,
                                suffix='.dot',
                                use_ext=False,
                                newpath=base_dir)
-    nx.write_dot(pklgraph, outfname)
+    if LooseVersion(nx.__version__) < LooseVersion('1.11'):                           
+        nx.write_dot(pklgraph, outfname)
+    else:
+        nx.drawing.nx_pydot.write_dot(pklgraph, outfname)
```
satra: @mwaskom - it's possible that they changed the api here completely.

@polosecki - could you please check if a former version of nx has this function, in which case you don't need the `if ... else ...``` clause.

mwaskom: It looks like the change just removed the direct import inside nx: https://github.com/networkx/networkx/pull/1929

polosecki:
Now the change is done without checking nx version, since the difference with older nx version is only that they removed the direct import but the write_dot function was always defined in the same place.

chrisfilo:
it seems backwards compatible with 1.10

PR #1360, opened 2016-02-10
[ENH] Addition of 3dHist from AFNI

oesteban:


chrisfilo:
wouldn't afni.Hist be more intuitive and consistent naming?

oesteban:
Ok

PR #1363, opened 2016-02-13
more robust access to neurodebian

chrisfilo:


oesteban:
There's a problematic point in that script: one apt-get install without the -y option (or similar, I don't recall exactly). It rarely fails, but sometimes it does. I wrote a fix: https://gist.github.com/oesteban/a0d8972f978339910bf4

satra:
@yarikoptic - hey yarik one change requested to neurodebian install script :)

satra:
@oesteban - you can submit a pull request to neurodebian

chrisfilo:
Yup - it's here:
https://github.com/neurodebian/neurodebian/blob/master/tools/neurodebian-travis.sh

PR #1365, opened 2016-02-15
BUG: Removes exists=True for optional output file in DenoiseImage

jessicaforbes:


oesteban:
I think this change may require the update of the autotests. Could you please run ```make specs``` and commit the auto test file corresponding to DenoiseImage?

chrisfilo:
Actually the "exists=True" parameter in OutputSpec does not denote mandatory (non-optional) outputs (sorry for the confusing syntax). I just ensure that if (and only if) the output will be set it will point to an existing file. I don't see any situation in which "exists=True"  should not be set in output files.

hjmjohnson:
There must be some other bug in this interface then.   If I set "save_noise=False" then the interface fails with an error stating that the output noise image does not exists.

satra:
there is a bug and it has to do with automated list_outputs!

hjmjohnson:
@jessicaforbes Jessica,  I don't understand Satra's comment regarding "automated list_outputs", but I hope you understand this statement and that you know how to fix it.

satra:
the problem is that there is nothing that indicates that the output `noise_image` is tied to `save_noise`. this is also an example where the `noise_image` input basically implies an output state, and should use the dual form or either `Bool or File(hasfiles=False)`

satra:
@jessicaforbes - would it be possible to make this interface be like:

https://github.com/nipy/nipype/blob/befd4f8bf7d81b2f33c65c4b6061cc49f8523aa9/nipype/interfaces/fsl/preprocess.py#L740

and then in format arg, you can simply use `noise_image`. i'll take a look at the automated `_list_outputs`

satra:
in this place: https://github.com/nipy/nipype/blob/befd4f8bf7d81b2f33c65c4b6061cc49f8523aa9/nipype/interfaces/base.py#L1636

we need to check if that output exists, and only then assign it. @oesteban - i know you are refactoring all of this, but does this logic make sense?

oesteban:
Yes, I was to mention that, @satra. In the refactoring I'm trying to move all the stuff related to inputs and outputs into the specification classes. Then, the name_source behavior will be more natural (and I will extend it with some other improvements, like handling multiple paths). Finally, _list_outputs will be deprecated.

Regarding the early discussion raised by @chrisfilo, I still haven't make any decision, but probably the outputs will also have the exists and mandatory outputs (I don't know if both or how). This way the logic underlying the list_outputs will be easier to implement (i.e. an output with mandatory=True will need to be defined, and I don't know what will happen with exists, if it really makes any sense).

In short, since this fix seems to be solving the problem and I will come over this interface very soon, I guess we can just merge it.

jessicaforbes:
Thank you all for your comments.

@oesteban I did run `make specs` before the pull request and there wasn't a change to test_auto_DenoiseImage.py. 

Since you are refactoring and planning to look over this interface soon, is there anything I should do now?

oesteban:
Thanks a lot. Yes, let's confirm that @satra and @chrisfilo agree on merging it. As @chrisfilo said, right now this attribute shouldn't have an impact... so I will investigate this in particular on that refactoring I'm preparing.

satra:
i'm happy for merging this knowing the refactoring that @oesteban is doing.

chrisfilo:
Shouldn't a fix for _list_outputs be introduced instead of merging this
workaround?

satra:
depends on how close @oesteban is - we have never had a policy of updating previous releases with patches.

oesteban:
Hi, well, the refactoring is pretty comprehensive. I will add some implementation to the deprecated methods like ```_list_outputs```, but I will do this at the very end when all tests are already passing (and therefore, it is theoretically not needed anymore). Since the workaround is solving a problem for @jessicaforbes, I'd let them go through it in this *lazy way*, and fix the issue on the refactor.

Regarding timing: I want to get the refactoring close to being merged by the end of this week. So pretty soon. I will probably ping you before making any decision alone.

PR #1372, opened 2016-02-23
ResourceMultiProc plugin and runtime profiler

dclark87:
1) @carolFrohlich changed the MultiProc plugin to be the ResourceMultiProc plugin. This plugin enables multithreading during workflow execution but also ensures that the workflow will not use more of the system resources than the user specifies with the plugin arguments 'n_procs' and 'memory' at the workflow level. Individual node-level resource usage can be specified via the node._interface.num_threads and node._interface.estimated_memory parameters (defaults are 1 thread and 1 gb, respectively).
2) @carolFrohlich added callback logger that will record estimated node resources and runtimes into a log file via json format.
3) @carolFrohlich added a draw_gantt_chart module which will create a gantt chart graphic html file of the resources utilized by each node during workflow execution from start to finish
4) @dclark87 added a runtime profiler using the psutil and memory_profiler python packages. This profiler will be enabled as long as those packages are installed, and disabled if not installed. The profiler records the actual number of threads and GB of RAM used by a node (as long as it has a nipype interface - will not work with util.function nodes). These values are stored as 'runtime_threads' and 'runtime_memory' in both the runtime Bunch object and result dictionary returned from running the node interface. Because of this, they are also written to the provenance file if that is enabled. The new callback logger will look for these values directly and log them as well.
5) All new features include additional unit tests added to the test folders within each sub-packagae

Thanks to @ccraddock for design and debug help


nipype/interfaces/base.py
```diff
@@ -764,6 +764,8 @@ def __init__(self, **inputs):
             raise Exception('No input_spec in class: %s' %
                             self.__class__.__name__)
         self.inputs = self.input_spec(**inputs)
+        self.estimated_memory = 1
```
chrisfilo: Can we rename this to `estimated_memory_gb` or `estimated_memory_mb` to be more explicit about units?

nipype/interfaces/base.py
```diff
@@ -1202,14 +1204,101 @@ def _read(self, drain):
...
+        the process to evaluate thead usage of
+
+    Returns
+    -------
+    num_threads : int
+        the number of threads that the process is using
+    """
+
+    # Import packages
+    import psutil
```
chrisfilo: Psutil needs to be added as a dependency to setup and documentation.

nipype/interfaces/base.py
```diff
@@ -1202,14 +1204,101 @@ def _read(self, drain):
...
+    # Import packages
+    from memory_profiler import _get_memory
+    import psutil
+
+    try:
+        mem_mb = max(mem_mb, _get_memory(proc.pid, include_children=True))
+        num_threads = max(num_threads, _get_num_threads(psutil.Process(proc.pid)))
+        if poll:
+            proc.poll()
+    except Exception as exc:
```
chrisfilo: Could we narrow down the range of expected exceptions?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -6,21 +6,63 @@
...
+    # Import packages
+    import datetime
+ 
+    # Init variables
     result = dict(result=None, traceback=None)
+
+    # 
     try:
-        result['result'] = node.run(updatehash=updatehash)
+        start = datetime.datetime.now()
```
chrisfilo: This is already calculated here: https://github.com/nipy/nipype/blob/47587b97c99025bc72a42125fc640521b44a3431/nipype/interfaces/base.py#L1035 and written to reports here: https://github.com/nipy/nipype/blob/1262955492fa63d3d84795683fd6cc5063f0aacf/nipype/pipeline/engine/nodes.py#L741. Why not we extend this mechanism and keep everything (duration, memory, threads) in one place (reports) instead of writing new JSON file in `callback_log`?

chrisfilo:
Thanks for sending this!

Couple more comments:

- As far as I understand ResourceMultiProc does everything that MultiProc did plus more. Instead of renaming MultiProc to ResourceMultiProc as you propose in this PR why not just keep the old name (for backward compatibility sake) with new functionality?
- Could you add some documentation on specifying resources for nodes and using the Gantt chart generation scripts. It would be good to add info about extra dependencies and limitations (from what I see in the code the resource estimation will only work on command line interfaces, but not on pure python).

All in all a great addition.

nipype/pipeline/plugins/multiproc.py
chrisfilo: I can see you already update the `runtime` object so my proposed change would only require you to add a couple of lines to print memory and threads in reports (here: https://github.com/nipy/nipype/blob/1262955492fa63d3d84795683fd6cc5063f0aacf/nipype/pipeline/engine/nodes.py#L741) and modify the Gantt generation script to use reports instead of JSON files.

nipype/interfaces/base.py
dclark87: Sure. Is there a preference between gb or mb for units?

chrisfilo: I don't have one. We can keep GB (that's what you use now yes?). It would
be just nice to make it explicit in the name.

chrisfilo:
In addition make sure that both CircleCI and Travis tests are passing :)

dclark87:
Yeah, I am looking through the failures now. What is the best way to enforce installing new Python dependencies on the CircleCI and Travis tests?

chrisfilo:
Firstly make sure that the tests pass even when optional dependencies are
not installed (it's important so things would not break fit people who are
just doing pip install). When that's passing you can add the optional
packages via pip in circle and travis yml files. Travis supports testing
against many different environment - it would be good to run tests with and
without the optional packages (current setup features something like that -
you can use it as an inspiration).

I hope this helps!

dclark87:
I'm having trouble diagnosing why the CI tests are failing. Any thoughts?

chrisfilo:
CircleCI had a temporary problem downloading test data. Travis seems to have been running a single test (that does not output anything) for more than 10 minutes. Long running tests should periodically print something on stdout.

But what surprises me is that any of your new tests run at all since psutils and memory_profiler are not installed and they should be skipped. Try creating a virtualenv without those packages and see how the tests behave. You can also try temporarily adding print statements inside your tests to figure out if they are the culprit.

chrisfilo:
PS I restarted the tests.

dclark87:
I just tested the PR in a virtualenv with only the nipype dependencies installed (not memory_profiler or psutil). I did the test via: "import nipype; nipype.test()" with 24 skips and no errors. I changed a logger message to avoid the word "Error" as it is a little misleading and re-pushed.

chrisfilo:
I have tried to increase test verbosity of travis here: https://github.com/nipy/nipype/pull/1385 it might help you track down the problem.

dclark87:
We're going to add some documentation to this per request as well as look everything over again to see if we can find the source of the error that's happening in TravisCI. I'll push again once that's done

chrisfilo:
Thanks!

satra:
@dclark87 - can you merge with current master and push.

dclark87:
I apologize this has taken so long to finalize. We are in the last stages of implementing the gantt-chart generator and fixing our number of threads estimation/runtime calculations. We expect this to be completed in the next week. Once that is done, we will pull in those changes and re-base with the current nipype master

dclark87:
Hey @satra @chrisfilo - I apologize for the delay in progress in getting this PR through. We have finished all of the features we wanted to add - however I see the CI tests are failing due to no response being detected for 10mins. Do you have any thoughts as to why this could be happening? Thanks.

satra:
@dclark87 - can you merge current master and push? it will give us a better sense of the failures.

dclark87:
As a documentation note, I added instructions on using the resource scheduler and runtime profiler: https://github.com/FCP-INDI/nipype/blob/resource_multiproc/doc/users/resource_sched_profiler.rst
I wasn't sure if there are steps to integrate this fully into the nipype documentation

nipype/pipeline/engine/nodes.py
```diff
@@ -737,15 +737,20 @@ def write_report(self, report_type=None, cwd=None):
...
                 return
             fp.writelines(write_rst_header('Runtime info', level=1))
+            # Init rst dictionary of runtime stats
+            rst_dict = {'hostname' : self.result.runtime.hostname,
+                        'duration' : self.result.runtime.duration}
+            # Try and insert memory/threads usage if available
+            try:
+                rst_dict['runtime_memory_gb'] = self.result.runtime.runtime_memory_gb
+                rst_dict['runtime_threads'] = self.result.runtime.runtime_threads
+            except:
```
chrisfilo: Please be specific about the exception you are expecting.

chrisfilo:
Here's a rendered version of the docs: https://circle-artifacts.com/gh/nipy/nipype/1139/artifacts/0/home/ubuntu/nipype/doc/_build/html/users/resource_sched_profiler.html

BTW if we will not be able to figure out why Travis hangs maybe we could skip this test on Travis (but still run it on CircleCI)?

nipype/interfaces/afni/base.py
```diff
@@ -157,6 +157,11 @@ def __init__(self, **inputs):
         else:
             self._output_update()
 
+        # Update num threads estimate from OMP_NUM_THREADS env var
+        # Default to 1 if not set
+        import os
+        self.num_threads = int(os.getenv('OMP_NUM_THREADS', 1))
```
chrisfilo: Shouldn't we do this other way around? If user sets num_threads to say 4 we would set the OMP_NUM_THREADS variable to 4? I think this is how it's done for ants: https://github.com/nipy/nipype/blob/master/nipype/interfaces/ants/base.py#L52

nipype/pipeline/plugins/callback_log.py
```diff
@@ -0,0 +1,56 @@
...
+    logger = logging.getLogger('callback')
+
+    # Check runtime profile stats
+    if node.result is not None:
+        try:
+            runtime = node.result.runtime
+            runtime_memory_gb = runtime.runtime_memory_gb
+            runtime_threads = runtime.runtime_threads
+        except:
+            runtime_memory_gb = runtime_threads = 'Unkown'
```
chrisfilo: Unknown

nipype/pipeline/plugins/callback_log.py
```diff
@@ -0,0 +1,56 @@
...
+    # Init variables
+    logger = logging.getLogger('callback')
+
+    # Check runtime profile stats
+    if node.result is not None:
+        try:
+            runtime = node.result.runtime
+            runtime_memory_gb = runtime.runtime_memory_gb
+            runtime_threads = runtime.runtime_threads
+        except:
```
chrisfilo: Please specify exception class.

nipype/pipeline/plugins/callback_log.py
```diff
@@ -0,0 +1,56 @@
...
+        ',"runtime_threads":' + '"'+ str(runtime_threads) + '"'+ \
+        ',"runtime_memory_gb":' + '"'+ str(runtime_memory_gb) + '"' + '}'
+
+        logger.debug(message)
+    # Other
+    else:
+        message  = '{"name":' + '"' + node.name + '"' + ',"id":' + '"' + \
+        node._id + '"' + ',"finish":' + '"' + str(datetime.datetime.now()) +\
+        '"' + ',"estimated_memory_gb":' + str(node._interface.estimated_memory_gb) + \
+        ',"num_threads":' + str(node._interface.num_threads) + ',"error":"True"}'
```
chrisfilo: The code would be much cleaner and less error prone if you used json library to create those error messages.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -42,33 +74,90 @@ class NonDaemonPool(pool.Pool):
...
+    thread_consuming_node.interface.num_threads = 16
+
+    The default number of threads and memory for a node is 1. 
+
+    Currently supported options are:
 
-    - n_procs : number of processes to use
     - non_daemon : boolean flag to execute as non-daemon processes
+    - num_threads: maximum number of threads to be executed in parallel
+    - estimated_memory_gb: maximum memory (in GB) that can be used at once.
```
chrisfilo: Shouldn't this be 'memory_gb'?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -100,3 +177,126 @@ def _report_crash(self, node, result=None):
 
     def _clear_task(self, taskid):
         del self._taskresult[taskid]
+
+    def _submit_job(self, node, updatehash=False):
+        self._taskid += 1
+        try:
+            if node.inputs.terminal_output == 'stream':
+                node.inputs.terminal_output = 'allatonce'
+        except:
```
chrisfilo: What kind of exception do you expect here? IMHO that if should always run...

nipype/pipeline/plugins/multiproc.py
```diff
@@ -100,3 +177,126 @@ def _report_crash(self, node, result=None):
...
+
+            if self.procs[jobid]._interface.estimated_memory_gb <= free_memory_gb and \
+               self.procs[jobid]._interface.num_threads <= free_processors:
+                logger.info('Executing: %s ID: %d' %(self.procs[jobid]._id, jobid))
+                executing_now.append(self.procs[jobid])
+
+                if isinstance(self.procs[jobid], MapNode):
+                    try:
+                        num_subnodes = self.procs[jobid].num_subnodes()
+                    except Exception:
```
chrisfilo: Any exception?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -100,3 +177,126 @@ def _report_crash(self, node, result=None):
...
+                        hash_exists, _, _, _ = self.procs[
+                            jobid].hash_exists()
+                        logger.debug('Hash exists %s' % str(hash_exists))
+                        if (hash_exists and (self.procs[jobid].overwrite == False or \
+                                             (self.procs[jobid].overwrite == None and \
+                                              not self.procs[jobid]._interface.always_run))):
+                            self._task_finished_cb(jobid)
+                            self._remove_node_dirs()
+                            continue
+                    except Exception:
```
chrisfilo: Any exception?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -100,3 +177,126 @@ def _report_crash(self, node, result=None):
...
+                        self.proc_pending[jobid] = False
+                        continue
+                logger.debug('Finished checking hash')
+
+                if self.procs[jobid].run_without_submitting:
+                    logger.debug('Running node %s on master thread' \
+                                 % self.procs[jobid])
+                    try:
+                        self.procs[jobid].run()
+                    except Exception:
```
chrisfilo: Any exception?

nipype/utils/draw_gantt_chart.py
```diff
@@ -0,0 +1,585 @@
...
+
+    # Import packages
+    import copy
+
+    events = {}
+    for node in nodes_list:
+        # Format node fields
+        try:
+            estimated_threads = float(node['num_threads'])
+        except:
```
chrisfilo: KeyError class?

nipype/utils/draw_gantt_chart.py
```diff
@@ -0,0 +1,585 @@
...
+            estimated_threads = 1
+        try:
+            estimated_memory_gb = float(node['estimated_memory_gb'])
+        except:
```
chrisfilo: KeyError class?

nipype/utils/draw_gantt_chart.py
```diff
@@ -0,0 +1,585 @@
...
+            estimated_memory_gb = 1.0
+        try:
+            runtime_threads = float(node['runtime_threads'])
+        except:
```
chrisfilo: KeyError class?

nipype/utils/draw_gantt_chart.py
```diff
@@ -0,0 +1,585 @@
...
+            runtime_threads = 0
+        try:
+            runtime_memory_gb = float(node['runtime_memory_gb'])
+        except:
```
chrisfilo: KeyError class?

chrisfilo: Actually each of the above can be replaced, byt a .get() with a default value http://stackoverflow.com/a/9359011

nipype/utils/draw_gantt_chart.py
```diff
@@ -0,0 +1,585 @@
...
+        content = content.read()
+        lines = content.split('\n')
+
+        for l in lines:
+            #try to parse each line and transform in a json dict.
+            #if the line has a bad format, just skip
+            node = None
+            try:
+                node = json.loads(l)
+            except Exception:
```
chrisfilo: Please be more specific about the exception class.

chrisfilo:
Ok I finished reading the code. Overall this is a great contribution. Apart from the things I mentioned above the resource_sched_profiler should be listed in https://raw.githubusercontent.com/nipy/nipype/master/doc/users/index.rst

Sorry for picking on the try-catch-all. Being specific about exception classes will pay off in the future. Silently dealing with all potential exceptions makes debugging incredibly hard (believe me, I learned this the hard way!).

satra:
@dclark87 - one question, is there any reason this information is being kept separate from the provenance record that nipype produces?

nipype/pipeline/plugins/multiproc.py
dclark87: I would imagine this exception would catch the same thing this one would: https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/multiproc.py#L20

So whatever specific exception that line is looking for we can use. But it is also vague

nipype/pipeline/plugins/multiproc.py
dclark87: I think sometimes it makes sense to be able to be prepared to catch any exception such that we can still clean and remove that node from the proc_pending queue. However, I am not sure what specific exception would come up here, could be an IOError from a missing filepath or perhaps a KeyError or AttributeError when trying to access various parameters of the procs dictionary. However, if you like, we can remove the try block altogether

nipype/pipeline/plugins/multiproc.py
dclark87: I think sometimes it makes sense to be able to be prepared to catch any exception such that we can still clean and remove that node from the proc_pending queue. However, I am not sure what specific exception would come up here, could be an IOError from a missing filepath or perhaps a KeyError or AttributeError when trying to access various parameters of the procs dictionary. However, if you like, we can remove the try block altogether

nipype/pipeline/plugins/multiproc.py
dclark87: Taken from here: https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/multiproc.py#L85

chrisfilo: Thanks. I don't see why this should be in a try catch. I would vote to remove it. @satra ?

nipype/pipeline/plugins/multiproc.py
chrisfilo: So maybe reraise the exception and add move the cleanup to a "finally" clause?

nipype/pipeline/plugins/multiproc.py
chrisfilo: Same as above. The problem is not the try-catch-all per se, but the fact that the information about the error is not logged anywhere making debugging very hard.

nipype/pipeline/plugins/multiproc.py
chrisfilo: Agreed, but note that the example you gave saves the traceback making debugging possible. Maybe a similar solution can be implemented here as well?

dclark87:
@satra- there is no reason this information is not in the provenance record. We currently store the runtime profiling info in the node's result.runtime as well as the callback logger, but it doesn't hurt to have it in more places. Where is the provenance code?

nipype/pipeline/plugins/multiproc.py
dclark87: So I found we took the code from: https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L354
So whatever exception is being looked for here - we are doing the same.

nipype/pipeline/plugins/multiproc.py
dclark87: Likewise here - https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L384

nipype/pipeline/plugins/multiproc.py
dclark87: Likewise - inherited from here: https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L396

chrisfilo: If you add a line to save the traceback like in the implementation you pointed to: https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/multiproc.py#L22 debugging capabilities will be greatly improved.

dclark87: @chrisfilo - what about something like this: https://github.com/FCP-INDI/nipype/blob/resource_multiproc/nipype/pipeline/plugins/multiproc.py#L240-246

chrisfilo: Works for me! Again - I'm not trying to picky here. Just want all of us to
avoid the situation when something breaks and we will keep scratching our
heads because the relevant exception was swallowed by one of those
"catch-all" snippets.

doc/users/resource_sched_profiler.rst
```diff
@@ -0,0 +1,160 @@
...
+	args_dict = {'n_procs' : 8, 'memory_gb' : 10, 'status_callback' : log_nodes_cb}
+	workflow.run(plugin='MultiProc', plugin_args=args_dict)
+	
+	# ...workflow finishes and writes callback log to '/home/user/run_stats.log'
+	
+	from nipype.utils.draw_gantt_chart import generate_gantt_chart
+	generate_gantt_chart('/home/user/run_stats.log', cores=8)
+	# ...creates gantt chart in '/home/user/run_stats.log.html'
+
+The `generate_gantt_chart`` function will create an html file that can be viewed
```
chrisfilo: Missing ` ?

chrisfilo:
LGTM!

PR #1374, opened 2016-02-24
WIP/ENH: Allow client args as IPython plugin_args

effigies:
I want to manage cluster processing outside of the default IPython profile, and this change allows nipype to accommodate that. While I'm only using the `profile` argument, this change accepts all of the valid parameters for `IPython.parallel.Client`. I think `url_file`, `profile` and `cluster_id` are probably a minimal set.

No tests yet, but throwing this up before I get too far, in case people have opinions on the approach, whether this is a good idea, etc.


satra:
@effigies - looks reasonable. a plugin can expose any args relevant to itself.

PR #1376, opened 2016-02-25
fix: Complete the documentation for ConMat

alexsavio:
The `tract_stat` input option is mandatory when you use `scalar_file` using Camino's ConMat.

I fixed the example so people don't forget to set it.


chrisfilo:
If it's mandatory we should add "mandatory=True" in the InputSpec.

PR #1379, opened 2016-02-29
ENH: Specify motion parameter source in modelgen

effigies:
This change adds an optional `parameter_source` input to the `SpecifyModel` classes, with the same options as available in [`ArtifactDetect`](https://github.com/nipy/nipype/blob/master/nipype/algorithms/rapidart.py#L163-L165), though only `FSFAST` causes any change from default behavior.

Per #706 and https://groups.google.com/d/msg/nipy-user/QzmvYr7x9B4/F30k7sET5lQJ, FSFAST motion parameters are in columns 1-6 (zero-indexed).

The default behavior is changed to choosing the first 6 columns of a parameter file, since the assumption appears to be rotation and translation parameters. Though it might be preferable to restore the old default, which is to take as many columns as are found.


chrisfilo:
@poline you might be interested in this

chrisfilo:
@satra @effigies I wonder if we should make the new input mandatory or set some defaults

effigies:
I was a little hesitant to make it mandatory because that would break existing scripts. As for defaults, it would probably make sense to have `SpecifySPMModel` use `parameter_source='SPM'`, but what would be a sensible default for the generic models?

oesteban:
Tests are not passing because of #1571. Sorry about that.

effigies:
Just going to re-up this, since it's behavior our scripts depend on. It'd be nice to keep it working without making a conversion script for our data or maintaining a lab-specific nipype patch set.

Is this something y'all are interested in including, and if so, what can I do to get it merge-ready? I don't see obvious tests to write, but let me know if you do.

nipype/algorithms/modelgen.py
```diff
@@ -187,6 +187,9 @@ class SpecifyModelInputSpec(BaseInterfaceInputSpec):
     realignment_parameters = InputMultiPath(File(exists=True),
                                             desc="Realignment parameters returned by motion correction algorithm",
                                             copyfile=False)
+    parameter_source = traits.Enum("SPM", "FSL", "AFNI", "NiPy", "FSFAST",
+                                   desc="Source of motion parameters",
+                                   mandatory=False)
```
satra: `mandatory=False` should be removed.

satra:
@effigies - we can include this for now (especially as it doesn't make any change to existing workflows), but this will likely be replaced with the new `SpecifyEvents` formulation that is being worked on in the context of BIDS and a forthcoming new project on naturalistic stimuli.

let's make sure this is merged with current master and all the tests pass.

effigies:
@satra Rebased on master, tests passing.

effigies:
Updated this with @chrisfilo's `normalize_mc_params` function. Also set `parameter_source` default to `SPM`, which `normalize_mc_params` passes through untouched.

Seem reasonable?

effigies:
Note that I removed Nipy because I couldn't find clear documentation on their parameters, and depending on `normalize_mc_params` implies that the parameters are presented in SPM-compatible format.

effigies:
@satra This is the PR I mentioned yesterday.

satra:
@effigies - for this one in particular, i would like to see the Nipy SpaceTimeRealign output handled, since we use it for our analysis.

to do that in an SPM-compatible way, the following would need to happen:

`nipype_params -> matrix representation -> extracting euler angles and translation parameters`

chrisfilo:
I could reuse that code in framewise displacement interface.

effigies:
@satra, @chrisfilo see 57740e1. Does this look reasonable?

I found a number of different suggestions on how to extract Euler angles, and whether there are multiple equivalent formulations or different starting assumptions for each, I was unable to make an intelligent choice among them. Given that [`aff2euler`](https://github.com/nipy/nipy/blob/master/nipy/algorithms/registration/scripting.py#L30-L43) is in the `nipy` catalogue, I sincerely hope that it assumes the format resulting from `to_matrix44`.

The [`mat2euler`](https://github.com/nipy/nipy/blob/master/nipy/algorithms/registration/scripting.py#L30-L43) function from nibabel returns rotations about z, y, and x, which is, from what I can tell, reverse from the SPM ordering.

I think probably the best way to test this is to remove [these lines](https://github.com/nipy/nipype/blob/master/nipype/algorithms/rapidart.py#L49-L52) and make sure RapidArt gives the same results (within some tolerance) for nipy motion parameters.

chrisfilo:
You can also run realigment with 4D and say mcflirt and compare FD. They should be roughly the same.

satra:
@effigies - the other piece i'm not sure about now is why i wrote this if: 

https://github.com/nipy/nipype/blob/master/nipype/algorithms/rapidart.py#L78

effigies:
@satra

From AFNI's `3dvolreg` manual:
```
       N.B.: The motion parameters are those needed to bring the sub-brick
             back into alignment with the base.  In 3drotate, it is as if
             the following options were applied to each input sub-brick:
              -rotate 'roll'I 'pitch'R 'yaw'A  -ashift 'dS'S 'dL'L 'dP'P
```

Assuming this applies them right-to-left on the CLI, then you have Rz -> Rx -> Ry -> T, which is what you've got. (But shouldn't the Rz be negative because they label the axis S-I, not I-S?)

Incidentally, [at least in FSL](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FLIRT/FAQ#What_is_the_format_of_the_matrix_used_by_FLIRT.2C_and_how_does_it_relate_to_the_transformation_parameters.3F), scale precedes skew. If the scaling matrix `S` is anisotropic, `S.dot(Sh)` there may be a bug here for FSL parameters.

I'm having trouble finding a detailed description of parameters and order-of-operations in the SPM manual.

nipype/utils/misc.py
```diff
@@ -248,10 +248,25 @@ def unflatten(in_list, prev_structure):
...
+    elif source.upper() in ('AFNI', 'FSFAST'):
         params = params[np.asarray([4, 5, 3, 1, 2, 0]) + (len(params) > 6)]
         params[3:] = params[3:] * np.pi / 180.
-    return params
\ No newline at end of file
+    elif source.upper() == 'NIPY':
+        from nipy.algorithms.registration import to_matrix44, aff2euler
+        matrix = to_matrix44(params)
+        params = np.zeros(6)
+        params[:3] = M[:3, 3]
```
chrisfilo: Does this work? "M" is undefined.

effigies: Woops. Fixed.

effigies:
Are you guys waiting on me for anything here, by the way? I'm just not sure what the status of this one is.

chrisfilo:
LGTM

PR #1390, opened 2016-03-03
ENH: Make interfaces.utility.Merge(1) merge a list of lists

effigies:
At present, `Merge()` does nothing, providing an undefined `outputs.out`. This change allows it to merge a list of lists (`inputs.in_lists`) into a single list, rather than enumerating them with `inputs.in1 .. inputs.inN`.

If no `inputs.in_lists` is set, `Merge()` still provides an undefined `outputs.out`, so we shouldn't break existing code.

`Merge(n)` for `n < 0` is unchanged.

### DIFF
--- a/nipype/interfaces/utility/base.py
+++ b/nipype/interfaces/utility/base.py
@@ -156,10 +156,8 @@ class Merge(IOBase):
                 else:
                     out.append(value)
         else:
-            for i in range(len(filename_to_list(values[0]))):
-                out.insert(i, [])
-                for value in values:
-                    out[i].append(filename_to_list(value)[i])
+            lists = [filename_to_list(val) for val in values]
+            out = [[val[i] for val in lists] for i in range(len(lists[0]))]
         if out:
             outputs['out'] = out
         return outputs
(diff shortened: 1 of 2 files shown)


effigies:
Slight amendment: `Merge(0, axis='hstack')` previously failed to run. It now behaves the same as `Merge(0, axis='vstack')`, if no `in_lists` is provided.

nipype/interfaces/utility/base.py
```diff
@@ -128,24 +128,36 @@ class Merge(IOBase):
     def __init__(self, numinputs=0, **inputs):
         super(Merge, self).__init__(**inputs)
         self._numinputs = numinputs
-        add_traits(self.inputs, ['in%d' % (i + 1) for i in range(numinputs)])
+        if numinputs > 0:
+            input_names = ['in%d' % (i + 1) for i in range(numinputs)]
+        elif numinputs == 0:
+            input_names = ['in_lists']
+        else:
```
satra: so the only question here is whether this should be for `Merge(1)` instead of `Merge(0)`

effigies: I can do `Merge(1)`. The goal was to leave the interface unmodified for any existing uses, but I guess nobody would actually do `Merge(1)` as it stands.

If we do, I'd suggest making `numinputs` default to 1, so that `Merge()` can be useful.

And I can update the doctests to be demonstrative.

satra:
sounds good.

Issue #1387, opened 2016-03-02
Rename interface can only use symlinks


billbrod:
The `utility.Rename` interface calls `utility.copyfile` to rename the files passed to it, however it always uses a symlink and doesn't allow the user to specify whether they would rather copy the file instead. I'm working on a CIFS filesystem, where symlinks are not allowed (at least, users are not allowed to create them; calling `os.symlink` always returns `[Errno 95] Operation not supported`), and so this prevents me from being able to use the Rename interface.

effigies:
As a stop-gap, you can use f3d1c9b57d5dc76ace6a98c8fac8c0e15e0a8967. I'm pretty sure that should work, but let me know.

billbrod:
Yup, that fix works perfectly. Thanks!

PR #1391, opened 2016-03-03
REF/FIX: Correct link/copy behavior

effigies:
Added a patch that addresses #1387 by falling back on `copyfile(copy=True)` when symlinking fails. But it seems it might be good to fully work out the cases, now that hard linking is an option, as well.

The current logic is:

flag:copy | flag:use_hardlink | sys:posix | cap:hardlink | action
-------|-------------------|-----------------------|------------------------|---------
T | T | * | T | hard link
T | F | * | * | copy
T | * | * | F| copy
F | * | T | * | symlink
F | T | F | T | hard link
F | * | F | F | copy
F | F | F | * | copy

I see two problems:

1. `os.name == 'posix'` does not guarantee symlinks work, so we need some other `cap:symlink`
    indicator.
2. `copy == False and use_hardlink == True` on a POSIX system uses a symlink when a hard link
    could work and be more in line with the flags.

The following seems like a more intuitive alternative:

flag:copy | flag:use_hardlink | cap:symlink | cap:hardlink | action
-------|-------------------|-----------------------|------------------------|---------
 * | T | * | T | hard link
 T | T | * | F | copy
 T | F | * | * | copy
 F | * | T | F | symlink
 * | * | F | F | copy

With exceptions to detect capabilities, this lends itself to a simple logic:

```Python
if use_hardlink:
    try:
        hardlink
    except:
        pass
if not copy:
    try:
        symlink
    except:
        pass
copyfile
```

Do people agree this is worth changing?

As to testing, is there a good way to create and mount a `vfat` filesystem without root? That would permit testing `cap:symlink` from a POSIX system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ before_install:
   echo 'include_dirs = /usr/include:/usr/include/X11' >> $HOME/.numpy-site.cfg;
   fi
 install:
+- sudo apt-get install fusefat
 - conda update --yes conda
 - conda create -n testenv --yes pip python=$TRAVIS_PYTHON_VERSION
 - source activate testenv
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,7 @@ Next release
 * ENH: New interfaces for interacting with AWS S3: S3DataSink and S3DataGrabber (https://github.com/nipy/nipype/pull/1201)
 * ENH: Interfaces for MINC tools (https://github.com/nipy/nipype/pull/1304)
 * FIX: Use realpath to determine hard link source (https://github.com/nipy/nipype/pull/1388)
+* FIX: Correct linking/copying fallback behavior (https://github.com/nipy/nipype/pull/1391)
 
 Release 0.11.0 (September 15, 2015)
 ============
--- a/circle.yml
+++ b/circle.yml
@@ -12,7 +12,7 @@ dependencies:
     - bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh)
   override:
     # Install apt packages
-    - sudo apt-get install -y fsl-core fsl-atlases fsl-mni152-templates fsl-feeds afni swig python-vtk xvfb
+    - sudo apt-get install -y fsl-core fsl-atlases fsl-mni152-templates fsl-feeds afni swig python-vtk xvfb fusefat
     - echo 'source /etc/fsl/fsl.sh' >> $HOME/.profile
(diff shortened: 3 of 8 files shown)


effigies:
Due to some rebasing, these commits are showing up out of order.

1. 14e8ac4e28fb84006b649c4cfb2f89f4d04a9667: Adds `TempFATFS` context manager that creates and mounts (FUSE) a small filesystem that doesn't support symlinks or hardlinks.
2. 16f830f6b8c23c9b2b8d2074d5f3d9df1b270b61: A failing test for #1387.
3. 29aa2baaec3e195292c31b69efed6ff69b62f2f0: The stop-gap fix offered for #1387. Test added in 16f830f6b8c23c9b2b8d2074d5f3d9df1b270b61 passes.
4. ff1b8e5: A simpler rewriting of the `copyfile` logic that follows the proposal above. Test still passes.

This final change separates the logic of checking and removing old files from that of linking, symlinking and copying. Additionally, the `copyfile` calls for associated files no longer pass `create_new`, since the new filename has already been created. This fixes a bug introduced in #1388.

satra:
@effigies - could you please add a line to CHANGES as well?

effigies:
Done. Going to add `fusefat` to `circle.yml` before you pull.

effigies:
Okay, hopefully Circle will pass now.

satra:
@effigies - could you please merge master to this. we were having some circle + sourceforge issues.

satra:
@effigies - any chance you can work on the failed test on circle? i'd like to get this one merged before release.

effigies:
Is there an easy way to reduce the circle tests to just the one? I can add information to the exception for debugging, but the time lag for feedback on Circle is prohibitive.

satra:
you can enable circle on your repo, reduce the examples that are run to just yours, and even ssh in to debug.

effigies:
Seems Circle does some weird caching, so we end up with a VM without a kernel modules directory, so fuse can't be loaded.

effigies:
Is there a preferred way to deal with this? I can catch the `IOError` in `test_copyfallback` and only run the tests if the filesystem could be mounted.

If it would be preferable to fix the Circle environment so that fuse filesystems can be used, that's beyond my ability to commit time right now.

effigies:
Thinking about it, having a hard dependency on FUSE/fusefat for testing seems too strong. Updated to skip tests if `fusefat` fails.

effigies:
Ready for review.

Issue #1404, opened 2016-03-11
sorted list of exec nodes?


alexsavio:
Hi,

Why do you use sorted in this line?
https://github.com/nipy/nipype/blob/master/nipype%2Fpipeline%2Fengine%2Fworkflows.py#L693

If I use a connect with an auxiliary function, e.g.:
`(params, stc, [(("slice_order", sum_one), "slice_order"),]),`

I get this error:
`TypeError('unorderable types: tuple() < str()',)`

Can't sort tuples against strings.

I removed the `sorted` because it doesn't make sense here and it works.

satra:
@alexsavio - this looks like a bug. in fact would be good to add this as a test and fix it.

PR #1405, opened 2016-03-12
fix:test & fix bug using aux functions in connect

alexsavio:
This is a PR related with issue #1404


alexsavio:
Sorry, I added the petpvc document improvement here. In the petpvc commit I don't change code, I am moving the docstring of the .py file inside the interface class declaration.

PR #1411, opened 2016-03-23
FIX: Permit relative path for concatenated_file input to Concatenate()

effigies:
The example as given in the [docs](http://nipy.org/nipype/interfaces/generated/nipype.interfaces.freesurfer.model.html#id1) will fail to run, raising:

    TraitError: The 'concatenated_file' trait of a ConcatenateOutputSpec instance must be an existing file name, but a value of 'bar.nii' <type 'str'> was specified.

While the `ConcatenateInputSpec` seems happy with a relative path, it seems `ConcatenateOutputSpec` needs an absolute path, so simply passing it through is insufficient. This change prepends the working directory. I've tested this on my own pipeline, and it resolves an error like the one above.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 Next release
 ============
+
 * ENH: Created interface for BrainSuite Cortical Surface Extraction command line tools (https://github.com/nipy/nipype/pull/1305)
 * FIX: job execution on systems/approaches where locale is undefined (https://github.com/nipy/nipype/pull/1401)
 * FIX: Clean up byte/unicode issues using subprocess (https://github.com/nipy/nipype/pull/1394)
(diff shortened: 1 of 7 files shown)


effigies:
Seeing FreeSurfer isn't installed in Circle, so tests can't run. Is this something that should change?

satra:
freesurfer is a nasty download 4.6GB. perhaps we can create a separate minimal freesurfer just for recon-all.  more recently i think i've seen an internal docker version. let me look into that.

in the meantime, let's not run this on circle. it would also take too long to run.

@dgellis90 - since you are running this on your cluster, perhaps you can keep us posted there.

oesteban:
I'm currently migrating to docker containers in some other projects. I think this could be the solution to the problem of downloading so large files.

chrisfilo:
Freesurfer inside a docker image vs outside of it is still a 4GB download.
We can cache it (or the docker image with it) though.

satra:
@chrisfilo is correct - that's why we need to restrict it to what recon-all needs - which is only about 350MB - but there is indeed an effort to again make freesurfer available through neurodebian - which may become the smallest distribution for the binaries.

chrisfilo:
I've tried to slim down FreeSurfer in this NeuroVault docker image
https://github.com/NeuroVault/NeuroVault/blob/master/fs_docker/Dockerfile,
but I never got even close to 350Mb...

effigies:
Okay. So, as far as this PR goes, I can report that `nosetests nipype.interfaces.freesurfer.tests` runs without error on my machine with FreeSurfer 5.3.0 installed.

PR #1416, opened 2016-04-02
FMRIB's ICA-based Xnoiseifier (aka FIX)

wmpauli:
Hi, 

I would say this represents a preliminary implementation of a wrapper for FIX. I will potentially improve in the near future while I am using this and other nipype features. Still pretty new, so this may be clunky.

Best,
Wolfgang


satra:
@wmpauli - you should also consider ICA-AROMA, which uses more automation.

wmpauli:
Thanks, I will look into it. I am mainly concerned with physiological noise
and scanner artifacts.

PR #1424, opened 2016-04-07
Dcm2nii

akeshavan:
* the dcm2nii interface was returning a list of 2 identical files. 
* if dcm2nii gave a warning, it was taken as a filename (which doesn't exist) 

This had to do with how stdout was parsed. My version of dcm2nii wrote the filename in stdout twice, so the interface picked it up twice. Also, warnings have the arrow ("->") which threw things off.


satra:
@akeshavan - check circle ci tests?

PR #1428, opened 2016-04-10
fsl.MotionOutliers output spec no longer requires out_file to exist

beausievers:
fsl.MotionOutliers output spec no longer requires out_file to exist. fsl_motion_outliers does not create an output file when no timepoints are above the threshold, and in this situation fsl.MotionOutliers would throw a FileNotFoundError. This pull request fixes that immediate problem.

### DIFF
--- a/nipype/interfaces/fsl/utils.py
+++ b/nipype/interfaces/fsl/utils.py
@@ -2045,7 +2045,7 @@ class MotionOutliersInputSpec(FSLCommandInputSpec):
 
 
 class MotionOutliersOutputSpec(TraitedSpec):
-    out_file = File(exists=True)
+    out_file = File()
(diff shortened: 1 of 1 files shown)


satra:
@oesteban, @chrisfilo - we need to come up with a reasonable solution before this release. turns out the file is not created when the threshold does not result in any outliers. i was going to write a list outputs for this interface, but do you think we could have the following pattern.

```
if metadata exists does not exist
   and file pattern created by name_template does not match an existing file
then keep output undefined
```

mgxd:
@satra @oesteban has this been solved or should we merge this?

effigies:
`exists=True` is a blight in output specs. I'm merging regardless.

Issue #1431, opened 2016-04-12
autorecon1 depends on SimpleITK


cni-md:
I am trying to run the FS create_reconall_workflow but I got stucked at the
`import SimpleITK as sitk` in autorecon1.py
SimpleITK is only available for windows and mac. easy_install did not work under Debian Jessy. 
git clone git://itk.org/SimpleITK.git, cmake..., make is still running but why?

The only use I found:
`img = sitk.ReadImage(t1)
size = img.GetSize()`

2h pain with installation for img.GetSize? :(

effigies:
I think that particular call to SimpleITK can be replaced with `nibabel.load(t1).shape`. The one in [workflows/smri/freesurfer/utils.py](https://github.com/nipy/nipype/blob/8cde742/nipype/workflows/smri/freesurfer/utils.py#L496-L510) seems a little less straightforward, as it's not immediately obvious what Get/SetOrigin are doing under the hood.

PR #1432, opened 2016-04-12
REF: Simplify checkT1s

effigies:
A number of cleanups I saw while looking at #1431.

* This uses nibabel instead of SimpleITK to determine image shape. SimpleITK is still a dependency in two functions in nipype/workflows/smri/freesurfer/utils.py.
* From what I could see, the conversion from `.mgz` to `.nii` was to allow SimpleITK to load the file, so I removed it.
* I used list/generator comprehensions to reduce indentation levels and separate unrelated calculations.

@dgellis90 Can you verify that this still behaves as expected in your tests?

Minor change, so forgoing a CHANGES entry unless otherwise instructed.

### DIFF
--- a/nipype/workflows/smri/freesurfer/autorecon1.py
+++ b/nipype/workflows/smri/freesurfer/autorecon1.py
@@ -6,52 +6,35 @@ import nipype.pipeline.engine as pe  # pypeline engine
 from nipype.interfaces.freesurfer import *
 from .utils import copy_file, copy_files
 
+
 def checkT1s(T1_files, cw256=False):
     """Verifying size of inputs and setting workflow parameters"""
-    import SimpleITK as sitk
-    import os
     import sys
-    # check that the files are in a list
-    if not type(T1_files) == list:
-        T1_files = [T1_files]
+    import nibabel as nib
+    from nipype.utils.filemanip import filename_to_list
+
+    T1_files = filename_to_list(T1_files)
     if len(T1_files) == 0:
         print("ERROR: No T1's Given")
         sys.exit(-1)
-    for i, t1 in enumerate(T1_files):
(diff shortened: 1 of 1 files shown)


dgellis90:
@effigies The workflow still performs as expected with your changes.

effigies:
Great! Thanks for the quick feedback.

Unrelated to this PR (except insofar as I want to get rid of the SimpleITK dependency), could I get some inputs and results to [center_volume](https://github.com/nipy/nipype/blob/master/nipype/workflows/smri/freesurfer/utils.py#L496-L510) and [recodeLabelMap](https://github.com/nipy/nipype/blob/master/nipype/workflows/smri/freesurfer/utils.py#L513-L567)? These should be pretty easy to rewrite with nibabel, as long as I can check my work.

dgellis90:
@effigies No problem! Here is an example with inputs and outputs that uses both: https://www.dropbox.com/s/ovmwprr71lcbi87/recodingexample.tar.gz?dl=0
Let me know if you need anything else.

effigies:
Do you have some inputs that have an affine matrix that doesn't look like this?

    1 0 0 a
    0 1 0 b
    0 0 1 c
    0 0 0 1

I want to understand `center_volume`. Right now it looks like the algorithm just sets `[a, b, c] = [±dim1/2, ±dim2/2, ±dim3/2]`, but that may simply be a consequence of there being no rotation or scaling.

Incidentally, this seems odd to me, and I want to verify that your goal really is just to adjust the affine matrix, and not to translate the contents of the data array such that you have a nice affine matrix. The consequence of `center_volume` seems to be to put it out of alignment with other files from the same subject. Admittedly, I haven't looked closely at where the resulting files are used, so this could be a dumb question.

PR #1435, opened 2016-04-15
[ENH] added interface for dcm2niix

mgxd:
support for @neurolabusc's dcm2niix converter - compatible upto 4.12.16 version


nipype/interfaces/dcm2nii.py
```diff
@@ -180,3 +180,127 @@ def _gen_filename(self, name):
             f.close()
             return config_file
         return None
+
+## dcm2niix update
```
satra: remove this line

nipype/interfaces/dcm2nii.py
```diff
@@ -180,3 +180,127 @@ def _gen_filename(self, name):
...
+    'dcm2niix -b y -z i -m n -f %q -o . -s n -v n functional_1.dcm'
+    """
+
+    input_spec = Dcm2niixInputSpec
+    output_spec = Dcm2niixOutputSpec
+    _cmd = 'dcm2niix'
+
+    def _format_arg(self, opt, spec, val):
+        if opt in ['bids_format', 'merge_imgs', 'single_file', 'verbose']:
+
```
satra: extra line

nipype/interfaces/dcm2nii.py
```diff
@@ -180,3 +180,127 @@ def _gen_filename(self, name):
...
+                else:
+                    pass
+
+                if out_file:
+                    files.append(out_file + ".nii.gz")
+
+                    if self.inputs.bids_format:
+                        bids.append(out_file + ".bids")
+
```
satra: remove extra white lines

satra:
@mgxd -  could you again merge with current upstream master? and also run `make  check-before-commit`

DimitriPapadopoulos:
Some initial comments:

* Under "Examples" the generated `cmdline` contains options `-b` and `-m` which are not available in versions of _dcm2niix_ distributed by NeuroDebian on Ubuntu 14.04 for example. This doesn't seem to bother _dcm2niix_ which seems to be skipping unknown options. No BIDS file is created though so in that case it's not part of the output. Not sure how to handle this.

* Also I would change `Uses MRICRON's dcm2niix` to `Uses Chris Rorden's dcm2niix` or at least `Uses MRIcroGL's dcm2niix`.

mgxd:
@DimitriPapadopoulos Thanks for the comments.

- If bids option is empty (or doesn't exist), it shouldn't be return (there's a check for an empty list). Since we'll be transitioning to bids format across all our datasets, it's a useful feature to include.

- Changed MRIcron's dcm2niix to Chris Rorden's dcm2niix

Let me know if you catch anything else!

DimitriPapadopoulos:
I'm not sure it can work in case DICOM file names are not passed as absolute paths.

I'm in a directory with hundreds of DICOM files:
```
$ cd /foo/bar
$ ls
[...]
20140614163106701.MR.dic
[...]
20140614163106761.MR.dic
[...]
$ 
```

Then:
```
>>> from nipype.interfaces.dcm2nii import Dcm2niix
>>> converter = Dcm2niix()
>>> converter.inputs.source_names = ['20140614163106701.MR.dic', '20140614163106761.MR.dic']
>>> converter.inputs.gzip_output = True
>>> converter.inputs.output_dir = '.'
>>> converter.cmdline
'dcm2nii -a y -c y -b config.ini -v y -d y -e y -g y -i n -n y -o . -p y -x n -f n 20140614163106701.MR.dic'
>>> 
```

However _dcm2niix_ doesn't seem to be processing paths relative to the current directory:
```
$ dcm2nii -a y -c y -b config.ini -v y -d y -e y -g y -i n -n y -o . -p y -x n -f n 20140614163106701.MR.dic
Chris Rorden's dcm2niiX version 9Sept2015 (64-bit)
Version 9Sept2015 (64-bit)
Error: unable to find any DICOM images in 
Conversion required 0.000067 seconds.
$ 
```
unlike _dcm2nii_:
```
$ dcm2nii -a y -c y -b config.ini -v y -d y -e y -g y -i n -n y -o . -p y -x n -f n 20140614163106701.MR.dic
Chris Rorden's dcm2nii :: 4AUGUST2014 (Debian) 64bit BSD License
reading preferences file config.ini
0 files in working directory /volatile/070000160913/TEST
Data will be exported to ./
Validating 552 potential DICOM images.
Found 551 DICOM images.
Converting 202/551  volumes: 202
20140614161725601.MR.dic->20140614_154657EPIfacess004a001.nii
[...]
$ 
```
What about `os.path.join(os.getcwd(), ...)` at least for file names?

DimitriPapadopoulos:
From my test directory with 2 DICOM sequences, _Dcm2nii_ and _Dcm2niix_ create 2 NIfTI files.

With _Dcm2nii_ default output file names used to be:
```
20140614_154657EPIfacess004a001.nii
20140614_154657EPIstopsignals005a001.nii
```

With _Dcm2niix_ default output file names are:
```
EP.nii.gz
EPA.nii.gz
```

I understand the default has changed between _dcm2nii_ and _dcm2niix_ but would it be worth changing the default output format of _Dcm2niix_ from `%q` to something more like _Dcm2nii_? Or is Nipype supposed to be as transparent as possible?

mgxd:
@DimitriPapadopoulos 

- When converting a single image, the -s flag should change to 'y'. I was able to successfully convert one of my dicoms:
`dcm2niix -b y -z i -m n -f %q -o . -s y -v n 996000-3-1.dcm
Compression will be faster with /usr/local/bin/pigz
Chris Rorden's dcm2niiX version 12Apr2016 (64-bit)
Version 12Apr2016 (64-bit)
Convert 1 DICOM as ./EP (108x108x65x1)
Conversion required 0.240000 seconds.`

I've updated the doctest to include the single file flag.

- As for default naming - I have no problem making the output look more like dcm2nii's. I've changed the default from `%q` to `%t%p` (time and protocol).

nipype/interfaces/dcm2nii.py
```diff
@@ -185,3 +185,112 @@ def _gen_filename(self, name):
...
+                spec.argstr += ' n'
+                val = True
+        if opt == 'source_names':
+            return spec.argstr % val[0]
+        return super(Dcm2niix, self)._format_arg(opt, spec, val)
+
+    def _run_interface(self, runtime):
+        new_runtime = super(Dcm2niix, self)._run_interface(runtime)
+        (self.output_files,
+         self.bvecs, self.bvals, self.bids) = self._parse_stdout(new_runtime.stdout)
```
satra: have you tested this with and without the bids option?

DimitriPapadopoulos:
@mgxd
I'm not certain about changeset 360352339552758af582d5e7fef41feb3d7f9c77:
Unless I'm mistaken it merely modifies the example and switches from the initial use case (converting a directory) that (still) doesn't work to a different use case (converting a single image) that works.
I would have expected the documentation to stay as is and the code to be modified so that the initial example does work (perhaps by adding `os.path.join(os.getcwd(), ...)` somewhere in the code).

Indeed in single file mode _dcm2niix_ is able to handle relative pathnames:
```
$ dcm2niix -b y -z i -m n -f %q -o . -s y -v n seminoma.dic
Chris Rorden's dcm2niiX version 9Sept2015 (64-bit)
Version 9Sept2015 (64-bit)
Convert 1 DICOM as ./EP (64x64x40x1)
Conversion required 0.027342 seconds.
$ 
```
but it cannot convert directories identified by a relative pathname:
```
$ dcm2niix -b y -z i -m n -f %q -o . -s n -v n seminoma.dic
Chris Rorden's dcm2niiX version 9Sept2015 (64-bit)
Version 9Sept2015 (64-bit)
Error: unable to find any DICOM images in 
Conversion required 0.000075 seconds.
$ 
```
unlike _dcm2nii_:
```
$ dcm2nii -a y -c y -v y -d y -e y -g y -i n -n y -o . -p y -x n -f n seminoma.dic
Chris Rorden's dcm2nii :: 4AUGUST2014 (Debian) 64bit BSD License
0 files in working directory /volatile/000045835265/TEST
Data will be exported to ./
Validating 397 potential DICOM images.
Found 393 DICOM images.
Converting 191/393  volumes: 191
[...]
$ 
```

I believe _Dcm2niix_ should accept relative pathames outside of single file mode, perhaps by adding `os.path.join(os.getcwd(), ...)` to the _Dcm2niix_ code (not to the example code). Then the initial example would work.

neurolabusc:
@DimitriPapadopoulos 
 I fixed the relative paths in the 22Apr2016 version of dcm2niix. The fix works on OSX and Linux, not sure about Windows (not sure if '.' is current working directory in Windows).
  https://github.com/neurolabusc/dcm2niix

DimitriPapadopoulos:
@mgxd 

- Since Chris confirms the relative path issue has been fixed in recent versions of _dcm2niix_, perhaps it's not worth working around it in _nypipe_. By the time a new version of _nipype_ is officially released, I guess the recent version of _dcm2niix_ might have made it to NeuroDebian.

- Also I'm not totally convinced about changing the default NIfTI output file name after all. I was just asking a question about consistency between _Dcm2nii_ and _Dcm2niix_, not firmly suggesting _Dcm2niix_ should not follow the default behaviour of _dcm2nii_. Perhaps this should be best discussed and handled at the _dcm2nii_-_dcm2niix_ level.

Perhaps 360352339552758af582d5e7fef41feb3d7f9c77 should be reverted then. What do you think?

mgxd:
@DimitriPapadopoulos 
- I agree - since it has been patched in the latest dcm2niix, I'd rather not add a change to this interface and then have to undo it. @satra what do you think?
- I actually like the new naming better, more information as a default isn't a bad idea. I arbitrarily set the default to sequence anyways.

satra:
@mgxd -  the one thing i would suggest is going back to the single file example in the doctest

nipype/interfaces/dcm2nii.py
```diff
@@ -185,3 +185,118 @@ def _gen_filename(self, name):
             f.close()
             return config_file
         return None
+
+
+class Dcm2niixInputSpec(CommandLineInputSpec):
+    source_names = InputMultiPath(File(exists=True), argstr="%s", position=-1,
+                                  copyfile=False, mandatory=True, xor=['source_dir'])
```
satra: `desc` missing for all inputs.

satra: i know dcm2nii didn't have any either. but that should be fixed as well.

nipype/interfaces/dcm2nii.py
```diff
@@ -185,3 +199,125 @@ def _gen_filename(self, name):
...
+    source_dir = Directory(exists=True, argstr="%s", position=-1, mandatory=True,
+                           xor=['source_names'])
+    out_filename = traits.Str('%t%p', argstr="-f %s", usedefault=True,
+                              desc="Output filename")
+    output_dir = Directory(exists=True, argstr='-o %s', genfile=True,
+                           desc="Output directory")
+    bids_format = traits.Bool(True, argstr='-b', usedefault=True,
+                              desc="BIDS sidecar")
```
satra: perhaps change to: `Create a BIDS sidecar file`

nipype/interfaces/dcm2nii.py
```diff
@@ -185,3 +199,125 @@ def _gen_filename(self, name):
...
+    compress = traits.Enum('i', ['y','i','n'], argstr='-z %s', usedefault=True,
+                           desc="Gzip compress images - [y=pigz, i=internal, n=no]")
+    merge_imgs = traits.Bool(False, argstr='-m', usedefault=True,
+                             desc="merge 2D slices from same series")
+    single_file = traits.Bool(False, argstr='-s', usedefault=True,
+                              desc="Convert only one image")
```
satra: is this always the first image? if so, this description should clarify.

PR #1436, opened 2016-04-15
WIP: Purge SimpleITK

### LINKED ISSUE #1431: autorecon1 depends on SimpleITK

effigies:
@dgellis90 For your validation.

`RecodeLabelMap` was straightforward. What `center_volume` was doing is somewhat ambiguous (see questions on #1432), so this is a provisional rewrite.

Fixes #1431.

### DIFF
--- a/nipype/workflows/smri/freesurfer/utils.py
+++ b/nipype/workflows/smri/freesurfer/utils.py
@@ -493,20 +493,22 @@ def checkenv(exitonfail=False):
     else:
         print("Warning: " + msg)
 
+
 def center_volume(in_file):
-    import SimpleITK as sitk
     import os
-    img = sitk.ReadImage(in_file)
-    size = img.GetSize()
-    origin = img.GetOrigin()
-    new_origin = [0,0,0]
-    for i, xx in enumerate(origin):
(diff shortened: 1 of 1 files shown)


dgellis90:
@effigies Sorry to get back to you late on this, but I think we should actually git rid of these methods from the master branch. They are not required for the main recon-all workflow itself, but were used as an add on to make some experimental tests easier.  If you'd like to keep them, I can review the changes and get back to you.

I'll make a PR for removing these here in a couple minutes.

effigies:
Nope, feel free to kill the functions. My only interest here is getting SimpleITK out, since it's not easily available for Linux users and it's not an official nipype dependency. Please put a "Closes #1436" in your description so this PR closes when yours is accepted.

PR #1437, opened 2016-04-15
FIX: Removes unnecessary methods from smri freesurfer workflow utils.

dgellis90:
Removes unnecessary methods from freesurfer workflows as well as SimpleITK. Closes #1436


satra:
Closes #1436

PR #1438, opened 2016-04-18
add nipype_crash_search command

alexsavio:


alexsavio:
There was a `CHANGES.orig` file I guess from a merge that someone else and accidentally added it.

PR #1444, opened 2016-04-20
fix: cleaned up provenance generation

satra:


satra:
@effigies - could you please take a look at the fix on filemanip, i had to add another exception on my os x system, because tempFATFS failed with unknown directory.

effigies:
Looks good to me.

PR #1460, opened 2016-04-27
New interfaces

dclark87:
Added new afni interfaces for: 3dECM, 3dDegreeCentrality, 3dLFCD, 3dClipLevel, 3dmask_tool, and 3dSeg as well as their unit tests


dclark87:
We need to add a couple of corrections here. @sgiavasis and I will let you guys know when these can be fully pulled in

chrisfilo:
Ok. Thanks for the update.

dclark87:
@chrisfilo - This is all patched up and ready now

dclark87:
@chrisfilo - any thoughts on this? Does this look good to merge?

chrisfilo:
Could you list the new interfaces in CHANGES file? Otherwise looks good.

dclark87:
Where in CHANGES? Here? https://github.com/fcp-indi/nipype/blob/new_interfaces/CHANGES#L34

chrisfilo:
That's the one! This way users can see what's new - it get's rendered here:
http://nipy.org/nipype/changes.html

PR #1461, opened 2016-04-28
FIX SGEGraph fixed to allow for varied file permissions

ostanley:
By not specifying bash in qsub commands you require the generated files to have executable permissions. Most linux systems create 644 files without these permissions. I corrected this by adding bash to the qsub command before the batchscript variable. If there is another proper way to do this please let me know.

Thank you,

Olivia

### DIFF
--- a/nipype/pipeline/plugins/sgegraph.py
+++ b/nipype/pipeline/plugins/sgegraph.py
@@ -136,7 +136,7 @@ class SGEGraphPlugin(GraphPluginBase):
                     if self._qsub_args.count('-o ') == 0:
                         stdoutFile = '-o {outFile}'.format(
                             outFile=batchscriptoutfile)
-                    full_line = '{jobNm}=$(qsub {outFileOption} {errFileOption} {extraQSubArgs} {dependantIndex} -N {jobNm} {batchscript} | awk \'{{print $3}}\')\n'.format(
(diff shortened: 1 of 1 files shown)


satra:
@hjmjohnson - can you take a look here?

nipype/pipeline/plugins/sgegraph.py
```diff
@@ -136,7 +136,7 @@ def make_job_name(jobnumber, nodeslist):
                     if self._qsub_args.count('-o ') == 0:
                         stdoutFile = '-o {outFile}'.format(
                             outFile=batchscriptoutfile)
-                    full_line = '{jobNm}=$(qsub {outFileOption} {errFileOption} {extraQSubArgs} {dependantIndex} -N {jobNm} {batchscript} | awk \'{{print $3}}\')\n'.format(
+                    full_line = '{jobNm}=$(qsub {outFileOption} {errFileOption} {extraQSubArgs} {dependantIndex} -N {jobNm} bash {batchscript} | awk \'{{print $3}}\')\n'.format(
```
hjmjohnson: @ostanley @satra I think it would be better to ensure that that bash script has execute permisssions.

ostanley: Forgive me if this is an obvious question but how would you do that? You can only use chmod after the file exists and the nipype creates the file just before run. Should I add an os.chmod() command to the script creation?

effigies:
Marking orphaned, but if someone comes back to this, you can check permissions with `os.access()`.

effigies:

Issue #1389, opened 2016-03-03
dcm2nii → dcm2niix


DimitriPapadopoulos:
We have found _dcm2nii_ is unable to convert some recent Philips DICOM files. Those files can be converted using _dcm2niix_ instead.

I suggest adding _dcm2niix_ to Nipype.

Because _dcm2niix_ is not yet mature software and does not support all the proprietary formats supported by _dcm2nii_, I don't think `dcm2nii` should be changed to `dcm2niix` in [nipype/interfaces/dcm2nii.py](https://github.com/nipy/nipype/blob/master/nipype/interfaces/dcm2nii.py). Should we provide a duplicate interface for _dcm2niix_? Add an option to switch between _dcm2nii_ and _dcm2niix_?

satra:
@DimitriPapadopoulos - i think dcm2niix should be a separate interface. we have found some instabilities in dcm2niix where it won't convert any of our dicoms!

DimitriPapadopoulos:
I totally agree both programs should be available, as I said _dcm2niix_ is not yet mature.

I was thinking of two options:
* Keep the same interface with an additional option to switch between `dcm2nii` (by default) and `dcm2niix` (only when the relevant option is used).
* Provide a new specific interface. Might be better indeed since _dcm2nii_ has options such as `-a` that seem to be missing from _dcm2niix_.

satra:
a new specific interface is better in this case.

cni-md:
You need a specific interface. The std output is different and the parser for e.g. bvec, bval will fail.
I tried to build a dcm2niix interface to be used with heudiconv. But dcm2niix -s will convert only one file. You can not convert a list of files needed by heudiconv. :( Convert all files in a dir is fine.

DimitriPapadopoulos:
One could imagine copying the relevant files to a temp dir to convert a subset of a dir. I don't know if it's worth the overhead though.

satra:
@mgxd is building an interface for this, which he says is almost ready for a PR.

mgxd:
I've submitted a PR - let me know what you guys think

neurolabusc:
[...]

PR #1463, opened 2016-05-02
output for DTI vectors changed (dcm2nii -> dcm2niix)

cni-md:
New: Saving 65 DTI gradients. Please validate if you are conducting DTI
analyses.

see https://github.com/nipy/nipype/issues/1389

### DIFF
--- a/nipype/interfaces/dcm2nii.py
+++ b/nipype/interfaces/dcm2nii.py
@@ -292,7 +292,7 @@ class Dcm2niix(CommandLine):
                         bvals.append(out_file + ".bval")
                         find_b = False
                 # next scan will have bvals/bvecs
-                elif 'DTI gradient directions' in line:
+                elif 'DTI gradients' in line:
                     find_b = True
                 else:
                     pass


nipype/interfaces/dcm2nii.py
```diff
@@ -292,7 +292,7 @@ def _parse_stdout(self, stdout):
                         bvals.append(out_file + ".bval")
                         find_b = False
                 # next scan will have bvals/bvecs
-                elif 'DTI gradient directions' in line:
+                elif 'DTI gradients' in line:
```
mgxd: would suggest changing to `elif 'DTI gradient' in line:` to make compatible with former output as well

cni-md: Oh,
that's nasty. I thought 'DTI gradient directions' was the old dcm2nii
syntax and copy&paste introduced this error, but now I realize that this
is "old" dcm2niiX output.

+1 for another elif statement


> In nipype/interfaces/dcm2nii.py
> <https://github.com/nipy/nipype/pull/1463#discussion_r61746680>:
> 
>> @@ -292,7 +292,7 @@ def _parse_stdout(self, stdout):
>>                          bvals.append(out_file + ".bval")
>>                          find_b = False
>>                  # next scan will have bvals/bvecs
>> -                elif 'DTI gradient directions' in line:
>> +                elif 'DTI gradients' in line:
> 
> would suggest changing to |elif 'DTI gradient' in line:| to make
> compatible with former output as well

satra:
i fixed this in the upstream master. let's consider this a beta interface since @neurosclab is changing things depending on user needs.

PR #1478, opened 2016-05-17
DOC:add instructions to install; fix doc build for python 3

shoshber:
small changes to start contributing!

### DIFF
--- a/doc/links_names.txt
+++ b/doc/links_names.txt
@@ -56,7 +56,7 @@
 .. _doctest-mode: http://www.cis.upenn.edu/~edloper/projects/doctestmode/
 .. _bazaar: http://bazaar-vcs.org/
 .. _subversion: http://subversion.tigris.org/
-.. _nose: http://somethingaboutorange.com/mrl/projects/nose
+.. _nose: http://nose.readthedocs.io/en/latest/
 .. _`python coverage tester`: http://nedbatchelder.com/code/modules/coverage.html
 
 .. Other python projects
--- a/doc/sphinxext/numpy_ext/numpydoc.py
+++ b/doc/sphinxext/numpy_ext/numpydoc.py
@@ -40,7 +40,7 @@ def mangle_docstrings(app, what, name, obj, options, lines,
 
     if what == 'module':
         # Strip top title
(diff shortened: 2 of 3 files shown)


satra:
@shoshber - welcome to nipype.

i was recently told that the install instructions are not very evident for a lay person. since you are new to nipype, and editing the file in this PR, could you please take a look?

shoshber:
@satra aye, that is exactly what I'm doing! Maybe not for a layperson, but I can provide perspective from someone who is new to the project.

shoshber:
I was planning to add clarifications to the install documentation in another pull request, but I can do it in this one--is that what you are asking?

satra:
@shoshber - since i saw an `install.rst` change, i thought it could be done with this. but having a separate PR for it would be good.

shoshber:
I'm not done testing my installation of nipype yet, but here are my thoughts so far:
1. nibabel 2.0.1 is already in nipype's requirements.txt file. Why do the instructions say to install it separately?
2. Setting up my development environment to build the documentation was unexpectedly difficult. Some ideas to help it go more smoothly for the next person: a) document packages that are required for development, say, in a file called requirements-developer.txt (by analogy to requirements.txt). sphinx, graphviz would be in the file. b) build documentation testing into circleci to prevent accidental commits of breaking changes

I updated install.rst with some clarifications/additional instructions. I'll add it to this pull request as soon as it finishes building.

Shoshana

chrisfilo:
1. `pip install nipype` should automatically install nibabel if it's not present. There is no need to install it manually. Maybe we should remove all dependencies that are installed automatically? They only confuse users.

1. Documentation is already built and tested for error on CircleCI. See https://github.com/nipy/nipype/blob/master/circle.yml#L49

mwaskom:
> There is no need to install it manually. Maybe we should remove all dependencies that are installed automatically? They only confuse users.

I think it's good to know what kind of dependencies a package has, though I agree it should be made clear what pip will take care of.

shoshber:
1. Yeah, the Mac OS X instructions make it seem like you must install nibabel separately. I'll clarify that in the instructions.
2. That's odd--the numpy doc would not build in python 3 until I made the changes in the commit. I'm not versed enough in circleci to check--perhaps it is only checking the build for python 2?

PR #1484, opened 2016-05-25
FIX: The GradientTable object should be immutable; the bvalues aren't

kesshijordan:
being passed into the object correctly, so I changed the function used
to create the GradientTable object

### DIFF
--- a/nipype/interfaces/dipy/base.py
+++ b/nipype/interfaces/dipy/base.py
@@ -59,13 +59,8 @@ class DipyDiffusionInterface(DipyBaseInterface):
     def _get_gradient_table(self):
         bval = np.loadtxt(self.inputs.in_bval)
         bvec = np.loadtxt(self.inputs.in_bvec).T
-        try:
-            from dipy.data import GradientTable
-            gtab = GradientTable(bvec)
-            gtab.bvals = bval
-        except NameError:
-            from dipy.core.gradients import gradient_table
-            gtab = gradient_table(bval, bvec)
+        from dipy.core.gradients import gradient_table
+        gtab = gradient_table(bval, bvec)
 
         gtab.b0_threshold = self.inputs.b0_thres
         return gtab


kesshijordan:
I was using this with a .bvecs and .bvals format input to the GradientTable object, and the units were off by quite a bit when I fit the tensor. It turns out, the bvalues weren't being used by the GradientTable object because they were being set as an attribute after the object was created. They should be input using one of the factory functions to ensure that they are handled correctly.

oesteban:
One question @arokem: is that possible that older versions of dipy only implement the ```GradientTable```?

If I'm not wrong, I implemented it that way because ```dipy.core.gradient_table``` didn't seem to be available in older versions of dipy.

Once confirmed, may I ask you @kesshijordan to go back to a try ... except structure, but using the recommended ```gradient_table``` as default?. Something like:

```
try:
    from dipy.core.gradients import gradient_table
    gtab = gradient_table(bval, bvec)
except NameError:
    from dipy.data import GradientTable
    gtab = GradientTable(bvec)
    gtab.bvals = bval
```

arokem:
This has been in place since version 0.6 (2012):
https://github.com/nipy/dipy/commit/1714e7c5059fd88a8f036b15a8e59953c7eff197

kesshijordan:
The way that is written, the bvals you are setting will not be reflected by the GradientTable object, and everything will by off by a factor of the bvalue. We could put just the vectors in, and let the user combine their bvals&bvecs in the except case (@mrbago, perhaps you could comment?)

oesteban:
Well, looking at the old code: https://github.com/nipy/dipy/commit/1714e7c5059fd88a8f036b15a8e59953c7eff197#diff-3d2aafe93b8f41be9cf3521aa9d5d194L5, maybe this would do:

```
try:
    from dipy.core.gradients import gradient_table
    gtab = gradient_table(bval, bvec)
except NameError:
    from dipy.data import GradientTable
    gtab = GradientTable(bval, bvecs=bvec)
```

Don't you think?

MrBago:
@oesteban GradientTable and gradient_table were added in the same release, the older version you're looking at was a WIP that was never made public. I just checked and dipy 0.5.0 does not have a `dipy.core.gradients` file at all.

The issue may have been because GradientTable is being imported from `dipy.data` instead of `dipy.core.gradients` (I'm not sure when we started importing `GradientTable` into `dipy.data`). I believe if you want a GradientTable from a bvec/bval pair, you should be using the helper functions.

Just for context, the helper functions do a bunch of checking for errors and corner cases then call `GradientTable(bval[:, None] * bvec)`.

oesteban:
@MrBago thanks for the details :). What do you think we should implement for nipype? Should we get rid of the try ... catch, and go use `gradient_table`?

MrBago:
I'm not sure about the try catch, why is it catching a NameError?

oesteban:
I included it because I was having that exception in earlier versions of dipy. It should be an ```ImportError``` , correct?

MrBago:
In that case I don't think you need the try catch.

arokem:
I think that what @MrBago means is that since the code proposed here imports from the right place (from `dipy.core.gradients`, this should work on older versions of Dipy as well. I am +1 for merging this PR.

PR #1485, opened 2016-05-25
DOC more documentation (tutorials) fixes/updates

shoshber:


doc/users/install.rst
```diff
@@ -76,8 +76,14 @@ nose_ installed, then do the following::
 
 you can also test with nosetests::
 
-    nosetests --with-doctest /software/nipy-repo/masternipype/nipype
-    --exclude=external --exclude=testing
+    nosetests --with-doctest <installation filepath>/nipype  --exclude=external --exclude=testing
```
satra: if nipype is already installed this can now be done with: `nosetests --with-doctest nipype`

doc/users/install.rst
```diff
@@ -119,7 +125,7 @@ Must Have
 Nibabel_ 1.0 - 1.4
```
satra: should probably allow for greater nibabel version. some of these are set with `*_MIN_VERSION` in info.py. we should just follow that i think.

doc/users/install.rst
```diff
@@ -119,7 +125,7 @@ Must Have
...
   Neuroimaging file i/o library.
 
-Python_ 2.7
+Python_ 2.7 or Python_ 3
```
satra: may need to be 3.3+

doc/users/install.rst
chrisfilo: Should we even list those dependencies? It gives a false impression that user needs to install those libraries manually when in truth all of this is done automatically by pip. I would simplify this to `pip install nipype`

shoshber: pip install nipype is here: https://github.com/nipy/nipype/pull/1485/files#diff-444fe9dbdb99f1db654e1abeebe704ebR40

The note above clarifies that I think. Or would you like to remove the section entirely?

satra: i agree. perhaps we can simply replace it with a link to the requirements file.

it would be good to simplify this to:

1. install with pip
2. if you don't have sudo privileges, 
    a. install a miniconda environment 
    b. install nipype into it.
3. if you don't want to do either talk to your sysadmin :)

doc/users/install.rst
shoshber: added it as an option

doc/users/install.rst
shoshber: removed "dependencies" section, added link to requirements.txt

doc/users/install.rst
shoshber: removed

shoshber:
I messed up my branches so the diffs are confusing--sorry about that

shoshber:
diffs are correct now

satra:
we should also update the initial bit in `install.rst` that points to older versions of nipype. perhaps it should simply say:

`
Current release
Development version
`

shoshber:
-> update the install.rst to point to current versions of nipype

done!

PR #1493, opened 2016-06-03
ENH Added grad_dev input to fsl.dti.bedpostx5 interface

mattfeld:
bedpostx expects a grad_dev.nii.gz file in subject directory if the -g flag is true.

I've tested these changes locally running a bedpostx workflow successfully on the HCP data.


nipype/interfaces/fsl/dti.py
```diff
@@ -348,6 +349,9 @@ def _run_interface(self, runtime):
                  os.path.join(subjectdir, 'bvals'))
         copyfile(self.inputs.bvecs,
                  os.path.join(subjectdir, 'bvecs'))
+        _, _, ext = split_filename(self.inputs.grad_dev)
```
satra:
```
if isdefined(self.inputs.grad_dev):
    _, _, ext = split_filename(self.inputs.grad_dev)
    copyfile(self.inputs.grad_dev, os.path.join(subjectdir, 'grad_dev' + ext))
```

satra:
thanks @mattfeld , the comment takes care of the situation when the grad_dev is not defined. also, if you could run make_specs and add the updated test_auto_bedpostx file that would be good.

mattfeld:
thanks @satra for your comment. I've added the if statement for times when grad_dev is not defined, ran the make specs and updated the test_auto_BEDPOSTX5.py file.  Let me know if there is anything else I need to do.

PR #1494, opened 2016-06-03
[ENH] Fill in_index automatically in ApplyTOPUP if not defined

oesteban:
Allow ```in_index``` to be undefined.

### DIFF
--- a/nipype/interfaces/fsl/epi.py
+++ b/nipype/interfaces/fsl/epi.py
@@ -378,7 +378,7 @@ class ApplyTOPUP(FSLCommand):
         # If not defined, assume index are the first N entries in the
         # parameters file, for N input images.
         if not isdefined(self.inputs.in_index):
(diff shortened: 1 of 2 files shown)


nipype/interfaces/fsl/epi.py
```diff
@@ -374,6 +371,17 @@ class ApplyTOPUP(FSLCommand):
...
     output_spec = ApplyTOPUPOutputSpec
 
+    def _parse_inputs(self, skip=None):
+        if skip is None:
+            skip = []
+
+        # If not defined, assume index are the first N entries in the
+        # parameters file, for N input images.
+        if not isdefined(self.inputs.in_index):
+            self.inputs.in_index = range(1, len(self.inputs.in_files) + 1)
```
satra: `range` -> `list(range(...))` in py3

Issue #1498, opened 2016-06-08
nipype.interfaces.fsl.Level1Design can't do gamma convolution


TheChymera:
[The docs](http://nipy.org/nipype/interfaces/generated/nipype.interfaces.fsl.model.html#level1design) seem to suggest the function could do gamma convolution.

I set the attribute at hand accrodingly:

```
level1design = pe.Node(interface=Level1Design(), name="level1design")
level1design.inputs.bases = {'gamma': {'derivs':False}}
```

And sure enough I got the following in my node's `_report/report.rst` file:

```
Original Inputs
---------------

* bases : {'gamma': {'derivs': False}}
* contrasts : [('allStim', 'T', ['s1', 's2', 's3', 's4', 's5', 's6'], [1.0, 1.0, 1.0, 1.0, 1.0, 1.0])]
* ignore_exception : False
* interscan_interval : 1.0
* model_serial_correlations : True
* session_info : [{'regress': [], 'cond': [{'duration': [30.0], 'onset': [172.77000000000001], 'name': 's1'}, {'duration': [30.0], 'onset': [322.76999999999998], 'name': 's2'}, {'duration': [30.0], 'onset': [472.76999999999998], 'name': 's3'}, {'duration': [30.0], 'onset': [622.76999999999998], 'name': 's4'}, {'duration': [30.0], 'onset': [772.76999999999998], 'name': 's5'}, {'duration': [30.0], 'onset': [922.76999999999998], 'name': 's6'}], 'hpf': 180.0, 'scans': '/mnt/data/NIdata/ofM.erc/GLM/level1_ext_gamma/first_level/_condition_ERC_ofM_subject_5503/_scan_type_EPI_CBV_alej/specify_model/corr_10_trans_filt.nii.gz'}]

Execution Inputs
----------------

* bases : {'gamma': {'derivs': False}}
* contrasts : [('allStim', 'T', ['s1', 's2', 's3', 's4', 's5', 's6'], [1.0, 1.0, 1.0, 1.0, 1.0, 1.0])]
* ignore_exception : False
* interscan_interval : 1.0
* model_serial_correlations : True
[...]

PR #1500, opened 2016-06-08
fixed fsl.Level1Design gamma convoluion

TheChymera:
This also provides a framework that is easily extensible to other convolution methods which we do not yet provide.

Fixes issue #1498

### DIFF
--- a/nipype/interfaces/fsl/model.py
+++ b/nipype/interfaces/fsl/model.py
@@ -124,8 +124,8 @@ class Level1Design(BaseInterface):
         f.close()
 
     def _create_ev_files(
-        self, cwd, runinfo, runidx, usetd, contrasts, no_bases,
-            do_tempfilter):
+        self, cwd, runinfo, runidx, usetd, contrasts,
+            do_tempfilter, basis_key):
         """Creates EV files from condition and regressor information.
 
            Parameters:
@@ -144,8 +144,9 @@ class Level1Design(BaseInterface):
         """
         conds = {}
         evname = []
-        ev_hrf = load_template('feat_ev_hrf.tcl')
-        ev_none = load_template('feat_ev_none.tcl')
+        if basis_key == "dgamma":
+            basis_key = "hrf"
+        ev_template = load_template('feat_ev_'+basis_key+'.tcl')
         ev_ortho = load_template('feat_ev_ortho.tcl')
         ev_txt = ''
         # generate sections for conditions and other nuisance
@@ -174,13 +175,13 @@ class Level1Design(BaseInterface):
                             evinfo.insert(j, [onset, cond['duration'][j], amp])
                         else:
                             evinfo.insert(j, [onset, cond['duration'][0], amp])
-                    if no_bases:
-                        ev_txt += ev_none.substitute(ev_num=num_evs[0],
+                    if basis_key == "none":
+                        ev_txt += ev_template.substitute(ev_num=num_evs[0],
                                                      ev_name=name,
                                                      tempfilt_yn=do_tempfilter,
                                                      cond_file=evfname)
                     else:
-                        ev_txt += ev_hrf.substitute(ev_num=num_evs[0],
+                        ev_txt += ev_template.substitute(ev_num=num_evs[0],
                                                     ev_name=name,
                                                     tempfilt_yn=do_tempfilter,
                                                     temporalderiv=usetd,
@@ -296,12 +297,9 @@ class Level1Design(BaseInterface):
         if isdefined(self.inputs.model_serial_correlations):
             prewhiten = int(self.inputs.model_serial_correlations)
         usetd = 0
-        no_bases = False
         basis_key = list(self.inputs.bases.keys())[0]
         if basis_key in ['dgamma', 'gamma']:
             usetd = int(self.inputs.bases[basis_key]['derivs'])
-        if basis_key == 'none':
-            no_bases = True
         session_info = self._format_session_info(self.inputs.session_info)
         func_files = self._get_func_files(session_info)
         n_tcon = 0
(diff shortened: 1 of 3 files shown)


satra:
thanks @TheChymera - could we add two more pieces?

1. an updated CHANGES file
2. a small test for the function that ensures that the hrf is set to the correct basis_key. the test could create a fake condition.

TheChymera:
1) where do I add a line to the `CHANGES` file?

Presumably not under the following header - right? Do I create a new header?

```
Release 0.12.0-rc1 (April 20, 2016)
============
```

2) Can you provide an example of how such a test should be formatted and where it should be put? On a side note, this might be a bit difficult - as the `Level1Design` node requires a `session_info` argument, which has to contain a path to an existing 4D NIfTI file. nipype doesn't ship with any of those - right?

satra:
@TheChymera - under Release 0.12.0-rc1 is fine, we will be updating that for the full release soon.

re: the test, i was thinking that the test could go into `../interfaces/fsl/tests/test_model`

this test would specifically use the `_create_ev_files` function of the interface, for which all i think you need to do is sth along the lines of:

```
from ...base import undefined
runinfo = dict(cond=[{'name': 'test_condition', 'onsets': [0, 10], durations=[10, 10]}])
runidx = 0
contrasts = undefined
usetd = False
do_tempfilter = False
```

`basis_key` will change for each type of hrf

and then check that the return from the function contains the right basis_key. 

this doesn't check the actual interface, which will be a separate repo of tests we will be developing.

nipype/interfaces/fsl/model.py
```diff
@@ -144,8 +144,9 @@ def _create_ev_files(
         """
         conds = {}
         evname = []
-        ev_hrf = load_template('feat_ev_hrf.tcl')
-        ev_none = load_template('feat_ev_none.tcl')
```
satra: `ev_none` is still needed for the regressors

TheChymera:
@satra I can't test my test script because no matter how I call it I get `ValueErrors` somewhere down the line due to the relative imports, e.g.:

```
chymera@quiethost ~/src/nipype/nipype/interfaces $ python -m fsl.tests.test_Level1Design.test_level1design
  File "/usr/lib64/python2.7/runpy.py", line 151, in _run_module_as_main
    mod_name, loader, code, fname = _get_module_details(mod_name)
  File "/usr/lib64/python2.7/runpy.py", line 101, in _get_module_details
    loader = get_loader(mod_name)
  File "/usr/lib64/python2.7/pkgutil.py", line 464, in get_loader
    return find_loader(fullname)
  File "/usr/lib64/python2.7/pkgutil.py", line 474, in find_loader
    for importer in iter_importers(fullname):
  File "/usr/lib64/python2.7/pkgutil.py", line 430, in iter_importers
    __import__(pkg)
  File "fsl/__init__.py", line 9, in <module>
    from .base import (FSLCommand, Info, check_fsl, no_fsl, no_fsl_course_data)
  File "fsl/base.py", line 28, in <module>
    from builtins import object
  File "/usr/lib64/python2.7/site-packages/builtins/__init__.py", line 8, in <module>
    from future.builtins import *
  File "/usr/lib64/python2.7/site-packages/future/builtins/__init__.py", line 13, in <module>
    from future.builtins.misc import (ascii, chr, hex, input, isinstance, next,
  File "/usr/lib64/python2.7/site-packages/future/builtins/misc.py", line 43, in <module>
    from io import open
  File "io.py", line 38, in <module>
    from .base import (TraitedSpec, traits, File, Directory,
ValueError: Attempted relative import in non-package
```

Hoe do you do it?

satra:
you can use:

`nosetests nipype.interfaces.fsl.tests.test_maths:test_meanimage`

but this requires nipype to be installed (`python setup.py install` or `python setup.py develop`)

satra:
the current updates still use ev_none here:

https://github.com/nipy/nipype/pull/1500/files#diff-eade534f1c2e76b62be0e071aae3324cR195

TheChymera:
> the current updates still use ev_none here:

Yes, isn't that supposed to happen for the regressors? I never used this feature, so I didn't even realize `ev_none` was still needed before you pointed it out. Now I reintroduced it.

>  this requires nipype to be installed

I won't test the right thing if I import from the installed nipype (I see the other manually written tests do) - because that way I am not using the updated `model.py`.

nipype/interfaces/fsl/tests/test_Level1Design.py
```diff
@@ -0,0 +1,12 @@
...
+from ..model import Level1Design
+
+def test_level1design():
+	l = Level1Design()
+	runinfo = dict(cond=[{'name': 'test_condition', 'onset': [0, 10], 'duration':[10, 10]}],regress=[])
+	runidx = 0
+	contrasts = Undefined
+	usetd = False
+	do_tempfilter = False
+	return Level1Design._create_ev_files(l,os.getcwd(),runinfo,runidx,usetd,contrasts,do_tempfilter,"hrf")
```
satra: this doesn't really test it. the following would run a test. `something` would need to be replaced with the expected string

```
output = Level1Design._create_ev_files(l,os.getcwd(),runinfo,runidx,usetd,contrasts,do_tempfilter,"none")
yield assert_true, 'something' in output
output = Level1Design._create_ev_files(l,os.getcwd(),runinfo,runidx,usetd,contrasts,do_tempfilter,"dgamma")
yield assert_true, 'something' in output
output = Level1Design._create_ev_files(l,os.getcwd(),runinfo,runidx,usetd,contrasts,do_tempfilter,"gamma")
yield assert_true, 'something' in output
```

TheChymera:
@satra could you please pul?

PR #1506, opened 2016-06-15
fslmaths -Tstd option added

Leoniela:
Hi, I added a class to fslmaths to calculate the standard deviation in a voxel. 

Here is the neurostars-thread about it:
https://neurostars.org/p/4004/

I already tried it out in my local nipype version in a workflow like this.
stdev = pe.Node(fsl.maths.StdImage(), name= 'Std_deviation_calculation')
The result looks good.

It is the first time I do something like this. Is this right? Also, I don't know, where I would have to add documentation. Feedback appreciated. Thanks, Leonie

### DIFF
--- a/nipype/interfaces/fsl/maths.py
+++ b/nipype/interfaces/fsl/maths.py
@@ -103,6 +103,17 @@ class Threshold(MathsCommand):
             return arg
         return super(Threshold, self)._format_arg(name, spec, value)
 
+class StdImageInput(MathsInput):
+
+    dimension = traits.Enum("T", "X", "Y", "Z", usedefault=True, argstr="-%sTstd", position=4,
+                            desc="dimension to standard deviate across")
+
+
+class StdImage(MathsCommand):
+    """Use fslmaths to generate a standard deviation in an image across a given dimension.
+    """
+    input_spec = StdImageInput
+_suffix = "_std"
 
 class MeanImageInput(MathsInput):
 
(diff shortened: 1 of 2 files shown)


nipype/interfaces/fsl/maths.py
```diff
@@ -103,6 +103,17 @@ def _format_arg(self, name, spec, value):
...
+
+    dimension = traits.Enum("T", "X", "Y", "Z", usedefault=True, argstr="-%sstd", position=4,
+                            desc="dimension to standard deviate across")
+
+
+class StdImage(MathsCommand):
+    """Use fslmaths to generate a standard deviation in an image across a given dimension.
+    """
+    input_spec = StdImageInput
+_suffix = "_std"
```
shoshber: Correct the indentation here to line up with ``input_spec``

shoshber:
Hi @Leoniela, nice to see you here too! Your contribution looks good to me. Before we merge your contribution, we need to make sure that the circleci test passes (see above). I would also like to see a test for your new code here: https://github.com/Leoniela/nipype/blob/9f491b56a3588bb074159cbf983b1bd6dcbe141b/nipype/interfaces/fsl/tests/test_maths.py

Again, you can probably follow ``-Tmax`` as a model.

Leoniela:
Hi @shoshber, thanks for the answer. I git pulled the master (updated mine), changed the indentation that had been messed up, and I added this test to the fsl/test/test_maths.py file (here I don't fully understand the code so I basically just copied the -Tmax snipped as you suggested and changed it accordingly - no warranty). Do I have to pull request again? How can I check this circleci test?

shoshber:
The circleci tests run automatically when you add to any pull request--it looks like they failed again. Don't worry about the testing for now, @satra and I are looking into it

satra:
i've created a separate pull-request (#1518) that replaces this one. hopefully once it goes through testing, it will be ready for merge

PR #1513, opened 2016-06-16
FIX/TST tests failing silently are bad

shoshber:


satra:
@shoshber - py3 tests are failing on travis.

Issue #1510, opened 2016-06-15
crash file overwriting


poldrack:
it appears that if two crashes occur within the same minute in the same workflow module, the crash file gets overwritten:

```
160615-11:32:18,49 workflow ERROR:
	 could not run node: fixedfx.l2model
160615-11:32:18,49 workflow INFO:
	 crashfile: /work/01329/poldrack/wrangler/code/fmri-handbook-2e-code/not
ebooks/DataPreparation/crash-20160615-113213-poldrack-l2model.pklz
160615-11:32:18,49 workflow INFO:
	 ***********************************
Traceback (most recent call last):
  File "DataPreparation-ds005-first+secondlevel.py", line 777, in <module>
    graph=fixed_fx.run() #plugin='MultiProc', plugin_args={'n_procs' : 16})
160615-11:32:18,49 workflow INFO:
	 ***********************************
160615-11:32:18,50 workflow ERROR:
	 could not run node: fixedfx.l2model
160615-11:32:18,50 workflow INFO:
	 crashfile: /work/01329/poldrack/wrangler/code/fmri-handbook-2e-code/not
ebooks/DataPreparation/crash-20160615-113213-poldrack-l2model.pklz
160615-11:32:18,50 workflow INFO:

```
should probably check for existing and add some kind of suffix for multiple files.

PR #1516, opened 2016-06-18
FIX/WIP unique crash files

shoshber:
per issue https://github.com/nipy/nipype/issues/1510


shoshber:
The build only fails in python 3.4, and the failure occurs before my changes are reached. I believe this was already broken, but this is the first test covering that code.

PR #1528, opened 2016-07-05
Fix broken format_node due to missing inspect import.

mvbnoel:
I tried to use the export feature of the workflow object, but it kept failing because of the import of the inspect module. The latter was indeed missing. Since inspect is not available for the early Python 3 and Python 2 releases, I also use the funcsigs module which backports the signature function.

I have tested it on my private fork, running Python 2.7 and the export now works. However I could not figure out how to specify the conditional install_requires on funcsigs inside the setup.py. You guys might want to merge this PR as it is and fix the setup.py later, or give me some pointers on how to do it.

cheers.

### DIFF
--- a/nipype/pipeline/engine/base.py
+++ b/nipype/pipeline/engine/base.py
@@ -15,6 +15,8 @@ The `EngineBase` class implements the more general view of a task.
 
 """
 
+from __future__ import absolute_import
+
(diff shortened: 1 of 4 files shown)


satra:
@mvbnoel - we only support 2.7, 3.3+ - so this fix should work.

Issue #1536, opened 2016-07-16
Create a test-docker-image


oesteban:
Software to be included in this image for tests:
- [x] FSL
- [x] FreeSurfer
- [x] SPM
- [x] ANTs
- [x] AFNI
- [ ] dcm2niix
- [x] nipy
- [x] C3D_affine_tool
- [ ] VTK

Environments:
- [x] nipypetest-2.7
- [x] nipypetest-3.4
- [x] nipypetest-3.5

oesteban:
This will be finished in #1815 - Closing this issue.

PR #1539, opened 2016-07-16
[TST] Add base and conda docker images #1536

oesteban:


satra:
shouldn't we just use the conda forge?

oesteban:
Yes, but the conda-forge packaging is still a bit experimental and @chrisfilo told me to wait yesterday ;)

PR #1541, opened 2016-07-18
[TST] Migrate CircleCI tests to docker

oesteban:


satra:
@oesteban - should we merge this? or will circle ci also build docker image from these dockerfiles?

oesteban:
@satra no, not yet. I am still fixing some final issues (i.e. sphinx was not installed). When the tests are equivalent to those we currently have I will remove the tag WIP in the title and let you know ;)

oesteban:
More tests are getting green :)

@satra, I have the impression that none of the tests uses the FSL course data. Do you remember if this is correct? It takes a while to download it and it'd be great if we could remove that dataset.

satra:
maybe this one does: https://github.com/nipy/nipype/blob/master/circle.yml#L57

from a certain perspective, unless we create a perfect matching workflow for feat, we can test on any data.

oesteban:
Well, that one uses the fsl feeds only, I guess... The fsl_course_data contains three data packages: ```fdt1```, ```fdt2``` and ```tbss```. The three of them are for diffusion and all the test workflows are for fmri...

satra:
we should have feeds only then for now instead of everything.

oesteban:
All green now! Some final details and this will be ready for merge.

oesteban:
@satra this is ready to be merged.

I expect to lower down a bit the 3h 30' it takes to run all the tests when the image caching starts to work. It seems that cache is taken from the master branch, and thus it is not working yet.

PR #1548, opened 2016-07-26
added 'tax' switch to ResponseSD interface _cmd to match latest specification of dwi2response

tclose:
The first argument to the latest specification of MRtrix 3's version of dwi2response is now a switch between 5 different methods for generating a diffusion response function. It appears that the ResponseSD interface in NiPype corresponds to the 'tax' option so I have added it into the _cmd attribute of the interface.

It may be desirable to include options for the different methods as well but I think these would work better as separate interfaces in any case.


oesteban:
Hi @tclose, would it be too hard to add a ```traits.Enum``` input (with default ```tax```) to support the remaining function modes? Would it imply changes in other parameters?

tclose:
Hi @oesteban, it would not be too hard to add an enum for the differnt switches but then there are about 3-4 options per switch that are specific to that algorithm.

I reckon the best approach would be to create a base class with the common options and then derived classes for each of the switches since they are really different commands under the hood.

effigies:

PR #1554, opened 2016-08-01
ENH: BrainSuite updates: Add interfaces for additional command line tools

jason-wg:
I have updated the BrainSuite Nipype interface to include classes for our BDP, SVReg and ThicknessPVC command line tools.


jason-wg:
@satra @chrisfilo 
Can you please take a look at my code updates, or refer this PR to someone who is available to review my code?

Thanks!

satra:
@jason-wg - quick observation - could you please try to get your code to be as close to PEP8 compliant as possible?  as an example see: #1593

jason-wg:
@satra 
I've made formatting fixes for PEP8 compliance, except that many lines run over 80 characters.
Does this look fine to you?

satra:
@jason-wg - could you please merge with current master, fix any conflicts and push?

jason-wg:
@satra 

A note regarding the failure that we see in the Travis CI build:
Per the log, a doctest fails: <code>Doctest: nipype.interfaces.brainsuite.brainsuite.BDP ... FAIL</code>
When I run the doctest to take a closer look, the failure message is:  <code>TraitError: Each element of the 'BVecBValPair' trait of a BDPInputSpec instance must be a value of type 'str', but a value of u'/directory/subdir/prefix.dwi.bvec' <type 'unicode'> was specified.</code>

For reference, the line containing the failing doctest is [here](https://github.com/nipy/nipype/blob/15e51c36d75e2a834186eb074e1774e26228cd29/nipype/interfaces/brainsuite/brainsuite.py#L1551)
[bdp.inputs.BVecBValPair](https://github.com/nipy/nipype/blob/15e51c36d75e2a834186eb074e1774e26228cd29/nipype/interfaces/brainsuite/brainsuite.py#L1088) is of type traits.List, requiring string inputs.
Based on the code in the doctest, I expect that both the items in the List are in fact strings.

I have tried to identify the cause of this failure, and I have found that when I take out the import of unicode_literals from [this line](https://github.com/nipy/nipype/blob/15e51c36d75e2a834186eb074e1774e26228cd29/nipype/interfaces/brainsuite/brainsuite.py#L2) that was recently added to my code in 632c1f1 , then the doctest runs fine.


When I add doctests for any other traits.List fields, I get the same error, <code>... instance must be a value of type 'str', but a value of ... <type 'unicode'> was specified</code>

I'm curious about why this error is appearing, so please do let me know if you have any insights on this.

As for my pull request, what do you suggest I do for my code?

satra:
@jason-wg - you will need to: `from ..traits_extension import Str`

all strings are now unicode in nipype.

jason-wg:
@satra 
Thanks for letting me know. I've added the import you suggested, and that resolved the problem I was seeing.
I'm curious, what was the motivation behind making all strings unicode in Nipype?

Regarding this pull request, does this look good to you?

satra:
@jason-wg - to support py3 where all strings are unicode.

Issue #1359, opened 2016-02-08
Simplified/alternative way to specify models


tyarkoni:
It would be nice to abstract some of the model specification details into a separate class that can handle high-level selection and filtering of regressors.

For example, consider a case where there are event files for 10 different conditions in a directory. A user might want to construct several different models, with different subsets of conditions, or even different subsets of individual events. At present, this preprocessing is left entirely up to the user (at least, I think that's the case--correct me if I've missed something). I.e., the user is expected to pass in either a Bunch of the exact names/onsets/durations (leaving them to pre-select and filter the inputs), or the exact set of event files. There's nothing wrong with this, but it can end up being quite verbose if one intends to fit many models, and clarity is reduced.

Now consider something like this:

```python
# Suppose there are files for subject 1 - 10, and conditions A - J (i.e., 100 files)
mf = ModelFactory('path/to/all/my/event_files*.txt')
# First model: conditions A, C, and E for all subjects
mod1 = mf.get_model(conditions=['A', 'C', 'E'])
# Second model: all event types for selected subjects
mod2 = mf.get_model(subjects=[1, 5, 8, 9]
# Third model: filter by time, selecting all events with onsets < 100s
mod3 = mf.get_model(query='onset < 100')
```

In each case, get_model() returns an initialized instance of class SpecifyModel that behaves normally. Nothing much changes internally, but the user gets to specify potentially complex models in just one or two lines of code, instead of having to do extensive preprocessing. This could also be implemented as a single method in modelgen.py that does it all in one shot--i.e., you pass in a list of lists, some Bunches, a pandas DataFrame, or a list of event files, and then various keyword arguments for filtering/selecting. It would return a SpecifyModel instance again.

If there's agreement that something like this would be useful, I'd be happy to give it a shot--I have a bunch of code that would be pretty easy to adapt.

tyarkoni:
On looking into this some more, I think a reasonable way to go about this would be to add a new interface called something like SpecifyEvents. This would typically slot in right before SpecifyModel. Right now SpecifyModel handles both (i) the conversion from discrete events to predictors in the design matrix, and (ii) the extraction of events from 1/2/3-column event files. The proposal is to refactor the code so that there's a clear distinction between these two. Specifically:

[...]

PR #1555, opened 2016-08-01
New SpecifyEvents interface

tyarkoni:
This is a first pass at a new interface for flexible transformation/processing of events, as discussed in #1359. It's a WIP and there are no tests yet--plus it depends on a JSON transformation spec that will need to be refined--but I'm opening a PR for discussion.

The basic idea is to introduce a new SpecifyEvents interface (meant to parallel SpecifyModel) that takes a set of event files (or a Bunch of onset/duration/amplitude) as input, performs an arbitrary set of transformations on the data (more on that below), and then outputs the new events as a Bunch compatible with the subject_info input of SpecifyModel. This is fully compatible with SpecifyModel, so users who don't need to manipulate event information in any way can simply ignore the new interface (though at some point it might be worth discussing whether SpecifyEvents should become the mandatory way to go from event files to a design matrix).

Here's a partial example of what SpecifyEvents usage looks like in context:

``` python
eventspec = pe.Node(interface=events.SpecifyEvents(), name="eventspec")
eventspec.inputs.input_units = 'secs'
eventspec.inputs.time_repetition = 3
eventspec.inputs.transformations = 'transformations.json'

# assume a standard fMRI workflow exists--pipe lists of event files to the
# event_files input, which can be either flat event files, or Bunch.
wf.connect(datasource, ('behav', check_behav_list), eventspec, 'event_files')
# Connect the output to the input of an instance of SpecifyModel
wf.connect(eventspec, 'subject_info', modelspec, 'subject_info')
```

Now suppose we have a list of conditions `["cond001", "cond002", "cond003", "cond004"]`. Then we can pass in a json transformation file that looks like this:

``` json
{
  "steps": [
    {
      "name": "orthogonalize",
      "input": [
        "cond001"
      ],
      "X_cols": [
        "cond002",
        "cond003",
        "cond004"
      ],
      "output": [
        "cond001_orth"
      ]
    },
    {
      "name": "standardize",
      "input": [
        "cond001_orth",
        "cond002",
        "cond003",
        "cond004"
      ]
    },
    {
      "name": "select",
      "input": [
        "cond001_orth",
        "cond003",
        "cond004"
      ]
    }
  ]
}
```

Here we're performing 3 different operations. First, we orthogonalize `cond001` with respect to the other 3 conditions. We rename the output `cond001_orth`. Next, we standardize all 4 columns (i.e., subtract mean and divide by SD). Finally, we decide to keep only 3 of the conditions for inclusion in the model. These will be passed as a Bunch to SpecifyModel, and any existing workflow should in principle work just fine (though this remains to be tested).

Note that there are many other implemented features not illustrated in the above example. For instance, one can multiply conditions or sets of conditions (e.g., to code interaction terms); dummy-code variable; apply grouping operations (e.g., standardize a given condition separately for each level of another condition); etc.

### DIFF
--- a/nipype/algorithms/events.py
+++ b/nipype/algorithms/events.py
@@ -130,7 +130,9 @@ class EventTransformer(object):
         self.data = func(self.data, *args, **kwargs)
 
     def _to_dense(self):
-
+        """ Convert the sparse [onset, duration, amplitude] representation
+        typical of event files to a dense matrix where each row represents
+        a fixed unit of time. """
         end = int((self.events['onset'] + self.events['duration']).max())
 
         targ_hz, orig_hz = self.target_hz, self.orig_hz
@@ -155,6 +157,11 @@ class EventTransformer(object):
         self.data.insert(0, 'onset', onsets)
 
     def resample(self, sampling_rate):
+        """
+        Resample the design matrix to the specified sampling rate. Primarily
+        useful for downsampling to match the TR, so as to export the design as
+        a n(TR) x n(conds) matrix.
+        """
         sampling_rate = np.round(sampling_rate * 1000)
         self.data.index = pd.to_datetime(self.data['onset'], unit='s')
         self.data = self.data.resample('%dL' % sampling_rate).mean()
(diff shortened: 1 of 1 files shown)


nipype/algorithms/events.py
```diff
@@ -0,0 +1,407 @@
...
+import re
+from glob import glob
+from os.path import basename
+import json
+iflogger = logging.getLogger('interface')
+
+# Eventually, move the following inside the interface, or wrap in an import check
+try:
+    import pandas as pd
+except ImportError:
```
satra: perhaps simply create a flag here: `no_pandas = True` and raise it in the `run_interface` function below

nipype/algorithms/events.py
```diff
@@ -0,0 +1,397 @@
...
+from os.path import basename
+import json
+iflogger = logging.getLogger('interface')
+
+
+have_pandas = True
+try:
+    import pandas as pd
+except:
+    have_nipy = False
```
qmac: Should this be `have_pandas`?

djarecka:
@tyarkoni - are you still working on this pr?

tyarkoni:
Not directly, no. But indirectly, yes. We're working on a pybids module that does all of this work, and  once it's done, the plan is to wrap that with a much lighter Nipype interface that mostly just passes a BIDS-Model spec through. But I'll close the PR, because the next iteration will probably be a radical departure from this code. Thanks for the reminder!

PR #1556, opened 2016-08-02
change 3dROIstats terminal output from allatonce to file

carolFrohlich:
Change 3dROIstats terminal output from "all at once" to "file".
If the terminal output is all at once, 3dROIstats hangs when psutil is installed.
Changing the terminal output solves the problem.


satra:
@carolFrohlich - this seems more like a patch than a fix. could we figure out what is it in the interaction between `all at once` and `psutil` that's causing the issue?

effigies:

PR #1557, opened 2016-08-02
Changed default for CommandLine._run_interface to not be mutable

blakedewey:
The correct_runtime_codes default is a list, which is mutable. This allows for the default to be changed in subsequent calls. Changed to a tuple


blakedewey:
Arguably, this is pedantic, but my IDE complains every time I overload _run_interface, so I thought I might correct it.

PR #1561, opened 2016-08-03
fix: remove psutils usage by default

### LINKED ISSUE #1532: psutils stopps pipeline
Hi,
I had the problem, that heudiconv -c dcm2niix was working on my office PC but not on the server. On the server the process dcm2niix was in sleepmode (status S+) and nothing happend.

Changing
convertnode.inputs.terminal_output = 'allatonce' to 'stream' solved this problem.

In addition interfaces.freesurfer.reconall stopped working too with psutils installed. No error message only a frozen output during the smoothing after approx 1h.

newest nipype, psutils 4.3.0

Joerg

satra:
closes #1532


satra:
and makes new release

PR #1562, opened 2016-08-04
BF: install/ship tests/use_resources script

yarikoptic:


yarikoptic:
actually this resolution was incorrect... sorry... this way it gets installed but not as a script, so would lack execution permissions and thus would not be directly usable and thus tests would fail...

checking solution given at http://stackoverflow.com/questions/18409296/package-data-files-with-executable-permissions  for now

PR #1565, opened 2016-08-05
BF(workaround): fix permissions for use_resources tests script during installation

yarikoptic:
I don't like it but it seems to work

ppl are saying that setuptools preserves permissions... might be worth migrating from plain distuils.setup

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -244,7 +244,29 @@ def _package_status(pkg_name, version, version_getter, checker):
         return 'low-version', have_version
     return 'satisfied', have_version
 
-cmdclass = {'build_py': get_comrec_build('nipype')}
+
+# Based on http://stackoverflow.com/a/18410671/1265472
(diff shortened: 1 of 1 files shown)


satra:
@yarikoptic -  we have got rid of `distutils` in relies directly on `setuptools`. can you see if this is still an issue, and if so if there is a `setuptools` way of doing this without using `distutils`?

oesteban:
I just tried this with a conda environment, installing from pypi and without any special umask and permissions after install are 0775.

djarecka:
@yarikoptic - are you still working on it? do we still experience some problems related to it?

djarecka:
@yarikoptic ping

oesteban:
I don't think this is necessary anymore after the refactoring of the resource monitor. Please correct me if I'm wrong.

PR #1572, opened 2016-08-09
[FIX] PY3 compatibility revisited

oesteban:


.pylintrc
```diff
@@ -0,0 +1,378 @@
+[MASTER]
```
satra: accidental inclusion of this file

oesteban:
Ok, we are growing too big here.

So, this is the situation: some parts of nipype are not python 3 compatible. The were passing unnoticed because the tests did not exercise these parts and CircleCI smoke tests were done in python 2.7.

In this PR, I changed the tests in CircleCI to run in python 3.5, with the following chain of issues:

- First problem were ```open()``` calls. I had to add a ```from builtins import open``` in all files using it.
- Now, these files would read/write unicode -> add ```from __future__ import unicode_literals```.
- Many interfaces stopped working. Particularly those using ```Directory``` and ```DictStrStr```. I fixed those, making ```Directory``` derive from ```traits.Unicode``` and ```DictStrStr``` with ```six.string_types```. I also disabled the fast C type checker for them.
- I removed the ```external``` code of ```six```. I don't see the point to maintain our version in parallel.

Finally, after too many modifications, the big issue is doctests. Python 2 will want to have a u-strings in tests, while Python 3 does not. There are two solutions:

1. Migrate to py.test (and enable ALLOW_UNICODE). However the test generators will not work, and we have plenty of those.
2. Tricks to fix the docstrings before nose:

  - https://dirkjan.ochtman.nl/writing/2014/07/06/single-source-python-23-doctests.html - Where should I place this code?
  - Use metaclasses and format the docstring before testing (the solution I'm testing right now, similar to the solution they deployed in IPython).

Any opinions, @satra?

oesteban:
@satra, @chrisfilo 

It seems to me like this PR is getting to a good endpoint. Summary of what is inside:

  - [X] A comprehensive review of code that was not totally python 3 compatible
  - [X] Migrated CircleCI tests to python 3.5
    - [ ] TODO before merge: the latest tests added by @effigies about file copy are failing
  - [X] Enabled many tests in CircleCI that were skipped for several reasons (the most important: bedpostx and eddy, skipped since I removed the fsl_course_data, now they work with a slim version of the course data)
  - [X] Fixed problem with coverage in travis (some errors showed up in the end of travis tests)

Some further work after this PR is finally merged would include:

  - Adding FreeSurfer to the docker image
  - Migrate travis tests to docker (#1546)

Regarding the failing tests in CircleCI, I can try to help @effigies tomorrow. I have invited you to the gitter channel. Let me know if that works for you :)

effigies:
Fixed test in #1586.

effigies:
Woops. CircleCI just rejected it. Looks like that wasn't quite the fix.

What's the easiest way to run a small test on CircleCI? I tried to do the SSH thing, but couldn't actually run tests because I couldn't get into a usable Python environment.

oesteban:
The easiest way to run I can think of right now is:
1. Run the build with SSH enabled and SSH when ready
2. Run the docker image in interactive mode:
```
docker run -it --rm -v /etc/localtime:/etc/localtime:ro -v ~/scratch:/scratch -w /root/src/nipype/doc nipype/nipype_test:py27
```
This will provide you the runtime container used in tests. Then you can ```cd /root/src/nipype```
and
3. Run the particular test:
``` nosetests nipype/utils/tests/test_filemanip.py```

If you modify the code outside the container, you will need to re-do the ```docker build``` line you find in the circle.yml file.

**Comment by oesteban** [comment]
Ok, so besides the [discussion about tests on symlinks](https://github.com/nipy/nipype/pull/1586#issuecomment-241067330), to help review this PR I suggest the following starting points:

  - nipype/utils/provenance.py
  - nipype/interfaces/base.py
  - nipype/interfaces/traits_ext.py
  - nipype/interfaces/filemanip.py

Most of other changes are just adding the encoding header (first line) and the following imports for backwards compatibility:
```
from __future__ import unicode_literals
from builtins import open, str, bytes
```

**Comment by effigies** [comment]
@oesteban #1586 passing. Sorry about how long that took.

**Comment by oesteban** [comment]
Thanks a lot @effigies. Don't worry it took long, this PR is going to need a great effort in reviewing so I don't expect it to be merged soon :)

**Comment by satra** [comment]
@oesteban - could you merge with master? and is this ready for review?

File: bin/fsl_imglob.py [hunk 2]
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python
+#   imglob - expand list of image filenames
+#   Stephen Smith, Mark Jenkinson and Matthew Webster FMRIB Image Analysis Group
+#   Copyright (C) 2009 University of Oxford
```
satra: why is this file being included? it's under the oxford license.

doc/sphinxext/numpy_ext/docscrape.py
```diff
@@ -13,7 +14,7 @@
 import pydoc
 from warnings import warn
 
-from nipype.external.six import StringIO
+from six import StringIO
```
satra: is six now a dependency?

nipype/caching/memory.py
```diff
@@ -293,7 +294,7 @@ def _clear_all_but(self, runs, warn=True):
             input.
         """
         rm_all_but(self.base_dir, set(runs.keys()), warn=warn)
-        for dir_name, job_names in runs.items():
+        for dir_name, job_names in list(runs.items()):
```
satra: is there a test that caught this?

nipype/interfaces/afni/preprocess.py
```diff
@@ -44,11 +53,10 @@ class BlurToFWHM(AFNICommand):
     >>> blur = afni.preprocess.BlurToFWHM()
     >>> blur.inputs.in_file = 'epi.nii'
     >>> blur.inputs.fwhm = 2.5
-    >>> blur.cmdline #doctest: +ELLIPSIS
+    >>> blur.cmdline #doctest: +ELLIPSIS +IGNORE_UNICODE
```
satra: should we ignore unicode, or simply return unicode for any cmdline.

nipype/interfaces/base.py
```diff
@@ -28,32 +26,24 @@
...
-                               has_metadata)
-from ..utils.filemanip import (md5, hash_infile, FileNotFoundError,
-                               hash_timestamp, save_json,
-                               split_filename)
-from ..utils.misc import is_container, trim, str2bool
-from ..utils.provenance import write_provenance
-from .. import config, logging, LooseVersion
-from .. import __version__
-from ..external.six import string_types, text_type
+from nipype import config, logging, LooseVersion, __version__
```
satra: why did these become absolute imports? py3 recommends relative imports for packages.

nipype/interfaces/cmtk/nx.py
```diff
@@ -10,7 +11,9 @@
 """
 
 from __future__ import division
-from builtins import range
+from __future__ import unicode_literals
+from builtins import str
+from builtins import open, range
```
satra: combine to one line

nipype/interfaces/dcm2nii.py
```diff
@@ -6,7 +7,9 @@
    >>> datadir = os.path.realpath(os.path.join(filepath, '../testing/data'))
    >>> os.chdir(datadir)
 """
-
+from __future__ import unicode_literals
+from builtins import str
+from builtins import open
```
satra: combine

nipype/interfaces/dipy/reconstruction.py
```diff
@@ -109,8 +117,8 @@ def _run_interface(self, runtime):
         # Estimate sigma required by RESTORE
         mean_std = np.median(noise_data.std(-1))
         try:
-            bias = (1. - np.sqrt(2. / (n - 1)) *
-                    (gamma(n / 2.) / gamma((n - 1) / 2.)))
+            bias = (1. - np.sqrt(old_div(2., (n - 1))) *
```
satra: `old_div` unnecessary here

nipype/interfaces/freesurfer/utils.py
```diff
@@ -736,7 +744,7 @@ def _list_outputs(self):
                 args = tuple([getattr(self.inputs, arg) for arg in stem_args])
                 stem = stem % args
         snapshots = ["%s-lat.tif", "%s-med.tif", "%s-dor.tif", "%s-ven.tif"]
-        if self.inputs.six_images:
+        if self.inputssix_images:
```
satra: seems incorrect here.

nipype/interfaces/io.py
```diff
@@ -35,18 +37,13 @@
...
-                   Undefined, BaseInterfaceInputSpec)
-from .. import config
-from ..external.six import string_types
-from ..utils.filemanip import (copyfile, list_to_filename,
-                               filename_to_list)
-from ..utils.misc import human_order_sorted
-from ..utils.misc import str2bool
-from .. import logging
-iflogger = logging.getLogger('interface')
+from nipype import config, logging
```
satra: why absolute imports?

satra:
@oesteban - this looks good. left a few comments - i think one i'd like to see reverted is how nipype sub-packages are imported inside nipype. using the `from .base` syntax makes sure that the system doesn't pick up from an installed version if one is developing.

bin/fsl_imglob.py
oesteban: Good point, I forgot to mention. This script was only python 2. But it is required by the eddy correct pipeline (and thus, it is necessary in the circleci tests).

What should we do with it? As I said, the only changes are for python 3 compatibility.

doc/sphinxext/numpy_ext/docscrape.py
oesteban: We could get rid of all dependencies from six. That'd be actually a good plan, with future and the standard_aliases is enough to make code compatible.

nipype/caching/memory.py
oesteban: I think this is one of the safe operations of running ```futurize --stage1```. So, answering you question, no, this did not come up from a test.

nipype/interfaces/afni/preprocess.py
oesteban: For python 3 this is not necessary. But, when using ```unicode_literals``` all the strings become unicode, and in python 2 ```repr(str)``` always has the ```u'``` prefix. This breaks doctests.

For this reason, to make doctests work in python 2 I added the doctest-ignore-unicode package, which requires that flag before any string doctest.

nipype/interfaces/base.py
oesteban: I had the opposite impression. Sorry about that. I will roll back all of them.

nipype/interfaces/dipy/reconstruction.py
oesteban: Yep. Actually I will revise all of them and use division from __future__.

nipype/interfaces/freesurfer/utils.py
oesteban: good catch!

oesteban:
@satra sorry about those absolute imports, I don't know where would I read that they were preferred.

satra:
@oscar - no worries - the `from .base import ...` is also called absolute import i believe. the terminology gets a little confusing, but yes python3 does not like `from base import ...` in case base is also an installed package.

shoshber:
Here is the PEP-8. This is more complicated than I had thought. It seems generally:

`from pkg.base import thing` > `from .base import thing` > `import thing`

but both the first too options are acceptable.

https://www.python.org/dev/peps/pep-0008/#imports

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path ):
> 
> import mypkg.sibling
> from mypkg import sibling
> from mypkg.sibling import example
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
> 
> from . import sibling
> from .sibling import example
> Standard library code should avoid complex package layouts and always use absolute imports.
> 
> Implicit relative imports should never be used and have been removed in Python 3.

satra:
@shoshber - thanks for the clarification.

@oesteban - i'm fine with keeping the absolute imports.

oesteban:
It's fine, I think the complexity of nipype justifies the relative imports. We miss out some readability but gain on control w.r.t. the installed version in development environments.

bin/fsl_imglob.py
oesteban: @satra let me know how should we do with this file

oesteban:
Ok, this is getting closer. Open issues:

  - One test is failing with python 3.4 in travis, need to find out the problem.
  - What to do with the FSL file?
  - I'm seeing this error when building the documentation (only python 2.7). Can we dismiss it?:

    ```
    OtsuThresholdImageFilter
      File "../tools/build_interface_docs.py", line 55, in <module>
    docwriter.write_api_docs(outdir)
      File "/home/oesteban/workspace/nipype/tools/interfacedocgen.py", line 472, in write_api_docs
    self.write_modules_api(modules, outdir)
      File "/home/oesteban/workspace/nipype/tools/interfacedocgen.py", line 439, in write_modules_api
       api_str = self.generate_api_doc(m)
      File "/home/oesteban/workspace/nipype/tools/interfacedocgen.py", line 311, in generate_api_doc
     ad += trim(classinst.help(returnhelp=True),
      File "/home/oesteban/workspace/nipype/nipype/interfaces/base.py", line 1618, in help
    allhelp = super(CommandLine, cls).help(returnhelp=True)
      File "/home/oesteban/workspace/nipype/nipype/interfaces/base.py", line 773, in help
    docstring = trim(cls.__doc__).split('\n') + ['']
      File "/home/oesteban/workspace/nipype/nipype/utils/misc.py", line 63, in trim
      return '\n'.join(trimmed)
   UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 8: ordinal not in range(128)
   ```

  - tests, does this sound good?:

    > - CircleCI: 17608 tests + 25 SKIP
    > - Travis:
    >  - py27, with deps: 17581 + 37 SKIP
    >  - py27, no deps: 16457 + 80 SKIP
    >  - py34, with deps: 17616 + 37 SKIP
    >  - py34, no deps: 16492 + 80 SKIP
    >  - py35, with deps: 17616 + 37 SKIP
    >  - py35, no deps: 16492 + 80 SKIP

bin/fsl_imglob.py
satra: two things:

1. we should send the patches back to fsl. their new release will be based on py3, so hopefully they have patched this as well.

2. we don't bundle with nipype but download from an external location till it's fixed in fsl. or include in nipype.externals and put a check in the test to use this.

putting in `bin` makes this available to users on their path, and they could be using this version without knowing.

satra:
this is because that docstring has non ascii characters:

cat -v slicer/legacy/filtering.py
```
...
N.Otsu, <E2>M-^@M-^X<E2>M-^@M-^XA threshold selection method from gray level histograms,<E2>M-^@M-^Y<E2>M-^@M-^Y IEEE Trans.Syst.ManCybe
rn.SMC-9,62<E2>M-^@M-^S66 1979.
...
```

for the moment we could simply fix this manually. but this issue also comes up also when matlab crashes and the exception handling of matlab also returns weird bytes. since py3 supports unicode we are ok, but run into trouble with py2.

oesteban:
Hi @satra, yes, this problem is related to the automatic generation of those interfaces (slicer, semtools, etc).

Could you give me some pointers on how these are built?. I'm pretty sure there is some lost open here or there which was not imported from builtins.

satra:
it's done with `.../interfaces/slicer/generate_classes.py` and applies to semtools and slicer.

oesteban:
@satra yes, but what is the origin of the templates that generate_classes uses?

satra:
the templates are inside generate classes, but gets populated with information from the `tool --xml` command.

satra:
that's why for now, perhaps we can hand fix this and then try to send information upstream.

oesteban:
@satra, you mean that:

  1. You need to have installed semtools and slicer (if there are any other packages I should list here, please let me know).
  2. Then you run a certain command that will take those ```--xml``` outputs and generate the inputs.

What is that command?, should I also include this generation of interfaces in the circleci?

oesteban:
Ok, let's leave the problem of auto-generated interfaces for another time. It should work well with the current version of code, since the files are to be written correctly encoded.

Before merging this, it is important to check on these:

  - [ ] File with FSL's license
  - [ ] Confirm that the number of tests and skipped tests are OK

satra:
@oesteban - tests are ok

oesteban:
Can we relicense the FSL file?

satra:
FSL licenses are very particular. but we should be able to use it during testing FSL. anyone who uses FSL using nipype is agreeing to use FSL with its license. i just didn't want a version in the bin directory which gets copied over and users don't know which version of this file they are using. if it's added in external, and we add external to the PATH, that would work.

oesteban:
-> FSL file moved to external.

This PR needs deep scrutiny before signing off. 
Do you want a second round, @satra? Can I ask you @shoshber and/or @chrisfilo for second opinions before merging? Of course, anyone not mentioned here is also very welcome to leave feedback :)

Thank you all.

satra:
@oesteban - thanks for this valiant effort - LGTM

nipype/caching/memory.py
oesteban: @satra - you meant here that we should have a little test to exercise this section of the code?

oesteban:
@satra - great thanks!

@shoshber - any thoughts? We can chat today about this, if you could make some room for it.

docker/nipype_test/Dockerfile_py35
```diff
@@ -42,21 +42,27 @@ ENV PATH /usr/local/miniconda/bin:$PATH
...
 COPY . /root/src/nipype
 
-RUN cd /root/src/nipype && \
+RUN rm -r /usr/local/miniconda/lib/python3.5/site-packages/nipype* && \
+    cd /root/src/nipype && \
     pip install -r requirements.txt && \
-    pip install -e .
-
+    pip install -e . && \
+    make specs
```
chrisfilo: This defies the purpose of autotests (making changes to the API visible) @satra

chrisfilo:
- Please reanable AWS tests
- build py27 image
- run nosetests in py27 (in addition to py35)
- run fmri_spm_nested MultiProc /root/examples/ level1 in py27 and py35 (both multiproc, remove the linear run, each container should have 2 cpus)

oesteban:
Ok, with the new additions to the tests we've discovered a couple of things:

- SPM fails in py27 because of unicode characters (will fix)
- Pickled files are not compatible between python 2 and 3 (meaning, if you run the workflow in one version and then you try to reuse the cached results with the other).

oesteban:
Ok, I have enabled some more tests (three from the runtime profiler), and now tests are run on both py27 and py35. Also, we are exercising MultiProc in both 2.7 and 3.5 for the fmri_spm_nested example.

It is a bit weird, but when I run it locally it does not skip the aws tests. In circle CI it does. I'm investigating this. Any ideas @chrisfilo ?

oesteban:
Hey @satra, @chrisfilo, I thought it'd be interesting to add some documentation with this PR. What do you think of this first draft for the addition https://241-17021985-gh.circle-artifacts.com/0//home/ubuntu/docs/devel/testing_nipype.html ?

satra:
@oesteban - sounds good.

oesteban:
this is finally ready!

satra:
the resource profiler should be disabled in current master

satra:
shall we merge this? it seems like a lot of things are dependent on this now.

oesteban:
@satra - I think so. Regarding the resource profiler: I will investigate the issue. Do you think that could be the explanation for MultiProc taking so much longer?

circle.yml
```diff
@@ -14,41 +14,49 @@ dependencies:
...
-    - docker run -v /etc/localtime:/etc/localtime:ro -v ~/examples:/root/examples:ro -v $(pwd)/scratch:/scratch -w /scratch nipype/nipype_test:py27 /usr/bin/run_examples.sh fmri_spm_nested Linear /root/examples/ level1
-    - docker run -v /etc/localtime:/etc/localtime:ro -v ~/examples:/root/examples:ro -v $(pwd)/scratch:/scratch -w /scratch nipype/nipype_test:py27 /usr/bin/run_examples.sh fmri_spm_nested Linear /root/examples/ l2pipeline
-    - docker run -e FSL_COURSE_DATA="/root/examples/nipype-fsl_course_data" -v /etc/localtime:/etc/localtime:ro -v ~/scratch:/scratch -w /scratch nipype/nipype_test:py27 /usr/bin/run_nosetests.sh :
-        timeout: 2600
+    - docker run -v /etc/localtime:/etc/localtime:ro -v ~/examples:/root/examples:ro -v ~/scratch:/scratch -w /scratch nipype/nipype_test:py35 /usr/bin/run_examples.sh fmri_fsl_reuse Linear /root/examples/ level1_workflow
+    - docker run -v /etc/localtime:/etc/localtime:ro -e NIPYPE_NUMBER_OF_CPUS=4 -v ~/examples:/root/examples:ro -v ~/scratch:/scratch -w /scratch nipype/nipype_test:py27 /usr/bin/run_examples.sh fmri_spm_nested MultiProc /root/examples/ level1
```
chrisfilo: Where is the NIPYPE_NUMBER_OF_CPUS variable used?

oesteban: https://github.com/oesteban/nipype/blob/fix/Py3UseBuiltinOpen/tools/run_examples.py#L23

chrisfilo: I see!

oesteban: It was the way to set it in run_examples.py with the least changes.

Issue #1569, opened 2016-08-08
ANTs registration initial_moving_transform requires invert_initial_moving_transform


ashgillman:
When I run `Registration` using `intial_moving_transform`, I get an error:

```
[160808-21:37:39,795 workflow INFO:
	 Executing node _registration_lvl21 in dir: /home/ash/proj/pylx/align/registration_lvl2/mapflow/_registration_lvl21
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nix/store/jn78f3vv1w96xlhhq0ps3n0xsfiizxv0-python3.4-nipype-0.12.0-rc1/lib/python3.4/site-packages/nipype/pipeline/engine/nodes.py", line 392, in run
    self._run_interface()
  File "/nix/store/jn78f3vv1w96xlhhq0ps3n0xsfiizxv0-python3.4-nipype-0.12.0-rc1/lib/python3.4/site-packages/nipype/pipeline/engine/nodes.py", line 502, in _run_interface
    self._result = self._run_command(execute)
  File "/nix/store/jn78f3vv1w96xlhhq0ps3n0xsfiizxv0-python3.4-nipype-0.12.0-rc1/lib/python3.4/site-packages/nipype/pipeline/engine/nodes.py", line 618, in _run_command
    cmd = self._interface.cmdline
  File "/nix/store/jn78f3vv1w96xlhhq0ps3n0xsfiizxv0-python3.4-nipype-0.12.0-rc1/lib/python3.4/site-packages/nipype/interfaces/base.py", line 1415, in cmdline
    allargs = self._parse_inputs()
  File "/nix/store/jn78f3vv1w96xlhhq0ps3n0xsfiizxv0-python3.4-nipype-0.12.0-rc1/lib/python3.4/site-packages/nipype/interfaces/base.py", line 1650, in _parse_inputs
    arg = self._format_arg(name, spec, value)
  File "/nix/store/jn78f3vv1w96xlhhq0ps3n0xsfiizxv0-python3.4-nipype-0.12.0-rc1/lib/python3.4/site-packages/nipype/interfaces/ants/registration.py", line 816, in _format_arg
    do_invert_transform = int(self.inputs.invert_initial_moving_transform)
TypeError: int() argument must be a string, a bytes-like object or a number, not '_Undefined']
```

[...]

PR #1577, opened 2016-08-11
Fix issue #1569

blakedewey:
Changed method for setting default for invert_initial_moving_transfor…m and initial_moving_transform_com for Registration.
Changed try/except loop on integer values to a check for defined inputs and added a test.


blakedewey:
@satra I don't know why this is failing CircleCI. It seems doc tests are failing, but on a lot of interfaces.

ashgillman:
FYI, #1569

PR #1582, opened 2016-08-16
fix: slurm job submission and result checking

satra:


dalejn:
Hi Satra,

There's still an issue with jobid_re looking for a string beginning with "Submitted batch job " 
Deleting that part of the string in line 41 of slurm.py allows it to run without problems for me

Best,
Dale

satra:
@dalejn - did you add your own custom expression:

`wf.run('SLURM', plugin_args={'jobid_re': '([0-9]*)'})`

dalejn:
Ah, I see-- I didn't.
I've restored your original fixes and included the custom expression, and it runs without any issues

Thanks

PR #1585, opened 2016-08-18
Fix #1584 color graphing to loop through blue, red, green.

### LINKED ISSUE #1584: Issues with color plotting for deep workflows
For deeply nested workflows (5 deep in my case), colors don't seem to operate as intended.

Sometimes (but perplexingly, not always) an error is thrown when `Workflow._get_dot` indexes beyond the length of `colorset`.

ashgillman:
suggested to Fix #1584

Colors now cycle, yellow for top-level, then B, R, G, B, R, G...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,12 @@ before_install:
   if $INSTALL_DEB_DEPENDECIES; then sudo ln -s /run/shm /dev/shm; fi &&
   bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh) &&
   sudo apt-get -y update &&
-  sudo apt-get -y install xvfb fusefat &&
+  sudo apt-get -y install xvfb fusefat graphviz &&
   if $INSTALL_DEB_DEPENDECIES; then travis_retry sudo apt-get install -y -qq
   fsl afni elastix fsl-atlases; fi &&
   if $INSTALL_DEB_DEPENDECIES; then
     source /etc/fsl/fsl.sh;
-    source /etc/afni/afni.sh; 
+    source /etc/afni/afni.sh;
     export FSLOUTPUTTYPE=NIFTI_GZ; fi }
 - travis_retry bef_inst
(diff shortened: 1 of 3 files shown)


ashgillman:
I realised this PR was very unclear, so I have added failing tests demonstrating the issue which now pass.

ashgillman:
Whoops, looks like CI test don't have dot. Is there a way I can test producing the dot file without running graphviz to produce a png?
Actually, I thought dot was only run if it was present on the system?

oesteban:
@ashgillman - apparently, the problem only affect travis. It makes sense since I think I included graphviz in the docker image (which is what circleci uses).

You'll need to install graphviz, in the .travis.yml file.

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
   conda config --add channels conda-forge &&
   conda update --yes conda &&
   conda update --all -y python=$TRAVIS_PYTHON_VERSION &&
-  conda install -y nipype &&
+  conda install -y nipype graphviz &&
```
oesteban: I think that should go here: https://github.com/ashgillman/nipype/blob/4c3a232b5d70da3638ff54dea173bcf3adf55acb/.travis.yml#L22 as an apt package.

ashgillman: Ah, yes that would make more sense wouldn't it

ashgillman:
Is the remaining error due to my code or a run time thing? I can't really tell

chrisfilo:
I have restarted the build hopefully the error will go away. We are working on a fix.

ashgillman:
Looks to have times out for Py27 (but not Py35) at test

    nipype.pipeline.plugins.tests.test_callback.test_callback_multiproc_exception('f_node', 'f_node') ... ok

Not sure if this test is problematic or whether that test just happened to be running during the timeout

ashgillman:
Also I don't know if the Travis has rerun since I updated travis.yml

PR #1570, opened 2016-08-08
FIX: Fix symlink test in copyfile

effigies:
Introduced #1391.

PR #1586, opened 2016-08-19
TEST: Do not test copy idempotence for unlinked files

effigies:
Not actually a bug, since there's no harm in what we were doing, but the test from #1570 caught unexpected behavior. Seen in [test failure](https://circleci.com/gh/effigies/nipype/79) for #1379.

The general rule when creating symlinks in `copyfile` is: if the new symlink will point to the same place as the old, do not replace. When creating a symlink from a symlink, though, it does not dereference the source symlink to check whether the end reference would be the same, and so always removes and recreates the destination symlink.

This brings the behavior with a symlink source file in line with behavior for a regular source file.

Separated the test to be more thorough and added error messages for debugging.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,7 +3,7 @@ Upcoming release 0.13
 
 * TST: reduce the size of docker images & use tags for images (https://github.com/nipy/nipype/pull/1564)
 * ENH: Implement missing inputs/outputs in FSL AvScale (https://github.com/nipy/nipype/pull/1563)
-* FIX: Fix symlink test in copyfile (https://github.com/nipy/nipype/pull/1570)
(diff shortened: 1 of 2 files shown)


effigies:
Okay, so my understanding was off. Symlinking symlinks was valid, so checking realpaths may not be correct.

That said, it's unclear to me why, on CircleCI (this hasn't been an issue on my local machines) the symlink to the symlink was being deleted and re-created. I'm trying to track that down.

It may simply be we don't care and should just skip on that case. Thoughts?

effigies:
Sorry. Screwed up again. The problem wasn't symlinks at all, but that copying could change the timestamp, which one of the `hashmethod`s uses to decide whether two files are the same. Reverted the symlink changes, checked both `hashmethod`s, and put in an exception for that case.

This should be ready to go, but I suppose we'll see whether I messed up yet again.

PR #1589, opened 2016-08-24
Add out_prefix input to interfaces.spm.Normalize12

mgclark:
This change allows the user to set the woptions.prefix field of an SPM12 normalization routine.


satra:
@mgclark - could you please do a `make specs` and add the auto test that this updates?

Issue #1272, opened 2015-11-13
SPM: NewSegment warping_regularization is a vector


alexsavio:
Hi,

The warping_regularization parametrization in SPM12 NewSegment should be a vector, right?

This is the piece of the job that I use from the GUI:

'''
....
matlabbatch{1}.spm.spatial.preproc.tissue(6).ngaus = 2;
matlabbatch{1}.spm.spatial.preproc.tissue(6).native = [0 0];
matlabbatch{1}.spm.spatial.preproc.tissue(6).warped = [0 0];
matlabbatch{1}.spm.spatial.preproc.warp.mrf = 1;
matlabbatch{1}.spm.spatial.preproc.warp.cleanup = 1;
**matlabbatch{1}.spm.spatial.preproc.warp.reg = [0 0.001 0.5 0.05 0.2];**
matlabbatch{1}.spm.spatial.preproc.warp.affreg = 'mni';
matlabbatch{1}.spm.spatial.preproc.warp.fwhm = 0;
matlabbatch{1}.spm.spatial.preproc.warp.samp = 3;
matlabbatch{1}.spm.spatial.preproc.warp.write = [1 1];
'''

PR #1590, opened 2016-08-25
[Issue #1272] Input a list of warping regularization parameters to interfaces.spm.NewSegment

mgclark:
Allows the user to submit a list of warping regularization parameters to an spm.NewSegment routine, which appears to be required for some SPM12 workflows. This is a fix for Issue #1272.

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -849,7 +849,7 @@ class NewSegmentInputSpec(SPMCommandInputSpec):
                                            field='warp.reg',
                                            desc=('Warping regularization '
                                                  'parameter(s). Accepts float '
-                                                 'or list of floats (the
+                                                 'or list of floats (the '
                                                  'latter is required by '
                                                  'SPM12)')
     sampling_distance = traits.Float(field='warp.samp',
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -843,10 +843,14 @@ class NewSegmentInputSpec(SPMCommandInputSpec):
                           field='tissue')
     affine_regularization = traits.Enum('mni', 'eastern', 'subj', 'none', field='warp.affreg',
                                         desc='mni, eastern, subj, none ')
-    warping_regularization = traits.Float(field='warp.reg',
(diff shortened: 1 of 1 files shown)


nipype/interfaces/spm/preprocess.py
```diff
@@ -845,6 +845,8 @@ class NewSegmentInputSpec(SPMCommandInputSpec):
                                         desc='mni, eastern, subj, none ')
     warping_regularization = traits.Float(field='warp.reg',
                                           desc='Aproximate distance between sampling points.')
+    warping_regularization_12 = traits.List(traits.Float(), minlen=5, maxlen=5, field='warp.reg',
```
satra: this should likely be either or with warping_regularization. a couple of alternatives:

you can make warping regularization use `traits.Either(traits.Float, traits.List(traits.Float, minlen, maxlen), desc= ...)`

that may keep the interface cleaner. in the description you can add that the list of 5 elements is only supported in SPM 12.

mgclark: Great suggestion. Added & committed.

nipype/interfaces/spm/preprocess.py
```diff
@@ -843,8 +843,14 @@ class NewSegmentInputSpec(SPMCommandInputSpec):
                           field='tissue')
     affine_regularization = traits.Enum('mni', 'eastern', 'subj', 'none', field='warp.affreg',
                                         desc='mni, eastern, subj, none ')
-    warping_regularization = traits.Float(field='warp.reg',
-                                          desc='Aproximate distance between sampling points.')
+    warping_regularization = traits.Either(traits.Float(),
+                                           traits.List(traits.Float(),
+                                                       minlen=5, maxlen=5),
+                                           field='warp.reg'
+                                           desc='Warping regularization \
```
satra: instead of the `\` you can simply create a tuple

```
desc=('Warping regularization parameter(s). '
      'Accepts float or list of floats (the '
      'latter being required by SPM12)')
```

mgclark:
Still new to the testing infrastructure. Unsure why CircleCI is failing. Looks like lots of isinstance() TypeFailures?

mgclark:
That worked! Tests passing.

Issue #1578, opened 2016-08-14
Convenient load of settings from JSON file at interface creation


oesteban:
It'd be great if we could create interfaces in this fashion:

```
reg = ants.Registration(from_file='settings.json')
```

where settings.json contains, for instance, something in this vein:

```
{
    "dimension": 3,
    "number_of_iterations": [ [ 50 ], [ 20 ] ],
    "use_histogram_matching": [ true, true ],
    "sampling_percentage": [ 0.2, 0.2 ],
    "metric": [ "Mattes", "Mattes" ],
    "winsorize_lower_quantile": 0.015,
    "metric_weight": [ 1.0, 1.0 ],
    "shrink_factors": [ [ 4 ], [ 1 ] ],
    "radius_or_number_of_bins": [ 64, 64 ],
    "winsorize_upper_quantile": 0.998,
    "sampling_strategy": [ "Random", "Random" ],
    "initial_moving_transform_com": 0,
    "convergence_threshold": [ 1e-07, 1e-08 ],
    "collapse_output_transforms": true,
    "convergence_window_size": [ 20, 10 ],
    "sigma_units": [ "mm", "mm" ],
    "transform_parameters": [ [ 0.5 ], [ 0.1 ] ],
    "transforms": [ "Translation", "Rigid" ],
    "use_estimate_learning_rate_once": [ true, true ],
    "smoothing_sigmas": [ [ 10.0 ], [ 2.0 ] ]
}
```

These parameters would nicely act as defaults for the interface, enabling to test an interface without too much verbosity and modifying one parameter at a time.

WDYT?

oesteban:
@satra @chrisfilo any objections to this?

satra:
@oesteban - as long as this is used to simply update the parameters of the interface it would be great. i would like to stay away from config files as alternate inputs (for example, like FNIRT). if all you are thinking is writing out or reading in the inputs from a json file, that would be completely fine with me. since the nipype inputs structure is a dictionary it should be easy to add that to inputspec base class, so you can do:

```
reg.load_inputs_from_json('parameter_set.json')
reg.save_inputs_from_json('parameter_set.json')
```

perhaps with better function names!

oesteban:
This should be relatively easy and the aim clear enough. Do you want to take on it, @shoshber ? Can I help you to touch base on it easy?

shoshber:
I don't see why not, though @satra, can you clarify what you mean by "config files as alternate inputs"? I have never used FNIRT.

[...]

PR #1591, opened 2016-08-26
[ENH] Convenient load/save of interface inputs

oesteban:
Implements #1578


nipype/interfaces/base.py
```diff
@@ -1148,6 +1155,34 @@ def version(self):
                                  self.__class__.__name__)
         return self._version
 
+    def load_inputs_from_json(self, json_file, overwrite=True):
+        """
+        A convenient way to load pre-set inputs from a JSON file.
+        """
+
+        with open(json_file) as fhandle:
+            inputs_dict = json.load(fhandle)
```
satra: this may be sufficient except for checks:

`self.inputs.update(**inputs_dict)`

nipype/interfaces/base.py
```diff
@@ -1148,6 +1155,34 @@ def version(self):
...
+                continue
+            val = getattr(self.inputs, key)
+            if overwrite or not isdefined(val):
+                setattr(self.inputs, key, newval)
+
+    def save_inputs_to_json(self, json_file):
+        """
+        A convenient way to save current inputs to a JSON file.
+        """
+        inputs = self.inputs.get()
```
satra: `input_dict = self.inputs.get_traitsfree()` returns a dictionary after removing all undefined traits.

nipype/interfaces/base.py
oesteban: mmm, I'm getting object does not have update. (?)

Anyways, I will use sets and get_traitsfree to implement this in a cleaner manner.

oesteban:
Please, do not merge before #1572 .

chrisfilo:
This seems like a great addition that a lot of people will appreciate. Could you add a brief mention of it to the documentation and/or use it in an example?

oesteban:
Sure thing :)

oesteban:
@chrisfilo : updated documentation here https://github.com/oesteban/nipype/blob/3f2612a46f27964e817d8f52122e88512b3c0d77/doc/users/interface_tutorial.rst

oesteban:
Added the use of the feature in the ANTs example.

oesteban:
Ok, let me fix the rendering of the example: https://297-17021985-gh.circle-artifacts.com/0//home/ubuntu/docs/users/examples/smri_ants_registration.html, and some minimal details in the docs: https://297-17021985-gh.circle-artifacts.com/0//home/ubuntu/docs/users/interface_tutorial.html

And this is ready to merge then.

oesteban:
Ok, I have generated them locally and they look good. If they look also fine as circle artifacts, then we are ready to merge.

PR #1596, opened 2016-08-31
A few fixes to bin scripts and a proposal of a new one

alexsavio:
- fix nipype_crash_search to print usage on no args.
- change nipype_crash_search shebang to #!python.
- add nipype_display_pklz to look into .pklz node files.

### DIFF
--- a/bin/nipype_crash_search
+++ b/bin/nipype_crash_search
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!python
 """Search for tracebacks inside a folder of nipype crash
 log files that match a given regular expression.
 
--- a/bin/nipype_crash_search
+++ b/bin/nipype_crash_search
@@ -6,6 +6,7 @@ Examples:
 nipype_crash_search -d nipype/wd/log -r '.*subject123.*'
 """
 import re
+import sys
 import os.path as op
 from glob import glob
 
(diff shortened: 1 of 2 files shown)


bin/nipype_crash_search
```diff
@@ -1,11 +1,12 @@
-#!/usr/bin/env python
+#!python
```
satra: could you please explain the rationale for this change?

alexsavio: "#!python" is being used in the other scripts, e.g., nipype_display_crash.
I tought about it and I am not sure if "#!/usr/bin/env python" works on Windows (if there is any Windows user here).

satra:
@alexsavio - i'll merge this, but i'm noting here that one thing we should do is have a single nipype command tool with subcommands, like aws, singularity, docker, etc.,.

alexsavio:
Hi Satra, thanks for the note.
I could do the subcommands for a `nipype` command. Would you have a dependency like `click` to make things easier or would you rather make something without dependencies?
I see something like:
`nipype display_crash` instead of `nipype_display_crash`,
`nipype crash_search` instead of `nipype_crash_search`,
`nipype run` instead of `nipype_cmd`, and
`nipype convert` instead of `nipype2boutiques`
?

satra:
let's go with `click` for now to start a refactor.

also perhaps `nipype display` and `nipype search`, which could be augmented later

alexsavio:
I have a first working approach. I moved `nipype_display_crash` to `nipype crash`,  `nipype_crash_search` to  `nipype search`,  and  `nipype_show_pklz` to `nipype show`.
nipype_cmd will be more involved since I have to process argv.
nipype2boutiques is straightforward, however I can't think on how to name it.

satra:
perhaps `nipype convert --format=boutiques`

PR #1598, opened 2016-08-31
[ENH] Minimize nipype test docker image

oesteban:
With all the meat in the base image


oesteban:
This adds a version number in the base docker-image tag (concretely 0.0.1), so we prevent those glitches when tests do not pass in circle but they pass in a local docker engine. I think this could be merged now, to introduce the docker image versioning. Then, we would merge #1572 when tests finally pass (circleCI is now running and travis are fixed). The docker image tag in #1572 is version 0.0.2, so it is already meant to keep the version tracking. WDYT @satra?

satra:
thanks @oesteban

Issue #1594, opened 2016-08-29
New CompCor interface


oesteban:
Create a new interface (under algorithm.misc ?) for the calculation of the [Component Based Noise Correction Method (CompCor)](http://dx.doi.org/10.1016%2Fj.neuroimage.2007.04.042) on r-fMRI.

The code is already here in nipype: 
  - https://github.com/nipy/nipype/blob/master/nipype/workflows/rsfmri/fsl/resting.py
  -  https://github.com/nipy/nipype/blob/c389d5d4c0f1eb5ff321de5553f815d1fb6fcc47/examples/rsfmri_vol_surface_preprocessing_nipy.py

So this issue would include updating both workflow and example to use the new interface.

- [x] implement
   - [x] CompCor
      - [x] remove regression residuals
   - [x] tCompCor
      - [x] threshold param
   - [x] aCompcor alias
- [x] use TSNR
   - [x] refactor TSNR
   - [x] move TSNR back into misc.py
   - [x] use in compcor.py
- [x] use interface in https://github.com/nipy/nipype/blob/master/nipype/workflows/rsfmri/fsl/resting.py
- [ ] use interface in https://github.com/nipy/nipype/blob/c389d5d4c0f1eb5ff321de5553f815d1fb6fcc47/examples/rsfmri_vol_surface_preprocessing_nipy.py
   - [x] refactor
   - [ ] runthrough
- [x] fix ups
   - [x] doctest
   - [x] use tempfile in test_tsnr.py
   - [x] explicit relative imports
   - [x] boolean flag for regress_poly

chrisfilo:
Big thumbs up!

chrisfilo:
It would be a "pure python" interface similar to https://github.com/nipy/nipype/blob/master/nipype/algorithms/icc.py#L25

satra:
the function implemented in the resting workflow assumes that something else computes the voxel mask over which the PCA is implemented. this allows the `a` and `t` pieces to be taken care of by other functions/workflows. for example, `fast` is used to get the tissue classes. i would keep that separation for this interface. you can then build a sequence of nodes for aCompCor and tCompCor.

shoshber:
are you saying that computing the `t` mask (top 2% tSTDs) is already implemented somewhere? That would be great

shoshber:
[...]

PR #1599, opened 2016-09-01
ENH CompCor

shoshber:
Putting this up early


shoshber:
Yeah, I wanted to ask what the best way to create test data/if I should create test data is. The ones I have in there are temporary.

nipype/algorithms/compcor.py
```diff
@@ -41,16 +43,16 @@ def _run_interface(self, runtime):
...
-            components = np.hstack((components, u[:, :num_components]))
+            components = np.hstack((components, u[:, :self.inputs.num_components]))
 
         components_file = os.path.join(os.getcwd(), 'noise_components.txt')
         np.savetxt(components_file, components, fmt="%.10f")
-        return components_file
+        return runtime
 
-class aCompCor(Node):
+class aCompCor(Workflow):
```
chrisfilo: Shouldn't this be BaseInterface?

nipype/algorithms/compcor.py
```diff
@@ -22,6 +22,15 @@ class CompCoreOutputSpec(TraitedSpec):
 class CompCore(BaseInterface):
     '''
     Node with core CompCor computation, used in aCompCor and tCompCor
+
+    Example
+    -------
+
+    >>> ccinterface = CompCore()
```
chrisfilo: You need to add this on top of the file: https://github.com/nipy/nipype/blob/master/nipype/interfaces/afni/preprocess.py#L5 for the doctests to run independently of CWD

shoshber: couldn't get that to work :(

chrisfilo: I don't see the file header I mention in my previous comment it your code. Did you try adding it? Also notice that the after adding the header the input files should be treated as if they were in the CWD: https://github.com/nipy/nipype/blob/master/nipype/interfaces/afni/preprocess.py#L45

shoshber: yes, I tried that earlier. It didn't work, I tried for ~10 minutes to get it to work, then decided it wasn't really worth it for now. Should I consider it worth it for now?

nipype/algorithms/compcor.py
shoshber: Like @satra 's comment on the issue (https://github.com/nipy/nipype/issues/1594), I am proposing to write an interface containing the core CompCor logic. tCompCor and aCompCor would be workflows that contain nodes for creating the masks, a node for the core CompCor logic, and possibly (?) nodes for applying the extracted noise components. What do you think?

chrisfilo: I see... Workflows go here:
https://github.com/nipy/nipype/tree/master/nipype/workflows and follow a
slightly different pattern (they are functions returning a workflow object
rather than sublasses of Workflow).

I personally would just create two Interfaces: aCompCor(BaseInterface) that
takes a mask, calculates PCA and return the components, and
tCompCor(aCompCor) that takes a percentile argument n, creates a mask that
includes only the n% most variable voxels and reuses the code from its
parent aCompCor to return components. This is as flexible as the workflows,
but easier to integrate with pipelines as well as implement. Let me know
what do you think.

On Fri, Sep 2, 2016 at 10:16 PM, Shoshana Berleant <notifications@github.com

nipype/algorithms/compcor.py
```diff
@@ -0,0 +1,81 @@
...
+class CompCoreInputSpec(BaseInterfaceInputSpec):
+    realigned_file = File(exists=True, mandatory=True, desc='already realigned brain image (4D)')
+    mask_file = File(exists=True, mandatory=True, desc='mask file that determines ROI (3D)')
+    num_components = traits.Int(6, usedefault=True) # 6 for BOLD, 4 for ASL
+    # additional_regressors??
+
+class CompCoreOutputSpec(TraitedSpec):
+    components_file = File(desc='text file containing the noise components', exists=True)
+
+class CompCore(BaseInterface):
```
satra: @shoshber - checking if this is a pun intended class name :) we should use the name from the paper just for searchability purposes. also see how doi's are now included to be able to generate a citation list for an interface.

mwaskom: Not the best place to mention this (I can make a separate issue) but is there a keyword search functionality in nipype? In other words it would be great if interfaces had keywords (like "motion correction", "brain masking"...it would be important to choose them carefully!) and then there was some way to search through them to find an interface that does what you want. Seems relevant here as it is best to have the name match what's in the paper, but then that name isn't actually all that descriptive about what the interface does!

chrisfilo:
1. Please rebase or  merge master to fix tests running.
2. Please get the doctests to run
3. I would suggest renaming CompCore to aCompCor and creating a child class tCompCor that takes percentile argument and does the variance based CompCor (tCompCor) reusing code from it's parent class.

nipype/algorithms/compcor.py
shoshber: Yes it's a pun :) I'll change it.

re doi's, is this what you're referring to? https://github.com/nipy/nipype/issues/1464

chrisfilo: Yup! It's just a question of adding `references_` field such as here:
https://github.com/nipy/nipype/blob/0f2cb9fce47dc10cfb29be7002193481c876f8e0/nipype/interfaces/spm/base.py#L239

oesteban:
I have the impression that absolute imports are messing up with the installed and the development versions in travis. I guess the easiest way to solve that is using explicit absolute imports.

chrisfilo:
I had a chat with @satra (we are both at #NHW16) about the temporal denoising. We think the best way to go forward is to add `regress_poly` (implemented like in https://github.com/nipy/nipype/blob/master/nipype/algorithms/misc.py#L312) to CompCor with defaults reflecting what is in the paper (I believe it's order 1 (linear) for aCompCor and order 2 (linear + qaudratic) for tCompCor, but please double check). It might be worthwhile to refactor tSNR to get the polynomial regression function shared between the interfaces.

shoshber:
@oesteban The travis tests are fixed. Did you mean absolute relative imports?

I've updated the issue to encompass the changes suggested by @chrisfilo

oesteban:
@shoshber it's sweet that tests are passing, after all it was the mask file right?. For the imports, I meant explicit relative imports (https://github.com/nipy/nipype/pull/1572#issuecomment-242160348) sorry about the mistake. But yeah, we should be consistent with that decision.

nipype/algorithms/compcor.py
```diff
@@ -0,0 +1,115 @@
...
+        # "Voxel time series from the noise ROI (either anatomical or tSTD) were
+        # placed in a matrix M of size Nxm, with time along the row dimension
+        # and voxels along the column dimension."
+        # voxel_timecourses.shape == [nvoxels, time]
+        M = voxel_timecourses.T
+        numvols = M.shape[0]
+        numvoxels = M.shape[1]
+
+        # "The constant and linear trends of the columns in the matrix M were removed ..."
+        timesteps = range(numvols)
```
satra: you may want to use legendre polynomials for this as shown here:

https://github.com/nipy/nipype/blob/master/nipype/algorithms/misc.py#L310

also this should become a boolean flag + order being another parameter that's set to some default value but user changeable.

nipype/testing/utils.py
```diff
@@ -98,3 +101,8 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             assert not os.path.exists(self.canary)
         self.dev_null.close()
         shutil.rmtree(self.tmpdir)
+
+def save_toy_nii(ndarray, filename):
```
oesteban: please save to temporal directory

nipype/algorithms/tests/test_tsnr.py
```diff
@@ -0,0 +1,49 @@
...
+import numpy as np
+import os
+
+class TestTSNR(unittest.TestCase):
+    ''' Note: Tests currently do a poor job of testing functionality '''
+    '''
+    in_file = InputMultiPath(File(exists=True), mandatory=True,
+    regress_poly = traits.Range(low=1, desc='Remove polynomials')
+    '''
+    in_file_name = 'tsnrinfile.nii'
```
oesteban: please create in a temporal folder

nipype/algorithms/compcor.py
shoshber: yeah, @chrisfilo suggested this earlier--I'm working on refactoring TSNR to share the code.

oesteban:
@satra, @chrisfilo - we are creating here too many files under algorithms (compcor.py, tsnr.py). Since the algorithms module does not load interfaces itself they will be hard to discover and users will require one import per interface.

I suggest creating only a submodule of algorithms called confounds.py (see #1604 and #1606) for all these algorithms. CompCorr would also go in that one. It is likely that workflows will use several of these interfaces, so with one ```from nipype.algorithms import confounds``` would suffice.

Regarding TSNR, I would prefer to keep it in misc as it was for two reasons. First, that way is easier to spot the changes @shoshber is doing in this PR. Second, it breaks compatibility with users importing it.

After the CompCorr is finished, then I would move TSNR to the confounds module if you want, provided with a class wrapper in misc that logs a warning of deprecation.

WDYT?

shoshber:
Thanks for the feedback. I am personally pro-small files--perhaps this is my Java background--but am willing to be persuaded that more than one interface per file is appropriate. I'm not sure how it compromises discoverability. Can you explain more, @oesteban ? Could a properly set up __init__.py file alleviate those concerns?

I don't think my moving TSNR to its own file breaks compatibility. At least, if worked for me when I tested it (https://github.com/nipy/nipype/pull/1599/files#diff-ae6a9061a228a455a45ce95549571550R43). Glad to be corrected, if that doesn't work.

oesteban:
Yes, an ```__init__.py``` would do, but python is rather obscure in that sense (in contrast with Java where you explicitly handle namespaces). And, it is harder to maintain: as a matter of fact some of the fsl interfaces are missing in the top init file of the fsl submodule (very isolated cases in fsl.utils and fsl.maths). This is because we forgot them at some point.

Additionally (@satra can correct me if I am wrong), I think in nipype we used to create submodules with several self-related features. Since Interfaces are rather verbose (that's a fact derived from the use of traits), you need really long files for wrapping few interfaces. I guess that we prefer that over having a countless number of files (I cannot imagine the fsl submodule with that single-feature approach). Of course, we can agree on moving towards that good practice.

Regarding TSNR, we were having the policy of explicitly warn the user about changes on nipype's API. Your current solution would work (sorry I missed that import) but is totally silent. An example is here: https://github.com/nipy/nipype/blob/master/nipype/algorithms/misc.py#L1478 .

Finally, it is still really hard to see the changes you are applying to the tsnr interface and how those could have second effects. It is not easy to review if you move it at the same time you modify it in a PR for a different feature. I admit I have done that bad practice and others (like, i.e. add PEP8 fixes in a PR for a feature) before, so I totally understand why you moved it. But I would appreciate to see the real changes highlighted by github...

shoshber:
Yeah, the fact that new files don't have good diffs are pretty annoying. At this point I don't want to spend time moving it back into misc.py if I'm just going to move it again to a confounds file, so for now you can look here https://github.com/shoshber/nipype/compare/33d2f309baf74303cc5e8f33061707f444a61c2b...shoshber:compcor1594
for the diff. The commit where I moved TSNR to it's own file was essentially copy-pasted.

oesteban:
That way you would import all the interface specs that we like to hide from the top view.

Thanks for the diff, it's easier that way. But still, I don't see the need to move code that was already there when you just need little adjustments for your interface...

I have submitted to you this PR: https://github.com/shoshber/nipype/pull/1 which tries to simplify the polynomial regression by reshaping data before and after estimation. Could you check that it gives the expected results and that is compatible with your needs? Now, you don't need to care about the dimensions of your input data, the last one will be always time.

shoshber:
`misc.py` is nearly 1500 lines long. Even if you are ok with longish files, that is too long! I figured since I was refactoring part of it (and had the tests to make sure I wasn't changing anything), I might as well start split out TSNR. Also, importing `misc` into `compcor.py` felt weird. Importing from `TSNR` into` compcor` feels weird too, but less. Let's wait until we hear from @satra or @chrisfilo to decide if we should make a `confounds.py`file (waiting to move compcor and tsnr to it for another pr).

chrisfilo:
I would leave moving TSNR to another PR - let's keep this one simple and focused. To increase code clarity you can propose a new module with functions shared by both TSNR and compcor - this should make imports seem more natural.

`confounds.py` sounds fine to me.

BTW Moving interfaces between modules has some compatibility consequences, but let's discuss this in the PR dedicated to refactoring TSNR into a separate module.

oesteban:
That said, I'm sure we all agree on @shoshber's point: ```nipype.algorithms.misc``` is sort of catch-all box particularly long and in need for a deep refactoring.

nipype/algorithms/tests/test_compcor.py
```diff
@@ -0,0 +1,92 @@
...
+import nibabel as nb
+import numpy as np
+import os
+
+class TestCompCor(unittest.TestCase):
+    ''' Note: Tests currently do a poor job of testing functionality '''
+
+    filenames = {'functionalnii': 'compcorfunc.nii',
+                 'masknii': 'compcormask.nii',
+                 'components_file': None}
```
chrisfilo: It would be better to create the test file int temp storage (see https://docs.python.org/3/library/tempfile.html) rather than current workign directory. Even though they are removed after tests run, if the testing suite crashes or gets interrupted the code directories will get polluted with temp files.

shoshber: Does this look right?: https://github.com/nipy/nipype/pull/1599/commits/08bccfacd7885905b0873abb998f486577eaee63

I based it off of nipype/interfaces/fsl/tests/test_model.py

It is probably worth modifying the testrunner itself, someday, to create a new tempdir for every test class that needs it. Good for namespacing, too, which I was having trouble with a few days ago!

satra:
@shoshber - for the temporary directory refactoring let's wait till we switch to pytest. it handles such things quite cleanly. 

also do merge with current master.

nipype/algorithms/tests/test_compcor.py
chrisfilo: Yup - this looks good.

nipype/algorithms/compcor.py
```diff
@@ -0,0 +1,126 @@
...
+        # of low-frequency nuisance terms (e.g., linear and quadratic drift)."
+
+        # "To construct the tSTD noise ROI, we sorted the voxels by their
+        # temporal standard deviation ..."
+        tSTD = self._compute_tSTD(time_voxels, 0)
+        sortSTD = np.sort(tSTD, axis=None) # flattened sorted matrix
+
+        # "... and retained a pre-specified upper fraction of the sorted voxels
+        # within each slice ... we chose a 2% threshold"
+        threshold = sortSTD[int(num_voxels * .98)]
```
chrisfilo: Could we make this an input parameter with default value?

nipype/algorithms/tests/test_tsnr.py
```diff
@@ -103,8 +109,8 @@ def assert_unchanged(self, expected_ranges):
             assert_almost_equal(np.amax(data), max_, decimal=1)
 
     def tearDown(self):
-        utils.remove_nii(self.in_filenames.values())
-        utils.remove_nii(self.out_filenames.values())
+        os.chdir(self.orig_dir)
+        shutil.rmtree(self.temp_dir)
```
oesteban: This is great 👍

nipype/algorithms/compcor.py
shoshber: Ok

nipype/algorithms/compcor.py
```diff
@@ -0,0 +1,178 @@
...
+            regressors = np.genfromtxt(self.inputs.extra_regressors)
+            return np.hstack((components, regressors))
+
+class TCompCorInputSpec(CompCorInputSpec):
+    # and all the fields in CompCorInputSpec
+    percentile_threshold = traits.Range(low=0., high=1., value=.02,
+                                        exclude_low=True, exclude_high=True,
+                                        usedefault=True, desc='the percentile '
+                                        'used to select highest-variance '
+                                        'voxels. By default the 2% of voxels '
```
chrisfilo: The default value is ".02" not "2%".

chrisfilo: there should be some clarity that this should be expressed as value ranging from 0.0-1.0 rather than 0-100

nipype/algorithms/compcor.py
```diff
@@ -0,0 +1,178 @@
...
+        outputs = self._outputs().get()
+        outputs['components_file'] = os.path.abspath(self.inputs.components_file)
+        return outputs
+
+    def _compute_tSTD(self, M, x):
+        stdM = np.std(M, axis=0)
+        # set bad values to x
+        stdM[stdM == 0] = x
+        stdM[np.isnan(stdM)] = x
+        stdM[np.isinf(stdM)] = x
```
chrisfilo: When does this happen? Why are those set to zero?

nipype/algorithms/compcor.py
```diff
@@ -0,0 +1,178 @@
...
+                           mandatory=False, usedefault=True,
+                           desc='filename to store physiological components')
+    num_components = traits.Int(6, usedefault=True) # 6 for BOLD, 4 for ASL
+    use_regress_poly = traits.Bool(True, usedefault=True,
+                                   desc='use polynomial regression'
+                                   'pre-component extraction')
+    regress_poly_degree = traits.Range(low=1, default=1, usedefault=True,
+                                       desc='the degree polynomial to use')
+    extra_regressors = File(exists=True, mandatory=False,
+                            desc='additional regressors to add')
```
chrisfilo: Do we need this? In preprocessing usually regressors from different sources are combined in a later step.

nipype/algorithms/tests/test_compcor.py
```diff
@@ -0,0 +1,115 @@
...
+                                         components_file='acc_components_file'),
+                                expected_components)
+
+        assert_equal(os.path.getsize(ccresult.outputs.components_file),
+                     os.path.getsize(accresult.outputs.components_file))
+
+    @mock.patch('nipype.algorithms.compcor.CompCor._add_extras')
+    def test_compcor_with_extra_regressors(self, mock_add_extras):
+        regressors_file ='regress.txt'
+        open(regressors_file, 'a').close() # make sure file exists
```
chrisfilo: Do we remove it?

nipype/algorithms/compcor.py
shoshber: I added `extra_regressors` here because https://github.com/nipy/nipype/blob/master/examples/rsfmri_vol_surface_preprocessing_nipy.py

had an `extra_regressors` input. I assumed that was normal ... I'm not averse to leaving it out, but then I'm not sure to refactor the above file.

nipype/algorithms/compcor.py
shoshber: I don't know! This was in the original implementation for aCompCor, though there they are set to 1. In tCompCor I set them to 0 because setting them to 0 guaranteed that they would not be selected for the ROI

nipype/algorithms/compcor.py
shoshber: I'll fix the documentation.

nipype/algorithms/compcor.py
chrisfilo: Let's keep it it "lean and mean" and remove this.

nipype/algorithms/compcor.py
chrisfilo: It's a bit counter intuitive. If the SD is infinite (as a numerical approximation of very very big) the voxel should be included. For the sake of safety I would remove this line.

chrisfilo:
Other than the two comments I made (inf and extra_regressors) I think this is ready to merge.

satra:
where did the duecredit reference go?

chrisfilo:
Good question - I saw it when I was reviewing this PR a few days ago. @shoshber ?

shoshber:
see https://github.com/nipy/nipype/pull/1630

PR #1604, opened 2016-09-08
[ENH] Add a FramewiseDisplacement interface

oesteban:
Calculates the FD following the implementation of fsl_motion_outliers,
that theoretically follows that of Power et al. 2012.

If ```save_plot``` is ```True```, then it will save an FD plot using
seaborn (but this is optional). An example of this FD plot is here:

plot

Includes a test which assesses that this implementation is equivalent
to the one in fsl_motion_outliers


chrisfilo:
It would be great to add option to plot and/or return the values in m/s (speed) instead of mm.

For plotting this could be just another axis (right) labelling.

oesteban:
If you set ```normalize=True``` then the plot is done in mm/s and the legends read correctly. But I can do what you propose.

chrisfilo:
My bad I missed that.

oesteban:
@satra, @chrisfilo I think this one is ready to go

Issue #1605, opened 2016-09-08
remove dependency on distutils


satra:
let's see if we can remove distutils dependencies 

```
DEPRECATION: Uninstalling a distutils installed project (nipype) has been deprecated 
and will be removed in a future version. This is due to the fact that uninstalling a 
distutils project will only partially uninstall the project.
```

oesteban:
I'm working on this here https://github.com/oesteban/nipype/tree/enh/MigrateToSetuptools

PR #1606, opened 2016-09-08
[ENH] New ComputeDVARS interface

oesteban:
Computes DVARS, standardized DVARS and voxel-wise standardized DVARS.

Additionally, correspondingly plots are generated:
sub-01_task-mixedgamblestask_run-01_bold_dvars_std


chrisfilo:
Could you add a test on top of the autogenerated one?

oesteban:
I'll see how, np.

satra:
merge with master

chrisfilo:
Comparison with @nicholst script sounds like a good idea.

oesteban:
@chrisfilo: the regression test it a bit permissive (MSE < 0.05) because the robust estimation of the SD in the @nicholsn's script and this implementation differ. For some reason the results from fslmaths and numpy are not exactly the same. Surprisingly the lag-1 coefficients are the same (using nitime and fsl), so I think we can go ahead with this implementation.

chrisfilo:
I think you meant @nicholst not @nicholsn ;)

oesteban:
My apologies to the mistakenly mentioned user. Ok, I had to add nitime. I guess that, as we move towards setuptools (#1605), we will need to add these semi-optional dependencies to the ```extras_require``` option of setup().

satra:
@oesteban - yule-walker should be easy to implement within the code, instead of another dependency. there is a set of nitime interaces in nipype, but we don't add nitime to requirements.txt.

oesteban:
You got it. For now I will skip the test if nitime is not found.

We could incorporate automated testing of dependencies into the Interfaces, using a class member like `dependencies_`.

nipype/algorithms/confounds.py
```diff
@@ -106,16 +270,121 @@ def _run_interface(self, runtime):
...
+    mask = nb.load(in_mask).get_data().astype(np.uint8)
+
+    if len(func.shape) != 4:
+        raise RuntimeError(
+            "Input fMRI dataset should be 4-dimensional")
+
+    # Robust standard deviation
+    func_sd = (np.percentile(func, 75, axis=3) -
+               np.percentile(func, 25, axis=3)) / 1.349
+    func_sd[mask <= 0] = 0
```
shoshber: Robust standard deviation sounds useful--we should consider moving it to another module so everyone can use it.

nipype/algorithms/confounds.py
```diff
@@ -106,16 +270,121 @@ def _run_interface(self, runtime):
...
+
+    if remove_zerovariance:
+        # Remove zero-variance voxels across time axis
+        mask = zero_variance(func, mask)
+
+    idx = np.where(mask > 0)
+    mfunc = func[idx[0], idx[1], idx[2], :]
+
+    # Demean
+    mfunc -= mfunc.mean(axis=1).astype(np.float32)[..., np.newaxis]
```
shoshber: I've seen this multiple places too

nipype/algorithms/confounds.py
```diff
@@ -106,16 +270,121 @@ def _run_interface(self, runtime):
...
+    """
+    idx = np.where(mask > 0)
+    func = func[idx[0], idx[1], idx[2], :]
+    tvariance = func.var(axis=1)
+    tv_mask = np.zeros_like(tvariance, dtype=np.uint8)
+    tv_mask[tvariance > 0] = 1
+
+    newmask = np.zeros_like(mask, dtype=np.uint8)
+    newmask[idx] = tv_mask
+    return newmask
```
shoshber: I did something very similar to this in compcor as well

nipype/algorithms/tests/test_confounds.py
```diff
@@ -1,12 +1,20 @@
...
+from nipype.testing import (assert_equal, example_data, skipif)
+from nipype.algorithms.confounds import FramewiseDisplacement, ComputeDVARS
 import numpy as np
 from tempfile import mkdtemp
 from shutil import rmtree
 
+nonitime = True
+try:
+    import nitime
+    nonitime = False
```
shoshber: I prefer boolean variable names to be positive, for ease of understanding.

nipype/algorithms/tests/test_confounds.py
```diff
@@ -16,3 +24,16 @@ def test_fd():
...
+    tempdir = mkdtemp()
+    ground_truth = np.loadtxt(example_data('ds003_sub-01_mc.DVARS'))
+    dvars = ComputeDVARS(in_file=example_data('ds003_sub-01_mc.nii.gz'),
+                         in_mask=example_data('ds003_sub-01_mc_brainmask.nii.gz'),
+                         save_all = True)
+    os.chdir(tempdir)
+    res = dvars.run()
+
+    dv1 = np.loadtxt(res.outputs.out_std)
+    yield assert_equal, (np.abs(dv1 - ground_truth).sum()/ len(dv1)) < 0.05, True
```
shoshber: for my own edification, since I've had trouble myself writing robust tests for nipype, how did you pick the 0.05 epsilon?

nipype/algorithms/confounds.py
oesteban: Well, I'd say that this is a quick and dirty way of calculating the SD. I am +1 if we had a really robust SD computation algorithm though.

nipype/algorithms/confounds.py
oesteban: I don't know what you mean, the masking of voxels with zero-variance or some other thing?

Anyways, thanks to this comment I spotted those ugly leftovers in ll. 326-328

nipype/algorithms/confounds.py
oesteban: So you mean, mask out all zero-variance pixels? There is this little function to do that now :)

nipype/algorithms/tests/test_confounds.py
oesteban: We generally use this convention for boolean variables that will be used in skip decorators.

nipype/algorithms/tests/test_confounds.py
oesteban: Honestly: the MSE I was getting was 0.048.

In detail: I run our implementation and @nicholst 's, stopping at every step and comparing outputs. Once I found out that they were exactly the same except for some differences in the *robust* SD estimation, I assumed that that epsilon is tolerable.

nipype/algorithms/confounds.py
shoshber: got it

nipype/algorithms/confounds.py
shoshber: I meant "demeaning":
```
$ grep -rIw "\-.*\.mean" .
./examples/rsfmri_vol_surface_preprocessing.py:        X = (X - np.mean(X, axis=0)) / stdX
./nipype/workflows/dmri/fsl/epi.py:    vsmmag_masked = vsmmag_masked - vsmmag_masked.mean()
./nipype/algorithms/#misc.py#:            signal = signal - signal.mean()
./nipype/algorithms/misc.py:            signal = signal - signal.mean()
./nipype/algorithms/compcor.py:        voxel_timecourses = voxel_timecourses - np.mean(voxel_timecourses,
./nipype/algorithms/compcor.py:        imgseries = imgseries - np.mean(imgseries, axis=1)[:, np.newaxis]
./nipype/algorithms/rapidart.py:        gz = (gz - np.mean(gz)) / np.std(gz)  # normalize the detrended signal
```

nipype/algorithms/confounds.py
shoshber: is this the right place to put it?

nipype/algorithms/tests/test_confounds.py
shoshber: ok

nipype/algorithms/tests/test_confounds.py
shoshber: got it!

nipype/algorithms/confounds.py
oesteban: Where else?

oesteban:
I think this is ready to go

PR #1609, opened 2016-09-08
replaced coveralls with codecov

chrisfilo:
It supports merging coverage reports from multiple builds.


satra:
is there a way we can test coverage from our examples as well?

chrisfilo:
There is, but that would involve running examples inside nose.

satra:
then let's wait till we move from nose to pytest.

oesteban:
LGTM and it is very much needed.

Issue #1612, opened 2016-09-09
Conditional keep outputs


ashgillman:
I have developed some interfaces that work with Interfiles, which are a two (or three) part file. One (or two) header(s) and one data file.

Currently, to prevent deletion of the data files I have to pass every file around twice, so I have `inputs.in1_header`, `inputs.in1_data`, etc. Whereas for these filetypes, if the header is required, the data file is always required. It would be great if I could specify in the `output_spec` that if `out` is a required output, the node should also not delete a file with the same name but a different extension that I could specify.

Similarly, I have an interface that takes a pattern as an input, rather than a list. If I want to pass an output that is actually a file pattern to this input, it would be good if I could keep files matching that pattern (e.g., `output_%d.hv`). This is a bit trickier, because it is arbitrary what type of pattern to use. My current workaround is either specify the pattern and the list as an output, and point the list somewhere so it is kept, or disabling cleaning for the Node.

Maybe you could specify a lambda in the output trait that matches files to keep? Or would this create pickling errors?

I'd be happy to help out with this, if it is something that the developers want.

Cheers

oesteban:
Quick question - Have you tried the ```keep_inputs``` and the ```remove_unnecessary_outputs``` options (see http://nipype.readthedocs.io/en/latest/users/config_file.html)?

satra:
@ashgillman - perhaps you can add to things here:

https://github.com/nipy/nipype/blob/master/nipype/utils/filemanip.py#L317

and here:

https://github.com/nipy/nipype/blob/master/nipype/utils/filemanip.py#L336

ashgillman:
Yes to `remove_unnecessary_outputs`, and I imagine it should also work with `keep_inputs`. But it would be nice to have these conditional on whether the other output is actually used.

ashgillman:
@satra Wow, I haven't come across those bits of code in my searches so far. That looks very similar to what I need.

mgxd:
@ashgillman did the PR clear this issue up?

ashgillman:
Yes, the important part at least.

For anyone interested, you can add to the related filetypes with:

```
nipype.utils.filemanip.related_filetype_sets.extend([
    ('.avh', '.hv', '.v'),
])
```

(but only as of master and presumably the next 0.13 releasse)

PR #1617, opened 2016-09-12
REF: define related filetypes once.

ashgillman:
In response to #1612

Refactors using DRY principal, define related files once.

Additionally, can be patched to extend the related filtypes with:

    import nipype
    nipype.utils.filemanip.related_filetype_sets.extend([
        ('.avh', '.hv', '.v'),
        ('.avs', '.hs', '.s'),
    ])

Local tests run, and I have checked the above patching with some of my code using Interfiles.

Potential caveats:
 - copyfile is recursive, my base condition is not to recurse if the destination file exists (i.e., has been copied). This could lead to infinite recursion if the copy fails. Suggestions are welcome (check for infinite recursion and throw instructive error? Add a new function argument defining whether to check related files so is only checked at top level?)

- behaviour is not identical to before. i.e., before `copyfile` on checked for related files for `.img` or `.BRIK`, and not if the other file part was used. I think this way is better, but some issues may come from using `.mat` files?

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -207,7 +207,8 @@ def hash_timestamp(afile):
 
 
 def copyfile(originalfile, newfile, copy=False, create_new=False,
-             hashmethod=None, use_hardlink=False):
+             hashmethod=None, use_hardlink=False,
+             copy_related_files=True):
     """Copy or link ``originalfile`` to ``newfile``.
 
     If ``use_hardlink`` is True, and the file can be hard-linked, then a
@@ -228,6 +229,9 @@ def copyfile(originalfile, newfile, copy=False, create_new=False,
     use_hardlink : Bool
         specifies whether to hard-link files, when able
         (Default=False), taking precedence over copy
+    copy_related_files : Bool
+        specifies whether to also operate on related files, as defined in
+        ``related_filetype_sets``
 
     Returns
     -------
@@ -320,30 +324,36 @@ def copyfile(originalfile, newfile, copy=False, create_new=False,
             fmlogger.warn(e.message)
 
     # Associated files
-    _, _, this_type = split_filename(originalfile)
-    for type_set in related_filetype_sets:
-        if this_type in type_set:
-            for alt_type in type_set:
-                alt_ofile = originalfile[:-len(this_type)] + alt_type
-                alt_nfile = newfile[:-len(this_type)] + alt_type
-                if os.path.exists(alt_ofile) and not os.path.exists(alt_nfile):
-                    copyfile(alt_ofile, alt_nfile, copy,
-                             hashmethod=hashmethod,
-                             use_hardlink=use_hardlink)
+    if copy_related_files:
+        related_file_pairs = (get_related_files(f, include_this_file=False)
+                              for f in (originalfile, newfile))
+        for alt_ofile, alt_nfile in zip(*related_file_pairs):
+            if os.path.exists(alt_ofile):
+                copyfile(alt_ofile, alt_nfile, copy, hashmethod=hashmethod,
+                         use_hardlink=use_hardlink, copy_related_files=False)
 
     return newfile
 
 
-def get_related_files(filename):
-    """Returns a list of related files for Nifti-Pair, Analyze (SPM) and AFNI
-    files
+def get_related_files(filename, include_this_file=True):
+    """Returns a list of related files, as defined in
+    ``related_filetype_sets``, for a filename. (e.g., Nifti-Pair, Analyze (SPM)
+    and AFNI files).
+
+    Parameters
+    ----------
+    filename : str
+        File name to find related filetypes of.
+    include_this_file : bool
+        If true, output includes the input filename.
     """
     related_files = []
-    path, name, ext = split_filename(filename)
+    path, name, this_type = split_filename(filename)
     for type_set in related_filetype_sets:
-        if ext in type_set:
-            for new_ext in type_set:
-                related_files.append(os.path.join(path, name + new_ext))
+        if this_type in type_set:
+            for related_type in type_set:
+                if include_this_file or related_type != this_type:
+                    related_files.append(os.path.join(path, name + related_type))
     if not len(related_files):
         related_files = [filename]
     return related_files
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -29,6 +29,12 @@ from ..interfaces.traits_extension import isdefined
 fmlogger = logging.getLogger("filemanip")
 
 
+related_filetype_sets = [
+    ('.hdr', '.img', '.mat'),
+    ('.BRIK', '.HEAD'),
+]
+
+
 class FileNotFoundError(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


nipype/utils/filemanip.py
```diff
@@ -29,6 +29,12 @@
 fmlogger = logging.getLogger("filemanip")
 
 
+related_filetype_sets = [
```
shoshber: Even DRYer--assuming all extensions start with a period, no need to specify the period here!

nipype/utils/filemanip.py
```diff
@@ -314,38 +320,30 @@ def copyfile(originalfile, newfile, copy=False, create_new=False,
...
-            related_files.append(os.path.join(path, name + ext))
-    elif filename.endswith(".BRIK") or filename.endswith(".HEAD"):
-        path, name, ext = split_filename(filename)
-        for ext in ['.BRIK', '.HEAD']:
-            related_files.append(os.path.join(path, name + ext))
+    path, name, ext = split_filename(filename)
+    for type_set in related_filetype_sets:
+        if ext in type_set:
+            for new_ext in type_set:
+                related_files.append(os.path.join(path, name + new_ext))
```
shoshber: lines 342-345 and 323-326 are the same--consider refactoring into a `get_type_set()` or `has_type_set()` function of some sort

nipype/utils/filemanip.py
```diff
@@ -314,38 +320,30 @@ def copyfile(originalfile, newfile, copy=False, create_new=False,
...
-        hdrnfile = newfile[:-5] + ".HEAD"
-        copyfile(hdrofile, hdrnfile, copy, hashmethod=hashmethod,
-                 use_hardlink=use_hardlink)
+    _, _, this_type = split_filename(originalfile)
+    for type_set in related_filetype_sets:
+        if this_type in type_set:
+            for alt_type in type_set:
+                alt_ofile = originalfile[:-len(this_type)] + alt_type
+                alt_nfile = newfile[:-len(this_type)] + alt_type
+                if os.path.exists(alt_ofile) and not os.path.exists(alt_nfile):
```
shoshber: this behavior feels a little weird to me. I don't have neuro background, so perhaps it isn't weird at all. But if `alt_ofile` AND `alt_nfile` exist, what does that mean? Should we throw a warning? Should we copy over `alt_nfile`? Can we perhaps use `get_related_files` in an intelligent way here?

shoshber:
I would love to see some unittests here

nipype/utils/filemanip.py
ashgillman: if original and new files exists, the function assumes we have already done the copy. This is therefore, the recursive base condition. This was my attempt at minimising changes to the function signature. Probably a better way is to change the signature:

    def copyfile(originalfile, newfile, copy=False, create_new=False,
                 hashmethod=None, use_hardlink=False, copy_related_files=True)

And then, when we recurse for related files, set `copy_related_files` to `False`

nipype/utils/filemanip.py
ashgillman: I take your point, but I think that leaving the periods makes the intent more obvious.

Would you suggest something like:

    related_filetype_sets = [tuple('.' + type_ext)
                             for type_ext in (
                                 ('hdr', 'img', 'mat'),
                                 ('BRIK', 'HEAD'),
                             )]

Or just handling it later (i.e., in a function `get_related_filetypes` that does what you suggest below)

I am still personally opting for leaving the dots in.

ashgillman:
Good suggestion on using `get_related_filetypes` from `copyfile`.

`copyfile` seems to be tested quite well (except only using analyze files), and I have added tests for `get_related_files`

nipype/utils/filemanip.py
shoshber: ah, I get it. Thanks for the explanation

nipype/utils/filemanip.py
shoshber: I was imagining handling it later--but I'm also not so much of a stickler as to insist on changing this

shoshber:
Good point on the infinite recursion--of your suggestions above, I think I prefer the add-extra-parameters option.

shoshber:
Argh, tiny phone big fingers!

ashgillman:
@shoshber Happens to the best of us. 4ab5554 includes the extra parameter option.

shoshber:
I see that! Looks good to me!

oesteban:
@effigies It looks like we have a comeback of the copy tests failure here (https://s3.amazonaws.com/archive.travis-ci.org/jobs/159763024/log.txt)

```
======================================================================
FAIL: nipype.utils.tests.test_filemanip.test_recopy(os.stat_result(st_mode=33204, st_ino=3540, st_dev=2049, st_nlink=1, st_uid=1000, st_gid=1000, st_size=0, st_atime=1473817459, st_mtime=1473817459, st_ctime=1473817459), os.stat_result(st_mode=33204, st_ino=3540, st_dev=2049, st_nlink=1, st_uid=1000, st_gid=1000, st_size=0, st_atime=1473817460, st_mtime=1473817459, st_ctime=1473817459), 'Symlink - OS: posix; Copy: True; Hardlink: False')
----------------------------------------------------------------------
  File "/home/travis/miniconda/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 315, in assert_equal
    assert_equal(actual[k], desired[k], 'item=%r\n%s' % (k, err_msg), verbose)
  File "/home/travis/miniconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 379, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
item=7
Symlink - OS: posix; Copy: True; Hardlink: False
 ACTUAL: 1473817459
 DESIRED: 1473817460

======================================================================
FAIL: nipype.utils.tests.test_filemanip.test_recopy(os.stat_result(st_mode=33204, st_ino=3541, st_dev=2049, st_nlink=1, st_uid=1000, st_gid=1000, st_size=0, st_atime=1473817459, st_mtime=1473817459, st_ctime=1473817459), os.stat_result(st_mode=33204, st_ino=3541, st_dev=2049, st_nlink=1, st_uid=1000, st_gid=1000, st_size=0, st_atime=1473817460, st_mtime=1473817459, st_ctime=1473817459), 'Symlink - OS: posix; Copy: True; Hardlink: False')
----------------------------------------------------------------------
  File "/home/travis/miniconda/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 315, in assert_equal
    assert_equal(actual[k], desired[k], 'item=%r\n%s' % (k, err_msg), verbose)
  File "/home/travis/miniconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 379, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
item=7
Symlink - OS: posix; Copy: True; Hardlink: False
 ACTUAL: 1473817459
 DESIRED: 1473817460
```

effigies:
Oh, I see the problem, now. The `atime` may be getting updated if the test straddles a clock tick, even though the files aren't being changed. Check out 05d6a05 (cherry-pickable). It adds a function to ignore `atime` and to wait for clock ticks to make the test deterministic.

That said, it increases the runtime of the test from <1s to ~13s, so you may want to satisfy yourselves that the logic works and then remove the `_wait_for_tick()` calls.

shoshber:
Instead of increasing test time by 13sec, can we just relax the equality requirement? Perhaps with `assert_almost_equal` and epsilon 1

effigies:
I wouldn't want to do `assert_almost_equal` on the whole thing, because inodes that differ by 1 are different inodes. I think removing `atime` from the comparison is the more conservative choice there. And that fully solves the issue here.

The clock-tick waits are basically the proof that the problem is solved, and the debugging technique I used to figure out the problem. The only advantage to leaving them in is that they will alert you if this problem crops up again. I'm perfectly comfortable removing them, if that's an unacceptable price.

shoshber:
Ah, I see. I vote to remove `_wait_for_tick` calls, like you suggest.

effigies:
Cool. Would you rather l rather I submit a separate PR to nipype, a PR to this branch, or just cherry-pick?

The commits to grab: 05d6a05 9573168

shoshber:
I guess a separate PR? Whatever you like best. I tried making it myself and cherrypicking but couldn't figure it out. I may not have the right permissions.

PR #1618, opened 2016-09-12
ignore tests in calculating coverage

chrisfilo:
Who's testing the tests? Who's policing the police?

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,10 @@
 coverage:
+  ignore:
+    - "test_*"
+    - "*/test_*"
+    - "*/*/test_*"
+    - "*/*/*/test_*"
+    - "*/*/*/*/test_*"
   status:
     project:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -1,4 +1,10 @@
 coverage:
+  ignore:
+    - "test_*"
+    - "*/test_*"
+    - "*/*/test_*"
+    - "*/*/*/test_*"
+    - "*/*/*/*/test_*"
```
shoshber: I'm surprised there's no way to make this just two lines a la 
```
ignore:
  - "*/test_*"
```
maybe using full-featured regex instead of simple wildcards?

chrisfilo: The documentation is very sparse. This was guaranteed to work.

oesteban:
Are we still policing the police?

PR #1624, opened 2016-09-12
[FIX] Encoding problems in SGE plugin

### LINKED ISSUE #1621: python 2.7 string issue in plugins
with recent updates, I get the following error using SGE plugin:
```
TraitError: Each key of the 'environ' trait of a CommandLineInputSpec instance must be a newstr or None, but a value of 'MDMSESSION' <type 'str'> was specified.
```
It seems that the traits definition uses the future newstr which does not match with what os.environ contains (old str). It is certainly the same for other plugins, and this is the only places where environ is fed to CommandLine in whole nipype.

Any idea how to fix this?
Thanks

### LINKED ISSUE #1625: [BUG] New initialization of interfaces with from_file fails with lists of lists of ints

oesteban:
Fixes #1621

### DIFF
--- a/nipype/pipeline/plugins/sge.py
+++ b/nipype/pipeline/plugins/sge.py
@@ -8,7 +8,6 @@ from builtins import object, str, bytes
 import os
 import pwd
 import re
-import subprocess
 import time
 
 import xml.dom.minidom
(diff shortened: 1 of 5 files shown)


nipype/pipeline/plugins/sge.py
```diff
@@ -147,7 +147,7 @@ def _qacct_verified_complete(taskid):
         while qacct_retries > 0:
             qacct_retries -= 1
             try:
-                strtaskid = '{0!d}'.format(taskid)
+                strtaskid = '{0:d}'.format(int(taskid))
                 proc = subprocess.Popen(
```
satra: i think i switched this to `CommandLine` in the slurm plugin - take a look there. `CommandLine` handles most encoding issues properly.

PR #1629, opened 2016-09-13
Fix dictstrstr

bpinsard:
The future strings in traits definition causes crashes for most interfaces using os.environ as an input.

### DIFF
--- a/nipype/interfaces/traits_extension.py
+++ b/nipype/interfaces/traits_extension.py
@@ -33,7 +33,7 @@ from traits.trait_base import _Undefined, class_of
 from traits.api import BaseUnicode
 from traits.api import Unicode
 
-DictStrStr = traits.Dict((bytes,str), (bytes, str))
+DictStrStr = traits.Dict((bytes,str,unicode), (bytes,str,unicode))
 Str = Unicode
 
 class BaseFile(BaseUnicode):
--- a/nipype/interfaces/traits_extension.py
+++ b/nipype/interfaces/traits_extension.py
@@ -33,7 +33,7 @@ from traits.trait_base import _Undefined, class_of
 from traits.api import BaseUnicode
 from traits.api import Unicode
 
-DictStrStr = traits.Dict(str, (bytes, str))
+DictStrStr = traits.Dict((bytes,str), (bytes, str))
 Str = Unicode
 
 class BaseFile(BaseUnicode):
--- a/nipype/interfaces/traits_extension.py
+++ b/nipype/interfaces/traits_extension.py
@@ -33,7 +33,7 @@ from traits.trait_base import _Undefined, class_of
 from traits.api import BaseUnicode
 from traits.api import Unicode
 
-DictStrStr = traits.Dict((bytes,str,unicode), (bytes,str,unicode))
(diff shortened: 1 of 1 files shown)


nipype/interfaces/traits_extension.py
```diff
@@ -33,7 +33,7 @@
 from traits.api import BaseUnicode
 from traits.api import Unicode
 
-DictStrStr = traits.Dict(str, (bytes, str))
+DictStrStr = traits.Dict((bytes,str,unicode), (bytes,str,unicode))
```
oesteban: unicode does not exist in python 3 anymore.

you can do

```
import sys

DictStrStr = traits.Dict(str, (bytes,str))
if sys.version_info[0] < 3:
    DictStrStr = traits.Dict((bytes,str,unicode), (bytes,str,unicode))
```

However, ```DictStrStr = traits.Dict((bytes, str), (bytes,str))``` does not solve your problem as well?

oesteban:
Ok just confirmed #1621 in a docker image with SGE.

Also confirmed that ```DictStrStr = traits.Dict((bytes, str), (bytes,str))``` fixes the issue.

nipype/interfaces/traits_extension.py
bpinsard: In the error, it seems that all my environ strings are in unicode, so without unicode it does not solve my problem.

chrisfilo: Maybe the solution is to typecast the environ somewhere in the base
interface class?

bpinsard: Does ```__future__``` or something else is changing the environ content? 
Because in any other context, the environ is all defined in str (=python3 bytes).

oesteban: Apparently, the problem is in the keys of environ, not the values.

Did it work for you using ```DictStrStr = traits.Dict((bytes, str), (bytes,str))``` instead the version with ```unicode```?

(i.e. adding bytes for the validation of keys in DictStrStr)

This should do (it worked on the SGE environment I created for testing, where I could reproduce the issue)

bpinsard: Without unicode, it does not work, weirdly my os.environ is all unicode inside nipype, outside it regular python str.
What is weird, I removed my bashrc, then the bug is fixed, I also try to remove a lot of stuff from my bashrc and it is also fixed.
The bashrc is not utf8 coded, and does not contains any special character, so it might be some sourcing that introduce this bug.
So the interaction of these sourced bash scripts with nipype and __future__ is causing os.environ to be all unicode in nipype only, why, I do not know?
Anyway, thanks for the help, I think we can keep the ```DictStrStr = traits.Dict((bytes, str), (bytes,str))``` then ?

satra:
@bpinsard - it's possible that you have some environmental setup in bashrc that's affecting things.

see: https://perlgeek.de/en/article/set-up-a-clean-utf8-environment

also, there are lines like this, which can affect input:

https://github.com/nipy/nipype/blob/master/nipype/interfaces/base.py#L1229

@oesteban - is `bytes` needed in DictStrStr

oesteban:
@satra - yes, I think this patch is necessary.

Issue #1628, opened 2016-09-13
Signal extraction interface


chrisfilo:
It's common to extract time series averaged over a tissue class or a brain region. Such signals could be used to perform:

- global signal regression (average over the whole brain)
- white matter and CSF signal regression (averaged over voxels belonging to the same tissue class)
- graph based analysis (average over voxels belonging to the same parcel)

It would be good to have a dedicated interface to perform such signal extractions. The interface would take:
- 4D file with fMRI data
- 3D label image (with 0 denoting background) or 4D file with probability maps
- a vector of class labels (as long as there are labels in the label image or in case the label image is 4D as long as the number of volumes in the 4D file with probability maps)

The output would be a .tsv file with as many columns as there were labels and as many rows as there were timepoints in the fMRI input data.

In addition we can consider adding an extra column "global" with mean signal across all labels.

[nilearn.input_data.NiftiLabelsMasker](http://nilearn.github.io/modules/generated/nilearn.input_data.NiftiLabelsMasker.html) can be very useful in implementing this Interface.

- [x] simplest case
- [x] 4D probability maps/partial voluming input
- [x] extra "global" column

satra:
this sounds like a python equivalent of an mri_segstats interface. it may be nice to just clone that in python. also keep in mind that most of these stat generating interfaces can also (optionally) take into account partial voluming.

oesteban:
@satra - wouldn't the _4D probability maps_ take into account the partial voluming?

shoshber:
@satra what do you mean by cloning mri_segstats? You mean re-implementing? Or writing it as a FreeSurfer interface that calls mri_segstats behind the scenes?

chrisfilo:
I think reimplementing this in python was what @satra meant by "cloning". FreeSurfer is a big dependency for such a simple functionality. In addition we can add some functionality in the future - for example using the first eigenvector instead of the mean.

chrisfilo:
I had some more discussions about the probabilistic tissue maps, and it seems more common for those to come as a set 3D files instead of a single 4D file. This is how FSL FAST outputs them for example. This is how we defined them in BIDS Derivatives a while ago (see https://docs.google.com/document/d/1Wwc4A6Mow4ZPPszDIWfCUCRNstn7d_zzaWPcfcHmgI4/edit#heading=h.6wlyh7diehqc).

I would vote for adding support for a list of 3D maps as the input for label_image. It's probably more common than the 4D case.

shoshber:
[...]

Issue #1613, opened 2016-09-09
Refactor algorithms/misc.py


shoshber:
see https://github.com/nipy/nipype/pull/1599

shoshber:
I'm going to start this unless anyone has any objections

oesteban:
I think #1628 is more interesting

shoshber:
definitely more interesting :)

PR #1631, opened 2016-09-13
REF refactor algorithms misc.py

shoshber:
https://github.com/nipy/nipype/issues/1613

Nice-to-have pre https://github.com/nipy/nipype/issues/1628 (so it will be neater)


nipype/algorithms/misc.py
```diff
@@ -35,6 +34,7 @@
                                BaseInterfaceInputSpec, isdefined,
                                DynamicTraitedSpec, Undefined)
 from ..utils.filemanip import fname_presuffix, split_filename
+from . import confounds
```
oesteban: This can be problematic. If you need some interface from confounds, please import that one interface only.

nipype/workflows/rsfmri/fsl/resting.py
```diff
@@ -7,8 +7,8 @@
 from ....interfaces import fsl as fsl          # fsl
 from ....interfaces import utility as util     # utility
 from ....pipeline import engine as pe          # pypeline engine
-from ....algorithms.misc import TSNR
-from ....algorithms import compcor as cc
+from ....algorithms.confounds import TSNR
```
oesteban: can we use ```confounds.TSNR``` along the file and remove this import?

shoshber: ok!

nipype/algorithms/misc.py
```diff
@@ -1520,3 +1418,14 @@ def __init__(self, **inputs):
...
                       DeprecationWarning)
+
+class TSNR(confounds.TSNR):
+    """
+    .. deprecated:: 0.12.1
+       Use :py:class:`nipype.algorithms.confounds.TSNR` instead
+    """
+    def __init__(self, **inputs):
+        super(confounds.TSNR, self).__init__(**inputs)
+        warnings.warn(("This interface has been deprecated since 0.12.0,"
```
chrisfilo: Could we say moved instead of deprecated? "deprecated" suggest that functionality has changed

shoshber: done

PR #1638, opened 2016-09-15
fixing unicode_literals issues

bpinsard:
see #1621 
fixes my problems


oesteban:
Let me add FreeSurfer to the testing image, we could be missing something we are not currently testing.

nipype/interfaces/freesurfer/base.py
```diff
@@ -211,7 +211,7 @@ def __init__(self, **inputs):
     def _num_threads_update(self):
         if self.inputs.num_threads:
             self.inputs.environ.update(
-                {'OMP_NUM_THREADS': str(self.inputs.num_threads)})
+                {'OMP_NUM_THREADS': bytes(self.inputs.num_threads)})
```
effigies: Python 2:

```Python
>>> bytes(5)
'5'
```

Python 3:
```Python
>>> bytes(5)
b'\x00\x00\x00\x00\x00'
```

I suspect what you want is:

```Python
>>> str(5).encode()
b'5'
```

satra:
@bpinsard - can you check if current master fixes your issues? i cannot get this line to error using current master in either py27 or py35:

```
python -c "from nipype.pipeline import engine as pe; from nipype.interfaces import freesurfer; n_tesselate = pe.Node(freesurfer.MRITessellate(label_value=1), name='tesselate')"
```

satra:
@bpinsard - can you let us know if this still an issue with current master?

oesteban:
@bpinsard did installing traits>=4.5 fix this problem (see #1809)?

bpinsard:
@oesteban yes it seems that it is fixed now.
Thanks.

PR #1627, opened 2016-09-13
[ENH] Abandon distutils

oesteban:
This PR removes the dependency on ```distutils``` and uses only ```setuptools```.

Following [the recommendations](https://packaging.python.org/requirements/) I have reorganized the way nipype installs. Reading through [this discussion](https://www.reddit.com/r/Python/comments/3uzl2a/setuppy_requirementstxt_or_a_combination/) it seems that ```requirements.txt``` should be written for development environments, and production (pypi, mostly) dependencies should be handled with ```setuptools```.

Therefore, in this PR we start using all the available possibilities (```setup_requires```, ```install_requires```, ```test_requires```, ```extra_requires```, etc.) to install nipype smoothly.

I am testing all the possibilities with fresh environments:

- Regular user install through pypi, with the duecredit extra dependency:

  ```
  pip install nipype[duecredit]
  ```

- Regular user install from master:

  ```
  python setup.py install
  ```

- Developer install:

  ```
  pip install -r requirements.txt
  pip install -e .[test,doc]
  ```

- Testing install:

  ```
  pip install -r requirements.txt
  pip install -e .[all]
  ```

The available extra dependencies are listed here: https://github.com/oesteban/nipype/blob/enh/MigrateToSetuptools/nipype/info.py#L157

This PR requires #1623 to be merged first. Closes #1605

PR #1639, opened 2016-09-16
[FIX] setup.py: remove future and configparse deps

oesteban:
When installing with pip install -e ., these needed to be already installed.


oesteban:
Why don't we include those tests in #1627 ? This is a minor fix to solve some issues when you don't handle the install process (i.e. automated build of libraries depending on nipype in RTD)

PR #1640, opened 2016-09-16
ENH SignalExtractionInterface

shoshber:
Done with the simplest case here: https://github.com/nipy/nipype/issues/1628


nipype/algorithms/stats.py
```diff
@@ -0,0 +1,88 @@
...
+from ..interfaces.base import (traits, TraitedSpec, BaseInterface,
+                               BaseInterfaceInputSpec, File)
+IFLOG = logging.getLogger('interface')
+
+class SignalExtractionInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='4-D fMRI nii file')
+    label_file = File(exists=True, mandatory=True,
+                      desc='a 3-D label image, with 0 denoting background, or '
+                      'a 4-D file of probability maps.')
+    class_labels = traits.List(mandatory=True,
```
oesteban: Are we sure we want this mandatory?, what if we have 100 labels (which is not a corner case)?

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,88 @@
...
+    >>> seinterface.inputs.class_labels = segments
+    >>> seinterface.inputs.stat = 'mean'
+    '''
+    input_spec = SignalExtractionInputSpec
+    output_spec = SignalExtractionOutputSpec
+
+    def _run_interface(self, runtime):
+        ins = self.inputs
+        if ins.stat == 'mean': # always true for now
+            nlmasker = nl.NiftiLabelsMasker(ins.label_file).fit()
```
oesteban: I think the whole point of adding another dependency only for NiftiLabelsMasker is to use all the features of that dependency. I would suggest to add, at least, inputs to set the ```standardize```, ```detrend```, and ```smoothing_fwhm``` arguments of NiftiLabelsMasker.

oesteban: Also, we could check if the segmentation input is a 3D volume (a segmentation) or a 4D construct (either 4D nifti or list of 3D niftis) and use ```NiftiMapsMasker``` in this second case. That would take into account the PVE @satra mentioned.

requirements.txt
```diff
@@ -12,3 +12,4 @@ xvfbwrapper
 psutil
 funcsigs
 configparser
+nilearn
```
oesteban: maybe as an extra dependency, as we do with nilearn?

nipype/algorithms/stats.py
oesteban: I see now that tp/vf maps were a requirement from the beginning in #1628

nipype/algorithms/stats.py
shoshber: That was my understanding from @chrisfilo . If someone really doesn't want human-readable labels, they could always just  make an empty list ["" for i in range(100)]`. Thoughts, Chris?

nipype/algorithms/stats.py
shoshber: `standardize`, `detrend`, and `smoothing_fwhm` were out of scope (from my understanding). Some clarity from @chrisfilo will help here.

I was planning to add tissue probability maps/partial voluming in a second pull request. (I like to keep PRs small  & quick to avoid merge headaches.)

requirements.txt
shoshber: I'm not sure I follow ... can you clarify?

shoshber:
Builds are failing because `sklearn`, a dependency of `nilearn`, is not getting installed for some reason

nipype/algorithms/stats.py
oesteban: Another example (which is not an unlikely use case) is the FreeSurfer parcellation -> https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/AnatomicalROI/FreeSurferColorLUT

I think we should allow the user reuse the labels from the segmentation --what if the labels are not consecutive numbers?--

requirements.txt
oesteban: Let's wait for @satra impressions on this new dependency

satra: haha! not in requirements.txt as far as i am concerned.  we don't depend on fsl, spm, nipy, matplotlib, etc.,.

so these tests should all have things like `@skip_if(no_nilearn)` but it does seem that masking is so trivial we could achieve it with nibabel alone.

nipype/algorithms/stats.py
shoshber: Oh, I didn't realize that could happen. The label file contain strings instead of integers? (or any object?)

nipype/algorithms/stats.py
chrisfilo: I think exposing detrending (similarly to CompCor) makes a lot of sense considering what this interface will be used for.

nipype/algorithms/stats.py
chrisfilo: @oesteban the problem is that the label_file image does not include the labels in it's header.
@shoshber the label_file always contains integers

The assumption is (and this should be explained in the docstring) that the class_labels list corresponds to class integers ordered in ascending fashion. We can also switch to using a dictionary to make it more explicit.

requirements.txt
shoshber: gotcha.

nipype/algorithms/stats.py
shoshber: will do

nipype/algorithms/stats.py
shoshber: will do

shoshber:
After thinking more about what satra said & learning about 4D label files/partial voluming I'm going to change this significantly, so I'm closing this pr.

PR #1645, opened 2016-09-18
WIP signal extraction interface

shoshber:
issue https://github.com/nipy/nipype/issues/1628


chrisfilo:
What's the justification for not using nilearn?

shoshber:
It's not a decision I made lightly. There are a few reasons:
1. It's too big a dependency for such a simple task (like @satra  mentioned). \
2. Also, nilearn worked really well for the simple case with a 3D label file, and even 4D label files with only two values, but I couldn't find a way to make it work for 4D label files with probabilities.
3. A well designed interface to work with 4D label files with floats, without nilearn, will work well, with minimal changes, to work with 3D label files.

Basically, I was thinking of 4D label files as a special case of 3D label files, but things are much clearer and work better if I think of 3D label files as a special case of 4D label files. Does all that make sense?

chrisfilo:
Nipype has many conditional dependencies that are not installed by default (for example dipy, nitime, and nipy). I think all @satra was suggesting is to treat nilearn the same way - not to stop using it.

Nilearn is not only well tested, but also (as far as I know) will handle resampling volumes if they happen to use different voxel sizes. It also has ways for extracting signals from classes defined with a single 3D image (where each class has a corresponding integer) as well as as a set of maps (where each class has a corresponding weight map). See http://nilearn.github.io/manipulating_images/masker_objects.html#extraction-of-signals-from-regions-niftilabelsmasker-niftimapsmasker for more details.

I would strongly recommend using nilearn.

chrisfilo:
It also provides hady functions for handling 4d volumes: http://nilearn.github.io/manipulating_images/manipulating_images.html#accessing-individual-volumes-in-4d-images

shoshber:
In that case, I'm somewhat confused as I did implement 4D label files using nilearn's NifitMapsMasker and it wasn't behaving the way I expected. This is what I had: https://github.com/nipy/nipype/compare/master...shoshber:partialvoluming?expand=1

chrisfilo:
How was it behaving?

chrisfilo:
It's probably because it is doing "least square regression" (see http://nilearn.github.io/modules/generated/nilearn.input_data.NiftiMapsMasker.html) using a model with all of the maps instead of calculating weighted mean from each map separately. You can do a little experiment and plot the output of NiftiMapsMasker (with no detrending or standardizing) when using **only one weight map** along a plot of timeseries you get using just the weighted mean. The plots should be similar and the time series should be highly correlated.

When I think about it the way nilearn handles signal extraction is actually better than simple weighted mean since it's only giving unique signal corresponding to each class. WDYT @oesteban @satra.
@GaelVaroquaux might chime in if he has time.

shoshber:
Ok--I thought we wanted weighted means. You also mentioned the possibility of allowing additional signal extraction functions (particularly PCA). I had a brief look at the nilearn source, and it doesn't look like the signal extraction function is intended to be customizable. The least-squares linear regression is baked in here: https://github.com/nilearn/nilearn/blob/b4ec94039a78faa4dc534c20c50b6d08519f4e38/nilearn/regions/signal_extraction.py#L266

Warning: extreme spitballing follows.

I think we could subclass NiftiMapsMasker and write a new _ExtractionFunctor but that seems like an abuse of nilearn's class structure. 

And/or we could add an optional lambda to NifitMapsMasker/_ExtractionFunctor/img_to_signals_maps (https://github.com/nilearn/nilearn/blob/b4ec94039a78faa4dc534c20c50b6d08519f4e38/nilearn/regions/signal_extraction.py#L198). But then we would need the approval of the nilearn folks.

Or maybe I'm missing something really basic and there's some python magick that would make this all really easy. I wonder if I can overload scipy.linalg.lstsq somehow.

chrisfilo:
Doing PCA to average signals over voxels would be nice to have in the future, but is not something we should focus on right now - especially if it means that we need to write a lot of custom code instead of reusing existing solutions. When it comes to needing PCA, it would be best to extend nilearn with this functionality if/when we need it (by send PR to nilearn). "Averaging" using PCA is also non trivial in the weighted case.

PS I would recommend you to play around with weighted mean and least square to get a better understanding of the two approaches.

shoshber:
I agree it would be best to extend nilearn for PCA in this case, and that it's not worth doing right at the moment.

I'll close this pull request and open one for the nilearn implementation--sorry for all the back and forth.

PR #1647, opened 2016-09-18
ENH signal extraction interface

shoshber:
-> using nilearn
-> incl. partial voluming
-> supports multiple 3d label files
-> option for demeaning
From code review:
- [x] move to interfaces module
- [x] ability to extract signal from each region independently
- [x] correct global signal implementation
- [x] global signal documentation


nipype/algorithms/stats.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def _run_interface(self, runtime):
+        masker = self._process_inputs()
+
+        region_signals = masker.fit_transform(self.inputs.in_file)
+
+        output = np.vstack((self.inputs.class_labels, region_signals.astype(str)))
+
+        # save output
+        np.savetxt(self.inputs.out_file, output, fmt=b'%s', delimiter='\t')
```
chrisfilo: Could you add a header row to this file based on the labels?

shoshber: yep, on line 71

chrisfilo: gotch ya!

shoshber:
issue #1628

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,136 @@
...
+        # check label list size
+        if len(self.inputs.class_labels) != n_labels:
+            raise ValueError('The length of class_labels {} does not '
+                             'match the number of regions {} found in '
+                             'label_files {}'.format(self.inputs.class_labels,
+                                                    n_labels,
+                                                    self.inputs.label_files))
+
+        if self.inputs.include_global:
+            all_ones_mask = nb.Nifti1Image(np.ones(self._label_data_shape()), np.eye(4))
```
chrisfilo: This will use voxels outside of the brain! Please use the combined mask from all of the classes instead of np.ones.

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,136 @@
...
+                               'class_labels must be equal to the number of '
+                               'segments (background excluded). This list '
+                               'corresponds to the class labels in label_file '
+                               'in ascending order')
+    out_file = File('signals.tsv', usedefault=True, exists=False,
+                    mandatory=False, desc='The name of the file to output to. '
+                    'signals.tsv by default')
+    include_global = traits.Bool(False, usedefault=True, mandatory=False,
+                                 desc='If True, include an extra column '
+                                 'labeled "global"')
```
chrisfilo: Please add an explanation how the values in this column would be calculated.

shoshber:
I can't see Chris's most recent review -_- github get it together!

nipype/algorithms/stats.py
shoshber: d'oh!

shoshber:
@chrisfilo take a look at the changes, tell me what you think

nipype/interfaces/nilearn.py
```diff
@@ -0,0 +1,138 @@
...
+    incl_shared_variance = traits.Bool(True, usedefault=True, mandatory=False, desc='By default '
+                                       '(True), returns simple time series calculated from each '
+                                       'region independently (e.g., for noise regression). If '
+                                       'False, returns unique signals for each region, discarding '
+                                       'shared variance (e.g., for connectivity)')
+    include_global = traits.Bool(False, usedefault=True, mandatory=False,
+                                 desc='If True, include an extra column '
+                                 'labeled "global", with values calculated from the entire brain '
+                                 '(instead of just regions). Only has effect with 4D probability '
+                                 'maps.')
```
chrisfilo: Why wouldn't we be able to do the same with a 3-d label image?

nipype/interfaces/nilearn.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # determine form of label files, choose appropriate nilearn masker
+        if np.amax(label_data.get_data()) > 1: # 3d label file
+            n_labels = np.amax(label_data.get_data())
+            maskers.append(nl.NiftiLabelsMasker(label_data))
+        else: # 4d labels
+            n_labels = label_data.get_data().shape[3]
+            if self.inputs.incl_shared_variance: # 4d labels, independent computation
+                for img in nli.iter_img(label_data):
+                    sortof_4d_img = nb.Nifti1Image(img.get_data()[:, :, :, np.newaxis], np.eye(4))
```
chrisfilo: Why is this necessary? If so please use the original image affine (`img.affine`) instead of `np.eye(4)`

nipype/interfaces/nilearn.py
```diff
@@ -0,0 +1,138 @@
...
+                             'label_files {}'.format(self.inputs.class_labels,
+                                                    n_labels,
+                                                    self.inputs.label_files))
+
+        if self.inputs.include_global:
+            global_label_data = label_data.get_data().sum(axis=3)
+            global_label_data = np.rint(global_label_data).astype(int).clip(0, 1)
+            global_label_data = global_label_data[:, :, :, np.newaxis] # add back 4th dimension
+            global_label_data = nb.Nifti1Image(global_label_data, np.eye(4))
+            global_masker = nl.NiftiMapsMasker(global_label_data, detrend=self.inputs.detrend)
```
chrisfilo: Since the global mask is binary wouldn't it make more sense to use NiftiLabelMasker here?

nipype/interfaces/nilearn.py
shoshber: Good to know about `img.affine`. `NiftiMapsMasker` complains if you input a 3-D nifti file, so that is my circuitous way of adding a 4th dimension (length 1). LMK if there is a native way to do it, I couldn't find one.

nipype/interfaces/nilearn.py
shoshber: Good catch, that edit was supposed to be for `incl_shared variance`

nipype/interfaces/nilearn.py
shoshber: The result is exactly the same as far as I can tell, so I changed it. What is the reason to prefer NiftiLabelMasker?

nipype/interfaces/tests/test_nilearn.py
```diff
@@ -0,0 +1,199 @@
...
+import tempfile
+import shutil
+
+import numpy as np
+
+from ...testing import (assert_equal, utils, assert_almost_equal, raises,
+                        skipif)
+from .. import nilearn as iface
+
+no_nilearn = True
```
satra: why is this check here and not in the nilearn interfaces file? see the nipy or dipy interfaces.

the check should be used inside run interface to indicate to the user that they need to install nilearn. at this point if they use this interface, it will result in an error deep in the code.

nipype/interfaces/tests/test_nilearn.py
```diff
@@ -0,0 +1,199 @@
...
+try:
+    __import__('nilearn')
```
satra: why this form of import - out of curiosity?

satra:
@chrisfilo - on future merges, please update the CHANGES file after a merge. see #1658

nipype/interfaces/tests/test_nilearn.py
shoshber: I was using the recent dvars testing file as a model (https://github.com/nipy/nipype/blob/master/nipype/algorithms/tests/test_confounds.py), which used nitime. I'm guessing that needs to be fixed too? Since this and the dvars interfaces have already been merged, I opened a separate issue for this: https://github.com/nipy/nipype/issues/1661

nipype/interfaces/tests/test_nilearn.py
shoshber: AFAIK, `import nilearn` would actually perform the import, while `__import__('nilearn')` only checks that `nilearn` is importable. Since `nilearn` wasn't used at all in this file, that seemed more appropriate. I don't feel strongly about it though.

Issue #1646, opened 2016-09-18
add demeaning and axis input to regress_poly function


satra:
the regress poly function can be made a bit more generic

https://github.com/nipy/nipype/blob/master/nipype/algorithms/confounds.py#L516

for example the following could be the default signature:

`def regress_poly(degree, remove_mean=False, axis=-1):`

shoshber:
closed by PR #1651

PR #1651, opened 2016-09-19
ENH Make regresspoly more universal

shoshber:
issue https://github.com/nipy/nipype/issues/1646

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -19,8 +19,8 @@ import os.path as op
 
 import nibabel as nb
 import numpy as np
+from numpy.polynomial import Legendre
 from scipy import linalg, signal
-from scipy.special import legendre
 
 from .. import logging
 from ..external.due import BibTeX
@@ -337,10 +337,8 @@ class CompCor(BaseInterface):
         # from paper:
         # "The constant and linear trends of the columns in the matrix M were
         # removed [prior to ...]"
-        if self.inputs.use_regress_poly:
-            voxel_timecourses = regress_poly(self.inputs.regress_poly_degree,
-                                             voxel_timecourses)
-        voxel_timecourses = regress_poly(0, voxel_timecourses, remove_mean=True)
+        degree = self.inputs.regress_poly_degree if self.inputs.use_regress_poly else 0
+        voxel_timecourses = regress_poly(degree, voxel_timecourses)
 
         # "Voxel time series from the noise ROI (either anatomical or tSTD) were
         # placed in a matrix M of size Nxm, with time along the row dimension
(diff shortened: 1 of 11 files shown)


nipype/algorithms/confounds.py
```diff
@@ -513,12 +509,15 @@ def _list_outputs(self):
...
+def regress_poly(degree, data, remove_mean=False, axis=-1):
     ''' returns data with degree polynomial regressed out.
-    The last dimension (i.e. data.shape[-1]) should be time.
+    Be default it is calculated along the last axis (usu. time)
     '''
     datashape = data.shape
-    timepoints = datashape[-1]
+    timepoints = datashape[axis]
+
+    if remove_mean: # i.e. regress_poly degree 0, which the following does not do
```
satra: actually the code for `regress_poly` was already estimating the beta for the mean, so if `remove_mean` is `True` one only needs to keep that beta in. see next comment.

satra:
```
if remove_mean:
    datahat = X[:, 1:].dot(betas[1:, ...]).T
else
    datahat = X.dot(betas).T
```

also in this function `if remove_mean and degree == 0` should raise an error - they are equivalently the same thing.

satra: also there is a polynomial package now in numpy, which we can use.

`from numpy.polynomial import Legendre` and then instead of `legendre(i)` use `Legendre.basis(i)`

this requires numpy 1.4 or greater

shoshber: Based on the second comment, I changed the default signature of regress_poly to be more intuitive (remove_mean defaults to True).

shoshber: Based on the third comment, I updated the code to use numpy.polynomial.Legendre.basis

shoshber: I just found a note in numpy's docs that Legendre.basis is new for numpy 1.7--should I change it back/upgrade our minimum numpy version in requirements.txt?

Regarding the first comment:
Ahh, I did not get that the first layer of `X` was all ones for a reason. I thought it was just a meaningless layer used to instantiate the array! I guess the writer did it that way because Legendre is time-expensive.

satra: @shoshber - my conditional is incorrect, it should be the other way around. 

`remove_mean`  should imply "please remove the mean signal!"

nipype/interfaces/base.py
```diff
@@ -1366,10 +1366,10 @@ def _get_ram_mb(pid, pyfunc=False):
         mem_mb = parent_mem/_MB
 
         # Iterate through child processes
-        for child in parent.children(recursive=False):
+        for child in parent.children(recursive=True):
```
oesteban: why don't you switch runtime_profiler off, because these changes are derived from that, right?

shoshber:
Switching runtime profiler off for circle and Travis is a larger
decision--and the other profiling tests seem to be fine. Please see the
corresponding issue (linked above) with a plan to get this test back online
when the bug is fixed.

shoshber:
Oh, wrong PR, sorry. That change should have been reverted for this pr

nipype/interfaces/base.py
shoshber: This was a reversion of the previous commit

shoshber:
@satra

PR #1653, opened 2016-09-20
Update dti.py

forwho:
Fix the bug of diffusion_toolkit interface.the attribute random_seed defined in the DTITrackerInputSpec Class has some mistake,the defination should be 
`random_seed = traits.Int(desc="use random location in a voxel instead of the center of the voxel to seed. can also define number of seed per voxel. default is 1", argstr="-rseed %d")`
not 
`random_seed = traits.Int(desc="use random location in a voxel instead of the center of the voxel to seed. can also define number of seed per voxel. default is 1", argstr="-rseed")`


mgxd:
@forwho can you merge with current master? and run `make specs`?

djarecka:
@forwho - thanks for the PR,  it needs to be merge with current master and auto tests should be updated. Do you need help with it?

djarecka:
@forwho - please let us know if you're planning to work on this PR, otherwise I'll try to review and finish.

PR #1654, opened 2016-09-20
wrapped ANTs function CreateJacobianDeterminantImage

dalejn:
https://neurostars.org/p/4236/


satra:
@dalejn - could you please merge with current master and push here.

PR #1656, opened 2016-09-21
[FIX] Several fixes related to unicode literals

### LINKED ISSUE #1644: [BUG] Parameterization in python 2 contains 'u' prefixes in parameter names
This is probably derived from using some sort of ```__repr__```

### LINKED ISSUE #1655: [BUG] Help string always contains References::
Even though ```self.references_``` is empty.

oesteban:
- Close #1655: fixes the References:: leftover in help() of interfaces
- Close #1644: fixes the parameterized paths generated with python 2
- All uses of ```repr()``` and re-definitions of ```__repr__``` have been revised
  (again, this is related to #1644, #1621, etc).

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -30,7 +30,7 @@ package_check('networkx', '1.3')
 
 import networkx as nx
 
-from ...utils.filemanip import (fname_presuffix, FileNotFoundError,
+from ...utils.filemanip import (fname_presuffix, FileNotFoundError, to_str,
                                 filename_to_list, get_related_files)
 from ...utils.misc import create_function_from_source, str2bool
 from ...interfaces.base import (CommandLine, isdefined, Undefined,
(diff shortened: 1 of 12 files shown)


nipype/pipeline/engine/utils.py
```diff
@@ -329,6 +330,8 @@ def _get_valid_pathstr(pathstr):
     Removes:  [][ (){}?:<>#!|"';]
     Replaces: ',' -> '.'
     """
+    if not isinstance(pathstr, (str, bytes)):
+        pathstr = to_str(pathstr)
```
oesteban: @satra, pathstr was converted to string using the ```__str__``` member. However, for lists, tuples, dict, etc, they fall back to ```repr```.  After adding unicode_literals everywere, ```repr('some string')``` is ```u'some string``` with the u. This u will be there after the pathstr is fixed for not valid characters.

oesteban:
@satra @chrisfilo this is ready and very much needed, please take a look

PR #1665, opened 2016-09-27
Apply verbose to ants command if terminal output is sat

Danielhiversen:
The antsRegistration command need --verbose 1 to give a terminal output.
So the terminal_output parameter is currently not working.
This PR will add --verbose 1 as a parameter if terminal_output is sat to 'stream', 'allatonce' or 'file'.

The tests are not updated yet

### DIFF
--- a/nipype/interfaces/ants/registration.py
+++ b/nipype/interfaces/ants/registration.py
@@ -216,6 +216,14 @@ class ANTS(ANTSCommand):
 
 
 class RegistrationInputSpec(ANTSCommandInputSpec):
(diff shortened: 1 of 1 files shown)


blakedewey:
@Danielhiversen This is something I have wanted but not had time to implement. It was left off a PR of mine some time ago because it would require version checking antsRegistration to ensure that the --verbose option is available. If you would like to do this please feel free.

Danielhiversen:
@blakedewey That is why I closed the PR. I am not sure how I should check if the antsRegistration version supports verbose?

PR #1669, opened 2016-10-04
fix: afni retroicor

mgxd:
make out_file optional, check if at least one of resp or card are defined, vim presets updated


oesteban:
Please merge with upstream/master, the download problem has been after your last commit. Thanks!

PR #1671, opened 2016-10-05
[FIX] Minor errors after migration to setuptools

### LINKED ISSUE #1670: nipype nosetests documentation broken
either we have to ask everyone to do their tests using docker containers, or we need to ensure the relevant nose components are present. for @djarecka and myself running the tests with instructions in the docs gave unicode errors.

a `pip install doctest-ignore-unicode` fixed the issue. based on our new setup.py perhaps we should update the install docs.

i've asked @djarecka to look into py.test conversion and support as part of her learning nipype.

oesteban:
- [x] Fixes #1670
- [x] Make sure matplotlib default backend is Agg in docker images
- [x] Do not use ```nipype.testing.assert_equal``` in ```nipype.interfaces.utility```
- [x] update new fsl_course_data link in docs


oesteban:
I am also going to use OSF for hosting the test data files, since google drive is shutting down the api for direct downloads.

oesteban:
@satra - please check these little changes and merge if ok

satra:
@oesteban - we should update the test command in the install docs as well. i can merge it right after that change since i know the tests pass. next up will be fixing doc building on readthedocs. we may need two separate requirements files.

oesteban:
@satra - I updated the makefile so that it runs the appropriate test command and also updated the documentation.

PR #1673, opened 2016-10-05
ENH: resets commandline arguments to default if set to Undefined

ellisdg:
This PR modifies the _parse_args method of the CommandLine class so that arguments are set to their default value if they were previously set to Undefined. This could be helpful when an upstream node passes an Undefined output to a CommandLine node input that has a default value. In nipype, as it stands now, the default value would be overwritten by the Undefined output from the upstream node. With this PR, the _parse_args method would realize that the input has been set to Undefined and then set that input back to its default value.

On the other hand, if the feature of being able to set arguments with a default value to Undefined is desired, this PR is a bad idea.


blakedewey:
This would save me a lot of hacking! I like it.

effigies:
So I'm in a position where it may actually be useful to disable an input, but there are undoubtedly scripts that depend on the implicit behavior encoded in nipype's choice of defaults. That said, it seems like reasonable behavior to have a connected port and assume that pushing `Undefined` is the same as not connecting.

How about this approach:

```Python
class _Disabled(_Undefined):
    _obj = None
    def __new__(cls):
        if cls._obj is None:
            cls._obj = object.__new__(cls)
        return cls._obj

Disabled = _Disabled()

def isdisabled(obj):
    return isinstance(obj, _Disabled)
```

Then:

```Python
>>> Disabled is Undefined
False
>>> (isdefined(Disabled), isdisabled(Disabled))
(False, True)
>>> (isdefined(Undefined), isdisabled(Undefined))
(False, False)
```

Then we could modify your check to:

```Python
if spec.usedefault and not isdisabled(value):
```

Alternately, we could turn it around and keep *all* existing behavior and create a `_Default` subclass of `_Undefined`, which would have the behavior you want.

satra:
question: how do we distinguish between intentional and non-intentional `Undefined`? so if i used Nipype to set a list of `iterables` such as `[Undefined, 1, 2, 3]` for a parameter that has a default value, is the expectation that the `Undefined` setting resets to default or not?

practically, we can do this using an execution configuration flag in the set_input function for a node. if the value that is being set is `Undefined`,  then skip  and do not set the corresponding input. also since config options can be global or node specific, this can be turned on/off on specific nodes.

ps. in the upcoming move from traits to traitlets (@djarecka is working on it), there is no semantically equivalent `Undefined` in traitlets. Instead we will be using None to indicate a value has not been set. but the same issue as raised here will come up there as well.

effigies:
I agree that we need to distinguish those, which is what my comment was trying to address with a special value which could have alternative semantics.

But assuming we want to distinguish them, I think the pertinent question becomes "What do I mean when I intentionally undefine?" And I think this has to be answered (at least by default) with what nipype has historically done when I purposely set `Undefined`, because we can't know what behavior people have come to depend on. So in the case of the traitlets approach, `None` will need to continue to override defaults.

That said, I think there is value in a separate special value that means "Use the default value even though this port is connected." But is there any way to create such a value in traitlets that won't trip the type constraints? Failing that, a per-node configuration flag for how to interpret `None` seems reasonable.

effigies:

PR #1676, opened 2016-10-06
WIP logging and validations in CompCor, SignalExtraction, ApplyTopUp

shoshber:


satra:
@shoshber - there seems to be a lot of changes happening here related to unicode drops. perhaps coming from a python 3 environment and make specs. can you do a make specs on the branch in a python 2 environment?

nipype/interfaces/fsl/epi.py
```diff
@@ -354,14 +354,14 @@ class ApplyTOPUP(FSLCommand):
 
     >>> from nipype.interfaces.fsl import ApplyTOPUP
     >>> applytopup = ApplyTOPUP()
-    >>> applytopup.inputs.in_files = ["epi.nii", "epi_rev.nii"]
+    >>> applytopup.inputs.in_files = ["ds003_sub-01_mc.nii.gz", "ds003_sub-01_mc.nii.gz"]
```
oesteban: why this? epi.nii and epi_rev.nii are empty files in testing/data

nipype/algorithms/confounds.py
```diff
@@ -402,6 +409,11 @@ class TCompCor(CompCor):
     def _run_interface(self, runtime):
         imgseries = nb.load(self.inputs.realigned_file).get_data()
 
+        if imgseries.ndim != 4:
```
oesteban: Why here you are using the numpy array ```.ndim``` property, and you have changed the same in your modification of ApplyTopup to use the nibabel header object? Which one is the appropriate option? If both are equivalent, I prefer this (numpy.ndim) rather than your change to ApplyTopup.

nipype/interfaces/fsl/epi.py
```diff
@@ -373,6 +373,12 @@ def _parse_inputs(self, skip=None):
         if skip is None:
             skip = []
 
+        for filename in self.inputs.in_files:
+            ndims = nib.load(filename).header['dim'][0]
```
oesteban: see comment above

oesteban:
For some reason, test_auto_CompCor.py got deleted in #1678.

satra:
@oesteban - since there are direct tests for CompCor `test_compcorr`, make specs deletes the auto test, which i think is fine.

shoshber:
-[ ] run `make specs` in python2
-[ ] consistent dimension checking

Regarding checking the dimensions of nifti files, how useful/difficult would it be to make a 4DNiftiFile trait? Are we worried about performance issues if we check dimensionality all the time?

nipype/algorithms/confounds.py
shoshber: The difference is that using numpy requires loading the data into numpy memory, which is expensive. In fact, in ApplyTopUp, it was erroring out. In TCompCor we are loading data into numpy memory anyway, so it is not an additional cost.

nipype/interfaces/ants/resampling.py
```diff
@@ -246,9 +246,13 @@ class ApplyTransformsInputSpec(ANTSCommandInputSpec):
     interpolation_parameters = traits.Either(traits.Tuple(traits.Int()),  # BSpline (order)
                                              traits.Tuple(traits.Float(),  # Gaussian/MultiLabel (sigma, alpha)
                                                           traits.Float())
-                                             )
-    transforms = InputMultiPath(
-        File(exists=True), argstr='%s', mandatory=True, desc='transform files: will be applied in reverse order. For example, the last specified transform will be applied first')
+                                            )
+    transforms = traits.Either(InputMultiPath(File(exists=True), desc='transform files: will be '
```
shoshber: I wanted to be able to pass an identity transform (use case: ds005 in fmriprep). This way, one can pass either a list of transform files OR the keyword 'identity'. It makes the code much cleaner in fmriprep, where I can say "transform the epi files with transformation ___"  for ds005 AND ds054, so I don't have to special-case. In some ways it is a workaround for not having conditional workflows in nipype.

nipype/interfaces/fsl/epi.py
shoshber: Bc I added the check for dimensionality, empty files were causing the tests to fail. `ds003_sub-01_mc.nii.gz` is an already-existing file in `testing/data` that has the appropriate dimensions. It is a little hacky but consistent with the rest of nipype's testing infrastructure.

blakedewey:
Are we sure we want to be doing these kinds of checks on CommandLine-derived interfaces? Maybe I missed the motivation for this but it seems like this kind of input checking should be left to the command that is being called. Does applytopup not give a useful error in this case? We have lots of interfaces here that require their input files to be in very specific forms but they are not checked to my knowledge. I can see doing it in a Pure Python module where no other checking is done. Thoughts?

nipype/interfaces/ants/resampling.py
oesteban: I think this is too specific of fmriprep to be included in nipype.

I propose you to create in fmriprep a wrapper of this interface adding this new 'identity' option, instead of including this modification in nipype, that will affect a lot of users. WDYT?

blakedewey: If you don't pass a --transforms option to antsApplyTransforms it uses the identity. Would this be a better option?

oesteban:
@blakedewey I think you are right in general, but _only_ for the case of TOPUP and ApplyTopup I would lean towards @shoshber's solutions. The reason is as follows: both interfaces have very specific inputs, with a lot of caveats and particular details (for example, TOPUP would not accept images with an odd number of voxels in x, y or z). The second reason is, as you mentioned, the errors that come out. These two tools have some meaningful crashes, but not always. If you play around the field you input (i.e. not generated with TOPUP) or the encoding file, etc. some errors are more like just a segfault. It is nice to have some extra checking on these (or at least, give more interpretable errors).

shoshber:
@blakedewey @oesteban @chrisfilo The reason I felt comfortable checking for 4-dimensionality here is that the description for `in_files` in the input spec specified 4-D niftis. However, I just looked in more detail at the FSL documentation and it is indeed valid, sometimes, for the inputs to be 3-dimensional. (http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/topup/ApplytopupUsersGuide)

In our case, we inputted a list of 3-dimensional nifti and received 3-D nifti file back, with no error whatsoever, which means some validation somewhere is necessary, though perhaps a more complicated one. Whether it should be a global (i.e. in nipype) validation or not is indeed a good question.

I await others' thoughts with bated breath.

nipype/interfaces/fsl/epi.py
oesteban: Oh I see. The ```_parse_inputs``` method is probably not the best place to this kind of check. I think it would be better placed the ```_run_interface``` method. Something like:

```
    ... your inputs sanity checks ...
    return super(ApplyTopup, self).__run_interface(runtime)
```
This way the doctests will pass with empty files and we don't alter the expected behavior of _parse_inputs.

blakedewey:
Yes I have used applytopup on 3D files. Honestly, I don't know anyone that can describe exactly what applytopup and topup will allow, every once in a while I get an error and have to look it up on the FSL mailing list. 

That being said, if you put in a 3D file for applytopup you should get a 3D file back (it is only performing the unwarping). 

I would be against global validation as the number of variations to cover would be extreme. For example, we aren't even limited to NIfTI files for FSL (although this is unlikely). Other ITK-based commands are even less limited than FSL.

nipype/interfaces/ants/resampling.py
shoshber: @blakedewey I think it is equivalent to what I did.
@oesteban I designed it to be backwards compatible. Did I miss something? I figured it is a change anyone solving a similar problem would appreciate.

shoshber:
I believe that this should be ready to merge, if @oesteban is satisfied with my answer above regarding identity transforms.

satra:
@shoshber - LGTM 

@oesteban - another pair of eyes would be helpful.

oesteban:
I think this is quite ready. My unique concern is the modification of the ants.Resample interface. It'd be worth checking @blakedewey's solution first (not passing any transform is understood as the identity transform), before we modify that interface.

Other than that there is an open directly from built-in. Tests are passing, but it doesn't hurt making sure that we enforce compatibility (and the solution is really quick).

Wrapping up: good job @shoshber !

shoshber:
> My unique concern is the modification of the ants.Resample interface. It'd be worth checking @blakedewey's solution first (not passing any transform is understood as the identity transform), before we modify that interface.

It turns out this isn't necessary for fmriprep, so I can take it out if you like.

> Other than that there is an open directly from built-in. Tests are passing, but it doesn't hurt making sure that we enforce compatibility (and the solution is really quick).

Could you tell me more about this?

oesteban:
@shoshber - regarding the open, there was a comment on that line, with a link to a reference on how to do it. Regarding the ants interface, I think we are better off not touching things that have been doing well this far.

nipype/algorithms/confounds.py
```diff
@@ -585,7 +626,7 @@ def compute_dvars(in_file, in_mask, remove_zerovariance=False):
 
     if remove_zerovariance:
```
oesteban: We took a closer look on this (w. @chrisfilo) and I realized that you were 100% right about this. Sorry about that.

Since this ```remove_zerovariance``` input does not make any sense, let's get rid of it (including the ```zero_remove``` function that is more readable inline).

Again, my apologies for not looking at this more thoroughly before.

nipype/algorithms/confounds.py
```diff
@@ -609,31 +650,28 @@ def compute_dvars(in_file, in_mask, remove_zerovariance=False):
...
-    func = func[idx[0], idx[1], idx[2], :]
-    tvariance = func.var(axis=1)
-    tv_mask = np.zeros_like(tvariance, dtype=np.uint8)
-    tv_mask[tvariance > 0] = 1
-
-    newmask = np.zeros_like(mask, dtype=np.uint8)
-    newmask[idx] = tv_mask
-    return newmask
+    new_mask = mask.copy()
+    new_mask[data == 0] = 0
```
oesteban: I think we can just do this directly on line 627

nipype/algorithms/tests/test_confounds.py
```diff
@@ -4,7 +4,9 @@
 from tempfile import mkdtemp
 from shutil import rmtree
 
-from nipype.testing import (assert_equal, example_data, skipif, assert_true)
+from io import open
+
+from nipype.testing import (assert_equal, example_data, skipif, assert_true, assert_in)
```
oesteban: :+1:

chrisfilo:
@shoshber are you planning to work on this more is is it ready to merge? (it still has "WIP" in the title).

PR #1689, opened 2016-10-20
fix semaphore acquire bug

carolFrohlich:
There was a bug on how the semaphore of the multiproc plugin was initialized and released.
The semaphore counter was initialized with 1, being acquired over and over again and never released.
That is because semaphore.acquire() was called more times than semaphore.release(), so the semaphore internal counter would always be > 0, leading to the master process to be active all the time.

We noticed that bug because the master process of the run was utilizing 100% cpu, even on the times it was supposed to be sleeping.

This fix makes the multiproc plugin have the expected behavior: the main process sleeps when it is waiting for jobs to finish.
Besides that, the multiproc plugin still works exactly the same as before.

### DIFF
--- a/nipype/pipeline/plugins/multiproc.py
+++ b/nipype/pipeline/plugins/multiproc.py
@@ -160,7 +160,6 @@ class MultiProcPlugin(DistributedPluginBase):
     def _wait(self):
         if len(self.pending_tasks) > 0:
             semaphore_singleton.semaphore.acquire()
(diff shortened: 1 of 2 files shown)


blakedewey:
+1 for this. Noticed this recently and it completely changed how I have to run my processes. I couldn't narrow it down though.

chrisfilo:
LGTM, but let's wait for the tests to finish. Thanks for sending this!

PR #1694, opened 2016-10-31
update PETPVC reference

alexsavio:
Updating the reference to PETPVC with their new publication.


alexsavio:
This is only updating the reference. The error shown by CircleCI is because of a timeout that happened LAST YEAR! :). I think this problem have already been solved by the latest changes?

PR #1696, opened 2016-10-31
Fixed type referencing non-existent attribute in QJobInfo

blakedewey:
This is an artifact from #1656. Causes an AttributeError when zombie processes are found.


blakedewey:
@oesteban, can you check to make sure that this does not hurt what you did. I think it is a simple typo, but wanted to make sure.

PR #1697, opened 2016-10-31
BF: install requires configparser

yarikoptic:
otherwise - kaboom:
```shell
$> virtualenv venv-tests
Running virtualenv with interpreter /usr/bin/python2
New python executable in /tmp/venv-tests/bin/python2
Also creating executable in /tmp/venv-tests/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
2 16589.....................................:Mon 31 Oct 2016 12:36:49 PM EDT:.
hopa:/tmp
$> source venv-tests/bin/activate
(venv-tests) 2 16590.....................................:Mon 31 Oct 2016 12:36:51 PM EDT:.
hopa:/tmp
$> pip install nipype
Collecting nipype
  Using cached nipype-0.12.1-py2.py3-none-any.whl
Collecting nibabel>=2.0.1 (from nipype)
Collecting traits>=4.3 (from nipype)
Collecting future>=0.15.2 (from nipype)
Collecting scipy>=0.11 (from nipype)
  Using cached scipy-0.18.1-cp27-cp27mu-manylinux1_x86_64.whl
Collecting xvfbwrapper (from nipype)
Collecting numpy>=1.6.2 (from nipype)
  Using cached numpy-1.11.2-cp27-cp27mu-manylinux1_x86_64.whl
Collecting python-dateutil>=1.5 (from nipype)
  Using cached python_dateutil-2.5.3-py2.py3-none-any.whl
Collecting prov>=1.4.0 (from nipype)
  Using cached prov-1.5.0-py2.py3-none-any.whl
Collecting simplejson>=3.8.0 (from nipype)
Collecting mock (from nipype)
  Using cached mock-2.0.0-py2.py3-none-any.whl
Collecting networkx>=1.7 (from nipype)
  Using cached networkx-1.11-py2.py3-none-any.whl
Collecting nose>=1.2 (from nipype)
  Using cached nose-1.3.7-py2-none-any.whl
Collecting six>=1.5 (from python-dateutil>=1.5->nipype)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting lxml (from prov>=1.4.0->nipype)
  Using cached lxml-3.6.4-cp27-cp27mu-manylinux1_x86_64.whl
Collecting rdflib>=4.2.1 (from prov>=1.4.0->nipype)
Collecting funcsigs>=1; python_version < "3.3" (from mock->nipype)
  Using cached funcsigs-1.0.2-py2.py3-none-any.whl
[...]

PR #1698, opened 2016-10-31
[FIX] Install requires configparser. Close #1697

oesteban:


oesteban:
Sorry, just realized that @yarikoptic had submitted a PR, not an issue. Closing this, as it is duplicated.

PR #1700, opened 2016-11-02
Add support for uncode filenames to filename_to_list()  in filemanip.py

danweflen:
Filemanip.py's filename_to_list() used to return None when passed a filename that is a unicode string. This commit allows it to behave correctly (return a list containing a unicode string) when passed a unicode string as a filename.

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -428,8 +428,8 @@ def copyfiles(filelist, dest, copy=False, create_new=False):
 def filename_to_list(filename):
     """Returns a list given either a string or a list
     """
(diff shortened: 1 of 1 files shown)


oesteban:
@danweflen - ```unicode``` has been deprecated, and that code fails in Python 3. All the string types should be correctly understand in that check since ```str``` and ```bytes``` are imported from ```future```. Could you provide a minimally working example of this issue?

djarecka:
@danweflen - are you still having any issues with strings? If yes, can you please provide an example.

djarecka:
@danweflen I'll close this PR for now, please reopen or create a new issue if you still have this problem.

PR #1702, opened 2016-11-04
ENH: Allows ReconAll.inputs.flags to take in a list of flags

ellisdg:
Allows for the flags input to ReconAll interface to be either a string or a list of strings. This makes it possible to set multiple flags on ReconAll without having to first convert them to a string as is the case here: https://neurostars.org/p/4314/#4320

### DIFF
--- a/nipype/interfaces/freesurfer/preprocess.py
+++ b/nipype/interfaces/freesurfer/preprocess.py
@@ -638,7 +638,7 @@ class ReconAllInputSpec(CommandLineInputSpec):
                   desc="Set parameters using expert file")
     subjects_dir = Directory(exists=True, argstr='-sd %s', hash_files=False,
                              desc='path to subjects directory', genfile=True)
-    flags = traits.Str(argstr='%s', desc='additional parameters')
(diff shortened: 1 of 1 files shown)


nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -624,7 +624,7 @@ class ReconAllInputSpec(CommandLineInputSpec):
                         desc="Number of processors to use in parallel")
     subjects_dir = Directory(exists=True, argstr='-sd %s', hash_files=False,
                              desc='path to subjects directory', genfile=True)
-    flags = traits.Str(argstr='%s', desc='additional parameters')
+    flags = traits.Either(traits.Str, traits.List, argstr='%s', desc='additional parameters')
```
satra: this is what we use `InputMultiPath` for . so the above would reduce to `InputMultiPath(Str, argstr=...)`

satra:
@effigies - if you approve this, please go ahead and merge. if you think this is redundant without your changes, let's close this.

effigies:
Looks reasonable to me.

PR #1708, opened 2016-11-09
fix: correctly join iterable expansion

mgxd:
This change ensures JoinNode connects only nodes with iterables (.a1, .a2, etc.).


mgxd:
after looking [at some of the tests](https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/tests/test_join.py#L288), this PR would do more harm than good..I'm closing this

PR #1693, opened 2016-10-29
Better FSL modelgen parameter handling and a few more EV file parameters

TheChymera:
The changes in parameter handling allow arguments to be dynamically selected from fields that are specified in the template files (e.g. `$whatever$`). This means that we can provide access in nipype to ALL EV file parameters just by editing the templates.

I also parameterized a few of the more interesting fields of the gamma template.

PR #1710, opened 2016-11-14
Better FSL modelgen parameter handling and a few more EV file parameters

TheChymera:
Rebased PR #1693


TheChymera:
As this PR's diff is quite wide-ranging, and as I cannot run the tests locally, I will be breaking this up into 2 PRs:

1) Parameter passing and parametrization: https://github.com/nipy/nipype/pull/1749

Issue #1667, opened 2016-09-30
Nipype and logging module of Python 2.7


sabfon:
I have also posted the question on Neurostars (http://neurostars.org/p/4239/).
I need to use Nipype for a program that I'm writing, but I've noticed a strange behavior when I import Nipype and the Python logging module together.
I'll try to explain my problem with more details.

```
import logging

if __name__ == '__main__':
    logging.basicConfig(filename='/path/log', level=logging.INFO)
    logging.warning('This is a Warning!')

```

With this simple piece of code, I obtain a log file containing the message "This is a warning".
If I import Nipype the situation changes.

```
import nipype
import logging

if __name__ == '__main__':
    logging.basicConfig(filename='/path/log', level=logging.INFO)
    logging.warning('This is a Warning!')
```

The log file is not created anymore and the program prints the message on the console. 

I've also tried something like 
`config.update_config({'logging': {'log_directory': '/path/log', 'log_to_file': True}})`

The problem is only for Python 2.7 (it works as expected on Python 3) and I'm using nipype 0.12.1.
I don't know if is a bug or maybe I'm doing something wrong and stupid.

Thank you for the help.

mgxd:
@sabfon I see you've tried updating the config - have you also tried updating [nipype's logging](http://nipype.readthedocs.io/en/0.12.1/users/config_file.html#enabling-logging-to-file)? 
```
from nipype import config, logging
config.update_config({'logging': {'log_directory': '/path/log',
                                  'log_to_file': True}})
logging.update_logging(config)
```

sabfon:
```
from nipype import config
from nipype import logging as nipype_logging
import logging as python_logging
import os

if __name__ == '__main__':

    config.update_config({'logging': {'log_directory': os.getcwd(), 'log_to_file': True}})
    nipype_logging.update_logging(config)


    python_logging.basicConfig(filename='/Users/sabfon/log', level=python_logging.INFO)
    python_logging.warning('hello')
```

I've tried this code, but still the python logging module doesn't work properly and starts to print everything on the console.

mgxd:
@sabfon this is really a case of defining logger's basicConfig first. Since nipype is also defining this when it's imported, you'll need to set this before importing nipype - however, this will disable the nipype logs.
```
import logging
logging.basicConfig(filename='/Users/sabfon/log', level=logging.INFO)
import nipype
```

anbai106:
@mgxd 
[...]

PR #1714, opened 2016-11-22
change for checking out FOV of T1

anbai106:
It is my first time to do the pull request, So i am not sure If i do it correctly, i just follow the steps from nipype..
Thanks in advance

### DIFF
--- a/nipype/workflows/smri/freesurfer/autorecon1.py
+++ b/nipype/workflows/smri/freesurfer/autorecon1.py
@@ -15,8 +15,10 @@ def checkT1s(T1_files, cw256=False):
     if len(T1_files) == 0:
         print("ERROR: No T1's Given")
         sys.exit(-1)
-
-    shape = nib.load(T1_files[0]).shape
+    f = nib.load((T1_files[0]))
+    shape = f.shape
+    voxel_size = f.header.get_zooms()
+    t1_size = f.header.get_data_shape()
     for t1 in T1_files[1:]:
         if nib.load(t1).shape != shape:
             print("ERROR: T1s not the same size. Cannot process {0} and {1} "
@@ -26,9 +28,11 @@ def checkT1s(T1_files, cw256=False):
     origvol_names = ["{0:03d}.mgz".format(i + 1) for i in range(len(T1_files))]
 
     # check if cw256 is set to crop the images if size is larger than 256
-    if not cw256 and any(dim > 256 for dim in shape):
+    if not cw256 and (voxel_size[0] * t1_size[0] > 256) or (voxel_size[1] * t1_size[1] > 256) or (voxel_size[2] * t1_size[2] > 256):
         print("Setting MRI Convert to crop images to 256 FOV")
         cw256 = True
+    else:
+        print("No need to add -cw256 flag")
 
     resample_type = 'cubic' if len(T1_files) > 1 else 'interpolate'
     return T1_files, cw256, resample_type, origvol_names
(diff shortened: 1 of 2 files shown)


oesteban:
I have quickly looked for the ```--cw256``` flag and did not see that it is specified in mm, therefore I guess it is specified in voxels. If the latter is the case, this PR would break the correct behavior of the interface.

Since I'm not an expert on freesurfer, please @satra, take a look.

nipype/workflows/smri/freesurfer/autorecon1.py
```diff
@@ -26,9 +28,11 @@ def checkT1s(T1_files, cw256=False):
     origvol_names = ["{0:03d}.mgz".format(i + 1) for i in range(len(T1_files))]
 
     # check if cw256 is set to crop the images if size is larger than 256
-    if not cw256 and any(dim > 256 for dim in shape):
+    if not cw256 and (voxel_size[0] * t1_size[0] > 256) or (voxel_size[1] * t1_size[1] > 256) or (voxel_size[2] * t1_size[2] > 256):
         print("Setting MRI Convert to crop images to 256 FOV")
         cw256 = True
+    else:
+        print("No need to add -cw256 flag")
```
anbai106: I have some data, which size of image is small than 256, but it still give me the error that we should use the cw256 flag, so I think the FOV is specified in mm, which corresponds to number of voxels multiply dimension of voxels. 
And for the nipype logging, I followed the site from nipype, and I know that in nipype, we can not use python logging module, which is a bug and you did not figure it out, and the nipyep logging [page()](http://nipype.readthedocs.io/en/latest/users/config_file.html) is not so clear to follow, so do you have some more elaborate page to follow???

nipype/workflows/smri/freesurfer/autorecon1.py
```diff
@@ -26,9 +28,11 @@ def checkT1s(T1_files, cw256=False):
     origvol_names = ["{0:03d}.mgz".format(i + 1) for i in range(len(T1_files))]
 
     # check if cw256 is set to crop the images if size is larger than 256
-    if not cw256 and any(dim > 256 for dim in shape):
+    if not cw256 and (voxel_size[0] * t1_size[0] > 256) or (voxel_size[1] * t1_size[1] > 256) or (voxel_size[2] * t1_size[2] > 256):
```
satra: @oesteban - that is correct.

nipype/workflows/smri/freesurfer/autorecon1.py
satra: was that data previously processed with `-cw256` flag? and would it be possible to put the `mri/orig/001.mgz` somewhere so we can check?

re: nipype logging - if there is no separate issue, could you please create one?

anbai106: For my data, for the first time that I ran recon-all, i got error that said to use -cw256, so in my pipeline I added this flag, this is the 001.mgz for one of my subjects: [https://www.dropbox.com/s/xpmlzlffhekese0/001.mgz?dl=0](url)

Sorry, i dont get the question for nipype logging, you mean I create another bug for nipype and python logging?
Thanks in advance

anbai106: Actually, For the logging bug, my colleague has posted this like 2 months ago, and you just closed this issue, here is the issue number [https://github.com/nipy/nipype/issues/1667](#1667)

effigies:

PR #1723, opened 2016-11-27
[FIX] FAST output files

shoshber:
NEW: `FAST._list_outputs()` was incorrectly listing output files in `cwd` when `out_basename` is not set. This PR fixes it.

OLD:The FAST interface is not outputting files into the correct directory when `out_basename` is not set. This solution feels hacky. I'm also concerned that this seemingly obvious bug hasn't been caught and fixed before now.

I'll fix up the tests once I have confidence this is the "right" solution.

### DIFF
--- a/nipype/interfaces/fsl/preprocess.py
+++ b/nipype/interfaces/fsl/preprocess.py
@@ -323,17 +323,20 @@ class FAST(FSLCommand):
             nclasses = self.inputs.number_classes
         # when using multichannel, results basename is based on last
         # input filename
+        _gen_fname_opts = {}
         if isdefined(self.inputs.out_basename):
-            basefile = self.inputs.out_basename
+            _gen_fname_opts['basename'] = self.inputs.out_basename
+            _gen_fname_opts['cwd'] = os.getcwd()
         else:
-            basefile = self.inputs.in_files[-1]
+            _gen_fname_opts['basename'] = self.inputs.in_files[-1]
+            _gen_fname_opts['cwd'], _, _ = split_filename(_gen_fname_opts['basename'])
 
-        outputs['tissue_class_map'] = self._gen_fname(basefile, suffix='_seg')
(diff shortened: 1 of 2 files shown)


nipype/interfaces/fsl/preprocess.py
```diff
@@ -315,6 +315,15 @@ def _format_arg(self, name, spec, value):
             formated = "-S %d %s" % (len(value), formated)
         return formated
 
+    def _parse_inputs(self, skip=None):
+        ''' ensures out_basename (argstring -o) always exists, otherwise fast
+        puts the outputs in the same folder as the inputs '''
```
satra: at the interface level, this is what it is supposed to do. within nipype we respect whatever the underlying interface does. 

however, at the workflow level, the files are copied/linked into the working directory, and fast should output the data in the correct place.

shoshber: So if I understand you correctly, this is the expected behavior of `fast`, so the FAST interface should follow this behavior. Then the problem is in `_list_outputs()`, which doesn't understand that behavior (and therefore the interface fails due to trait errors when it is run with `.run()`)

shoshber: thanks for clarifying the bug!

shoshber:
BTW since github collapsed the conversation I'm copying it here:
satra a day ago NIPY developers member
at the interface level, this is what it is supposed to do. within nipype we respect whatever the underlying interface does.

however, at the workflow level, the files are copied/linked into the working directory, and fast should output the data in the correct place.
       @shoshber
shoshber a day ago collaborator
So if I understand you correctly, this is the expected behavior of fast, so the FAST interface should follow this behavior. Then the problem is in _list_outputs(), which doesn't understand that behavior (and therefore the interface fails due to trait errors when it is run with .run())
       @satra
satra a day ago NIPY developers member
that would then be a bug.

PR #1724, opened 2016-11-29
changes slice_order type from Int to Float

miykael:
With the newer SPM12, it is now possible to use "slice onsets" in milliseconds, instead of the good old simple slice order.

I've tested if there is any difference between slice_order [0 1 2 3,...] or [0.0 1.0 2.0 3.0,...] and with my test data there was none.

SPM12 and SPM8, both accept the float point approach, but SPM8 crashes if one tries to use actual onsets, i.e. [0.0 66.66 133.33, 200.0...].

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -47,11 +47,13 @@ class SliceTimingInputSpec(SPMCommandInputSpec):
                                           'calculated as TR-(TR/num_slices)'),
                                     mandatory=True)
     slice_order = traits.List(traits.Float(), field='so',
-                              desc=('1-based order or onset in which slices are '
-                                    'acquired'),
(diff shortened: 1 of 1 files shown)


nipype/interfaces/spm/preprocess.py
```diff
@@ -46,8 +46,8 @@ class SliceTimingInputSpec(SPMCommandInputSpec):
                                     desc=('time of volume acquisition. usually'
                                           'calculated as TR-(TR/num_slices)'),
                                     mandatory=True)
-    slice_order = traits.List(traits.Int(), field='so',
-                              desc=('1-based order in which slices are '
+    slice_order = traits.List(traits.Float(), field='so',
+                              desc=('1-based order or onset in which slices are '
```
chrisfilo: could you confirm SPM requires onsets in milliseconds and make it explicit in this description?

Issue #1692, opened 2016-10-28
MulitProc Workflows locking up


blakedewey:
I think this may be a result of #1689. When I run a large MultiProc workflow, it runs for one or two nodes, then hangs. There is nothing reported and the nodes have finished successfully. I am also unable to use SIGTERM to end the process and have to manually close the terminal window. Moving back before this change seems to fix the issue.

satra:
@carolFrohlich - could you please take a look at this?

chrisfilo:
CCing @ccraddock who promised to help. We have experienced those as well. Deadlocks are stochastic in nature so there are no steps to reproduce. Here are nipype tests which timed out due to deadlocks in multiproc:

https://circleci.com/gh/nipy/nipype/1690
https://circleci.com/gh/nipy/nipype/1686
https://circleci.com/gh/nipy/nipype/1686
https://circleci.com/gh/nipy/nipype/1685
https://circleci.com/gh/nipy/nipype/1682
https://circleci.com/gh/nipy/nipype/1676
https://circleci.com/gh/nipy/nipype/1674

the problem seems to affect Python 2.7 more often than python 3.5

chrisfilo:
It seems to be affecting travis tests as well:

https://api.travis-ci.org/jobs/180558248/log.txt?deansi=true
```
nipype.pipeline.plugins.tests.test_linear.test_run_in_series([1, 1], [1, 1]) ... ok


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

On a [successful run](https://api.travis-ci.org/jobs/179964049/log.txt?deansi=true) the next test after `test_run_in_series` is `nipype.pipeline.plugins.tests.test_multiproc.test_run_multiproc`

djarecka:
As far as I remember my problems with travis were always in the second run only (python2.7; INSTALL_DEB_DEPENDECIES=false), e.g. https://travis-ci.org/djarecka/nipype/jobs/180111821
Checking now my CircleCI builds I also can't find problems with python3.

jvarada:
Is there a reason for using the multiprocessing library and handling locks for the main thread manually rather than using concurrent.futures (or futures in py2.7), sending threads out as Futures, and waiting on the Futures instead?

chrisfilo:
[...]

PR #1733, opened 2016-12-04
skip python2 multiproc tests until the deadlock issue will be resolved

chrisfilo:
Temporary measure until: https://github.com/nipy/nipype/issues/1692 is resolved

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -35,9 +35,9 @@ dependencies:
 test:
   override:
     - docker run -v /etc/localtime:/etc/localtime:ro -v ~/scratch:/scratch -w /root/src/nipype/doc nipype/nipype_test:py35 /usr/bin/run_builddocs.sh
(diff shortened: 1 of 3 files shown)


satra:
@chrisfilo - i'm worried that this doesn't solve the fundamental problem when people use multiproc. i'm going to take a look at the multiproc pieces, and specifically suppress the semaphore components and resource multiproc use.

chrisfilo:
This is not intended to be a solution but a temporary measure to restore
reliability of our CI tests.

chrisfilo:
To avoid confusing contributors with multiproc tests timing out I would recommend merging this ASAP and turning those test back on when solution will be found.

chrisfilo:
Happened to me just now on another PR: https://circleci.com/gh/nipy/nipype/1699

satra:
only a temporary merge - let's revert as soon as we figure out the issue. because many people still use multiproc to run things.

chrisfilo:
For sure!

PR #1739, opened 2016-12-07
multiproc test

carolFrohlich:
Please don't accept the pull request, this is just a test.


chrisfilo:
Just to make it clear - the issue reported in #1692 is not confined just to tests and is not a problem with tests themselves. It has also been reported by users in their production pipelines (I also have experienced it). The relevant multiproc tests should not be reenabled until a solution the the multiproc bug will be found.

ccraddock:
Do you have any examples that might help us reproduce the problem? We do not see these issues with our pipeline and were only able to reproduce it with nosetests.

PR #1741, opened 2016-12-08
[FIX] config setting of TOPUP should not usedefault

oesteban:
Default settings of TOPUP are really good. Setting this default b02b0.cnf config basically breaks any workflow including this interface.


chrisfilo:
auto tests :)

oesteban:
Ok, hold on on this. I am wrong:

> As part of the topup distribution we supply a predefined config file: b02b0.cnf. It contains parameters that have been found to be useful for registering a set of good quality b=0 images. Together with the override facility this is probably the starting (an quite possible finishing) point for most users. Note that b02b0.cnf uses sub-sampling to speed up the estimation and that topup requires that the image-size is a multiple of the sub-sampling level. So, for example if you want to use sub-sampling by a factor of 2 the image-size should be a multiple of 2 in all directions (e.g. 96x96x48).

PR #1742, opened 2016-12-10
[WIP]add new files to ants brain extraction output spec and list outputs

rwblair:
For now exposing all files generated by ants brain extraction. Of particular interest are the n4 bias correction and segmentation mask.


nipype/interfaces/ants/segmentation.py
```diff
@@ -701,7 +701,7 @@ def _list_outputs(self):
                                                        self.inputs.out_prefix +
                                                        'BrainExtractionBrain.' +
                                                        self.inputs.image_suffix)
-        if isdefined(self.inputs.keep_temporary_files) && self.inputs.keep_temporary_files != 0:
+        if isdefined(self.inputs.keep_temporary_files) and self.inputs.keep_temporary_files != 0:
```
oesteban: hehehe I was to say that I preferred an "and" (the word) there.

rwblair: thats what I get for having syntax highlighting turned off...

chrisfilo:
LGTM - anything else missing here?

satra:
@chrisfilo - we should start building out regression tests soon for interfaces - so we may ask people who augment or add interfaces to provide a regression test.

chrisfilo:
By that you mean a test that runs the software a particular interface is
wrapping? Running those will take a lot of time...

satra:
yes - essentially as part of repronim we want to start doing regression tests for the underlying software. these won't run by default, but we will be creating various containers to test the underlying tools across versions of the tools and there will be a dashboard reflecting these tests.

PR #1743, opened 2016-12-10
potential fixes

carolFrohlich:
We have narrowed the problem down pretty well. The completion signal is being received from the threadpool worker process, but once the distributedplugin base class's run loop checks the return value of the node, it appears to have not completed. The loop then puts itself back to sleep, but since the signal has already been sent, it will never be woken up again. This appears to be some weird interaction within the threadpool and how it calls async jobs. We have simplified the implementation of the multiproc plugin to remove as many places were deadlocking can occur, and it still happens every now and then. Best we can tell this is an issue within the threadpool itself, that has been fixed in python 3 by replacing a mutex lock with an event. 

To fix the symptom, we have switched the scheduling semaphore with an event so that we can have a timeout in the wait. So no more deadlock.

### DIFF
--- a/nipype/pipeline/plugins/base.py
+++ b/nipype/pipeline/plugins/base.py
@@ -218,7 +218,6 @@ class DistributedPluginBase(PluginBase):
         self.proc_done = None
         self.proc_pending = None
         self.max_jobs = np.inf
-        self.last_pending_task = []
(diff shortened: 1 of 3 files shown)


chrisfilo:
Closed in favour of #1754

PR #1745, opened 2016-12-11
Fixes for tCompCor

### LINKED ISSUE #1744: tCompCor ignores the input mask
### Summary
tCompCor does not restrict the percentile calculation to the input mask instead it ignores it.

### Actual behavior
Input mask is ignored.

### Expected behavior
Percentile calculation should be restricted to voxels withing the mask if a mask is provided.

Run tCompCor with different masks. Results should not change

### Platform details:
n/a

chrisfilo:
- input mask is not being ignored anymore (fixes #1744)
- use np.percentile instead of implementing it by hand
- output the estimated high variance mask (useful for debugging)

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -372,8 +372,8 @@ class CompCor(BaseInterface):
         outputs['components_file'] = os.path.abspath(self.inputs.components_file)
         return outputs
 
-    def _compute_tSTD(self, M, x):
-        stdM = np.std(M, axis=0)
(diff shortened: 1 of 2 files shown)


djarecka:
did you change  nipype/algorithms/tests/test_auto_TCompCor_test.py  running `python tools/checkspecs.py`? I'm getting a different file when running `python tools/checkspecs.py` or `make specs`:
https://github.com/djarecka/nipype/commit/ac65ea42b52988789382eff697e4b750db7e99e8

PR #1722, opened 2016-11-26
Pytest

djarecka:
I removed `nose` and `unittest` libraries; and rewrote the tests to `pytest`. I still kept `mock` library in a few tests. I tried to not change the tests coverage, they should tests exactly the same functionality. However,` py.test` stops execution after the first `assert` statement that fails. 

This is still not completely finished:
* there are some notes (search for "NOTE_dj") that should be removed, but just kept them for now to think and discuss. 
* there are 2 places that I marked as "FAIL", I had to comment them to not get Fails... not sure if I should change the tests or the tested function.
* a few tests are marked as `xfail`,  they were `skipif(True)` before.
* there are tests that are always skipped (I used the test docker containers)
* `circle.yml` runs `run_pytests.py` instead of `run_nosetests.py`, `run_pytests.py` needs to be improved

I also slightly adjusted docstrings so doctests with `pytest` can be run.

PR #1751, opened 2016-12-16
Pytest follow-up

djarecka:
This is a follow-up to my previous Pytes PR (#1722), this time I've read the Satra review :) 
* removed some files from tools directory that are not used anymore
* removed fixes
* changed .coveragerc (workflows is not omitted) 
* removed coverage from travis, so it is now calculated from CircleCI only 
* added one auto tests `test_auto_ACompCor.py` and changing `test_auto_TCompCor.py`, so it agrees with checkspec.py output
* recovered `test_BEDPOSTX.py` and `test_Xfibres.py`
* removed nosetes from `__init__.py` 
* changed one test from `test_math.py` that was failing
* some smaller changes suggested in the review 

Still TODO:
* tests `test_BEDPOSTX.py` and `test_Xfibres.py` don't affect anything IMO. Should I create `test_BEDPOSTX5.py` and `test_Xfibres5.py`, and delete  `test_auto_BEDPOSTX5.py` and `test_auto_Xfibres5.py`?
* `__init__.py` is not finished, `nipype.test()` doesn't work


nipype/interfaces/fsl/tests/test_maths.py
```diff
@@ -207,10 +207,8 @@ def test_stdimage(create_files_in_directory):
         assert stder.cmdline == cmdline%dim
 
     # Test the auto naming
-    stder = fsl.StdImage(in_file="a.nii")
-    #NOTE_dj, FAIL: this is failing (even the original version of the test with pytest)
-    #NOTE_dj: not sure if this should pass, it uses cmdline from interface.base.CommandLine
-    #assert stder.cmdline == "fslmaths a.nii -Tstd %s"%os.path.join(testdir, "a_std.nii")
+    stder = fsl.StdImage(in_file="a.nii", output_type='NIFTI')
+    assert stder.cmdline == "fslmaths a.nii -Tstd %s"%os.path.join(testdir, "a_std.nii")
```
satra: use `.format()`

satra:
@djarecka - `test_bedpostx` and `test_xfibres` simply prevent the autocreation of the auto tests for those which would duplicate the `test_auto_*5` at a point in the past we had two versions of these tools. so your current changes are fine and it doesn't need `test_*5.py`

satra:
@djarecka - if you can finish the `nipype.test()` this should be ready. i'll also send a PR to your branch with some updated other tests.

djarecka:
@satra even without `test_bedpostx` and `test_xfibres` auto tests are NOT created.

djarecka:
@satra I'm a bit stuck with `nipype.test()`. If you want to merge this ASAP, it might be better if I move `nipype.test()` to another PR, it somehow independent issue.

satra:
@djarecka - you are correct - the autotests are not created because they are the same exact class as the `*5.py` classes in this case. so yes, in this case those two tests can be deleted.

satra:
`test_BEDPOSTX.py` and `test_Xfibres.py`

djarecka:
@satra ok, I've removed the extra tests. Just wasn't sure, if you want to disable auto tests for `BEDPOSTX5` and `XFibres5`.

I also changed earlier `%s` to `format()`  in `test_math` in the places where I could do it without changing the code itself (you're using `%s` style in `nipype.interfaces.fsl.maths`)

djarecka:
@satra can you also check this assert? https://github.com/nipy/nipype/blob/master/nipype/interfaces/tests/test_base.py#L137
I've removed it since I'm only getting a UserWarning

satra:
@djarecka -  i can't seem to figure out why that should raise a `TypeError` - ok to remove for now.

satra:
regarding `.format` to slowly start moving towards current python standards - so a new change should always reflect current coding standards.

PR #1754, opened 2016-12-17
FIX MultiProc deadlock

ccraddock:
Reduced the complexity of the MultiProc processing to reduce likelihood for deadlock by:
- removing semaphores and replacing with event, this allows for a timeout in the wait that safeguards against deadlock
- setting results directly from apply_async callback rather than two stage processing, where a signal is sent from the callback, that wakes the multiproc.run() thread, which then sets the results -- this is where i think the previous deadlock was happening

Found deadlock where events would be cleared before they were handled, creating scenarios where thread waits for a signal that will never come -- fixed by changing the location from before the wait to after

Added missing close thread pool

### DIFF
--- a/nipype/pipeline/plugins/multiproc.py
+++ b/nipype/pipeline/plugins/multiproc.py
@@ -139,7 +139,7 @@ class MultiProcPlugin(DistributedPluginBase):
         self.processors = cpu_count()
         self.memory_gb = get_system_total_memory_gb()*0.9 # 90% of system memory
 
-        self._timout=2.0
+        self._timeout=2.0
         self._event = threading.Event()
 
 
--- a/nipype/pipeline/plugins/multiproc.py
+++ b/nipype/pipeline/plugins/multiproc.py
@@ -11,6 +11,7 @@ from builtins import open
 
 # Import packages
 from multiprocessing import Process, Pool, cpu_count, pool
+import threading
 from traceback import format_exception
 import sys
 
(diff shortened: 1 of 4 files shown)


nipype/pipeline/plugins/multiproc.py
```diff
@@ -178,44 +193,54 @@ def _report_crash(self, node, result=None):
...
+
     def _send_procs_to_workers(self, updatehash=False, graph=None):
         """ Sends jobs to workers when system resources are available.
             Check memory (gb) and cores usage before running jobs.
         """
         executing_now = []
 
         # Check to see if a job is available
-        jobids = np.flatnonzero((self.proc_pending == True) & \
+        currently_running_jobids = np.flatnonzero((self.proc_pending == True) & \
```
satra: these are not running yet.

chrisfilo:
There are few more places where tests need to be reenabled:
https://github.com/nipy/nipype/search?utf8=%E2%9C%93&q=issues%2F1692

nipype/pipeline/plugins/multiproc.py
ccraddock: according to this they should be, or am I confused?

https://github.com/nipy/nipype/blob/a746633d72d3e861a4b1a52038dd40a590792e24/nipype/pipeline/plugins/base.py#L205

satra: that should be revised.

the following line sets a flag when a job is submitted.

https://github.com/nipy/nipype/blob/a746633d72d3e861a4b1a52038dd40a590792e24/nipype/pipeline/plugins/base.py#L371

ccraddock: :-)

satra:
@ccraddock - we have enabled pytest now in master. can you merge with master and resolve the `test_multiproc*` files

PR #1757, opened 2016-12-18
Fix/misc test fixes

satra:
this includes changes from #1751


satra:
@chrisfilo, @oesteban, @djarecka  - any help would be much appreciated  here:

for the moment ignore that this PR passes - it doesn't have the correct test. as you can see py3 gives the correct answer, py2 does not. note the error pattern. A `Function` interface is dynamic and uses `DynamicTraits` for it's `inputspec`.

test function:
```

def test_mapnode_crash3():
    """Test mapnode crash when mapnode is embedded in a workflow"""
    def myfunction(value):
        return value + 1
    node = pe.MapNode(niu.Function(input_names=['WRONG'],
                                   output_names=['newstring'],
                                   function=myfunction),
                      iterfield=['WRONG'],
                      name='myfunc')
    node.inputs.WRONG = ['mystring{}'.format(i) for i in range(3)]
    wf = pe.Workflow('testmapnodecrash')
    wf.add_nodes([node])
    wf.run(plugin='Linear')
```

py27 output:
```
Exception: Subnodes of node: myfunc failed:
Subnode 0 failed
Error:
myfunction() got an unexpected keyword argument 'WRONG'
Interface Function failed to run. 
Subnode 1 failed
Error:
'FunctionInputSpec' object has no attribute 'read'
Interface Function failed to run. 
Subnode 2 failed
Error:
'FunctionInputSpec' object has no attribute 'read'
Interface Function failed to run. 
```

py35 output
```
Subnode 0 failed
Error:
myfunction() got an unexpected keyword argument 'WRONG'
Interface Function failed to run. 
Subnode 1 failed
Error:
myfunction() got an unexpected keyword argument 'WRONG'
Interface Function failed to run. 
Subnode 2 failed
Error:
myfunction() got an unexpected keyword argument 'WRONG'
Interface Function failed to run. 
```

djarecka:
@satra: can't repeat your output from py.27. My output from py27 looks like output from py5

satra:
@djarecka - finally found the culprit, but no solution yet.  see the second line - if you comment it then the code runs properly.

```
from __future__ import print_function, division, absolute_import
from __future__ import unicode_literals # <---- this is the culprit in py27
from builtins import range, open

def test_mapnode_crash3():#tmpdir):
    """Test mapnode crash when mapnode is embedded in a workflow"""
    def myfunction(value):
        return value + 1
    node = pe.MapNode(niu.Function(input_names=['WRONG'],
                                   output_names=['newstring'],
                                   function=myfunction),
                      iterfield=['WRONG'],
                      name='myfunc')
    node.inputs.WRONG = ['mystring{}'.format(i) for i in range(3)]
    wf = pe.Workflow('testmapnodecrash')
    wf.add_nodes([node])
    #wf.base_dir = str(tmpdir)
    #with pytest.raises(RuntimeError):
    wf.run(plugin='Linear')

if __name__ == "__main__":
    test_mapnode_crash3()
```

djarecka:
@satra : that would explain, why I didn't get the same output for py27 yesterday, since I simply copy your code that was without this line.
Do you understand why `unicode_literals` breaks the code in these specific tests?  

regarding solution (at least temporary): `unicode_literals` is actually not needed for these tests...

satra:
@djarecka  - needed for other tests. that's something we cannot change immediately. for the moment i've pushed some code that should enable all of these tests to pass. but we need to clean up traits sometime soon.

PR #1764, opened 2016-12-28
improve python 3 compatibility when parameterize_dirs is False

### LINKED ISSUE #1762: parameterize_dirs = False broken in python 3
### Summary
Using hashed dir parameterization is not working as expected.

### Actual behavior
```
  File "/Users/wtriplett/git/nipype-py3/nipype/pipeline/engine/workflows.py", line 580, in run
    self._configure_exec_nodes(execgraph)
  File "/Users/wtriplett/git/nipype-py3/nipype/pipeline/engine/workflows.py", line 683, in _configure_exec_nodes
    (op.join(edge[0].output_dir(),
  File "/Users/wtriplett/git/nipype-py3/nipype/pipeline/engine/nodes.py", line 209, in output_dir
    params_str = [self._parameterization_dir(p) for p in params_str]
  File "/Users/wtriplett/git/nipype-py3/nipype/pipeline/engine/nodes.py", line 209, in <listcomp>
    params_str = [self._parameterization_dir(p) for p in params_str]
  File "/Users/wtriplett/git/nipype-py3/nipype/pipeline/engine/nodes.py", line 394, in _parameterization_dir
    return sha1(param).hexdigest()
TypeError: Unicode-objects must be encoded before hashing
```
### Expected behavior
```
... normal execution output...
```
```
    config.update_config({'execution': {'parameterize_dirs': False}})
```
### Platform details:
```
{'nibabel_version': '2.1.0', 'scipy_version': '0.18.1', 'networkx_version': '1.11', 'commit_source': 'repository', 'pkg_path': '/Users/wtriplett/
git/nipype-py3/nipype', 'traits_version': '4.6.0', 'numpy_version': '1.11.1', 'sys_platform': 'darwin', 'sys_executable': '/Users/wtriplett/anaco
nda3/bin/python', 'sys_version': '3.5.2 |Anaconda 4.2.0 (x86_64)| (default, Jul  2 2016, 17:52:12) \n[GCC 4.2.1 Compatible Apple LLVM 4.2 (clang-
425.0.28)]', 'commit_hash': '2d4726f'}
```

wtriplett:
Tested in python 3.5 and 2.7.12 before and after fix. Used non-specific/empty `encode()` based on what looks like the common usage elsewhere in code.

Fixes #1762


oesteban:
LGTM - thanks @wtriplett

satra:
@wtriplett - could you please add #1762 as a test case? essentially something like the following which should fail before your fix is applied:

```
def test_unicode_parameterization():
    wf = some workflow
    wf.config['execution']['parameterize_dirs'] = True
    error_not_raised = True
    try:
       wf.run()
    except UnicodeError:
        error_not_raised = False
    assert error_not_raised
```

wtriplett:
@satra: Ok, I can do that.

wtriplett:
The CircleCI tests are also running here without the fix applied. The expected results would be pass on python2.7 and fail on 3.5.

https://circleci.com/gh/wtriplett/nipype/2

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -626,6 +626,38 @@ def func1(in1):
     assert not error_raised
 
 
+def test_parameterize_dirs_false(tmpdir):
+    from .... import config
+    from ....interfaces.utility import IdentityInterface
+    from ....testing import example_data
+
+    config.update_config({'execution': {'parameterize_dirs': False}})
```
satra: make this a local change rather than a global change.

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -626,6 +626,38 @@ def func1(in1):
...
+
+    n1 = pe.Node(EngineTestInterface(), name="Node1")
+    n1.iterables = ('input_file', (input_file, input_file))
+    n1.interface.inputs.input1 = 1
+
+    n2 = pe.Node(IdentityInterface(fields='in1'), name="Node2")
+
+    wf = pe.Workflow(name="Test")
+    wf.base_dir = wd
+    wf.connect([(n1, n2, [('output1', 'in1')])])
```
satra: local change for wf: `wf.config['execution']['parameterize_dirs'] = False`

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -626,6 +626,38 @@ def func1(in1):
...
 
+def test_parameterize_dirs_false(tmpdir):
+    from .... import config
+    from ....interfaces.utility import IdentityInterface
+    from ....testing import example_data
+
+    config.update_config({'execution': {'parameterize_dirs': False}})
+
+    wd = str(tmpdir)
+    os.chdir(wd)
```
satra: these two statements are not required for a workflow since you set `wf.base_dir = wd` later.

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -626,6 +626,33 @@ def func1(in1):
...
+
+    wf = pe.Workflow(name='Test')
+    wf.base_dir = str(tmpdir)
+    wf.config['execution']['parameterize_dirs'] = False
+    wf.connect([(n1, n2, [('output1', 'in1')])])
+
+    error_raised = False
+    try:
+        wf.run()
+    except Exception as ex:
```
wtriplett: Good question. I was looking at the other tests, and they all seemed to catch generic exceptions. It dawned on me that if this is a test of a configuration specific execution pathway, that maybe it was better to catch a general exception in case there was another reason for failure and the assertion would still work rather than bailing out.

satra: in this case since we know it's a `TypeError` we should use that. the intent of a Test should be to catch the most specific error, because the other types of errors may require different handling.

for many of the tests in this file, they do raise a generic exception, but this should also be improved ( ping @djarecka )

wtriplett: Roger that - I'll revert back to TypeError.

djarecka: @satra : I don't understand this try/except block, since we expect `wr.run()` to work, and if we get ANY exception,  the test should fail (and pytest will inform what exception has been raised). IMO, we should remove try/except block at all, just keep `wr.run()`. The same is true about all test that have `assert not error_raised`.  And the test that have `assert error_raised` can be rewritten using ` with pytest.raises(Exception)`.

If you agree with me, I can clean whole file after this PR is merged (unless you prefer to do it before).

satra:
let's do it after it's merged. but yes - we should clean up logic where we can.

Issue #1752, opened 2016-12-16
JoinNode bug appears only on a different computer


bpinsard:
### Summary

I have a simple code that have an iterable connected to a  join node.
```
    subjects_info = pe.Node(
        utility.IdentityInterface(fields=['sub']),
        run_without_submitting=True,
        name='subjects_info')
    subjects_info.iterables = [('sub', subject_ids)]


    template_args = dict(
        t1 = [['sub','ses','anat/sub-*_ses-SSV*_acq-MEMPRAGE_T1w.nii.gz']],
        t2 = [['sub','ses','anat/sub-*_ses-SSV*_T2w.nii.gz']],
        fmri = [['sub','ses','func/*.nii.gz']],
        fieldmap_mag=[['sub','ses','fmap/*phasediff.nii.gz']], # names are swapped !!                                         
        fieldmap_field=[['sub','ses','fmap/*magnitude1.nii.gz']],) # names are swapped !!                                     
    sessions_info = pe.Node(
        nio.DataGrabber(infields=['sub', 'ses'],
                        outfields=template_args.keys(),
			sort_filelist = True,
			raise_on_empty = False,
                        base_directory = mri_data_dir,
                        template=''),
	run_without_submitting=True,
        name='sessions_info')
    sessions_info.inputs.template = 'sub-%07d/ses-SSV%d/%s'
    sessions_info.inputs.template_args = template_args

    sessions_info.iterables = [('ses',range(1,15))]

    all_sessions = pe.JoinNode(
	utility.IdentityInterface(fields=template_args.keys()),
	joinsource = 'sessions_info',
        run_without_submitting=True,
        unique=False,
        name='all_sessions')
    w = pe.Workflow(name='moco_multiband')
    w.base_dir = proc_dir

    w.connect([
        (subjects_info, sessions_info,[('sub',)*2]),
        (sessions_info, all_sessions, [(k,k) for k in template_args.keys()])
    ])

```

I tried to install nipype (master branch) on a HPC, and use the same pipeline as before, but when I try to run I get the following error:

### Actual behavior

```
AttributeError                            Traceback (most recent call last)
<ipython-input-5-b9a5896f20e5> in <module>()
----> 1 w.run()

/home/bpinsard/.conda/envs/prod/lib/python2.7/site-packages/nipype-0.13.0_gff36588.dev-py2.7.egg/nipype/pipeline/engine/workfl
ows.pyc in run(self, plugin, plugin_args, updatehash)
[...]

PR #1768, opened 2016-12-29
rel: create release 0.13.0-rc1

satra:


mgxd:
@satra perhaps we should limit traits <4.6 til we track down #1752?

PR #1769, opened 2016-12-30
Make CircleCI use all 4 containers

conormcd:
- Cache both docker containers.
- Split the tests across all 4 containers.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -20,43 +20,27 @@ dependencies:
     - sudo apt-get -y update && sudo apt-get install -y wget bzip2
 
   override:
-    - mkdir -p ~/examples ~/scratch/pytest ~/scratch/logs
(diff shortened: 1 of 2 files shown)


chrisfilo:
So far we have been using spare containers to run tests on PRs. What's the speed improvement when switching to using 4 containers (when counting on test time)?

conormcd:
The tests are currently set to 4x parallelism so three out of the four containers are idle and unusable for every build. This PR brings the time for a test run from 3.5-4 hours down to 1.5-2 hours. So, approximately a 2 hour saving per build.

https://circleci.com/gh/conormcd/nipype/8 and https://circleci.com/gh/conormcd/nipype/9 are example builds with this change.

chrisfilo:
Ah yes - you are right.

PR #1773, opened 2017-01-04
simplifying tests, removing some try/except blocks

djarecka:
* I left the try/except blocks in `test_node_init` and  `test_io_subclass`: it looks like the tests should fail only for some exceptions, and others are ok

* I removed also `logger.info('Exception: %s' % str(e))` parts, most other test don't have it anyway. I can bring it back if needed.

### DIFF
--- a/nipype/pipeline/engine/tests/test_engine.py
+++ b/nipype/pipeline/engine/tests/test_engine.py
@@ -512,11 +512,16 @@ def test_node_hash(tmpdir):
     # create dummy distributed plugin class
     from nipype.pipeline.plugins.base import DistributedPluginBase
 
+    # create a custom exception 
+    class EngineTestException(Exception):
+        pass
+
     class RaiseError(DistributedPluginBase):
         def _submit_job(self, node, updatehash=False):
-            raise Exception('Submit called')
+            raise EngineTestException('Submit called')
 
-    with pytest.raises(Exception) as excinfo:
+    # check if a proper exception is raised
+    with pytest.raises(EngineTestException) as excinfo:
         w1.run(plugin=RaiseError())
     assert 'Submit called' == str(excinfo.value)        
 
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -518,35 +509,26 @@ def func2(a):
...
-    error_raised = False
     # create dummy distributed plugin class
     from nipype.pipeline.plugins.base import DistributedPluginBase
 
     class RaiseError(DistributedPluginBase):
         def _submit_job(self, node, updatehash=False):
             raise Exception('Submit called')
-    try:
+
+    with pytest.raises(Exception) as excinfo:
```
satra: can we make this a specific exception, but changing what exception is raised. here it could be a specific exception, like `EngineTestException`.

djarecka: I'm not sure if I understand. Do you want pytest to check specific exceptions or you want to define a class `EngineTestException(Exception)`.

I sometimes ask pytest to check a specific error or exception, but in this case the exception from line 517 was raised (i.e. `raise Exception('Submit called')`), so to be more specific I could only check the message and I've done it in the line 521 `assert 'Submit called' == str(excinfo.value)`.

satra: in this particular case define a specific class and check for that. we shouldn't raise generic exceptions unless absolutely necessary.

djarecka: ok, can do it, but the class `RaiseError(DistributedPluginBase)` is defined within the test function, so it will only influence this specific test and  not the general code behavior.

satra: correct. most of nipype is ok - we have a couple of general exception catches, mostly because we don't know how python functions or external tools called via subprocess are going to crash

djarecka: ok, understand now

PR #1776, opened 2017-01-10
re-add neurostars link

mgxd:


djarecka:
@mgxd : i've restarted travis to check if the problem would appear again. Travis looks ok now, but I'm not sure if this is problem with travis or with tests... had similar problem 2 weeks ago (but different test): https://travis-ci.org/djarecka/nipype/jobs/189379626

mgxd:
@djarecka my early build passed on travis but not circle - although I'm not sure that was due to tests or some docker/conda error

djarecka:
@mgxd : your last build only passed travis after I restarted it (not talking about previous problems with CI). Travis doesn't send you emails about broken builds? I'm not able to check this now (or don't know how), but link from emails should still show the previous build that ended with segfault.

djarecka:
sorry, probably I shouldn't have restarted but was wondering if it would be OK after restarting, as it was OK in my case 2 weeks ago...

PR #1780, opened 2017-01-16
FSL FIX wrapper

chrisfilo:


wmpauli:
Is there anything I need to do here? It doesn't seem like the first test (continuous-integration/travis-ci/pr) failed due to a reason unrelated to the fix.py file.

satra:
@djarecka - the segfault is back!

djarecka:
@wmpauli - can you please merge your repo with the new nipy:master? We changed the testing framework a few months ago, CI is testing according to your branch and tries to run nosetest, that is not supported anymore. 

@satra - travis is "fixed" :) but it looks like the segfault is real, and since wpauli's master is pretty old, it looks like this segfault might have been with us for a while... we should ask people to run travis 50times before accepting PR ;-) Will have to try to reproduce on my laptop, in 3 cases I saw it was always py27. 
btw. Do you understand why travis is running py.test and CI uses wpauli master to run nose? Is it expected behavior?

wmpauli:
@djarecka Sorry about that. I merged, but it still segfaults for the python 2.7 build, without installing debs.

djarecka:
@wmpauli - thanks for merging. The segfault is not related to your changes, I've seen it recently in other branches. It doesn't occur every single time, and can jump from one environment to another, just stays within py27. We have to figure this out.

@satra - i didn't get that this is THE segfault... On my laptop I'm getting it for every run

chrisfilo:
I rerun the tests and everything seems green.

wmpauli:
Hooray! Thank you for merging.

PR #1781, opened 2017-01-16
Fix small bug/typo in misc.CreateNifti

alexsavio:
Hi,

I've found a bug in misc.CreateNifti while running Camino tractography workflows.
This seems to fix it.

Cheers,
Alex

### DIFF
--- a/nipype/algorithms/misc.py
+++ b/nipype/algorithms/misc.py
@@ -245,7 +245,7 @@ class CreateNifti(BaseInterface):
         else:
             affine = None
 
-        with open(self.inputs.header_file, 'rb') as data_file:
+        with open(self.inputs.data_file, 'rb') as data_file:
             data = hdr.data_from_fileobj(data_file)
 
         img = nb.Nifti1Image(data, affine, hdr)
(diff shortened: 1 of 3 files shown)


nipype/algorithms/misc.py
```diff
@@ -245,7 +245,7 @@ def _run_interface(self, runtime):
         else:
             affine = None
 
-        with open(self.inputs.header_file, 'rb') as data_file:
+        with open(self.inputs.data_file, 'rb') as data_file:
```
satra: this interface seems to only work with `hdr/img` pair files. in that space, shouldn't this be the `header_file`?

effigies: I think @alexsavio is right, here. These two lines [were originally](https://github.com/nipy/nipype/commit/a16b7eb356b7f1560f1327eceb0e422ed12fd9b5#diff-ae6a9061a228a455a45ce95549571550L247):

```Python
data = hdr.data_from_fileobj(open(self.inputs.data_file, 'rb'))
```

alexsavio: Thanks @effigies. @satra, my workflows were working until I got this update, it only worked when I fixed this. A Nipype Camino tractography example also uses this, it probably won't work without this. I was thinking on adding a test for this case, but I would need Analyze files, somehow. Any hint?

effigies: You can generate them with nibabel easily enough. e.g.

```Python3
analyze = nb.AnalyzeImage(np.zeros((3,4,5,6)), np.eye(4))
analyze.to_filename('test.hdr')  # Should create test.hdr/test.img
```

alexsavio:
Hi, This PR is ready for review. Thanks!

PR #1790, opened 2017-01-27
[ENH] Update ReconAll interface for v6

### LINKED ISSUE #1782: FreeSurfer v6.0 entorhinal cortex labeling
### Summary
In FreeSurfer v6.0 ([release candidate](https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.0-release-candidate/)), `-(no)label-exvivo-ec` are no longer valid options. Entorhinal cortex labeling appears to be done by `-balabels` (see the final line of the [recon-all table](https://surfer.nmr.mgh.harvard.edu/fswiki/ReconAllTableStableV6.0#line-175)).

I'm guessing the solution will be to check `$FREESURFER_HOME/build-stamp.txt` for the version and modify [`ReconAll._steps`](https://github.com/nipy/nipype/blob/cc04eb0ca2b00a0cb5296538686fa9ee56faa6e2/nipype/interfaces/freesurfer/preprocess.py#L659-L737) accordingly.

### Actual behavior
`'-(no)label-exvivo-ec'` options are added to the `ReconAll` cmdline.

### Expected behavior
`'-(no)label-exvivo-ec'` options are omitted for FreeSurfer v6.0+.

1. Install FreeSurfer v6.0:
```Shell
curl -sSL https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.0-release-candidate/freesurfer-Linux-centos6_x86_64-stable-pub-v6.0.0.tar.gz \
    | tar zx -C /opt \
    --exclude='freesurfer/trctrain' \
    --exclude='freesurfer/subjects/fsaverage_sym' \
    --exclude='freesurfer/subjects/fsaverage3' \
    --exclude='freesurfer/subjects/fsaverage4' \
    --exclude='freesurfer/subjects/fsaverage5' \
    --exclude='freesurfer/subjects/fsaverage6' \
    --exclude='freesurfer/subjects/cvs_avg35' \
    --exclude='freesurfer/subjects/cvs_avg35_inMNI152' \
    --exclude='freesurfer/subjects/V1_average' \
    --exclude='freesurfer/average/mult-comp-cor' \
    --exclude='freesurfer/lib/cuda' \
    --exclude='freesurfer/lib/qt'
```

2. Place license file in `/opt/freesurfer/license.txt`

3. Run `ReconAll` on `bert`:

```Python
>>> import os
>>> from nipype.interfaces.freesurfer import ReconAll
>>> interface = ReconAll(subject_id='bert', directive='all', subjects_dir=os.getenv('SUBJECTS_DIR'))
>>> 'label-exvivo-ec' in interface.cmdline
True
```

### Platform details:
```Python
{'commit_hash': '<not found>',
 'sys_platform': 'linux',
 'networkx_version': '1.11',
 'commit_source': '(none found)',
 'numpy_version': '1.11.3',
 'sys_executable': '/usr/local/miniconda/bin/python',
 'nibabel_version': '2.1.0',
 'traits_version': '4.6.0',
 'sys_version': '3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:53:06) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]',
 'scipy_version': '0.18.1',
 'pkg_path': '/usr/local/miniconda/lib/python3.5/site-packages/nipype'
}
```

effigies:
This is more of an RFC with some attached code than a PR, at the moment.

Below are (most of) the changes to `ReconAll._steps` needed to reflect the FreeSurfer v6.0 [recon-all table](https://surfer.nmr.mgh.harvard.edu/fswiki/ReconAllTableStableV6.0), rather than the v5.3 [recon-all table](https://surfer.nmr.mgh.harvard.edu/fswiki/ReconAllTableStableV5.3).

There are some additions, deletions as well as different outputs. So a few questions to begin the discussion:

1. Is the `_steps` approach still good? Note that `aseg.mgz` is the output of two different steps, which means the latter almost certainly won't be run without actually checking whether outputs are newer than inputs.
2. Do we want to include `-T2pial/-FLAIRpial` or leave it as something people can specify in `ReconAll.inputs.args`?
3. Given that there exists a Makefile-based recon-all maintained by the FreeSurfer devs, does it make sense to do our own dependency management?

---
Fixes #1782.


satra:
@effigies - thanks for bringing up these questions

> 1. Is the _steps approach still good? Note that aseg.mgz is the output of two different steps, which means the latter almost certainly won't be run without actually checking whether outputs are newer than inputs.

the steps approach was there to restart at the appropriate point in the recon-all workflow and there was something off about simply using the make option, but if the make option is cleaner and it works i would support removing steps. 

> 2. Do we want to include -T2pial/-FLAIRpial or leave it as something people can specify in ReconAll.inputs.args?

https://github.com/nipy/nipype/blob/master/nipype/interfaces/freesurfer/preprocess.py#L621

> 3. Given that there exists a Makefile-based recon-all maintained by the FreeSurfer devs, does it make sense to do our own dependency management?

not if it works.

on a related note, there will be some work towards moving the recon-all script to a more maintainable mode  by the freesurfer folks, and we are definitely hoping to converge towards sth that we can use to generate a parallelizable nipype workflow. the intent behind the work from iowa was to figure out ways to improve or replace pieces of recon-all (for example, replacing n3biascorrection with n4 or taking out mni tools, etc.,.).

effigies:
Ah, looked right at the `-T2pial` and didn't see it.

So, looking at `recon-all -make`, setting `MAKEFLAGS="-j -l <NCORES>"` causes it to try to do things out of order (starts trying to smooth non-existent surfaces), presumably due to an incomplete dependency graph. That's not great. Still, it does pass through `-parallel -openmp <NCORES>`, so we wouldn't be losing anything by going through `make`. I'll see if I can make that work and resume properly in NiPype.

The simplest alternative I see is to add an optional `dependencies` field to steps, checking `mtime`s if present. That way, at least, if people need to re-run after manual edits, there'll be an existing data structure to update as needed.

One other thing about steps, though, is that, if you ever complete a step in `autorecon3`, the directive will always be `-autorecon3`, regardless of whether you've touched up a file and need to rerun parts of `autorecon2` (see https://surfer.nmr.mgh.harvard.edu/fswiki/recon-all#SpecifyingDirectives for examples of times you might want that). Fixing this may be a bit out of scope for just getting this so you can run `ReconAll` from start-to-finish, and probably a poor use of time if there's work toward re-architecting. Would be glad to participate in existing efforts, there.

effigies:
Nope. `-make` doesn't handle `-no*` arguments. Updating `steps`, and added a third field for optionally specifying dependencies. `check_deps` tests target existence and whether targets are newer than dependencies. If this is an acceptable addition, I'll write some tests.

I've also removed the `directive` override. Reasoning from commit:

    Assuming any directive maps to '-all' with all previous steps completed
    is incorrect. Further, the directive update to 'autorecon2' precludes
    the completion of 'autorecon3', which 'all' includes.

    Taking the existing directive and removing completed steps is a more
    conservative approach, with less chance of unexpected behavior.

If you want to put something else to replace it, I can, but I think we risk getting too complicated, here. Following the user's directive, and just removing the completed bits (extraneous `-no*` flags won't modify behavior) seems like the desired behavior.

satra:
@effigies - just to be sure this interface will work with 5.3 and 6 correct?

effigies:
Ah, no. At present, it just replaces the 5.3 `_steps`. I'll re-add, and make a version check.

effigies:
Travis [passing](https://travis-ci.org/effigies/nipype/builds/197178379) on my account. The error there is one of those random segfaults.

@satra Ready for review.

chrisfilo:
Don't we also need to add the `-parallel` flag?

effigies:
Good call. For some reason I got it in my head that `-openmp` implied `-parallel`. Any other flags worth giving inputspec entries? Such as `-hires`, `-use-cuda`/`-use-gpu`, or `-clean`?

chrisfilo:
We'll need `-hires` eventually, but this can wait for another PR.

Issue #1788, opened 2017-01-26
segfault running /pipeline/engine/tests/test_utils.py 


djarecka:
### Summary
Segfault when running tests from /pipeline/engine/tests/test_utils.py

### Actual behavior
See e.g. https://travis-ci.org/nipy/nipype/builds/195258843
appears randomly, but never seen outside py2.7
It's probably the same problem as described in #1757, but the temporary @satra  solution doesn't always work(?). 

Not sure, if my recent failure in different tests might be related: https://travis-ci.org/djarecka/nipype/jobs/189379626

### How to replicate the behavior
From my experience, it might appear more often on some systems than on the others. It should be python2.7 (other travis setting are not important IMO). You can use the docker container `nipype/nipype_test:py27` and run ` pytest -v nipype/pipeline/engine/tests/test_utils.py` (after installing pytest) multiple times, 10 should be enough. 

It usually returns segfault when running `test_mapnode_crash2` or `test_mapnode_crash2`, but running any of these tests separately do not give segfault (at least not as often).

satra:
and those particular tests will pass if `unicode_literals` is turned off. so it's some interaction between the traits and unicode literals

oesteban:
traits has C "fast" checking on some types. It looks to me that could be the source of the segfault

djarecka:
looks better after updating traits to 4.6, but have to test more.

djarecka:
Traits 4.6 either solves the problem or hides it pretty well so the segfault occurs very rare (having problem with getting one). Since I don't understand this segfault, can't really say which one is true.

btw. I've just found that this is not the first time you have problems with traits&unicode_literals  errors that are not reported by Travis (at least not in every run) #1621

satra:
@djarecka - the change you made in `interfaces.base` in dynamic traits was based on traits 4.6. perhaps we should set that as the minimum version.

djarecka:
yes, all test should pass with traits 4.6. I've created a PR, but would keep the issue open

djarecka:
it looks like travis with traits4.6 still have sagfault

mgxd:
@djarecka @satra how do we progress with this? Will moving from traits to traitlets help this?

djarecka:
@mgxd  - I will check traitlets as @satra suggested.

oesteban:
[...]

PR #1791, opened 2017-01-30
changing the traits minimum version

djarecka:
The new traits version probably solves problems from #1788.


satra:
@oesteban - do you recollect why our docker builds don't use miniconda?

djarecka:
Ok, so Python 2.7 from my OSX and from the nipype_test container with traits 4.6 worked very well, but it looks like py27 from Travis works even worse than before...

satra:
@djarecka - perhaps i can put this on your plate - we had long considered switching from traits to traitlets (by the ipython folks) - perhaps we can start looking into that again. right now the ci failure has to do with building traits since it's not using conda. but the py27 failure is strange and also why it happens only in one branch of the test.

djarecka:
@satra -  CircleCI has indeed problem with installation (I manually installed with conda for testing), but Travis has segfault.

What do you mean by writing that it happens in one branch of the test?

djarecka:
@satra - I can read about traitlets, never heard before

oesteban:
@satra we used miniconda to get thinner images, but we could perfectly jump into anaconda, it is much more reliable.

Right now we don't rely that much on traits (nothing traitlets couldn't do). If we don't intend to make a larger use of traits (their interfacing abilities, linking traits, etc), we are probably better off moving into traitlets.

oesteban:
@djarecka, hope this helps:

- CircleCI now works with parallelism, you'll need to set 4x in https://circleci.com/gh/djarecka/nipype/edit#parallel-builds

- To force traits>=4.6 you'll need to update the version here https://github.com/nipy/nipype/blob/master/nipype/info.py#L109, and here https://github.com/nipy/nipype/blob/master/requirements.txt#L4 (why you only changed the rtd_requirements.txt, which is only used by readthedocs?)

djarecka:
@oesteban - thanks for the suggestions. In the last commit I simply removed my changes from `requirements.txt` since was trying to understand the new import error (#1798), but I will update `requirements.txt`. Just wanted to wait for your PR (#1796) to be accepted. 

I understood that @satra asked about miniconda in docker, since I was getting  an error when I was trying to import new traits (see https://circleci.com/gh/djarecka/nipype/138). At the end I'm not sure if we're planning to move to miniconda for docker builds.

oesteban:
Got you.

Let's start with the second bit: [your build](https://circleci.com/gh/djarecka/nipype/138) is failing because the docker image does not have gcc installed to compile stuff. Best alternative is use conda to install it (not pip) since it will install only binaries.

All about docker images needs a cleanup, I'll try to submit a PR later today or tomorrow. I will remove all those unnecessary docker files and update the necessary ones to be more lightweight and clearer.

Regarding the icu installation, we merged #1808 on Friday. That should get you going without the import error.

My PR #1796 should not affect you much (however, I've made some fixes on the travis file that you probably want to have asap).

To summarize this PR:
  - You need an update of the docker images (so that you can add traits in the appropriate conda install)
  - You need #1808 merged into this branch.
  - Then there is this two-fold dependency management issue: 1) updating the ```info.py``` file (for the pypi distribution) and 2) updating the requirements.txt (for the development environments). Since we are using conda in travis and circle, we need to add travis>=4.6 manually (until the conda-forge installation of nipype sets that dependency). Maybe @chrisfilo could chime in and give us some lines about how to pin package versions for nipype in the conda-forge channel. I guess we need to update nipype there as well.

satra:
@oesteban - if we can make these docker updates, we can update nipype on conda-forge. we should definitely add icu as a dependency for the lxml library if it's not already there.

satra:
@oesteban - we should release 0.13 soon - it's been sitting in release candidate mode, but should really merge in these latest fixes and as long as the builds pass we release.

djarecka:
@oesteban - great, I'll wait for a new docker files. 

I also have general question about the docker images. Please correct me if I'm wrong, but we're building docker images in CircleCI based on the current version of docker files. However, if we ask people to test "locally" they have to pull images from docker hub, and this was not updated for the last 5 months https://hub.docker.com/r/nipype/nipype_test/~/dockerfile/. So currently I have to always install pytest (and few other libraries) if I really want to use those images for testing. Should we also think about docker hub versions or I'm the only one using those images?

oesteban:
Good point. @chrisfilo has done a great investigation on that. The idea would be pushing (automatically, from CircleCI) a new image when a release is done.

oesteban:
For developers, I'd say they should build their own docker image from the dockerfile. The built image should be up-to-date if all nipype dependencies are well-captured (meaning, no need to install pytest for instance).

djarecka:
didn't happen right away, but travis still might return segfault in py27. 
Was also able to  get a timeout in nipype/interfaces/afni/preprocess.py in 3.4.

@oesteban - any idea what is wrong with my CircleCI?

djarecka:
having various problem with CircleCI build: https://circleci.com/gh/djarecka/nipype/147 or https://circleci.com/gh/djarecka/nipype/146. 
@oesteban - do you have any suggestions?

djarecka:
@satra - I merged with master, so Circle CI works. This PR does not solve the segfault problem IMO, but the segfault might happen less often.

oesteban:
LGTM, after merging this, we can close #1809

nipype/pipeline/engine/tests/test_utils.py
```diff
@@ -18,6 +18,11 @@
 from ..utils import merge_dict, clean_working_directory, write_workflow_prov
 
 
+def test_traits_version():
+    #just a temprorary test, testing CI
+    import traits
+    assert traits.__version__ >= "4.6.0"
```
satra: merged this a little too quickly, while this works most of the time, we should use LooseVersion, which does a better check.

oesteban: Agreed. Actually, we should have like a pre-install hook (we've talked about it) to check versions.

djarecka: Actually that was my temporary test, when wanted to be sure that Travis indeed uses traits4.6 and still gives the segfault. 
I should probably just remove this test...

PR #1794, opened 2017-02-02
fix: typo

mgxd:


mgxd:
ran into `ImportError: libicui18n.so.56: cannot open shared object file: No such file or directory` in travis builds - adding icu 56.1 seemed to fix this.

mgxd:
@oesteban wdyt?

oesteban:
Sorry @mgxd , yes, that should do :+1:

Issue #1802, opened 2017-02-08
Failure on MRTrix2TrackVis with nipype 0.13 but not the older 0.12


dorianps:
### Summary
The same code fails if I use the developer version of nipype 0.13 but works with the older release 0.12.
Sorry for not giving too much details, it's my first time using nipype and I did not keep track of everything (who would have thought that it was the nipype version).

### Actual behavior
The ".trk" file produce with the new version is 1kb with nothing in it. It loads in TrackVis but has not streamlines to show.

### Expected behavior
The ".trk" file should be dependent on the number of streamlines, and larger than 1kb. It should also be loadable in TrackVis.

### How to replicate the behavior
```import nipype.interfaces.mrtrix as mrt
tck2trk = mrt.MRTrix2TrackVis()
tck2trk.inputs.in_file = 'temp.tck'
tck2trk.inputs.image_file = 'meanb0.nii.gz'
tck2trk.inputs.out_filename = 'temp.trk'
tck2trk.run()
```

### Platform details:
please paste the output of: `python -c "import nipype; print(nipype.get_info())"`
This is what's working well.
{'nibabel_version': '2.1.0', 'networkx_version': '1.11', 'numpy_version': '1.12.0', 'sys_platform': 'linux2', 'sys_version': '2.7.13 |Anaconda 4.3.0 (64-bit)| (default, Dec 20 2016, 23:09:15) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]', 'commit_source': 'installation', 'commit_hash': u'a4d1e07', 'pkg_path': '/data/jag/dpustina/engine/anaconda/lib/python2.7/site-packages/nipype', 'sys_executable': '/data/jag/dpustina/engine/anaconda/bin/python', 'traits_version': '4.6.0', 'scipy_version': '0.18.1'}

I am trying to upload a zip file to help you replicate the error but it's not allowing it.

mgxd:
Hi @dorianps,
This seems weird, if you could share some data with me to replicate this I'd like to take a look

dorianps:
Follow up: I have installed nipype in 3-4 computers in the last two days, and I finally realized that all packages need to be updated for the Trackvis conversion to work. That said, once everything works on nipype 0.12.1 upgrading to 0.13 brings a charset error, as below.

[...]

PR #1804, opened 2017-02-09
fix: unicode problem with MRTrix2TrackVis

mgxd:
due to #1802

@satra @chrisfilo @oesteban is there a better solution than just ignoring the problematic byte?


oesteban:
I'd say that the trk file should be opened as bytes (`'rb'`). Then, decode the strings (`line.decode()`). WDYT @satra ?

I would not advise skipping errors, that way we cannot reconstruct what really happened. It'd be better wrap it into a try catch that logged a warning, at the very least.

BTW, the import of open at the top should be from io, instead of builtins.

satra:
@oesteban - i agree on both accounts.

satra:
@mgxd - now the file object is a byte-stream, as @oesteban suggests wherever it's read, you will likely have to do `read().decode()`

mgxd:
@satra When I tried that I was running into an error saying `ascii codec can't decode byte` - I'm assume the read() method is flexible enough to take in the byte-stream as it's not throwing an error without the trailing `.decode()`

mgxd:
still running into `libicui18n.so.58` error, despite adding `icu === 56.1` in #1794

satra:
@mgxd - did you merge with master?

also i'm not getting the error in py3 when i just load the file directly using:
```
In [1]: from io import open

In [2]: with open('/Users/satra/Downloads/LeftTempSingleSeed_tracks.tck', 'rb') as fp:
   ...:     data = fp.read()
   ...:     

In [3]: data[:619].decode().split('\n')
Out[3]: 
['mrtrix tracks    ',
 'downsample_factor: 3',
 'fod_power: 0.25',
 'init_threshold: 0.100000001',
 'lmax: 8',
 'max_angle: 60',
 'max_dist: 243.00621',
 'max_num_attempts: 200000',
 'max_num_tracks: 2000',
 'max_seed_attempts: 1',
 'max_trials: 1000',
 'method: iFOD2',
 'min_dist: 12.1503105',
 'mrtrix_version: 0.3.15-285-g68589079',
 'output_step_size: 1',
 'rk4: 0',
 'samples_per_step: 4',
 'sh_precomputed: 1',
 'source: WMfod.mif',
 'step_size: 1',
 'stop_on_all_include: 0',
 'threshold: 0.100000001',
 'timestamp: 1486486988.2982976437',
 'unidirectional: 0',
 'roi: seed LeftTempSingleSeed.mif',
 'roi: mask bmask.nii.gz',
 'roi: mask bmask.nii.gz',
 'datatype: Float32LE',
 'file: . 640',
 'count: 2000',
 'total_count: 2925',
 'END',
 '']
```

satra:
however i see the error trying to decode beyond the END marker:
```
In [4]: data[:1000].decode().split('\n')
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-4-cd5b79147f7f> in <module>()
----> 1 data[:1000].decode().split('\n')

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf5 in position 640: invalid start byte
```
i believe the rest is binary and shouldn't require decoding. @mgxd - could you take a look at the functions and the mrtrix file format to see how they represent things?

mgxd:
@satra this should be good to go

PR #1811, opened 2017-02-13
ENH: Enable new BBRegister init options for FSv6+

effigies:
`bbregister` has new initialization options, including an implicit default of using `mri_coreg`, as of version 6.0.0.

This change updates the `init` trait if FreeSurfer is version 6 or newer, following the example from [FSL](https://github.com/nipy/nipype/blob/f5ed4d68fc922fa3eea6a43036b09f79bfa94acb/nipype/interfaces/fsl/model.py#L589).

Is there a good way to test this?

### DIFF
--- a/nipype/interfaces/freesurfer/preprocess.py
+++ b/nipype/interfaces/freesurfer/preprocess.py
@@ -942,6 +942,12 @@ class BBRegisterInputSpec(FSTraitedSpec):
                                     desc='output warped sourcefile either True or filename')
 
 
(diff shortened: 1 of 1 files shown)


chrisfilo:
The PR looks good, but it's missing those silly `auto_` tests. Could `make specs`?

effigies:
Doesn't seem to result in any changes to `auto_` tests.

chrisfilo:
I guess it's not picking up the freesurfer as version 6. It should be fine to keep it as it is.

PR #1812, opened 2017-02-14
[WIP/ENH] Add antsMotionCorr and related utilities

rwblair:
Couple of interfaces to use Ants motion correction, and its statistics program. Only included inputs for the affine registration for motion correction.


oesteban:
Hey @rwblair, would it be very hard to re-bump this? Can you merge current master in? (maybe after S3 recovers from the hiccups)

rwblair:
@oesteban will do

oesteban:
While https://github.com/stnava/ANTs/issues/419 is not solved, can we generate the ITK matrices manually?:

  - CSV file has 6 parameters - generate an Euler3DTransform file (see the ```# Transform <index>``` entry, that would be associated with the timepoint):
    ```
    #Insight Transform File V1.0
    #Transform 0
    Transform: Euler3DTransform_double_3_3
    Parameters: -0.28025704622268 -0.0066158948466181755 -0.0005159910651855171  0.4772340655326843 4.694962024688721 -9.615015983581543
    FixedParameters: 3.5478134155273438 14.090387344360352 0.7471063733100891

    #Transform 1
    Transform: Euler3DTransform_double_3_3
    Parameters: -0.18025704622268 -0.016158948466181755 -0.0002159910651855171  1.4772340655326843 2.694962024688721 -19.615015983581543
    FixedParameters: 3.5478134155273438 14.090387344360352 0.7471063733100891
    ```
  - CSV file has 12 parameters:
    ```
    #Insight Transform File V1.0
    #Transform 0
    Transform: AffineTransform_double_3_3
    Parameters: 0.9997254014015198 -0.01978115923702717 -0.012558568269014359 0.022292371839284897 0.9680354595184326 0.24982060492038727 0.007215398363769054 -0.250031977891922 0.9682106971740723 -2.5178310871124268 4.0089569091796875 -32.092098236083984
    FixedParameters: 3.5478134155273438 14.090387344360352 0.7471063733100891
    ```

The ```FixedParameters``` can be calculated with ndimage.

I will be able to handle you some code to include it.

oesteban:
https://github.com/poldracklab/fmriprep/blob/f99cdf13f3c1cfb9d36cc869f5f02a26b544b0d6/fmriprep/interfaces/hmc.py#L170

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,305 @@
...
+        desc=("Specify the shrink factor for the virtual domain (typically "
+              "the fixed image) at each level.")
+    )
+    n_images = traits.Int(
+        argstr="-n %d",
+        desc=("This option sets the number of images to use to construct the "
+              "template image.")
+    )
+
+    use_fixed_reference_image = traits.Bool(
```
oesteban: - does ```-u``` need a value? I think this is just a switch (```argstr='-u'```).
- Please, set the default value using positional arguments (```traits.Bool(True, argstr=...)```)

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,305 @@
...
+
+class AntsMotionCorrInputSpec(ANTSCommandInputSpec):
+    '''Input spec for the antsMotionCorr command.'''
+    dimension_desc = (
+        "This option forces the image to be treated as a "
+        "specified-dimensional image. If not specified, N4 tries to infer "
+        "the dimensionality from the input image."
+    )
+    dimensionality = traits.Enum(3, 2, argstr='-d %d', usedefault=True,
+                                 position=0, desc=dimension_desc, default=3)
```
oesteban: remove the final ```default=3```

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,305 @@
...
+
+    average_image = File(argstr='-a %s', position=1, exists=False,
+                         desc="Average the input time series image.")
+
+    output_average_image = File(hash_files=False, desc="", argstr="%s",
```
oesteban: There's no default value, but ```usedefault=True```. Please fix (either remove ```usedefault``` or set a default value)

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,305 @@
...
+    use_fixed_reference_image = traits.Bool(
+        argstr="-u %d",
+        default=True,
+        desc=("use a fixed reference image instead of the neighor in the time "
+              "series.")
+    )
+
+    use_scales_estimator = traits.Bool(
+        argstr="-e %d",
+        default=True,
```
rwblair: -e has an argument here:
https://github.com/stnava/ANTs/blob/master/Scripts/antsMotionCorrExample

nipype/interfaces/ants/preprocess.py
rwblair: -u has an argument here:
https://github.com/stnava/ANTs/blob/master/Scripts/antsMotionCorrExample

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,333 @@
...
+
+    transformation_model = traits.Enum("Affine", "Rigid", argstr="%s")
+    gradient_step_length = traits.Float(requires=['transformation_model'],
+                                        desc='')
+
+    iterations = traits.Int(
+        argstr="-i %d",
+        desc="Specify the number of iterations at each level."
+    )
+    smoothing_sigmas = traits.Int(
```
oesteban: as per command usage, smoothing_sigmas should accept the format MxNx0. This can be achieved with the ```sep``` meta argument:
```
smoothing_sigmas = traits.List(traits.Float, argstr='-s %f', sep='x', desc='...')
```

Please add a doctest exercising these multiple arguments

Please, also check that the command takes floats for the smoothing_sigmas or if it is in voxels (then ints).

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,333 @@
...
+        argstr="-s %d",
+        desc="Specify the amount of smoothing at each level."
+    )
+    shrink_factors = traits.Int(
```
oesteban: as per command usage, shrink_factors should accept the format MxNx0. This can be achieved with the ```sep``` meta argument:
```
shrink_factors = traits.List(traits.Int, argstr='-f %d', sep='x', desc='...')
```

Please add a doctest exercising these multiple arguments

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,333 @@
...
+        desc=("use a fixed reference image instead of the neighor in the time "
+              "series.")
+    )
+
+    use_scales_estimator = traits.Bool(
+        True,
+        argstr="-e %d",
+        desc="use the scale estimator to control optimization."
+    )
+
```
oesteban: Please add the option ```-l, --use-estimate-learning-rate-once``` (bool)

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+        exists=True,
+        desc='Motion crrection matrices to be converted into FSL style motion parameters',
+        mandatory=True
+    )
+
+class AntsMatrixConversionOutputSpec(TraitedSpec):
+    parameters = File(exists=True, desc="parameters to be output")
+
+
+class AntsMatrixConversion(BaseInterface):
```
oesteban: I'd be more specific here: something like `MotionParameters2FSL` instead of `AntsMatrixConversion`

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+    average_image = File(exists=True, desc="Average of an image")
+    composite_transform = File(desc="Composite transform file")
+    inverse_composite_transform = File(desc="Inverse composite transform file")
+    warped_image = File(desc="Outputs warped image")
+    inverse_warped_image = File(desc="Outputs the inverse of the warped image")
+    save_state = File(desc="The saved registration state to be restored")
+    displacement_field = File(desc=("4D displacement field that captures the "
+                                    "affine induced motion at each voxel"))
+
+class AntsMotionCorr(ANTSCommand):
```
oesteban: Since this will live under `nipype.interfaces.ants` I'd say the "Ants" prefix is unnecessary. What about `MotionCorrection` instead of `AntsMotionCorr`? (similarly to `antsRegistration` which is just `Registration` in nipype).

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+import math
+import os
+
+from ..base import (BaseInterface, BaseInterfaceInputSpec, TraitedSpec, File,
+                    traits, isdefined, Str)
+from .base import ANTSCommand, ANTSCommandInputSpec
+from ...utils.filemanip import split_filename
+
+
+def _extant(field):
```
oesteban: Is this really necessary? Since these are `File` traits, you can't set `None` on them. Why not using `isdefined` exclusively?

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+                            desc="do framwise summarywise stats")
+    output_spatial_map = File(argstr='-s %s', hash_files=False,
+                              desc="File to output displacement magnitude to.")
+
+class AntsMotionCorrStatsOutputSpec(TraitedSpec):
+    ''' Output spec for the antsMotionCorrStats command '''
+    spatial_map = File(desc="output image of displacement magnitude", exists=True)
+    output = File(desc="CSV file containg motion correction statistics", exists=True)
+
+class AntsMotionCorrStats(ANTSCommand):
```
oesteban: See comment below. I'd suggest `MotionCorrectionStats` instead of `AntsMotionCorrStats`

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+class AntsMatrixConversion(BaseInterface):
+    '''
+    Take antsMotionCorr motion output as input, convert to FSL style
+    parameter files. Currently does not output origin of rotation.
+    '''
+    input_spec = AntsMatrixConversionInputSpec
+    output_spec = AntsMatrixConversionOutputSpec
+
+    def _run_interface(self, runtime):
+        in_fp = open(self.inputs.matrix)
```
oesteban: I would use here `np.loadtxt`, defining the `sep=','`. This way you get an numpy array, removing all the float casts down below (just one `.astype(float)` would be necessary).

Additionally, you can use a `with` here right?, is the handle necessary at all during the method?

oesteban:
Please add the interfaces (`MotionCorr`, `Matrix2FSLParams` and `MotionCorrStats`) in the [module `__init__.py`](https://github.com/nipy/nipype/blob/10b9675488ac9714c5e7b53a428f14a419fa790e/nipype/interfaces/ants/__init__.py) so that you can do `from nipype.interfaces.ants import MotionCorr`

oesteban:
Tests are failing:

```
Expected:
    "antsJointFusion -a 0.1 -g ['rc1s1.nii', 'rc1s2.nii'] -l segmentation0.nii.gz -b 2.0 -o ants_fusion_label_output.nii -s 3x3x3 -t ['im1.nii']"
Got:
    u'antsJointFusion -a 0.1 -g [rc1s1.nii,rc1s2.nii] -t [segmentation0.nii.gz] -b 2.0 -o ants_fusion_label_output.nii -s 3x3x3 -t [im1.nii]'
```
I think you'll need to remove those single quotes from the test string :)

oesteban:
Sorry, just need merge upstream: #1862

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,355 @@
...
+        in_data = csv.reader(in_fp)
+        pars = []
+
+        # Ants motion correction output has a single line header that we ignore
+        next(in_data)
+
+        for x in in_data:
```
oesteban: is this doing the same [`nibabel.euleragnles.mat2euler`](http://nipy.org/nibabel/reference/nibabel.eulerangles.html#nibabel.eulerangles.mat2euler) is doing?

rwblair: Looks like the main difference is in inputs and outputs. The ants converter takes in and spits out files. While mat2euler takes a numpy array and returns the three rotation values. We could ditch this code and write a wrapper for the nibabel functions that reads in and writes out the files for it?

oesteban: I think that is safer than re-implementing the same thing. I guess here you would just need to reshape the parameters into a matrix before feeding it into `mat2euler`

nipype/interfaces/ants/preprocess.py
```diff
@@ -330,12 +332,13 @@ def _run_interface(self, runtime):
         next(in_data)
 
         for x in in_data:
-            t1 = math.atan2(float(x[7]), float(x[10]))
-            c2 = math.sqrt((float(x[2]) * float(x[2])) + (float(x[3]) * float(x[3])))
-            t2 = math.atan2(-float(x[4]), c2)
-            t3 = math.atan2(float(x[3]), float(x[2]))
+            mat = numpy.zeroes((3, 3))
```
oesteban: why don't loading data with numpy.loadtxt, and then use numpy fancy indexing for all of this?

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+    '''
+    Take antsMotionCorr motion output as input, convert to FSL style
+    parameter files. Currently does not output origin of rotation.
+    '''
+    input_spec = Matrix2FSLParamsInputSpec
+    output_spec = Matrix2FSLParamsOutputSpec
+
+    def _run_interface(self, runtime):
+        in_fp = open(self.inputs.matrix)
+        in_data = csv.reader(in_fp)
```
effigies: A lot of this (parsing, reordering, formatting) can be handled by numpy. How about this?

```Python
# Ants motion correction output has a single line header that we ignore
# The first column is XXX (volume index?)
in_data = np.loadtxt(self.inputs.matrix, delimiter=',', skiprows=1, usecols=range(1, 14))

translations = in_data[:, 9:]

# There might be a slightly cleaner way to do this
rotations = np.zeros((in_data.shape[0], 3))
for i, row in enumerate(in_data[:, :9]):
    rotations[i] = mat2euler(row.reshape(3, 3))

# Don't duplicate this from `_list_outputs`
parameters = self._list_outputs()['parameters']
np.savetxt(parameters, np.hstack((rotations, translations)), fmt='%.8f')
return runtime
```

nipype/interfaces/ants/preprocess.py
```diff
@@ -0,0 +1,358 @@
...
+            fname = '{}Warp.nii.gz'.format(self.inputs.output_transform_prefix)
+            outputs['displacement_field'] = os.path.abspath(fname)
+        return outputs
+
+class Matrix2FSLParamsInputSpec(BaseInterfaceInputSpec):
+    matrix = File(
+        exists=True,
+        desc='Motion crrection matrices to be converted into FSL style motion parameters',
+        mandatory=True
+    )
```
effigies: Sorry for suggesting API changes, but better before merge than after. Feel free to ignore this if I'm being too fiddly.

1. `Matrix2FSLParams` doesn't really imply "ANTs MC" to me. I know this is in `ants/preprocess.py`, but maybe indicate the source a little more clearly? `MotionCorr2FSLParams`?

2. There's a really handy nipype feature:

```Python
class Matrix2FSLParamsInputSpec:
    ants_matrix = File(exists=True, mandatory=True,
                       desc='Motion correction matrices to be converted into FSL-style motion parameters')
    fsl_params = File(name_template='%s.par', name_source='ants_matrix',
                      desc='FSL parameter file')

class Matrix2FSLParamsInputSpec:
    fsl_params = File(exists=True, desc='FSL parameter file')
```

This lets you entirely drop the `_list_outputs` function. (You'd need to make the entry in the output spec match the one in the input spec.)

3. Also, there's a typo in "correction". And I'd use "FSL-style" instead of "FSL style".

oesteban: @effigies I think the `name_source` only works for command line interfaces. (I remember working on this myself, but I don't recall having it merged).

nipype/interfaces/ants/tests/test_MotionCorr2FSLParams.py
```diff
@@ -1,34 +1,53 @@
...
 from nipype.interfaces.ants.preprocess import MotionCorr2FSLParams
+from nipype.interfaces.fsl.utils import AvScale
 from nipype.utils.tmpdirs import InTemporaryDirectory
 from nipype.utils.filemanip import split_filename
 
 def test_MotionCorr2FSLParams():
     with InTemporaryDirectory():
         cwd = os.getcwd()
+        fsl_mat_fname = "fsl_style.mat"
+        fp = open(fsl_mat_fname, 'w+')
```
oesteban: please use numpy.savetxt to workaround encoding issues (py2/3)

nipype/interfaces/ants/tests/test_MotionCorr2FSLParams.py
```diff
@@ -1,34 +1,53 @@
...
+        fp.write(
+            "1.000000 0.000000 -0.000935 0.062539\n"
+            "0.000001 0.999999 0.001470 -0.162467\n"
+            "0.000935 -0.001470 0.999999 -0.279038\n"
+            "0.000000 0.000000 0.000000 1.000000\n"
+        )
+        fp.close()
+
         in_filename = os.path.join(cwd, 'in_file.csv')
         fp = open(in_filename, 'w+')
```
oesteban: same as before, please use numpy.savetxt

effigies:
@rwblair Do you think you could merge master and update the auto-tests?

effigies:

PR #1813, opened 2017-02-14
BF: Run `make specs` with CWD in PYTHONPATH

### LINKED ISSUE #1348: `make specs` uses wrong version of nipype
`make specs` uses the installed version of nipype instead of the one in the CWD. Unless the developer has CWD in PYTHONPATH wrong tests will be generated.

effigies:
Not an expert at Makefiles, so this may not be the best way to go about this, but it does seem to set the PYTHONPATH so that `make specs` is run on the working directory, rather than an installed `nipype`.

Fixes #1348.


effigies:
Might as well run `make specs` and include the updated tests with this patch.

effigies:
@satra Do you want to have a quick look? This is just a quick fixup to make `make specs` work properly for devs with an installed copy of nipype and/or working in Python 3.

nipype/interfaces/fsl/tests/test_auto_Eddy.py
```diff
@@ -38,9 +38,9 @@ def test_Eddy_inputs():
     mandatory=True,
     ),
     in_topup_fieldcoef=dict(argstr='--topup=%s',
-    requires=['in_topup_movpar'],
+    requires=[u'in_topup_movpar'],
```
satra: the primary reason we were using python 2 was because of unicode issues. however, if we include `unicode_literals` at the top of template, we can generate the same tests from either py2 or py3 by modifying the test generator a bit.

the main point of these auto-tests were to create an easy visual check for changes in input/outputspec api and to instantiate each object in the absence of consistent doctests. the former we can capture through a good PR review. the latter would require some work to go through all interfaces. so if we can achieve these two things somehow, it would be good.

effigies: 4ea783c makes py2/py3 behavior equivalent. I'm not entirely sure what you're suggesting for instantiating each object... Does the current system not do that?

PR #1817, opened 2017-02-14
Unset default init for BBRegister in FS 6+

effigies:
Instead of replicating FS default of `--init-coreg`, set `init` as non-mandatory.

Follow-up to #1811.


effigies:
Following discussion with @chrisfilo, default interface is for FreeSurfer 6, so spec tests are set to non-mandatory.

satra:
@effigies - can't we support both with something along the lines of what we do in FSL:

https://github.com/nipy/nipype/blob/f5ed4d68fc922fa3eea6a43036b09f79bfa94acb/nipype/interfaces/fsl/model.py#L589

and depending on local version the appropriate inputspec gets used?

satra:
this is out current measure before we move to versioned interfaces. also each input parameter can be version controlled with `min_ver` and `max_ver` (see: http://nipype.readthedocs.io/en/latest/devel/interface_specs.html#common)

Issue #1403, opened 2016-03-11
xvfb-run not ending gracefully for CommandLine interfaces


blakedewey:
I just updated my testing environment to the latest release of Nipype and I have been getting errors from my `_redirect_x=True` CommandLine interfaces. The CLI completes without error, but then `xvfb-run` errors out with:
`/usr/bin/xvfb-run: line 171: kill: (25939) - No such process`

Some quick Googling seems to point to the fact that `xvfb-run` was not able to kill the display because it was assigned the same display as another instance. My thought is since this only happens on MultiProc/SGE runs of my workflows is that `xvfb-run` is used almost simultaneously and hasn't acquired a lock for the display in use and the `-a` option still reuses the same display for multiple commands. This way when any command after the first finishes, there is no display to kill, leading to the error. I rolled back my code to use @chrisfilo's hack with subprocess and `Xvfb` and the problem resolved itself. I think we should start the `-a` option from a random starting value using the `-n` option. 

Has anyone else seen this issue? Thoughts?

oesteban:
Yes, the virtual framebuffer is indeed a problem. I included ```xvfb-run``` (#1156) because previously the behavior was a bit problematic.

However, ```xvfb-run``` only works for ```CommandLineInterface```-derived interfaces. And, the script creates a new virtual display for every interface with ```_redirect_x=True```. As you noted, ```xvfb-run``` is not reliable when opening many virtual displays.

I have commented this issue with @satra some times, and the conclusions are:

- The optimal solution would be opening **only one** virtual display before running. This would require just one call to ```Xvfb``` before the workflow is run, and redirecting all interfaces with ```_redirect_x``` to that new display. However, ```Xvfb``` has some other problems (#1400).
- If you are directly using interfaces (without the workflow wrapping) then we should detect that no other virtual display is open before creating a new one. Otherwise, attach to an existing one.
- IMHO, ```xvfb-run``` should be removed from nipype ASAP because it only works for command line interfaces and it is hard to handle from nipype. I wonder if it also has the same problems that ```Xvfb``` has in mac OSX.

satra:
@blakedewey @oesteban - should this be one `xvfb` per process or one `xvfb` per physical node? i'm thinking of multiple processes started by a scheduler, rather than the multiproc plugin.

blakedewey:
It should be one per physical node. I was thinking of keeping track of the xvfb processes in the parent process, where we have a dictionary of physical node names and xvfb display numbers.

[...]

PR #1819, opened 2017-02-15
removed obligatory xvfbwrapper dependency

TheChymera:
On an otherwise headless server, xvfbwrapper pulls in a significant stack totalling at least 42 dependencies:

```
[ebuild  N     ] sys-libs/mtdev-1.1.5::gentoo  USE="-static-libs" 268 KiB
[ebuild  N     ] media-libs/libepoxy-1.4.0::gentoo  USE="{-test}" ABI_X86="(64) -32 (-x32)" 301 KiB
[ebuild  N     ] x11-proto/fontsproto-2.1.3::gentoo  USE="-doc" 151 KiB
[ebuild  N     ] x11-proto/xineramaproto-1.2.1-r1::gentoo  ABI_X86="(64) -32 (-x32)" 94 KiB
[ebuild  N     ] x11-proto/randrproto-1.5.0::gentoo  ABI_X86="(64) -32 (-x32)" 139 KiB
[ebuild  N     ] x11-proto/videoproto-2.3.3::gentoo  ABI_X86="(64) -32 (-x32)" 119 KiB
[ebuild  N     ] x11-libs/libxkbfile-1.0.9-r1::gentoo  USE="-static-libs" ABI_X86="(64) -32 (-x32)" 340 KiB
[ebuild  N     ] x11-apps/xauth-1.0.10::gentoo  USE="ipv6 {-test}" 159 KiB
[ebuild  N     ] x11-proto/resourceproto-1.2.0::gentoo  101 KiB
[ebuild  N     ] x11-proto/scrnsaverproto-1.2.2-r1::gentoo  USE="-doc" ABI_X86="(64) -32 (-x32)" 117 KiB
[ebuild  N     ] x11-proto/trapproto-3.4.3::gentoo  48 KiB
[ebuild  N     ] x11-proto/compositeproto-0.4.2-r1::gentoo  ABI_X86="(64) -32 (-x32)" 99 KiB
[ebuild  N     ] x11-proto/bigreqsproto-1.1.2::gentoo  USE="-doc" 111 KiB
[ebuild  N     ] x11-proto/xf86dgaproto-2.1-r2::gentoo  ABI_X86="(64) -32 (-x32)" 84 KiB
[ebuild  N     ] x11-proto/xf86rushproto-1.1.2-r1::gentoo  37 KiB
[ebuild  N     ] x11-apps/rgb-1.0.6::gentoo  136 KiB
[ebuild  N     ] x11-misc/xkeyboard-config-2.20::gentoo  956 KiB
[ebuild  N     ] x11-proto/recordproto-1.14.2-r1::gentoo  USE="-doc" ABI_X86="(64) -32 (-x32)" 121 KiB
[ebuild  N     ] x11-proto/xcmiscproto-1.2.2::gentoo  USE="-doc" 110 KiB
[ebuild  N     ] x11-apps/iceauth-1.0.7::gentoo  133 KiB
[ebuild  N     ] dev-libs/libevdev-1.5.6::gentoo  USE="-static-libs" ABI_X86="(64) -32 (-x32)" 398 KiB
[ebuild  N     ] x11-apps/xrdb-1.1.0::gentoo  135 KiB
[ebuild  N     ] x11-libs/libfontenc-1.1.3::gentoo  USE="-static-libs" 295 KiB
[ebuild  N     ] x11-libs/libXfont2-2.0.1::gentoo  USE="bzip2 ipv6 -doc -static-libs -truetype" 462 KiB
[ebuild  N     ] x11-apps/xkbcomp-1.3.1::gentoo  242 KiB
[ebuild  N     ] x11-apps/xinit-1.3.4-r1::gentoo  USE="minimal -systemd" 162 KiB
[ebuild  N     ] x11-libs/libXScrnSaver-1.2.2-r1::gentoo  USE="-static-libs" ABI_X86="(64) -32 (-x32)" 284 KiB
[ebuild     UD ] sys-fs/udev-230-r1::gentoo [232-r2::gentoo] USE="acl kmod (-selinux) -static-libs%" ABI_X86="($4) -32 (-x32)" 4'188 KiB
[ebuild  N     ] dev-libs/libinput-1.6.1:0/10::gentoo  USE="{-test}" INPUT_DEVICES="-wacom" 906 KiB
[ebuild  N     ] x11-base/xorg-server-1.19.1:0/1.19.1::gentoo  USE="glamor ipv6 suid udev xorg xvfb -debug -dmx
-doc -kdrive -libressl -minimal (-selinux) -static-libs -systemd -tslib -unwind -wayland -xephyr -xnest" 5'901 $iB
[ebuild  N     ] x11-base/xorg-drivers-1.19::gentoo  INPUT_DEVICES="keyboard libinput mouse -acecad -aiptek -el$graphics -evdev -fpit -hyperpen -joystick -mutouch -penmount -synaptics -tslib -vmmouse -void -wacom" VIDEO_CAR$S="amdgpu dummy fbdev intel nouveau radeon radeonsi vesa -apm -ark -ast -chips -cirrus -epson (-fglrx) (-freedr$no) (-geode) -glint -i128 (-i740) -i965 -mach64 -mga -neomagic (-newport) -nv -nvidia (-omap) (-omapfb) -qxl -r$28 -rendition -s3 -s3virge -savage -siliconmotion -sis -sisusb (-sunbw2) (-suncg14) (-suncg3) (-suncg6) (-sunff$) (-sunleo) (-suntcx) -tdfx (-tegra) -tga -trident -tseng -via -virtualbox -vmware (-voodoo)" 0 KiB
[ebuild  N     ] x11-drivers/xf86-video-dummy-0.3.8::gentoo  USE="-dga" 296 KiB
[ebuild  N     ] x11-drivers/xf86-input-libinput-0.24.0::gentoo  359 KiB
[ebuild  N     ] x11-drivers/xf86-video-fbdev-0.4.4::gentoo  287 KiB
[ebuild  N     ] x11-drivers/xf86-video-intel-2.99.917_p20161206::gentoo  USE="dri sna udev -debug -dri3 -uxa -$vmc" 924 KiB
[ebuild  N     ] x11-drivers/xf86-video-vesa-2.3.4::gentoo  279 KiB
[ebuild  N     ] x11-drivers/xf86-input-mouse-1.9.2::gentoo  369 KiB
[ebuild  N     ] x11-drivers/xf86-video-nouveau-1.0.13::gentoo  603 KiB
[ebuild  N     ] x11-drivers/xf86-video-amdgpu-1.2.0::gentoo  USE="-glamor" 380 KiB
[ebuild  N     ] x11-drivers/xf86-input-keyboard-1.9.0::gentoo  334 KiB
[ebuild  N     ] x11-drivers/xf86-video-ati-7.8.0::gentoo  USE="glamor -udev" 826 KiB
[ebuild  N     ] dev-python/xvfbwrapper-0.2.8::gentoo  USE="{-test}" PYTHON_TARGETS="python2_7 python3_4 -pypy"
5 KiB
```

A number of these packages might be nontrivial to handle depending on how and whether there is any video support/emulation on the machine. This is not good if we want to be lightweight and well suited for compact reproducible data analysis environments.

The only xvfbwrapper import is this:

```
nipype/interfaces/base.py
1014-    def _run_wrapper(self, runtime):
1015-        sysdisplay = os.getenv('DISPLAY')
1016-        if self._redirect_x:
1017-            try:
1018:                from xvfbwrapper import Xvfb
```

I don't have a quantitative estimate, but only very few interfaces actually use `_redirect_x`. Similarly to FSL, SPM, AFNI, and many other dependencies, xvfbwrapper is thus:

* nontrivial to build and carries numerous dependencies of its own
* not needed by very many users

Accordingly, I think it should be treated similarly to these packages and not list it as a mandatory dependency.

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -164,6 +164,7 @@ EXTRA_REQUIRES = {
     'fmri': ['nitime', 'nilearn', 'dipy', 'nipy', 'matplotlib'],
     'profiler': ['psutil'],
     'duecredit': ['duecredit'],
+    'xvfbwrapper': ['xvfbwrapper'],
     # 'mesh': ['mayavi']  # Enable when it works
 }
 
@@ -171,4 +172,3 @@ EXTRA_REQUIRES = {
 EXTRA_REQUIRES['all'] = [val for _, val in list(EXTRA_REQUIRES.items())]
 
 STATUS = 'stable'
-
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -143,7 +143,6 @@ REQUIRES = [
     'simplejson>=%s' % SIMPLEJSON_MIN_VERSION,
     'prov>=%s' % PROV_MIN_VERSION,
     'click>=%s' % CLICK_MIN_VERSION,
-    'xvfbwrapper',
     'funcsigs',
     'configparser',
     'pytest>=%s' % PYTEST_MIN_VERSION
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,6 @@ future>=0.15.2
 simplejson>=3.8.0
 prov>=1.4.0
 click>=6.6.0
-xvfbwrapper
 psutil
 funcsigs
 configparser
--- a/rtd_requirements.txt
+++ b/rtd_requirements.txt
@@ -9,7 +9,6 @@ pytest-cov
 future>=0.15.2
 simplejson>=3.8.0
 prov>=1.4.0
-xvfbwrapper
 psutil
 funcsigs
 configparser


satra:
@oesteban, @chrisfilo - what do you think - i'm fine with this line of thinking.

chrisfilo:
I'm OK with this as long as we document well when users should install xvfb.

oesteban:
Well, nipype could be optimized in many directions as regards dependencies. We try to keep them as limited as we can, but many dependencies could be workaround besides the xvfb.

The only dependency (beyond the tiny ```xvfbwrapper``` package and the ```xvfb``` binaries) that is different from a headless system to a system with display is ```xorg-server``` (this is a big one, that is true). However, in the latter you 99% of the times have it installed because you want to use XWindows. So little difference here.

In sum, the only use case for which xvfb is useless is: when 1) nipype runs in a headless system and 2) the pipeline does not involve AFNI at all (since for some reason they compiled AFNI linking to graphical libraries and use the ```$DISPLAY``` in all binaries), some FSL interfaces, and any VTK-based interface (possibly some part of FreeSurfer is included here too). I've seen that developers are trying to reduce those AFNI dependencies when they are not really justified, but I can't tell beyond ```3dSkullStrip``` how far they got on this with their latest version.

I would agree on moving xvfbwrapper as an extra dependency. This way, the user could control if they don't want it to be installed.

The way we call xvfbwrapper is largely optimizable. I also agree on this. First of all we should check that xvfb is available:

```
    def _run_wrapper(self, runtime):
        sysdisplay = os.getenv('DISPLAY')
        if self._redirect_x:
            try:
                from xvfbwrapper import Xvfb
            except ImportError:
                iflogger.error('Xvfb wrapper could not be imported')
                raise

            vdisp = Xvfb(nolisten='tcp')
            vdisp.start()
            try:
                vdisp_num = vdisp.new_display
            except AttributeError:  # outdated version of xvfbwrapper
                vdisp_num = vdisp.vdisplay_num

            iflogger.info('Redirecting X to :%d' % vdisp_num)
            runtime.environ['DISPLAY'] = ':%d' % vdisp_num

        if self._redirect_x:
            vdisp.stop()

        return runtime
```

Particularly, if we remove the dependency, that raise should be replaced by a more reliable alternative. We can't just stop if the dependency is not mandatory. Also, these lines would crash if you have xvfbwrapper but you don't have xvfb installed. Maybe it is better leave the responsibility of crashing to the wrapped software, maybe it does not fail!. Nipype can definitely use any effort in improving that bit of code.  Of course there is only one import for xvfbwrapper, but all interfaces can use it...

The code of xvfbwrapper is pretty simple itself, indeed it was part of nipype's code some time ago. We could bring it back, but I honestly prefer to defer maintenance to the python community as it is right now.

Finally, and regarding this particular PR:

  - Removing xvfbwrapper from rtd_requirements.txt does not have any practical implication (actually it could break the RTD build), so I'd say it should stay on that file.
  - Removing xvfbwrapper from requirements.txt only affects the developers, I would not mind about it since developers can workaround the missing dependency.
  - In info.py I would keep it where it is since I think the corner case here is only those users running nipype with a display. If we want to call xvfb only when it is really necessary (and we would need to discuss on how to redirect to the existing $DISPLAY, for those softwares creating windows and stuff), then we need to think of deeper changes on interfaces/base.py.

HTH

oesteban:
Forgot to mention that we should create the virtual frame buffer only once (i.e. a MapNode with 1000 iterations would call xvfb 1000 times when only one for all is needed). There is an issue about this somewhere in the past

blakedewey:
#1403

TheChymera:
@oesteban 

>  2) the pipeline does not involve AFNI at all

This is not true for me, I am using AFNI nodes in my preprocessing (`3dFWHMx`) and it works without requiring xorg.

> Particularly, if we remove the dependency, that raise should be replaced by a more reliable alternative.

To me the code looks like it works as it is. If xvfbwrapper is not installed or working properly, an error is raised, and the message is logged.

> In info.py I would keep it where it is since I think the corner case here is only those users running nipype with a display.

I don't think I understand this. What is the downside to xvfbwrapper missing from `info.py`?

satra:
@oesteban - are we good with this?

oesteban:
I think we should keep ```xvfbwrapper``` as an extra_dependency, at least.

TheChymera:
@oesteban what do you mean exactly? I could not find `extra_dependency` in our source code, nor in the python docs.

oesteban:
Here: https://github.com/nipy/nipype/blob/master/nipype/info.py#L162.

Sorry I misspelled the name [extra_requires](https://github.com/nipy/nipype/blob/master/setup.py#L140).

TheChymera:
@oesteban looking good? 

Can someone start travis again?

TheChymera:
@satra strange, I still can't use nipypecli on my headless machines:

```
chymera@neurohost ~/NIdata/ofM.dr/preprocessing/crashdump $ nipypecli crash crash-20170226-183106-chymera-get_f_scan.aI.a02.b16-05cf1c41-d646-496e-befc-47f394f85c4e.pklz
  File "/usr/lib/python-exec/python2.7/nipypecli", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 3030, in <module>
    @_call_aside
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 3014, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 3043, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 654, in _build_master
    ws.require(__requires__)
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 968, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 854, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'xvfbwrapper' distribution was not found and is required by nipype
```

The traceback comes from files not provided by nipype - so I have no idea what's going on here. Why is it that it's still checking for xvfbwrapper? Is it because it's in `EXTRA_REQUIRES`? I actually doubt that. DiPy is also in there, and it doesn't complain about that :-/

PR #1824, opened 2017-02-17
FIX: Correctly detect completed ReconAll steps

effigies:
Mislabeled locations caused some parts of ReconAll to be re-done unnecessarily in v6, specifically `parcstats*` and `balabels`.

Affects both FreeSurfer 5.3 and 6. I missed this before because these are fairly short recon steps, and I was mostly focused on `mri/` and `surf/`. Some steps were un-/misdocumented, as well, so this PR adds them to match actual behavior.

If recon-all decides there's nothing to do, it returns with an exit code of 1, so in this case I've switched to an informative echo. Added a `force_run` instance variable for use with non-standard flags (e.g. `-hippo-subfields` or `-clean-*`).

Tested on a completely reconstructed subject in 5.3. Will test on 6.0 momentarily.

Follow-up to #1790.


oesteban:
CircleCI should be fixed after merging #1825, sorry about that.

effigies:
No worries. This one needs to be tested by hand, anyway.

effigies:
Okay. I've tested this on FSv5.3 and FSv6.0. Rebased to test on new circle config.

PR #1828, opened 2017-02-21
[ENH] Refactoring of nipype.interfaces.utility

oesteban:
This PR splits the interfaces in nipype.interfaces.utility to make this code easier to maintain. The new three sub-modules are:

- nipype.interfaces.utility.base -> general concept of utilities (IdentityInterface, Merge, Select, Split, etc)
- nipype.interfaces.utility.csv -> CSV handling. Now it only contains the ReadCSV interface, but we could move in here some of the csv handling interfaces in nipype.algorithms.misc
- nipype.interfaces.utility.wrappers -> the ```Function``` interface for now. I am working on a Workflow wrapper that will be placed here.

All the interfaces are loaded at module import, so the new module should behave the same way it was doing before.


nipype/interfaces/utility/base.py
```diff
@@ -1,16 +1,21 @@
...
+"""
+Various utilities
 
    Change directory to provide relative paths for doctests
-   >>> import os
-   >>> filepath = os.path.dirname( os.path.realpath( __file__ ) )
-   >>> datadir = os.path.realpath(os.path.join(filepath, '../testing/data'))
-   >>> os.chdir(datadir)
+
+   .. testsetup::
```
satra: is the older form not necessary anymore?

nipype/interfaces/utility/csv.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+"""CSV Handling utilities
+
+   Change directory to provide relative paths for doctests
+
+   .. testsetup::
+      import os
```
satra: same question here.

nipype/testing/fixtures.py
```diff
@@ -25,12 +30,13 @@ def analyze_pair_image_files(outdir, filelist, shape):
 
 
 def nifti_image_files(outdir, filelist, shape):
+    if not isinstance(filelist, (list, tuple)):
```
satra: this could be replaced with `filename2list` from `filemanip`

nipype/interfaces/utility/base.py
oesteban: I thought we were printing those code-blocks in the documentation, but now I see we must be manually parsing them out, right?.

The ```.. testsetup::``` directive just makes sure that the code is run before doctests but not printed out in the documentation.

I guess this would be more appropriate, since sphinx has a directive for this use case of docstrings. But I don't mind to roll it back if you want :).

nipype/testing/fixtures.py
oesteban: Will do, thanks!

oesteban:
Note: this can be merged after #1833 (actually, I'd say we should merge #1833 when ready, before any other PR since circle ci tests are failing silently right now)

PR #1831, opened 2017-02-21
FIX: Import NUMPY_MMAP inside functions

effigies:
Follow-up to #1796. Some imports were at the file level when they should have been at the function level. My criterion for moving was whether there are other function-local imports.

Addresses issue discovered in poldracklab/fmriprep#362 (and poldracklab/fmriprep#363).


effigies:
cc @oesteban @satra for review. (Apparently requesting review is a privileged act? Weird.)

PR #1835, opened 2017-02-22
[ENH/WIP] WorkflowInterface

oesteban:
This workflow creates a new type of interface that wraps workflows. A naive addendum has been done to the BaseWorkflow class, in order to post-process the executed graph to retrieve the outputs that were connected to an "outputnode". This addition only happens for synchronous runners. For now, only the Linear and MultiProc plugins are set as synchronous.

The inner workflow is then run using either Linear or Multiproc.

The doctest reads like:
```
    >>> from nipype.pipeline import engine as pe
    >>> from nipype.interfaces.utility.base import IdentityInterface
    >>> wf = pe.Workflow('test_workflow')
    >>> node1 = pe.Node(IdentityInterface(fields=['a', 'b']), name='inputnode')
    >>> func = 'def func(a, b): return a + b'
    >>> node2 = pe.Node(Function(input_names=['a', 'b'], output_names=['out'],
    ...                 function=func), name='strconcat')
    >>> node3 = pe.Node(IdentityInterface(fields=['out']), name='outputnode')
    >>> wf.connect([
    ...     (node1, node2, [('a', 'a'), ('b', 'b')]),
    ...     (node2, node3, [('out', 'out')])
    ... ])
    >>> wi = WorkflowInterface(workflow=wf)
    >>> wi.inputs.a = 'b input '
    >>> wi.inputs.b = 'was concatenated to a'
    >>> res = wi.run()
    >>> res.outputs.out
    'b input was concatenated to a'
```

Before this PR can be merged, there are these tasks still open:

  - [ ] Merge #1828 (this refactoring was a preliminary work to this PR)
  - [ ] Write some additional documentation (besides the Function interface)
  - [ ] Pass a config object to the inner workflow (probably overwriting the ```stop_at_first_crash``` option to ```True```, maybe @satra can think of other options that should be set and potential problems of this approach)
  - [ ] Add more tests:
    - [ ] workflow that fails for some traits setting errors
    - [ ] workflow that fails for some runtime error (maybe a Function interface that just raises)
    - [ ] embed the new interface in a mapnode
    - [ ] run the interface with iterables


satra:
@oesteban - the new api we are working on basically has the following subclassing structure `Node -> Workflow`, so by definition a `Workflow` can be a `Node`. we have a discussion with @effigies and @djarecka scheduled for tomorrow and hoping we have a draft to present after that.

we are considering `input_map` and `output_map` as parameters to `Workflow` that can basically route `inputs` and `outputs`, without creating explicit `inputspec` and `outputspec` identity nodes. by default `input_map` and `output_map` would be a flattened key_value object for all nodes of the workflow, but a user could constrain that space by explicitly specifying these mappings, or sub-selecting from them.

however, this would be a good interface to implement that functionality.

effigies:
@oesteban is this one orphaned?

satra:
@effigies and @oesteban - i wouldn't mind seeing this in the 1.0 branch, but this would be explicitly supported in 2.0 engine.

oesteban:
I can recheck in two weeks. If re-syncing is not too hard (better said: very easy), we can add it as a patch for 1.0. As Satra mentioned, this does not make sense in 2.0.

PR #1840, opened 2017-02-23
Support for motion parameters produced by AFNI (FD)

chrisfilo:


nipype/algorithms/confounds.py
```diff
@@ -195,7 +195,8 @@ def _list_outputs(self):
 
 
 class FramewiseDisplacementInputSpec(BaseInterfaceInputSpec):
-    in_plots = File(exists=True, mandatory=True, desc='motion parameters as written by FSL MCFLIRT')
+    in_file = File(exists=True, mandatory=True, desc='motion parameters as written by FSL MCFLIRT or AFNI 3dvolreg')
+    format = traits.Enum("FSL", "AFNI", desc="Format of the motion parameters file: FSL (radians), AFNI (degrees)")
```
satra: why not use the same set of support that's in `ArtifactDetect`?

https://github.com/nipy/nipype/blob/86cd8ffb6dae1362a07ed8e9c79d583db87780eb/nipype/algorithms/rapidart.py#L163

and parameter interpretation is done here.

https://github.com/nipy/nipype/blob/86cd8ffb6dae1362a07ed8e9c79d583db87780eb/nipype/algorithms/rapidart.py#L42

ps. the composite norm output of artifact detect is another way of measuring motion.

chrisfilo:
All good after last commit @satra ?

nipype/algorithms/confounds.py
```diff
@@ -249,9 +252,11 @@ class FramewiseDisplacement(BaseInterface):
     }]
 
     def _run_interface(self, runtime):
-        mpars = np.loadtxt(self.inputs.in_plots)  # mpars is N_t x 6
+        mpars = np.loadtxt(self.inputs.in_file)  # mpars is N_t x 6
         diff = mpars[:-1, :] - mpars[1:, :]
         diff[:, :3] *= self.inputs.radius
+        if self.inputs.parameter_source == "AFNI":
+            diff[:, :3] *= (np.pi / 180)
         fd_res = np.abs(diff).sum(axis=1)
```
effigies: I'd suggest modifying the `mpars` rather than the diff, just to make it a little clearer what's going on. Following also adds FSFAST support.

```Python
mpars = np.loadtxt(self.inputs.in_file)  # mpars is N_t x 6 (N_t x 7 for FSFAST)
if self.inputs.parameter_source in ('AFNI', 'FSFAST'):
    mpars = mpars[np.asarray([4, 5, 3, 1, 2, 0]) + (mpars.shape[1] > 6)]
    mpars[:, 3:] *= np.pi / 180
diff = np.diff(mpars, axis=0)
diff[:, :3] *= self.inputs.radius
fd_res = np.abs(diff).sum(axis=1)

Comment (effigies) [inline nipype/algorithms/confounds.py]: UGH. Do not use the above, which re-sorts the parameters. This can be done more simply by taking the last 6 columns.

```Python
mpars = np.loadtxt(self.inputs.in_file)[:, -6:]  # mpars is N_t x 6
if self.inputs.parameter_source in ('AFNI', 'FSFAST'):
    mpars[:, :3] *= np.pi / 180
diff = np.diff(mpars, axis=0)
diff[:, :3] *= self.inputs.radius
fd_res = np.abs(diff).sum(axis=1)
```

And this doesn't handle SPM, contra my previous comment, because SPM puts its rotation parameters first. (Though my confused comments may support the idea of solving the "load normalized motion params" issue once and reusing.)

satra:
@effigies - if you hadn't seen my comments to chris earlier, here is the other place in nipype we deal with this:

https://github.com/nipy/nipype/blob/86cd8ffb6dae1362a07ed8e9c79d583db87780eb/nipype/algorithms/rapidart.py#L42

effigies:
@satra I did, and that's what I was working off of for my first comment (and completely skipped over the FSL/SPM difference). Which makes this (if we count #1379) three places where we (at least partially) solve the same problem of normalizing for parameter source. That one only handles a single row, so it's not quite a drop-in replacement, but still.

chrisfilo:
How does this new refactor look to you?

effigies:
Looks reasonable.

PR #1845, opened 2017-02-24
Add ANTs' KellyKapowski interface

alexsavio:
Hi,

This is the command for DiReCT cortical thickness measuring. Instead of using the whole antsCorticalThickness.sh script (which includes bias-field correction, warping, tissue segmentation, etc..).

This study shows that other combinations using only DiReCT can lead to more robust results:
http://dx.doi.org/10.1016/j.nicl.2016.05.017


alexsavio:
this is something related with the output of the example. maybe I have to add the 'u' in the beginning of the output string? thanks!

nipype/interfaces/ants/segmentation.py
```diff
@@ -1153,3 +1154,166 @@ def _list_outputs(self):
...
+
+    def _parse_inputs(self, skip=None):
+        if skip is None:
+            skip = []
+        skip += ['warped_white_matter', 'gray_matter_label', 'white_matter_label']
+        return super(KellyKapowski, self)._parse_inputs(skip=skip)
+
+    def _list_outputs(self):
+        outputs = self._outputs().get()
+        outputs['cortical_thickness']  = os.path.abspath(self._gen_filename('cortical_thickness'))
```
satra: could we use namesource and template for these two?

alexsavio: do you mean to change the name of the output parameters?

satra: e.g., like this https://github.com/nipy/nipype/blob/0620abaff5776b6b19268fab5ac3192e37779da0/nipype/interfaces/freesurfer/preprocess.py#L1506

then you can get rid of the list_outputs function.

alexsavio:
done. thanks for the review.

nipype/interfaces/ants/segmentation.py
```diff
@@ -1153,3 +1154,160 @@ def _list_outputs(self):
...
+                                        desc="Sets the option for B-spline smoothing of the velocity field.")
+
+    number_integration_points = traits.Int(10, argstr="--number-of-integration-points %d",
+                                           desc="Number of compositions of the diffeomorphism per iteration.")
+
+    max_invert_displacement_field_iters = traits.Int(20, argstr="--maximum-number-of-invert-displacement-field-iterations %d",
+                                                     desc="Maximum number of iterations for estimating the invert \n"
+                                                          "displacement field.")
+
+    cortical_thickness = File(argstr='--output "%s"', genfile=True, keep_extension=True,
```
satra: i don't think genfile is required here.

nipype/interfaces/ants/segmentation.py
```diff
@@ -1153,3 +1154,160 @@ def _list_outputs(self):
...
+                    'tags': ['implementation'],
+                    }]
+
+    def _parse_inputs(self, skip=None):
+        if skip is None:
+            skip = []
+        skip += ['warped_white_matter', 'gray_matter_label', 'white_matter_label']
+        return super(KellyKapowski, self)._parse_inputs(skip=skip)
+
+    def _gen_filename(self, name):
```
satra: i think you can delete this function as well.

alexsavio: you are right. fixed.

alexsavio: If I delete this function, how do I format the `output_file` parameter for the commandline in the `_format_arg` function?

alexsavio: I had to do this because the `output` command argument is a 2-file argument with the values of `cortical_thickness` and `warped_white_matter`:
```
--output "[segmentation0_cortical_thickness.nii.gz,segmentation0_warped_white_matter.nii.gz]"
```
Is there a better way to do this? This is the scheme of I could find from other ANTs interfaces.

satra:
@alexsavio - could you please merge with master and push - we have been trying to clean up travis/circle tests.

PR #1848, opened 2017-02-24
draft of the simplest one and twon nodes tests

djarecka:
@satra @effigies - I added the simplest test for one and two nodes, mapper includes one or two variables only. Before I start implementing more tests or fill the code, we should agree on this, so let me know.

I also included several questions related to either general concepts, or implementation (that is less important, but just to let you know what I was assuming)


nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    N.run()
+    assert (N.outputs["out"] == expected_output).all()
+
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a":[3, 1, 8], "b":[0, 1, 2]}, [0, 1, 16]),
+        ({"a":[3, 1, 8], "b":[2]}, [6, 2, 16]),
+        ])
+def test_single_node_3(inputs_dict, expected_output):
+    N = Node(interface=fun3, mapper='(a,b)')
```
satra: mapper = ('a', 'b')

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a":[3, 1], "b":[1, 2, 4]}, [[3, 6, 12], [1, 2, 4]]),
+        ({"a":[[3, 1]], "b":[1, 2, 4]},[[[3, 6, 12], [1, 2, 4]]]),
+        ({"a":[[3, 1], [30, 10]], "b":[1, 2, 4]},
+         [[[3, 6, 12], [1, 2, 4]],[[30, 60, 120],[10, 20, 40]]]),
+        ({"a":[3, 1], "b":[2]}, np.array([[6], [2]])),
+        ])
+def test_single_node_4(inputs_dict, expected_output):
+    sn = Node(interface=fun3, mapper='[a,b]')
```
satra: `mapper = ['a', 'b']` 

docstring should note that is a cross-product.

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.mark.parametrize("inputs_dict", [
+        {"a":[[3, 1], [0,0]], "b":[1, 2, 0]},
+        {"a":[[3, 1], [0,0], [1, 1]], "b":[1, 2, 0]}, # think if this should work
+        {"a":[[3, 1, 1], [0,0, 0]], "b":[1, 2, 0]},  # think if this should work
+        ])
+def test_single_node_wrong_input(inputs_dict):
+    with pytest.raises(Exception):
+        sn = Node(interface=fun3, mapper='(a,b)')
```
satra: same comments as before.

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+        sn.inputs = inputs_dict
+        sn.run()
+
+def test_single_node_wrong_key():
+    with pytest.raises(Exception):
+        sn = Node(interface=fun3, mapper='(a,b)')
```
satra: same comments as before.

nipype/pipeline/engine/test_newengine/test_multi_nodes.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.mark.parametrize("inputs_dict1, inputs_dict2, expected_output", [
+        ({"a": [3, 4, 5]}, {"b": [10, 20, 30]}, [19, 36, 55]),
+          ])
+def test_2nodes_2(inputs_dict1, inputs_dict2, expected_output):
+    wf = Workflow()
+    N1  = Node(inputs=inputs_dict1, mapper="a", interface=fun1)
+    wf.add_nodes(N1)
+
+    N2 =  Node(inputs=inputs_dict2, mapper="(N1.a,b)", interface=fun4)
```
satra: `... mapper=('N1.a', 'b'), ...`

nipype/pipeline/engine/test_newengine/test_multi_nodes.py
```diff
@@ -0,0 +1,89 @@
...
+         [[11, 23, 39, 67], [11, 24, 46, 104], [11, 25, 55, 165]]),
+          ])
+def test_2nodes_3(inputs_dict1, inputs_dict2, expected_output):
+    wf = Workflow()
+    N1  = Node(inputs=inputs_dict2, mapper="a", interface=fun2)
+    wf.add_nodes(N1)
+
+    # IMPORTANT: I understand that if b is used in a mapper with A,
+    # I do not anymore assume a x b, i.e. no "state input mapping"
+    N2 =  Node(inputs=inputs_dict2, mapper="(A,b)", interface=fun4)
```
satra: `... mapper=['A', 'b'], ...`

nipype/pipeline/engine/test_newengine/test_multi_nodes.py
```diff
@@ -0,0 +1,89 @@
...
+          [[11, 21, 31, 41], [15, 25, 35, 45], [35, 45, 55, 65], [135, 145, 155, 165]]]),
+          ])
+def test_2nodes_3(inputs_dict1, inputs_dict2, expected_output):
+    wf = Workflow()
+    N1  = Node(inputs=inputs_dict2, mapper="a", interface=fun2)
+    wf.add_nodes(N1)
+
+    # IMPORTANT: I understand that if b is used in a mapper with A, 
+    # I do not anymore assume a x b, i.e. no "state input mapping"       
+    N2 =  Node(inputs=inputs_dict2, mapper="[A,b]", interface=fun4)
```
satra: `... mapper= ['A', 'b'], ...`

satra:
@djarecka - both files should also have examples for `groupByIndex` and `groupByValue`

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    return a**pow
+
+def fun3(a, b, **dict):
+    return a * b
+
+
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [ 
+        ({"a": [3, 4, 5]}, [9, 16, 25]),
+        ({"a": [[3, 4, 5], [6, 7, 8]]}, [[9, 16, 25], [36, 49, 64]]),
```
satra: i think this will raise an exception since `[3, 4, 5] ** 2` is not valid in python. 

in this context, each input to fun1 is a list.

satra: whereas this would work:

`({"a": [np.array([3, 4, 5]), np.array([6, 7, 8])]}, [np.array([9, 16, 25]), np.array([36, 49, 64])]),`

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    assert (N.outputs["out"] == expected_output).all()
+    
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a": [3, 4, 5]}, [9, 16, 25]),
+        ({"a": [[3, 4, 5], [6, 7, 8]]}, [[9, 16, 25], [36, 49, 64]]),
+        ({"a": np.array([3, 4, 5])}, [9, 16, 25]),
+        ])
+def test_singlenode_1a(inputs_dict, expected_output):
+    N  = Node(inputs=inputs_dict, mapper="a", interface=fun1)
```
satra: should remove `inputs_dict` from here.

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    # testing if you can set inputs outside __init__
+    N.inputs = inputs_dic
```
satra: spelling.

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    N.run()
+    assert (N.outputs["out"] == expected_output).all()
+
+
+def test_single_node_1b():
+    # how to name or rename (from a default name, e.g. "out") outputs 
+    sn = Node(interface=my_function_1, mapper='a', output_name=["out_1"])
```
satra: this is not necessary for now. let's assume that there is default output key (`out`) if no output dictionary is provided by the function or interface.

nipype/pipeline/engine/test_newengine/test_single_node.py
satra: or unless fun1 became:

```python
def fun1(a):
    import numpy as np
    return np.power(a, 2)
```

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
+from supernode import Node # or whatever
+
+import numpy as np
+
+import pytest, pdb
+
+def fun1(a, **dict):
```
satra: why do these functions have `**dict`

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    # how to name or rename (from a default name, e.g. "out") outputs 
+    sn = Node(interface=my_function_1, mapper='a', output_name=["out_1"])
+    sn.inputs = {"a" : [3, 1, 8]}
+    sn.run()
+    assert (sn.outputs["out_1"] == [9, 1, 64]).all()
+
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a": [3, 4, 5]}, [[1, 3, 9, 27], [1, 4, 16, 64], [1, 5, 25, 125]]),
+        ({"a": [[3, 4, 5], [6, 7, 8]]},
```
satra: this will raise exception

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,103 @@
...
+    sn.run()
+    assert (sn.outputs["out_1"] == [9, 1, 64]).all()
+
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a": [3, 4, 5]}, [[1, 3, 9, 27], [1, 4, 16, 64], [1, 5, 25, 125]]),
+        ({"a": [[3, 4, 5], [6, 7, 8]]}, 
+         [[[1, 3, 9, 27], [1, 4, 16, 64], [1, 5, 25, 125]], 
+          [[1, 6, 36, 196], [1, 7, 49, 343], [1, 8, 64, 512]]]),
+        ({"a": np.array([3, 4, 5])}, [[1, 3, 9, 27], [1, 4, 16, 64], [1, 5, 25, 125]]),
```
satra: so will this.

satra:
in these examples since there is no groupby, the outputs would all have state information.

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: ok, i changed the mapper that it's not a single string. I understand that you think about docstrings in the Node class

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: In my previous drafts of the Node class I always converted list to numpy array during input set, so the user could provide either list or numpy array, as it is in matplotlib etc., see various inputs_dict in `test_singlenode_1`. Of course, this can be change if you don't like it.

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: yes

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: that is what I assumed in all other examples. I can xfail this test for now.

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: I was calling function like this `self.Interface(**self.inputs)` and was assuming that self.inputs can have other keys, but not sure if we will need it, depends on the final implementation, e.g. not sure where we will store "N1.a" variable.

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: why? I thought that when I'm mapping over "a", I'm sending to the fun2 one element at a time, so I execute `fun2(3)`, `fun2(4)` and `fun2(5)`.

djarecka:
@satra - I will make groupby examples, but should I understand that you want node.output to look like `{"out": [("a=3", 9), ("a=4", 16)]}` or  `{"out": [(3, 9), (4, 16)]}`. Do I understand correctly that you just want to provide this info when a user is asking, i.e. linking the elements of `self._outputs` to `self.inputs["a"]` in a get output method?

nipype/pipeline/engine/test_newengine/test_single_node.py
satra: since the map is over the outer elements of a, not over inner elements.

djarecka: you mean the second example? I see what you mean, so we only map the most "outer axis" and don't want to go deeper. will remove it.

djarecka:
@satra @effigies - tried to include suggestions and add some groupByIndex examples, again don't want to add too many tests before checking with you some basic syntax (that is really not as straightforward when you have all those option of mapping and reducing). FYI, the comments are included in order to explain what I had in mind, but will be remove (or replaced). 

I have one bigger issue. I agreed yesterday that groupBy can be over those inner elements of a node output (e.g. `groupByInd="A"` in the (`test_2nodes_3`)[ https://github.com/nipy/nipype/pull/1848/files#diff-404586ac72a92b8db175bb0334e0276cR99]),  but again I feel this is not such a great idea...  Elements of `A` can be related to various values of contrast or something else, but values of `A_i` are the combinations of contrast and N1.a, so if we group by index we will not get unique values of `A` in one group.  And if we decide to groupingByValue we will get completely different grouping and we might group together elements that are related to different values of contrast. If this is really what you want? 
If contrast (not `A`) is a meaningful variable, I would think that contrast really should be a state input. See 2 approaches below,  they should do the same calculations at the end, but in the second approach contrast is a state input and you can always group/reduce using N1.contrast.

1)  
```python
def fun_N1(subject):
    list_of_contrast = np.array([1, 2, 3])
    return list_of_contrast * subject.something

def fun_N2(other_variable, from_fun_N1):
    return something

wf = Workflow()
N1 = Node(inputs={"subject": ["A", "B", "C"]}, interface=fun_N1, mapper="subject")
wf.add_nodes(N1)
N2 = Node(inputs={"other_variable": [a, b, c]}, interface=fun_N2, mapper=["from_fun_N1", "other_variable"])
wf.connect(N1, N1.out, N2, N2.from_fun_N1)
```

2)
```python
def fun_N1(subject, contrast):
    return contrast * subject.something

def fun_N2(other_variable, from_fun_N1):
    return something

wf = Workflow()
N1 = Node(inputs={"subject": ["A", "B", "C"], "contrast":[1,2,3]}, interface=fun_N1, mapper=["subject", "contrast"])
wf.add_nodes(N1)
N2 = Node(inputs={"other_variable": [a, b, c]}, interface=fun_N2, mapper="other_variable")
wf.connect(N1, N1.out, N2, N2.from_fun_N1)
```

I remember you said that you're also using already implemented interfaces (so you can't easily change the implementation), but if the interface returns an array that every element is related to a specific value of some meaningful variable (that is a list), e.g. contrast, you should have access to this variable. Otherwise, that wouldn't be a meaningful variable if you don't care about values. Please, correct me if I'm wrong, but if you do have access to this variable, we can create an extra input state that can be later used as a grouper/reducer.

nipype/pipeline/engine/test_newengine/test_single_node.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a":[3, 1], "b":[1, 2, 4]}, [[(["a=3","b=1"], 3), (["a=3","b=2"], 6), (["a=3","b=4"], 12)], 
+                                       [(["a=1","b=1"], 1), (["a=1","b=2"], 2), (["a=1","b=4"], 4)]]),
+        ({"a":[[3, 1], [30, 10]], "b":[1, 2, 4]},
+         [[[(["a=3","b=1"], 3), (["a=3","b=2"], 6), (["a=3","b=4"], 12)], 
+           [(["a=1","b=1"], 1), (["a=1","b=2"], 2), (["a=1","b=4"], 4)]],
+          [[(["a=30","b=1"], 30), (["a=30","b=2"], 60), (["a=30","b=4"], 120)],
+           [(["a=10","b=1"], 10), (["a=10","b=2"], 20), (["a=10","b=4"], 40)]]]),
```
effigies: By my understanding, this should be:
```Python
({"a":[[3, 1], [30, 10]], "b":[1, 2, 4]},
 [[(["a=[3, 1]","b=1"], [3, 1]), (["a=[3, 1]","b=2"], [3, 1, 3, 1]), (["a=[3, 1]","b=4"], [3, 1, 3, 1, 3, 1, 3, 1])], 
  [(["a=[30, 10]","b=1"], [30, 10]), (["a=[30, 10]","b=2"], [30, 10, 30, 10]), (["a=[30, 10]","b=4"], [30, 10, 30, 10, 30, 10, 30, 10])]])
```

nipype/pipeline/engine/test_newengine/test_groupby.py
```diff
@@ -0,0 +1,163 @@
...
+        assert out[0] == expected_output[i][0] 
+        assert (out[1][1] == expected_output[i][1][1]).all() #still assuming that values will be returned as arrays 
+
+    
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a":[3, 1, 8], "b":[0, 1, 2]}, 
+         [(["a=3"], (["b=0"], 0)), (["a=1"], (["b=1"], 1)), (["a=8"], (["b=2"], 16))]), # since there is a dot product, after grouping over "a" we have still only single values (but they are connected to values of "b") 
+        ])
+def test_single_groupbyind_2(inputs_dict, expected_output):
+    N = Node(interface=fun3, mapper=('a','b'), groupByInd="a")
```
djarecka: @effigies - very sorry for late answer, i completely missed your review 2 weeks ago... you're right that `groupyBy*` replacing temporary `reducer` (I think we didn't decide on the final name). I was suggesting `groupBy` notation, since in our examples we're not "reducing" anything but "grouping" output based on some criteria. 
I was concentrating mostly on "GroupByInd", since when you think about NumPy arrays it can be coded easier and I believed that this is the operation that you usually need. Generally speaking, if the input state contains only unique values, there is only one option. But if your input state looks like `a = [2, 3, 4, 2]` then `GroupByInd` gives four elements and `GroupByValue` gives three elements (combining output for `a[0]` and `a[3]`)

nipype/pipeline/engine/test_newengine/test_single_node.py
djarecka: I'm not sure about it... The first element `(["a=[3, 1]","b=1"], [3, 1])` looked a bit like mine, but I just broke into pieces, since I assumed that the output should be 3-dim array, that every element is related to one value of `a` and one element of `b`. I also don't know why your second element has a 4-elements list `[3, 1, 3, 1]`.
I will review these examples today, so will try to understand better what you had in mind.

nipype/pipeline/engine/test_newengine/test_groupby.py
```diff
@@ -0,0 +1,188 @@
...
+
+    
+@pytest.mark.parametrize("inputs_dict, expected_output", [
+        ({"a":[3, 1, 8], "b":[0, 1, 2]}, 
+         # output is 1dimensional, out[0] is for a[0]=3, etc.
+         # since mapper was a.b out[0][1] has only one element, for b[0]=0 
+         [({"a":3}, ({"b":0}, 0)), ({"a":1}, ({"b":1}, 1)), ({"a":8}, ({"b":2}, 16))]), # since there is a dot product, after grouping over "a" we have still only single values (but they are connected to values of "b") 
+        ])
+def test_single_groupbyind_2(inputs_dict, expected_output):
+    N = Node(interface=fun3, mapper=('a','b'), groupByInd="a")
```
oesteban: @satra, what would it be the coding-style guideline here for keyword arguments? camel-case like here for `groupByInd` or underscores like the `inputs_names` of the Function interface?

satra: i would suggest going with `_` since pandas, numpy, scipy follows that as well.

satra:
@djarecka has been looking at the api for a `State` class that will become an attribute of the `Node` object

roughly speaking the `State` class will take care of the mapping and grouping functions.

the mapping will result in a generator
the grouping will result in, possibly, a new `State` object

it will also have properties like `shape` and indexing into the `State` object

so if mapper creates a two dimensional array (e.g., subjects x algorithms),

state_obj.shape == (5, 3)
state_obj[2, 1] will return a variety of information about that particular cell. does it have output, what output it has, did it error, etc.,.

thus one Node can ask for the state of things of nodes it receives inputs from and create new states if requested. the object needs to be efficient enough not to increase memory requirements as one goes through the workflow. 

we hope to work towards some tests regarding this object this week.

djarecka:
@satra happy to talk about State class this week. I wrote some pieces, but wanted to discuss a few things before moving forward.

PR #1854, opened 2017-02-28
[WIP/ENH] Revise how docker build is run in CircleCI

oesteban:
Adds the ```--rm=false``` arguments so that docker does not remove
intermediate layers after build. This should speed up the build process
and allow Docker to export intermediate layers into the cache.

Also, the ~/docker folder has been added as an artifact to check if
the cache tarfile is generated.

Following up here: https://discuss.circleci.com/t/docker-layer-caching/10068/8?u=oesteban


oesteban:
Seems like there's no way to get around cicle's issues when caching docker. Many BTRFS features are not available during build and this idea cannot work. Closing.

PR #1856, opened 2017-02-28
fix: remove xor from EstimateResponseSH inputs

### LINKED ISSUE #1849: EstimateResponseSH dipy interface can not change default 'auto' input to False
### Summary
I can't find a way to set response estimation to recursive
### Actual behavior
Only auto_response can be used 
```Python
import nipype.interfaces.dipy as ndp

estimate = ndp.EstimateResponseSH
estimate.inputs.auto = False
```
throughs the following error
>IOError: Input "recursive" is mutually exclusive with input "auto", which is already set

and similarly with
```Python
estimate.inputs.recursive = True
```

### Platform details:
{'nibabel_version': '2.1.0', 'networkx_version': '1.11', 'numpy_version': '1.11.2', 'sys_platform': 'linux2', 'sys_version': '2.7.6 (default, Oct 26 2016, 20:30:19) \n[GCC 4.8.4]', 'commit_source': 'installation', 'commit_hash': 'a4d1e07', 'pkg_path': '/home/bougacha/.local/lib/python2.7/site-packages/nipype', 'sys_executable': '/usr/bin/python', 'traits_version': '4.6.0', 'scipy_version': '0.18.1'}
0.12.1

mgxd:
fix #1849 

- `xor` arguments and defaults for `auto` and `recursive` response estimators didn't allow the use of `recursive`


satra:
@mgxd - i would take the usedefault's away and keep the `xor`. those parameters are mutually exclusive (can you check).

mgxd:
@satra although there are checks inside the function as well, that sounds like a good idea to avoid confusion.

PR #1857, opened 2017-03-01
ENH: Add -hires and -expert flags to ReconAll

effigies:
FreeSurfer 6 supports a `-hires` flag for [submillimeter reconstruction](http://freesurfer.net/fswiki/SubmillimeterRecon). Currently, it's recommended that an [expert options file](https://surfer.nmr.mgh.harvard.edu/fswiki/recon-all#ExpertOptionsFile) be provided to modify default `mris_inflate` parameters, so I've also added the `-expert` flag, though it is not new.


effigies:
At present, this assumes people will create their own expert files. I could add a new interface to be used like:

```Python
expert_file = pe.Node(ExpertFile(mris_inflate='-n 15'), name='ExpertFile')
```

That produces the file

```Shell
mris_inflate -n 15
```

As a sketch of the specs limiting the options to programs that will accept expert files (there might be a better way):

```Python
class ExpertFileInputSpec:
    talairach = traits.Str(desc="Flags to pass to talairach commands")
    ...
    mri_aparc2aseg = traits.Str(desc="Flags to pass to mri_aparc2aseg commands")

class ExpertFileOutputSpec:
    out_file = File(exists=True, desc="Expert options file")
```

effigies:
Just to note: Segfault still [occurring](https://travis-ci.org/effigies/nipype/jobs/206623506) sometimes on Travis.

effigies:
Added the `ExpertOptions` interface since I wanted to experiment with using it. Probably won't need it after all (will just manually create the file), so I'll only keep it if it seems like a good idea in general.

chrisfilo:
Isn't the ExpertOptions interface a little bit of an overkill? I mean it only produces one config file and it will only be connected to ReconAll node (correct me if I am wrong). Wouldn't it be more straightforward to fold this into ReconAll itself?

We have created on demand config files on other occasions https://github.com/nipy/nipype/blob/master/nipype/interfaces/dcm2nii.py#L195

effigies:
@chrisfilo How do you feel about this fold-in? It accepts either an existing file or a dictionary of `{binary: options_string}` and creates a file if needed. (This doesn't check for valid binaries, unlike the spec'd interface.)

I'm also okay with simply dropping file creation.

nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -632,6 +633,10 @@ class ReconAllInputSpec(CommandLineInputSpec):
                         desc="Number of processors to use in parallel")
     parallel = traits.Bool(argstr="-parallel",
                            desc="Enable parallel execution")
+    hires = traits.Bool(argstr="-hires", min_ver='6.0.0',
+                        desc="Conform to minimum voxel size (for voxels < 1mm)")
+    expert = traits.Either(File, DictStrStr, argstr='-expert %s',
```
chrisfilo: `File(exists=True)`?

chrisfilo:
Hmm... now we are not exposing all of the possible dict keys. What about adding all of those inputs to ReconAll https://github.com/nipy/nipype/pull/1857/commits/1941987503926dcccf3c1a456090a670358ba9c3#diff-08eb19cbbabcb3f9cef1bb3197b661ecL2885 and xoring on the expert_file input? This would have an advantage of people knowing straight away what are the possible options WDYT?

nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -933,6 +980,24 @@ def cmdline(self):
         iflogger.info('resume recon-all : %s' % cmd)
         return cmd
 
+    def _prep_expert_file(self):
+        if isdefined(self.inputs.extra):
```
chrisfilo: extra?

effigies: What I get for pushing before testing. Updated.

effigies:
@satra How do you feel about this interface? @chrisfilo is right that the expert file is `recon-all`-specific, but this definitely adds a lot to the input spec.

PR #1858, opened 2017-03-02
[FIX] CircleCI tests - error codes are not propagated

oesteban:
The former structure of the tests.sh file in which all the tests are called
did not exit with error if one of the test commands errored, since the
last exit code was from codecov, and no exit code tracking or propagation
was done.

This PR fixes that issue. Additionally, the scripts related to CircleCI
have been split up into a new .circle/ folder.


oesteban:
@chrisfilo I have some concerns with the collection of coverage results. Looking at the artifacts, it seems like only the node 0 is generating coverage xmls from the workflows.

My question is: are the rest of nodes not supposed to generate coverage results?

Other than that (and it could be fixed in another PR since we probably want this to be merged ASAP), this is ready to merge in.

chrisfilo:
Since unittests are only run on node 0 it is the only one generating coverage XML files. Other nodes run smoke tests and generate other XML files (which from what I can see are also uploaded).

oesteban:
@chrisfilo shouldn't this line at the end of the build output of node 2 generate an xml file?:

```
coverage xml -o /scratch/smoketest_fmri_spm_nested_MultiProc__root_examples__l2pipeline.xml
```

Apparently that is not happening now...

PR #1861, opened 2017-03-04
[ENH] Fix paths for coverage in circleci

oesteban:
For Circle to understand the test outputs, the xml files should be placed in a subfolder of ${CIRCLE_TEST_REPORTS} (see https://circleci.com/docs/1.0/test-metadata/#metadata-collection-in-custom-test-steps)

-- Realized right after #1858 was merged, sorry about that


satra:
@oesteban - can we work on merging this today? i'm free to chat/help after 1.30 eastern. 

also let's make sure that we do not use `/root` for anything. let's do everything in `/src` (nipype, workdir, etc.,.)

oesteban:
Hey @satra, this was failing (why merge?)... I'll try to fix it early tomorrow.

satra:
it had all green! but now looks like an accidental pass. if you look at the report above (with view details, you'll see all 6 checks passed.

PR #1862, opened 2017-03-05
FIX: AntsJointFusion unicode command line formatting

hjmjohnson:
When strings are formatted in unicode, then the format
of the AntsJointFusion -g, -t flags is not the same
as the python string representations of lists.  These
flags need to be unpacked one-by-one and a new string
representation needs to be provided on the command line.

### DIFF
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -1129,11 +1129,28 @@ ants_joint_fusion_posterior_%d.nii.gz, ants_joint_fusion_voting_weight_%d.nii.gz
             retval = ''
             if not isdefined(self.inputs.out_label_fusion):
                 retval = '-o {0}'.format(self.inputs.out_intensity_fusion_name_format)
+        elif opt == 'atlas_image':
+            atlas_image_cmd=" ".join(
+                [ '-g [{0}]'.format(",".join(
+                    fn for fn in ai) ) for ai in self.inputs.atlas_image]
+            )
+            retval = atlas_image_cmd
+        elif opt == 'target_image':
+            target_image_cmd=" ".join(
+                [ '-t [{0}]'.format(",".join(
+                    fn for fn in ai) ) for ai in self.inputs.target_image]
+            )
+            retval = target_image_cmd
+        elif opt == 'atlas_segmentation_image':
+            assert len(val) == len(self.inputs.atlas_image), "Number of specified " \
+                "segmentations should be identical to the number of atlas image " \
+                "sets {0}!={1}".format(len(val), len(self.inputs.atlas_image))
+            atlas_segmentation_image_cmd=" ".join(
+                [ '-t [{0}]'.format( fn ) for fn in self.inputs.atlas_segmentation_image ]
+            )
+            retval = atlas_segmentation_image_cmd
         else:
-            if opt == 'atlas_segmentation_image':
-                assert len(val) == len(self.inputs.atlas_image), "Number of specified " \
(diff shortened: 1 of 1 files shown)


satra:
@hjmjohnson - we should not be having any `u` -prefixed strings even though we use unicode in python 2. the `unicode_literals` import should have taken care of this.

if this is happening in the base command line class we should fix it there.

nipype/interfaces/ants/segmentation.py
```diff
@@ -1129,11 +1129,28 @@ def _format_arg(self, opt, spec, val):
...
+                    "'{0}'".format(fn) for fn in ai) ) for ai in self.inputs.atlas_image]
+            )
+            retval = atlas_image_cmd
+        elif opt == 'target_image':
+            target_image_cmd=" ".join(
+                [ '-t [{0}]'.format(", ".join(
+                    "'{0}'".format(fn) for fn in ai) ) for ai in self.inputs.target_image]
+            )
+            retval = target_image_cmd
+        elif opt == 'atlas_segmentation_image':
```
satra: this one should happen automatically.

nipype/interfaces/ants/segmentation.py
```diff
@@ -1129,11 +1129,28 @@ def _format_arg(self, opt, spec, val):
             retval = ''
             if not isdefined(self.inputs.out_label_fusion):
                 retval = '-o {0}'.format(self.inputs.out_intensity_fusion_name_format)
+        elif opt == 'atlas_image':
```
satra: for this flag is the sets of images intentional?

hjmjohnson: I think so.  My goal for this patch was to make the minimal number of changes necessary return AntsJointFusion to a working state.  The previous command line generated had problems with formatting the command lines due to the way that lists of strings were being created.  i.e. "-g %s" for [u'/file_a', u'/file_b'] was being presented as "-g [u/file_a, u/file_b]" on the command line.

satra:
@hjmjohnson - let me see if there is a general solution to this. i'm trying to figure out if this is something happening just in this ants interface or in general for any input spec handled by the commandline base class

satra:
ok looks like these are ants specific changes that need to be made. the base command class appears to be ok.

```python
In [25]: class Foo(nib.CommandLineInputSpec):
    ...:     foo = nib.InputMultiPath(nib.File, argstr='-t [%s]', sep=',')
    ...:     

In [26]: class CMD(nib.CommandLine):
    ...:     input_spec = Foo
    ...:     

In [27]: cmd = CMD('echo')

In [28]: cmd.inputs.foo = [u'a', u'b']

In [29]: cmd.cmdline
Out[29]: u'echo -t [a,b]'
```

satra:
and just for fun, here is a version with quotes on arguments:

```python
In [35]: class Foo(nib.CommandLineInputSpec):
    ...:     foo = nib.InputMultiPath(nib.File, argstr='-t ["%s"]', sep='", "')
    ...:     

In [36]: class CMD(nib.CommandLine):
    ...:     input_spec = Foo
    ...:     

In [37]: cmd = CMD('echo')

In [38]: cmd.inputs.foo = [u'a', u'b']

In [39]: cmd.cmdline
Out[39]: u'echo -t ["a", "b"]'
```

satra:
so your changes look reasonable - can you merge with upstream master and push?

hjmjohnson:
@satra Thank you for doing the extra exploration.   I will merge.

mgxd:
@hjmjohnson could you update the doctests to reflect the new changes?

PR #1866, opened 2017-03-07
fixing tests and install errors

satra:
- removing unnecessary autotests
- adding brukerdir to installer


satra:
address #1866 

@milesseidel - if you would like to test this yourself, you can try installing my branch while this pull request goes through review.

`pip install --upgrade https://github.com/satra/nipype/archive/fix/bru2.zip`

nipype/algorithms/tests/test_auto_CompCor.py
```diff
@@ -1,37 +0,0 @@
-# AUTO-GENERATED by tools/checkspecs.py - DO NOT EDIT
-from __future__ import unicode_literals
-from ..confounds import CompCor
-
-
-def test_CompCor_inputs():
```
oesteban: why these need to go away? We are continuously adding and removing them...

satra: because autotests are only added for an interface if a named test doesn't exist. and we have `test_compcor.py`

oesteban: I see, thanks!

milesseidel: @satra Thanks! 

$ pip install --upgrade https://github.com/satra/nipype/archive/fix/bru2.zip
$ python -c "import nipype; nipype.test()"
...
===== 585 passed, 1 skipped in 22.58 seconds ====

setup.py
```diff
@@ -130,7 +133,7 @@ def main():
         install_requires=ldict['REQUIRES'],
         setup_requires=['future', 'configparser'],
         provides=ldict['PROVIDES'],
-        packages=find_packages(exclude=['*.tests']),
+        packages=find_packages(),
```
satra: @oesteban, @chrisfilo - any idea why the tests directories were removed from setup? i'm putting them back in.

nipype/interfaces/fsl/tests/test_auto_ApplyXfm.py
```diff
@@ -1,9 +1,9 @@
 # AUTO-GENERATED by tools/checkspecs.py - DO NOT EDIT
 from __future__ import unicode_literals
-from ..preprocess import ApplyXfm
+from ..preprocess import ApplyXFM
```
effigies: `nipype/interfaces/fsl/tests/test_auto_ApplyXFM.py` already exists. Should this file just be removed?

satra: i think this PR has taken care of it. the issue is with OSX and case-insensitive installations, which gave rise to the same named file with different content. the older interface which was deprecated in 0.12.1 has now been removed, so we should be ok.

setup.py
oesteban: It is a bit blurry in my memory now, but I think this was added to avoid tests to be run from the installed version of nipype and make sure they run from the source  I guess py.test takes care of this, so it should be fine re-adding them.

Issue #1750, opened 2016-12-15
To consider: add retry check in the File trait (workflows more tolerant to NFS delays)


oesteban:
I am gauging the interest in this idea.

### Summary
In NFS systems, sometimes files are not synchronized fast enough and workflows may fail when propagating the outputs of a node into the inputs of the next.

### Actual behavior
The File trait with metadata ```exists=True``` only checks once, instantly, when the interface is run or the command line queried.

### Expected behavior
Modulated by a nipype config setting, add retries to the File trait validation with a certain delay. This way nipype will be more tolerant to NFS delays.

### How to replicate the behavior
N/A

### Platform details:
Nipype workflow running in any platform with a base_dir in an NFS mount.

satra:
@oesteban - this is what the timeout setting in the execution config is for.

oesteban:
@satra - you mean ```job_finished_timeout``` ? We are using the linear plugin, and it crashes for this reason (file not found). Then you check yourself and the files are there.

satra:
indeed. yes the linear plugin does not use that flag, but could be augmented to do the same thing that's done in the parallel execution plugin.

shoshber:
What about building a retry mechanism for node failures?

chrisfilo:
Seems relevant: http://stackoverflow.com/a/3206187/616300

oesteban:
@satra I've been looking at the `job_finished_timeout` and only found this use: https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L527

We are seeing these synchronization problems at two points:

- when checking traits (the original reason for opening this issue)
- when moving the `_unfinished.json` to finished `.json` here: https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/nodes.py#L370 (see the poldracklab/mriqc issue that Chris attached).

I don't think the `job_finished_timeout` is currently covering these, right?

chrisfilo:
Definitely not in MultiProc.

oesteban:
Yep, that would affect any plugin, virtually. I think Joke was running mriqc with Linear.

chrisfilo:
I'm saying that `job_finished_timeout` only influences SGE like plugins and most likely solves NFS sync issues for them. From what I see in the code changing `job_finished_timeout` currently does not influence, Linear or MultiProc.

satra:
@oesteban - with linear or multiproc there shouldn't be any file sync issue as the files are local (even if mounted via nfs)

[...]

PR #1867, opened 2017-03-07
[WIP] Potential workaround for NFS sync issues

chrisfilo:
See #1750

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -34,6 +34,7 @@ from shutil import rmtree
 import sys
 from tempfile import mkdtemp
 from hashlib import sha1
+from time import time
 
 from ... import config, logging
 from ...utils.misc import (flatten, unflatten, str2bool)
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -367,7 +368,22 @@ def run(self, updatehash=False):
...
                 os.remove(hashfile_unfinished)
                 raise
-            shutil.move(hashfile_unfinished, hashfile)
+
+            t = time()
+            timeout = float(self._config['execution']['job_finished_timeout'])
+            timed_out = True
+            while True:
+                try:
+                    shutil.move(hashfile_unfinished, hashfile)
```
satra: this is taking place on the running node - shouldn't require a timeout.

nipype/pipeline/engine/nodes.py
```diff
@@ -367,7 +368,22 @@ def run(self, updatehash=False):
...
+                    timed_out = False
+                    break
+                except Exception as e:
```
oesteban: Besides this, we should check if `hashfile` is there. If so, the move has been successful and we should not raise.

nipype/pipeline/engine/nodes.py
oesteban: maybe we can apply the EAFP principle here? Even though theoretically this shouldn't require a timeout?

nipype/pipeline/engine/nodes.py
chrisfilo: I don't follow. Are you saying that `move` can throw an exception, but work correctly nonetheless?

oesteban: Yes, otherwise we have a race condition between two processes/threads in a section where only one process is intended to be running. And I can't see how that could happen.

oesteban: And in fact, the target file is in place when you go back after the error

chrisfilo: So just to recap: the exception is thrown because **source file** `hashfile_unfinished` cannot be found, but the **target file** `hashfile` is where it should be as if the copying succeeded? If that the case we are likely dealing with two move operations happening at the same time. We should try to figure out why this was happening.

oesteban:
I think we can close this PR for now, and reopen it if the hypothesis of the simultaneous use of the workdir turns out false.

PR #1869, opened 2017-03-08
Header files (.mat files) as spm realign output for 4D files.

himito:
Hello,

I'm working with 4D files and SPM and I didn't find a way to copy the header files generated by spm realign into the output folder of my workflow.

Also, I found that when these header files are not in the working directory of another node (eg, normalization), then the results of Nipype are different from those when the same workflow is executed in SPM Matlab. So, I added an input to copy "auxiliary files" into the working directory of a node.

I hope these modifications are useful.

Cheers,
Jaime


satra:
@himito - it has to be used in a workflow though.

himito:
Hi @satra,
Thank you for your response. Here, there is an example where the .mat files generated by realign are not copied to the working folder of normalize12. 
https://gist.github.com/himito/438146bbe7d4361cc250ccc604f9cdd3

satra:
let me run this a little later today and get back to you. in general this function should copy the files properly:

https://github.com/nipy/nipype/blob/master/nipype/utils/filemanip.py#L357
https://github.com/nipy/nipype/blob/master/nipype/utils/filemanip.py#L368

however, i can see if you are dealing with 4D nii files, how the mat files would be left out.

nipype/interfaces/spm/preprocess.py
```diff
@@ -177,6 +177,10 @@ class RealignOutputSpec(TraitedSpec):
                                              desc=('Estimated translation and '
                                                    'rotation parameters'))
 
+    header_files = OutputMultiPath(traits.Either(
```
satra: for spm, let's just call these `estimated_transform_files`, as these would be true for coregister as well when using 4d files.

himito:
Thank you  @satra for your time :)

himito:
Hi @satra,

I found also the way to solve the problem of copying related files for 4D only by adding 

```python
related_filetype_sets = [
    ('.hdr', '.img', '.mat', '.nii'),  # I added '.nii'
    ('.BRIK', '.HEAD'),
]
```

this variable is used by the function `get_related_files` in `nipype/utils/filemanip.py`.

satra:
@himito - i prefer this solution as it doesn't require changing the interfaces. but we need to make sure it works fine.

can you either change this pull-request or send a new one?

Issue #1728, opened 2016-12-01
OMP_NUM_THREADS is set default to 1 for all nodes because of AFNI interfaces


bpinsard:
I have problem having multi-threaded (numpy MKL) code running on multiple cores when launched through nipype, but not outside nipype.
The faulty line here changes environment variable, thus affecting all other interface, even when submitted through plugin such as SGE:
https://github.com/nipy/nipype/blob/a746633d72d3e861a4b1a52038dd40a590792e24/nipype/interfaces/afni/base.py#L165

I temporarily circumvented it, but how should it be changed?

Thanks.

chrisfilo:
Good catch. To contain the change to one interface I would suggest replacing this line with:

`self.inputs.environ['OMP_NUM_THREADS'] = str(self.num_threads)`

PR #1870, opened 2017-03-08
fix: contain environmental variable to interface

mgxd:
#1728


mgxd:
@oesteban any idea what cause this error?
```
e=1 && for i in {1..5}; do docker build --rm=false -f docker/Dockerfile_py27 -t nipype/nipype_test:py27 . && e=0 && break || sleep 15; done && [ "$e" -eq "0" ] died unexpectedly
```

oesteban:
No idea. I'm working on caching the docker images but seems a no-go for Circle 1.0

nipype/interfaces/afni/base.py
```diff
@@ -162,7 +162,7 @@ def __init__(self, **inputs):
 
         # Update num threads estimate from OMP_NUM_THREADS env var
         # Default to 1 if not set
-        os.environ['OMP_NUM_THREADS'] = str(self.num_threads)
+        self.inputs.environ['OMP_NUM_THREADS'] = str(self.num_threads)
```
satra: what if this changes between initialization and run. this should happen in the `_run_interface` function.

PR #1711, opened 2016-11-14
enh: additional eddy inputs

mgxd:
Additional args:
- is_shelled
- field and field_mat
- use_gpu (allows eddy to run with openmp/cuda )

PR #1871, opened 2017-03-09
FIX: Select Eddy command at runtime

effigies:
Follow-up to #1711

### DIFF
--- a/nipype/interfaces/fsl/epi.py
+++ b/nipype/interfaces/fsl/epi.py
@@ -498,7 +498,6 @@ class Eddy(FSLCommand):
     >>> res = eddy.run() # doctest: +SKIP
 
     """
-    _cmd = 'eddy_openmp'
     input_spec = EddyInputSpec
     output_spec = EddyOutputSpec
 
@@ -507,8 +506,6 @@ class Eddy(FSLCommand):
     def __init__(self, **inputs):
         super(Eddy, self).__init__(**inputs)
         self.inputs.on_trait_change(self._num_threads_update, 'num_threads')
-        if isdefined(self.inputs.use_cuda):
-            self._use_cuda()
         if not isdefined(self.inputs.num_threads):
             self.inputs.num_threads = self._num_threads
         else:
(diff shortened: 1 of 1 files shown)


satra:
lgtm

mgxd:
whoops, good catch!

effigies:
Actually, if `use_cuda` is in the `input_spec`, it should be editable after instantiation, IMO. Pushed a change that just defines `_cmd` as a property. Is that reasonable, or does something need for `_cmd` to be accessible for the class as well?

satra:
@effigies - for many things we have class level assignments and instance level assignments. we could potentially support both in this case, but it's less of an issue here. i'm fine with your current changes.

alexsavio:
Thanks for this. I will remove the other issue I've just created: #1876. 

How are you installing Eddy? The copy I've got is from the Debian version of the Neurodebian package:
http://neuro.debian.net/pkgs/fsl-5.0-eddy-nonfree.html

satra:
@yarikoptic, @mih  - does the neurodebian version still distribute eddy separately? the fsl installer no longer does.

effigies:
What's the best way to do a version check? Input spec? And what would be the check? The latest release seems to be 5.0.9, which is what's in Neurodebian.

If it's easier, someone feel free to build on my branch and start a new PR.

alexsavio:
For version checking, the `interfaces.fsl.Info.version` static method is your best friend. :)

satra:
@alexsavio - do you not have an `eddy_openmp` command with neurodebian?

alexsavio:
@satra - no I don't. I looked for it, I tried to update my installed package...nothing. I am using the Ubuntu xenial version.

satra:
@alexsavio - it looks like indeed `eddy_openmp` is not distributed via neurodebian or osx, which have `eddy`. however, it is available as part of the fsl centos download.

@effigies - we should think about adjusting this interface.

effigies:
@satra is `eddy` equivalent to `eddy_openmp`? And can we assume or require a properly defined `$FSL_BIN`?

nipype/interfaces/fsl/epi.py
```diff
@@ -523,11 +521,27 @@ def _num_threads_update(self):
...
-            _cmd = 'eddy_openmp'
+    @staticmethod
+    def _eddycmd(use_cuda):
+        logger = logging.getLogger('interface')
+        if 'FSLDIR' not in os.environ:
+            logger.warn("FSLDIR not set: assuming command 'eddy'")
+            return 'eddy'
+
+        FSLDIR = os.environ['FSLDIR']
+        if use_cuda and os.path.exists(os.path.join(FSLDIR, 'eddy_cuda')):
```
satra: shouldn't the `and` clause be on a separate line? otherwise if cuda is not found it will default to one of openmp or plain eddy.

satra: i see how this is problematic for a lot of things, but the only thing i was hoping for was to replace `eddy_openmp` with `eddy` on systems that have eddy but not eddy_openmp at runtime. (so in an overloaded run_interface call.

the previous deleted `_use_cuda` seems reasonable here and can also be used in doctests without requiring fsl to be present.

satra:
looks good

alexsavio:
:+1: LGTM

PR #1872, opened 2017-03-09
enh: updates to support py36

satra:


satra:
@oesteban - some more dockerfile changes here to support py36. 

we really should converge on a streamlined docker protocol as much as possible. we are using dockerhub to create `nipype/nipype:latest`, which will take a while to build. perhaps we can have a hangout to chat about this.

also would like to streamline between @miykael 's dockerfiles and the master repo files.

oesteban:
@satra - sure. We stopped using dockerhub because the builds were too long. As it is now, CircleCI buids the image and pushes it to dockerhub when the commit is a tag.

I'm also working on the cache of CircleCI, but found impossible to get around the BTRFS they use. The problem is that docker uses BTRFS snapshots for the image layers. When they are tar.gz'ed by the cache collector (if you point the cache to the docker cache root) the structure is flattened. When the cache is restored, the BTRFS snapshots structure is not there anymore, and docker is able to use the cache but fails to perform any change. - see it here: https://circleci.com/gh/oesteban/mriqc/tree/enh%2FDockerCache

I'm up to chat about this whenever you want :)

oesteban:
More on the docker caching issue here: https://discuss.circleci.com/t/docker-layer-caching-on-1-0/10814

docker/Dockerfile_py27
```diff
@@ -30,18 +30,15 @@ FROM nipype/nipype:latest
 MAINTAINER The nipype developers https://github.com/nipy/nipype
 
 COPY docker/files/run_* /usr/bin/
-RUN chmod +x /usr/bin/run_*
+RUN chmod +x /usr/bin/run_* && rm /usr/bin/fsl_imglob.py
```
oesteban: The softlink of `${FSLDIR}/bin/imglob` to `/usr/bin/fsl_imglob.py` exists at this point. Not sure if we want to move the imglob replacement to the base image.

satra:
closing for now - py36 support to be included in @oesteban's PR.

PR #1874, opened 2017-03-09
Estimated transform files in spm realign output directory for 4D files

himito:
Hello,

I'm working with 4D files and SPM and I didn't find a way to copy the header files generated by spm realign into the output folder of my workflow.

Also, I found that when these header files are not in the working directory of another node (eg, normalization), then the results of Nipype are different from those when the same workflow is executed in SPM Matlab. 

I hope these modifications are useful.

Cheers,
Jaime

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -30,7 +30,7 @@ fmlogger = logging.getLogger("filemanip")
 
 
 related_filetype_sets = [
-    ('.hdr', '.img', '.mat'),
+    ('.hdr', '.img', '.mat', '.nii'),
     ('.BRIK', '.HEAD'),
 ]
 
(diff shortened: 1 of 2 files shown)


himito:
Hi @satra, 

here the new solution :)

thank you for your time and help.

Jaime

effigies:
Want to update the test?

https://github.com/nipy/nipype/blob/master/nipype/utils/tests/test_filemanip.py#L323-L324

nipype/utils/filemanip.py
```diff
@@ -30,7 +30,7 @@
 
 
 related_filetype_sets = [
-    ('.hdr', '.img', '.mat'),
+    ('.hdr', '.img', '.mat', '.nii'),
```
satra: perhaps add an additional row here:

('.nii', '.mat'),

satra: and remove the `.nii` from the previous set.

himito:
voilà @satra @effigies  :)
Thank you for your help :)
Jaime

Issue #1317, opened 2016-01-12
FLIRT interface does not allow setting apply_xfm with uses_qform option


brennerd11:
Hi all,

the FLIRT interface does not allow to set _apply_xfm =True_ in conjunction with _uses_qform=True_ as _apply_xfm_ requires _in_matrix_file_
Anyways this should be possible as apply_xfm and uses_qform (without providing an additional matrix file) together allow resampling one image to the reference space just by utilizing the affine matrix stored in the header (see also http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FLIRT/FAQ).
I am unsure what the best way would be to fix that - just remove the _required=['in_matrix_file']_ from apply_xfm? Or is there a way to set require one of two inputs?

Cheers,

Daniel

mwaskom:
There's also the [`nipype.fsl.ApplyXfm`](http://www.mit.edu/~satra/nipype-nightly/interfaces/generated/nipype.interfaces.fsl.preprocess.html#applyxfm) interface. I'm not sure if it will work better for you, but might be worth trying.

brennerd11:
Thanks for the hint! I was not aware of that.
But unfortunately the ApplyXfm interface has the same issue - requires “in_matrix_file” when “apply_xfm” is True (default) which does not allow the comibation of only -usesqform and -applyxfm.

danella:
I am having a similar problem. I need to register my fieldmap to my EPI (using flirt or ApplyXfm) to use in FUGUE - and am unable to do it because of this glitch. Any fixes or ways around this? Thanks!

mgxd:
@brennerd11 @danella this should be fixed in the above PR, sorry about the delay!

PR #1875, opened 2017-03-09
enh: allow uses_qform in FLIRT, ApplyXFM interfaces

mgxd:
Addresses #1317 - allows use of `apply_xfm` with either `in_matrix_file` or `uses_qform` args


nipype/interfaces/fsl/preprocess.py
```diff
@@ -561,13 +562,18 @@ def _parse_inputs(self, skip=None):
         if isdefined(self.inputs.save_log) and self.inputs.save_log:
             if not isdefined(self.inputs.verbose) or self.inputs.verbose == 0:
                 self.inputs.verbose = 1
+        if isdefined(self.inputs.apply_xfm) and self.inputs.apply_xfm:
+            if not self.inputs.in_matrix_file and not self.inputs.uses_qform:
```
satra: shouldn't this be `if not isdefined(self.inputs.in_matrix_file) and not (isdefined(self.inputs.uses_qform)  and not self.inputs.uses_qform)`

mgxd: is `isdefined` necessary in this case? I added a test for both cases of using apply_xfm

satra: what if someone sets them to `Undefined`

```
In [1]: from traits.api import Undefined

In [2]: not Undefined
Out[2]: False
```

satra: however the check with not appears to work now

satra:
```
In [20]: foo.inputs.wmnorms = Undefined

In [21]: not foo.inputs.wmnorms
Out[21]: True
```
haven't figured out why yet

mgxd: @satra should this be changed?

PR #1879, opened 2017-03-13
[FIX] Doctest of antsJointFusion failing

oesteban:
The doctest of antsJointFusion was failing since #1862, these changes
hopefully fix this problem without breaking the correct functionality
of the interface.

Also corrected what I suspect is an error: -l was replaced by -t for
the segmentations.


oesteban:
@hjmjohnson could you confirm this is OK?

I think this is what @mgxd requested here https://github.com/nipy/nipype/pull/1862#issuecomment-286177241

hjmjohnson:
@satra Sorry,  I am out of town this week.  I'll test when I return next week.

PR #1882, opened 2017-03-14
[DOC] Improve readability of modelgen examples.

tsalo:


satra:
@tsalo - although this helps a bit, it still doesn't clean things up fully. if you have the time to tackle this once more, it would be great.

PR #1886, opened 2017-03-16
[RTM] Support GIFTI outputs in SampleToSurface

effigies:
This PR creates "implicit" output types for Freesurfer utilities, where no explicit `--out_type` is supported, but a filename with the associated extension will be correctly produced as that type.

At present, I've only tested this with GIFTI on `mri_vol2surf`. I expect `mri_surf2surf` will have the same level of support. I also suspect that any of the `mris_convert` outputs would work fine.

Note that `.gii.gz` actually produces a `.mgz` file with `.gii.gz` extension, and so any additional output types need to be checked for correctness. Similarly, not all files seem to produce a correct GIFTI; for example, `--srchit test.gii` produces a valid GIFTI file with completely different data.

### DIFF
--- a/nipype/interfaces/freesurfer/utils.py
+++ b/nipype/interfaces/freesurfer/utils.py
@@ -29,12 +29,13 @@ filemap = dict(cor='cor', mgh='mgh', mgz='mgz', minc='mnc',
                afni='brik', brik='brik', bshort='bshort',
                spm='img', analyze='img', analyze4d='img',
                bfloat='bfloat', nifti1='img', nii='nii',
-               niigz='nii.gz')
+               niigz='nii.gz', gii='gii')
 
 filetypes = ['cor', 'mgh', 'mgz', 'minc', 'analyze',
              'analyze4d', 'spm', 'afni', 'brik', 'bshort',
              'bfloat', 'sdt', 'outline', 'otl', 'gdf',
              'nifti1', 'nii', 'niigz']
+implicit_filetypes = ['gii']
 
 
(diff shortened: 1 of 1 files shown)


effigies:
I've done some playing, and `mri_vol2surf` with `.gii` file as output produces identical data as `mri_vol2surf` with `.mgz` output, followed by `mris_convert`.

Also, `mri_surf2surf` similarly accepts `.gii` in output filenames, so I've updated the `SurfaceTransform` interface.

effigies:
I guess one issue here is that, currently, setting output type allows you to get away with non-matching names. Here, setting the target type to `gii` would not set an output type, but would change the filename, provided that an output filename is not given. If output filename is set, setting the target type would have no effect.

So if this change is to be accepted, I should notify users. Would something in the interface docstring, trait description, or a runtime check for output filename and target type `== 'gii'` and emitting a warning? All of the above?

chrisfilo:
We can throw an error in format_args if the type and extensions are inconsistent. Nothing good can come out of such combinations.

BTW is `.gii.gz` a thing?

effigies:
[`.gii.gz`](https://www.nitrc.org/frs/?group_id=75). I suspect this is just for compression on NITRC, not an officially supported extension. Since .gii can contain gzip-base64 encoded blocks, it's probably seen as a bit pointless.

For the error, I'd say a warning if they mismatch, an error if they mismatch and the extension belongs to a supported type. There are some extension-less FreeSurfer files, and catching all of those could be tricky.

chrisfilo:
sounds good to me

satra:
if there are no other changes requested, this looks good to me. feel free to merge.

PR #1888, opened 2017-03-16
[FIX] CircleCI tests

oesteban:
This PR fixes most of the problems generated after the early merge of #1861. I opened a revert PR, but merging this it won't be necessary anymore, so close #1887.

After merging this, the situation of circle tests is as follows:

  * The examples are run wrapped inside `coverage run` ~~but apparently no data is collected~~ (discussing this right now in gitter).
  * The tests are all passing, and xml reports placed correctly (so CircleCI will fail for two reasons: if the exit code is dismissed by mistake --situation that happened before and I tried to narrow down--, any failing test will show up in the report and CircleCI will correctly show the failure).
  * Docker infrastructure:
    * There is a base image `/docker/base.Dockerfile`, pulling together all the external software which is built and pushed manually to Docker Hub. This image is pulled during tear up. I tried to get it cached, as we are doing in other projects but it may be too big for the `docker save/load` trick.
     * Then there is the `/Dockerfile`, which accepts two `--build-arg` to define the python version that should be used (`PYTHON_VERSION_MAJOR`, `PYTHON_VERSION_MINOR`).

Future:

  * @satra suggested that we could pull all the software in Circle (outside docker) and mount all these dependencies during run-time. This would solve most of the problems with unreliability of builds and caching we are having now. That would also enable more elaborated ways to fetch the dependencies in the future. This could be addressed in a separate PR.
  * deploy to Pypi from circleci?

Please, look carefully at:

  * This PR modifies the way we traditionally have built the `__version__` string.
  * This PR introduces the `-v` flag when running `py.test`. Please, make sure we are not skipping any important test.


oesteban:
hehehe, of course, the good old segfault! (ง ͠° ͟ل͜ ͡°)ง

oesteban:
Please don't merge immediately, I will push some small commits before merge.

oesteban:
Ok, this should be fine. I'm not 100% sure of being submitting the coverage results to codecov from builds 1-3, but the files are there and everything looks good. Let's fix master and possibly move the codecov integration right after the tests.

PR #1890, opened 2017-03-17
fix: address config parser differences between py2 and 3

satra:


satra:
this resolves `nipype.get_info()` errors when the package is installed via pip in the following manner: `pip install [--upgrade] https://github.com/nipy/nipype/archive/master.zip`

PR #1893, opened 2017-03-20
ENH: Add mris_expand interface

effigies:
This utility is the [Fischl-endorsed](https://mail.nmr.mgh.harvard.edu/pipermail//freesurfer/2016-May/045409.html) way to get a midthickness/graymid surface.


nipype/interfaces/freesurfer/utils.py
```diff
@@ -2879,3 +2879,41 @@ def _list_outputs(self):
...
+class MRIsExpand(FSCommand):
+    """
+    Expands a surface (typically ?h.white) outwards while maintaining
+    smoothness and self-intersection constraints.
+
+    Examples
+    ========
+    >>> from nipype.interfaces.freesurfer import MRIsExpand
+    >>> mris_expand = MRIsExpand(thickness=True, distance=0.5)
+    >>> mris_expand.inputs.in_file = 'lh.white'
```
effigies: Travis issue is `exists=True`. Not really clear why this should be causing an error while the `Apas2Aseg` (immediately above) interface is fine.

satra: there is no `lh.white` dummy file in the testing directory, the above interface uses `lh.pial`

effigies: Ah, should have thought of that. I thought there was some voodoo in the test setup.

chrisfilo: The only VooDoo is in our correlations.

effigies:
Was having issues with actually running until I realized it was dropping the output in the subject's surf directory by default. Updated out_file functionality to be more contained, and added all of the options.

There's one that went in post-6.0 release that I added in its own commit. Not sure how to set `min_ver` for a `dev` version. Maybe just leave it as something to add to `args` until it's in a numbered release?

satra:
@effigies - we need to augment min_ver and max_ver to allow for `><=` combinations

regarding things leaving it in the surf directory, if that's the default behavior of the underlying interface, then that should be retained at the interface level. however the node(interface) should always leave things inside the node's working directory. generally we use copyfile to achieve this, but some interfaces also require recreating directory structures locally.

effigies:
Augmenting `min_ver` and`max_ver` is a bit beyond the scope here; I'll just comment that out for now, and we can revisit once `*_ver` are modified.

The directory structure isn't needed, but in this case, if we copy the surface file, we also need to copy the associated `?h.thickness` and `?h.pial` files. I think I'll need to override `_get_filecopy_info`.

effigies:
@satra Would appreciate a look at the latest commit. The goal is to support file names (or specifiers, really) as `mris_expand` takes them, which required a little wrangling that might be seen as abuse of the input spec.

PR #1894, opened 2017-03-20
Split over-eager globs in FreeSurferSource

effigies:
If you don't specify `hemi`, `FreeSurferSource` matches `*white` and `*pial` for white matter / pial surfaces, which picks up `?h.jacobian_white` for `white` and `?h.{area,curv}.pial` for `pial`.

This restricts the glob prefix for `hemi='both'` to `?h.`, and adds `jacobian_white`, `area_pial` and `curv_pial` entries, which is almost certainly more useful to somebody trying to access any of these entities.


nipype/interfaces/io.py
```diff
@@ -1511,12 +1511,21 @@ class FSSourceOutputSpec(TraitedSpec):
...
                            loc='surf')
+    avg_curv = OutputMultiPath(
+        File(exists=True), desc='Average atlas curvature, sampled to subject',
+        loc='surf')
     inflated = OutputMultiPath(
         File(exists=True), desc='Inflated surface meshes',
         loc='surf')
     pial = OutputMultiPath(
-        File(exists=True), desc='Gray matter/pia mater surface meshes',
+        File(exists=True), desc='Gray matter/pia matter surface meshes',
```
ltirrell: This (along with lines 1524 and 1527) should stay as [pia mater](https://en.wikipedia.org/wiki/Pia_mater), the name for the innermost membrane surrounding the brain

effigies: Ah, thanks. Fixed. And hopefully gave reasonable descriptions for the other two.

PR #1895, opened 2017-03-20
Revert "[ENH] Move codecov to the end of each test #1889"

oesteban:
CircleCI environment variables were not set in the docker container, and it wasn't working. This PR reverts that situation (and also tries to make the coverage reporting more reliable).


oesteban:
@chrisfilo, since you were fighting with codecov some time ago, could you please take a look on this one? I think the codecov-io comment above is actually related to the fact that coverage has not been updated in master for a while.

satra:
@chrisfilo and @oesteban - codecov is again doing something weird - it's down to 46%

Issue #1526, opened 2016-06-27
"Resource Scheduling and Profiling with Nipype" is not listed in the index


chrisfilo:
http://nipype.readthedocs.io/en/latest/users/resource_sched_profiler.html is not linked anywhere

PR #1896, opened 2017-03-21
[ENH] Sphinx extension to plot workflows

oesteban:
Adds a sphinx extension to the nipype installation. After adding `nipype.sphinxext.plot_workflow` to the `extensions` list in the `doc/conf.py`, sphinx will parse blocks like:

```
.. workflow ::
    :graph2use: flat
    :simple_form: no

    from nipype.workflows.dmri.camino.connectivity_mapping import create_connectivity_pipeline
    wf = create_connectivity_pipeline()
```

The graph is embedded at the point where the block is placed in the docstring.

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -16,6 +16,7 @@ import gzip
 import hashlib
 from hashlib import md5
 import os
+from errno import EEXIST
 import re
 import shutil
 import posixpath
@@ -424,6 +425,18 @@ def copyfiles(filelist, dest, copy=False, create_new=False):
             destfile = copyfile(f, destfile, copy, create_new=create_new)
             newfiles.insert(i, destfile)
     return newfiles
+
+def mkdirp(folder):
+    """
+    Equivalent to bash's mkdir -p
+    """
+    if not os.path.exists(folder):
+        try:
+            os.makedirs(folder)
+        except OSError as exc:
+            if not exc.errno == EEXIST:
+                raise
+    return folder
 
 
 def filename_to_list(filename):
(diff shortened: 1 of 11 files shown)


oesteban:
Example:
https://551-17021985-gh.circle-artifacts.com/0/home/ubuntu/work/docs/users/sphinx_ext.html

Also links the resource profile documentation (https://551-17021985-gh.circle-artifacts.com/0/home/ubuntu/work/docs/users/resource_sched_profiler.html) in the user documentation index (#1526) .

nipype/utils/filemanip.py
```diff
@@ -425,6 +426,18 @@ def copyfiles(filelist, dest, copy=False, create_new=False):
             newfiles.insert(i, destfile)
     return newfiles
 
+def mkdirp(folder):
+    """
+    Equivalent to bash's mkdir -p
+    """
+    if not os.path.exists(folder):
```
effigies: You might want to use `os.path.isdir`, which will throw an `EEXIST` if a non-directory exists.

nipype/utils/filemanip.py
```diff
@@ -425,6 +426,18 @@ def copyfiles(filelist, dest, copy=False, create_new=False):
             newfiles.insert(i, destfile)
     return newfiles
 
+def mkdirp(folder):
+    """
+    Equivalent to bash's mkdir -p
+    """
+    try:
+        os.makedirs(folder)
```
satra: on py3 we should just use `os.makedirs(exist_ok=True)`, (also see bug that was fixed here:  https://bugs.python.org/issue21082)

for py2, i'm still worried about race conditions, given that this is something people can start using. how about, just make this a hidden function for the moment and use it in the extension?

PR #1908, opened 2017-03-25
MRG: allow more support for cli

satra:
consider using this: https://pypi.python.org/pypi/traitscli


satra:
@chrisfilo - a review here would be helpful - currently many of the cli break because we don't have full support. this is an attempt to improve that a bit.

nipype/scripts/utils.py
```diff
@@ -68,13 +68,35 @@ def add_args_options(arg_parser, interface):
...
+        if not spec.inner_traits:
+            trait_type = type(spec.trait_type.default_value)
+            if trait_type in (str, int, float):
+                args["type"] = trait_type
+        elif len(spec.inner_traits) == 1:
+            trait_type = type(spec.inner_traits[0].trait_type.default_value)
+            if trait_type in (bool, str, int, float):
+                args["type"] = trait_type
+
         if hasattr(spec, "mandatory") and spec.mandatory:
```
oesteban:
```
if getattr(spec, "mandatory", False):
```

nipype/scripts/utils.py
```diff
@@ -68,13 +68,35 @@ def add_args_options(arg_parser, interface):
             args["default"] = getattr(inputs, name)
             args["action"] = 'store_true'
 
+        # current support is for simple trait types
+        if not spec.inner_traits:
+            trait_type = type(spec.trait_type.default_value)
+            if trait_type in (str, int, float):
+                args["type"] = trait_type
+        elif len(spec.inner_traits) == 1:
```
oesteban: What if `len(spec.inner_traits) > 1`? is that a possible case?. If so, we should raise an error

nipype/scripts/utils.py
```diff
@@ -68,13 +68,35 @@ def add_args_options(arg_parser, interface):
             args["default"] = getattr(inputs, name)
             args["action"] = 'store_true'
 
+        # current support is for simple trait types
+        if not spec.inner_traits:
+            trait_type = type(spec.trait_type.default_value)
+            if trait_type in (str, int, float):
```
oesteban: For compatibility, we should add an `from builtins import str, bytes`. Also, add `bytes` to this check.

Finally, change the type if `trait_type` is `bytes` to be `str`

nipype/scripts/utils.py
```diff
@@ -68,13 +68,35 @@ def add_args_options(arg_parser, interface):
...
+                args["type"] = trait_type
+        elif len(spec.inner_traits) == 1:
+            trait_type = type(spec.inner_traits[0].trait_type.default_value)
+            if trait_type in (bool, str, int, float):
```
oesteban: Same as before with `bytes`

PR #1911, opened 2017-03-29
Adding niftyseg

byvernault:
Dear nipype team,

We are ready to add niftyseg to nipype. I copied all the interfaces we generated for this package as well as the test.

I tested it with python 2 and python 3. It's working on my computer.

Let me know if you see anything not compatible or that I should change.

Kind Regards,

Ben


byvernault:
Hi,

I think we are finally good with this branch to add niftyseg interface to nipype.

Let me know if you have any question.

Kind Regards,
Ben

oesteban:
I'll look at this after #1913 is merged :+1:

nipype/interfaces/niftyseg/label_fusion.py
```diff
@@ -0,0 +1,333 @@
...
+                                       self.inputs.file_to_seg,
+                                       self.inputs.template_file)
+
+    def get_staple_args(self, ranking):
+        classtype = self.inputs.classifier_type
+        if classtype not in ['STAPLE', 'MV']:
+            return None
+
+        if ranking == 'ALL':
+            return '-%s' % ranking
```
oesteban: woudn't `return '-ALL'` be more readable?

nipype/interfaces/niftyseg/em.py
```diff
@@ -0,0 +1,146 @@
...
+                          desc='Number of classes to use without prior',
+                          xor=['prior_4D', 'priors'])
+
+    prior_4D = File(argstr='-prior4D %s',
+                    exists=True,
+                    mandatory=True,
+                    desc='4D file containing the priors',
+                    xor=['no_prior', 'priors'])
+
+    desc = 'List of priors filepaths.'
```
oesteban: why not setting this string directly on the argument? (this holds for many traits defined below)

nipype/interfaces/niftyseg/label_fusion.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def aggregate_outputs(self, runtime=None, needed_outputs=None):
+        outputs = self._outputs()
+        # local caching for backward compatibility
+        outfile = os.path.join(os.getcwd(), 'CalcTopNCC.json')
+        if runtime is None:
+            try:
+                out_files = load_json(outfile)['files']
+            except IOError:
+                return self.run().outputs
```
oesteban: I am doubtful of this `self.run()`, WDYT @satra?

nipype/interfaces/niftyseg/label_fusion.py
```diff
@@ -0,0 +1,333 @@
...
+    _cmd = get_custom_path('seg_CalcTopNCC')
+    _suffix = '_topNCC'
+    input_spec = CalcTopNCCInputSpec
+    output_spec = CalcTopNCCOutputSpec
+
+    def aggregate_outputs(self, runtime=None, needed_outputs=None):
+        outputs = self._outputs()
+        # local caching for backward compatibility
+        outfile = os.path.join(os.getcwd(), 'CalcTopNCC.json')
+        if runtime is None:
```
oesteban: Wouldn't be better to check for non-empty runtime.stdout as well?:

```
if runtime is None or not runtime.stdout:
...
```

nipype/interfaces/niftyseg/maths.py
```diff
@@ -0,0 +1,427 @@
...
+    out_file = File(name_source=['in_file'],
+                    name_template='%s',
+                    position=-2,
+                    argstr='%s',
+                    desc='image to write')
+
+    _dtypes = ['float', 'char', 'int', 'short', 'double', 'input']
+
+    desc = 'datatype to use for output (default uses input type)'
+    output_datatype = traits.Enum(*_dtypes,
```
oesteban: This is totally stylistic, but _dtypes is not used anymore right?, why not having the options in the Enum initialization?

nipype/interfaces/niftyseg/stats.py
```diff
@@ -0,0 +1,213 @@
...
+    For Documentation, see:
+        http://cmictig.cs.ucl.ac.uk/wiki/index.php/NiftySeg_documentation
+
+    Examples
+    --------
+    >>> from nipype.interfaces import niftyseg
+    >>> node = niftyseg.UnaryStats()
+    >>> node.inputs.in_file = 'im1.nii'
+    >>> node.inputs.operation = 'v'
+    >>> node.cmdline  # doctest: +ALLOW_UNICODE
```
oesteban: I think all these interfaces would deserve some more doctests with more operations, so that we are sure that they work and that we don't break them in future maintenance.

nipype/interfaces/niftyseg/maths.py
```diff
@@ -0,0 +1,427 @@
...
+    Examples
+    --------
+    >>> from nipype.interfaces import niftyseg
+    >>> node = niftyseg.BinaryMaths()
+    >>> node.inputs.in_file = 'im1.nii'
+    >>> node.inputs.operation = 'sub'
+    >>> node.inputs.operand_file = 'im2.nii'
+    >>> node.inputs.output_datatype = 'float'
+    >>> node.cmdline  # doctest: +ALLOW_UNICODE
+    'seg_maths im1.nii -sub im2.nii -odt float im1_sub.nii'
```
oesteban: I think all these interfaces would deserve some more doctests with more operations, so that we are sure that they work and that we don't break them in future maintenance.

nipype/interfaces/setup.py
```diff
@@ -24,6 +24,7 @@ def configuration(parent_package='', top_path=None):
     config.add_subpackage('mne')
     config.add_subpackage('mrtrix')
     config.add_subpackage('mrtrix3')
+    config.add_subpackage('niftyseg')
```
oesteban: I would recommend having only one subpackage. I'll comment on this on the summary.

nipype/interfaces/niftyseg/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+
+"""
+The niftyseg module provides classes for interfacing with the `NIFTYSEG
+<https://sourceforge.net/projects/niftyseg/>`_ command line tools.
+
+Top-level namespace for niftyseg.
+"""
+
+from .base import no_niftyseg, get_custom_path
```
byvernault: hmmm I don't think so. I will remove them.

nipype/interfaces/niftyseg/label_fusion.py
byvernault: ahah yes we can do that.

nipype/interfaces/niftyseg/label_fusion.py
byvernault: sure. Thanks.

nipype/interfaces/niftyseg/em.py
byvernault: yes, it was for style and not go over 80 chars when they are very long but I should put this one back there as well as the other one.

nipype/interfaces/niftyseg/maths.py
byvernault: sure.

nipype/interfaces/niftyseg/maths.py
byvernault: Do you think I can follow fslmaths doctests to edit those ones?

nipype/interfaces/setup.py
byvernault: Niftyreg/Niftyseg/Niftyfit are different packages that you install separately. Do you think having only one package won't be odd for the user? What do you think?

I could remove probably the base and use the same commands from niftyreg.base in this package. Or do you mean moving all the code into the same folder?

oesteban: Ok, I don't have an special interest for having only one package. Let's observe that preference (reg/seg/fit packages) :+1: .

So if we keep the 3 packages structure, then I'm suggesting here to import from niftyreg.base and remove the duplicated code from niftyseg and niftyfit.

WDYT?

byvernault: Yes sounds good. I jut did the changes and I renamed the function in niftyreg base for something more general like no_nifty_package.

What do you think about the changes? 

Also do you prefer to import NiftyRegCommand like this:

import ..niftyreg.base.NiftyRegCommand as NiftySegCommand 

Or do as I did?

I will do then the edits on niftyfit when you agree with those changes.

Kind Regards,
Ben

byvernault: Oh there are some issues for the test since most executables in niftyseg don't have a --version options that is working.

I will work on fixing this later today.

byvernault: I add back the base.py just to create NiftySegCommand because of the version that isn't available for all the executables.

I also edited slightly the base.py for NiftyRegCommand.
The test should pass now.

Let me know what you think.
 
Cheers,
Ben

nipype/interfaces/niftyseg/maths.py
oesteban: What I had in mind is something closer to the documentation of `nipype.interfaces.ants.registration.Registration`: https://github.com/nipy/nipype/blob/master/nipype/interfaces/ants/registration.py#L432-L640

Here, you'll see how the inputs to the interface are varied and then the new cmdline tested.

byvernault: I can do that but for those interfaces it's pretty easy, only the operation changed. It's kind of the same than fslmaths.

I added different exemples for some operation but by doing so I realised that I am not checking the inputs for some argument (for example pow, thr, uthr, ... use only a float or min uses only a file). I am adding that to the interface.

Let me know if you approve this and you really want to see one test per operation.I can do that but for those interfaces it's pretty easy, only the operation changed. It's kind of the same than fslmaths.

I added different exemples for some operation but by doing so I realised that I am not checking the inputs for some argument (for example pow, thr, uthr, ... use only a float or min uses only a file). I am adding that to the interface.

Let me know if you approve this and you really want to see one test per operation.

PR #1912, opened 2017-03-29
Minor QoL change for write_graphs()

adamkimbler:
changed behavior of write_graph() slightly. If workflow name is defined but not base_dir the dot graphs be stored in a directory named after the workflow name, rather than just dumping to the current working directory. If neither name nor base_dir is defined then it should default to current behavior.

### DIFF
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -421,6 +421,8 @@ connected.
                 base_dir = self.base_dir
                 if self.name:
                     base_dir = op.join(base_dir, self.name)
+            elif self.name:
+                base_dir =  self.name
             else:
                 base_dir = os.getcwd()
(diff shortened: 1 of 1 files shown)


oesteban:
@adamkimbler all nipype workflows must have a name, so the last case is not possible.

Current default is writing the graph in a temporal folder (which is ok, IMHO). If you set `base_dir` to current directory, then you'll get the graph under `$PWD/<workflow_name>/`, if I'm not wrong.

adamkimbler:
@oesteban As far as I can tell, the current behavior ignores the workflow name when defining the base_dir for the write_graph() function if the workflow base_dir is initially undefined. Resulting in the graphs being output to the current working directory of the pipeline script itself, instead of $PWD/<workflow_directory> as you suggested.

Issue #1783, opened 2017-01-18
Datagrabber bug with the flag raise_on_empty


anbai106:
Hello:
Recently, I tried datagrabber to grab some files from my dataset, but I found a problem which datagrabber does not work so well, here is the case:

here is the debug code:

```
import nipype.interfaces.io as nio
input_dir = '/Volumes/dataARAMIS/users/junhao.wen/dataset/PREVDEMALS/GENFI'
datagrabbernode = nio.DataGrabber(
                    infields=['subject_list', 'session_list', 'subject_repeat', 'session_repeat'],
                    outfields=['anat_t1'])
datagrabbernode.inputs.base_directory = input_dir
datagrabbernode.inputs.template = '*'
datagrabbernode.inputs.field_template = dict(anat_t1='%s/%s/anat/%s_%s_T1w.nii.gz') # This is just to adapt to the output anat_t1
# just for test with my home nii files
# datagrabbernode.inputs.field_template = dict(anat_t1='sub-%s/ses-%s/anat/sub-%s_ses-%s_T1w.nii')
datagrabbernode.inputs.template_args = dict(anat_t1=[['subject_list', 'session_list', 'subject_repeat',
                                                      'session_repeat']])
datagrabbernode.inputs.sort_filelist = False
datagrabbernode.inputs.subject_list = ['sub-PREVDEMALS0010001BE', 'sub-PREVDEMALS0010002VH']
datagrabbernode.inputs.session_list = ['ses-M0', 'ses-M0']
datagrabbernode.inputs.subject_repeat = ['sub-PREVDEMALS0010001BE', 'sub-PREVDEMALS0010002VH']
datagrabbernode.inputs.session_repeat = ['ses-M0', 'ses-M0']
datagrabbernode.inputs.raise_on_empty = False
datagrabbernode.run()
```

For example, I have a dataset which has two subjects, and one of them has the the nifti image .nii, and another one does not have the image, so what i want is to not let my workflow die in the datagrabber even one of the image is missing, so I put the flag **_'raise_on_empty'_**, so what I expected to give me back the output like:

> {'anat_t1': [None, '/Volumes/dataARAMIS/users/junhao.wen/dataset/PREVDEMALS/GENFI/sub-PREVDEMALS0010002VH/ses-M0/anat/sub-PREVDEMALS0010002VH_ses-M0_T1w.nii.gz']}

But in your source code, if one of the image is missing, you will give back **_None_**, which i think does not make sense.

[...]

PR #1915, opened 2017-03-29
bug for datagrabber

anbai106:
There are two other classes in the io.py file, they maybe have the same problem, if you think my change works, you can also change the other two classes.

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -1242,8 +1242,11 @@ class DataGrabber(IOBase):
                         if self.inputs.sort_filelist:
                             outfiles = human_order_sorted(outfiles)
                         outputs[key].append(list_to_filename(outfiles))
-            if any([val is None for val in outputs[key]]):
-                outputs[key] = []
+            if self.inputs.raise_on_empty: # if the raise_on_empty is True by default, make sure no None in outputs
+                if any([val is None for val in outputs[key]]):
+                    outputs[key] = []
+            else: # if the raise_on_empty is False remove the Nones in outputs
+                outputs[key] = [[] if x is None else x for x in outputs[key]]
             if len(outputs[key]) == 0:
                 outputs[key] = None
             elif len(outputs[key]) == 1:
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -1242,11 +1242,11 @@ class DataGrabber(IOBase):
                         if self.inputs.sort_filelist:
                             outfiles = human_order_sorted(outfiles)
                         outputs[key].append(list_to_filename(outfiles))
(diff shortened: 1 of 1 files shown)


satra:
@mgxd - could you please review this and make sure it functions as intended in the description for raise_on_empty?

mgxd:
@anbai106 maybe we can add a new input field `drop_blank_outputs` that performs this check on top of `raise_on_empty`.

mgxd:
@anbai106 the behavior you want would require a new input for datagrabber - something along the lines of 

```python
...
        if self.inputs.drop_blank_outputs:
            outputs[key] = [x for x in outputs[key] if x is not None]
        else:
            if any([val is None for val in outputs[key]]):
                outputs[key] = []
...
```

this behavior will potentially return lists of different sizes, which is dangerous when working with mapnodes / iterables across runs - thus, better to make a new input rather than build off of `raise_on_empty` 

(linking to related issue #1783)

This is very close to merge!

effigies:
Hi @anbai106, are you going to have time to finish this up, soon?

anbai106:
@effigies sorry for the delay, not so familiar with git pull request, i hope now everything is OK.

mgxd:
@anbai106 we're just missing two things:

could add a new input for `DataGrabber` here?
https://github.com/nipy/nipype/blob/166e06cf5e8ec76cbbb16642f6526f093c28bfe5/nipype/interfaces/io.py#L1039
and make the changes from the [earlier comment](https://github.com/nipy/nipype/pull/1915#issuecomment-356078102)?

If you don't find time, we could make the changes for you, if that would be easier. Thanks!

anbai106:
@mgxd , I am doing it now, if I make it wrong this time, you will be in charge of it, haha

anbai106:
@mgxd I corrected like you proposed and merge my branch into the master branch, can you check if everything is ok?

mgxd:
@anbai106 you'll have to `git push` your branch for us to see the changes

effigies:
Hi @anbai106, I think you merged into your own master branch, which isn't going to show up here. Try the following:

```Shell
git remote add upstream git@github.com:nipy/nipype.git
git fetch upstream
```

This will make the official master branch available to you as `upstream/master`. Then we need to merge your current changes into the `anbai` branch, which is connected to this pull request.

```Shell
git checkout anbai
git merge master
```

Finally, I would recommend merging the current master, just to update the tests, then push:

```Shell
git merge upstream/master
git push
```

anbai106:
@effigies finally, I think it is done :)
Thanks for your help

effigies:
@anbai106 For some reason that latest version had some commit issues in it. I rebased your PR against the current master to remove the noise. Can you please check the [diff](https://github.com/nipy/nipype/pull/1915/files) and verify that it looks the way you intended?

effigies:
@mgxd I think this is ready, unless you need tests.

PR #1918, opened 2017-03-30
ENH: Auto-derive input_names in Function

### LINKED ISSUE #1386: Simplifying the Function interface
The arguments of a function `f` can be listed: `f.func_code.co_varnames[:f.func_code.co_argcount]`.

Currently, I'm decorating functions like so:

``` Python
def nipype_function(*output_names, **kwargs):
    def decorator(func):
        kwargs.setdefault('output_names', output_names or ['output'])
        kwargs.update({
            'function': func,
            'input_names': func.func_code.co_varnames[:func.func_code.co_argcount]})
        fn = util.Function(**kwargs)
        fn.inputs.function_str = u'\n'.join(
            fn.inputs.function_str.split(u'\n')[1:])
        return fn
    return decorator

@nipype_decorator('out_list')
def mergelists(lists):
    return [elem for lst in lists for elem in lst]
```

So two questions:
1. Would something like this decorator be useful to include in nipype?
2. Would it be useful to permit `nipype.interfaces.utility.Function` to derive `input_names` from the function or function string? I would probably also set a default output name. Thus you could generally write:

``` Python
def func(arg1, arg2):
    return arg1 + arg2

fi = Function(function=func)
fi.inputs.arg1 = 2
fi.inputs.arg2 = 3
res = fi.run()
assert res.outputs.out == 5
```

I know I'm kind of throwing a flurry of things at y'all without polishing each first, so thanks for your continued attention.

effigies:
Found this branch sitting around. Will be glad to update (and improve) if there's interest.

Closes #1386.

### DIFF
--- a/nipype/interfaces/utility/tests/test_wrappers.py
+++ b/nipype/interfaces/utility/tests/test_wrappers.py
@@ -24,7 +24,7 @@ def test_function(tmpdir):
     def increment_array(in_array):
         return in_array + 1
 
-    f2 = pe.MapNode(utility.Function(input_names=['in_array'], output_names=['out_array'], function=increment_array), name='increment_array', iterfield=['in_array'])
+    f2 = pe.MapNode(utility.Function(function=increment_array), name='increment_array', iterfield=['in_array'])
 
     wf.connect(f1, 'random_array', f2, 'in_array')
     wf.run()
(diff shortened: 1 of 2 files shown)


satra:
simplifying the Function node would be useful and would minimize typos. i'm also thinking of using cloudpickle to pickle the function and see if we can move nested functions more easily (but separate PR for that).

effigies:
Circle tests pending, but ready for review.

PR #1928, opened 2017-04-02
Make sure the new FreeSurfer dev version scheme plays nice

chrisfilo:


effigies:
Probably a good idea to get this behavior in for 0.13.

For context, the current dev release build stamp is `freesurfer-Linux-centos6_x86_64-dev-20170412-df76142`, so this changes `FSVersion` from `df76142-dev` to `6.0.0.df76142-dev`. I don't have a dev build from between 5.3 and 6.0, so I'm not sure exactly what the old behavior was.

This seems fine for the very short term, but once 6.0.1 comes out, then dev will be evaluated to `<` stable. I'm not sure what the best way to handle this would be. We do get dates, so we could start maintaining a table, where `> 20170118 => 6.0.0.$HASH-dev`, and add an entry with the 6.0.1 release date, and so on.

Or we could say `dev => 7.0.0-dev`, and not worry about people who are sitting on dev builds that are older than latest stable.

satra:
can we move to semantic versioning for this - so `6.0.0-dev.gdf76142` building roughly on semantic versioning + `git describe`

PR #1929, opened 2017-04-02
Add 'bold' and 'dti' contrast type to bbregister

chrisfilo:


effigies:
```
$ bbregister --help
[...]
	Contrast Arguments (one required)
	--t1
		Assume t1 contrast, ie, White Matter brighter than Grey Matter
	--t2
		Assume t2 contrast, ie, Gray Matter brighter than White Matter
	--bold
		Same as --t2
	--dti
		Same as --t2
```

If we're going to add bold, I figure we should add dti; but is there any particular reason to add aliases?

chrisfilo:
You are right it is an alias. I've added 'dti'. It does not hurt to have those and it makes easier for new nipype users accustomed to certain flags.

PR #1930, opened 2017-04-02
[DOC] Add duecredit references for AFNI and FSL.

tsalo:


nipype/interfaces/base.py
```diff
@@ -1387,7 +1387,7 @@ def _get_ram_mb(pid, pyfunc=False):
 def get_max_resources_used(pid, mem_mb, num_threads, pyfunc=False):
     """Function to get the RAM and threads usage of a process
 
-    Paramters
+    Parameters
     ---------
```
oesteban: Just a tiny change: can you add a hyphen so sphinx does not raise a warning?

nipype/interfaces/fsl/base.py
```diff
@@ -157,6 +158,48 @@ class FSLCommand(CommandLine):
     input_spec = FSLCommandInputSpec
     _output_type = None
 
+    references_ = [{'entry': BibTeX('@article{WoolrichJbabdiPatenaudeChappellMakniBehrens'
```
oesteban: I'd say we should place only the Neuroimage 2012 reference at the top level, and then have the specific references with the corresponding interfaces.

satra:
@tsalo  - during the workshop we realized that the afni interfaces need some more tlc. for example `1deval` doesn't have a flag `-prefix` for output. it uses the `1deval ... > out_file` option. also `outliercount` is not using `namesource` properly and needs an `out_file` input to work. (while this is not the right PR to fix this, if you know those interfaces, could you please submit a separate PR to fix them?)

tsalo:
@satra I can try to fix those problems.

tsalo:
@satra And @neuroumbrage, @62442katieb, and @adamkimbler say they'll help!

satra:
> And @neuroumbrage, @62442katieb, and @adamkimbler say they'll help!

awesome!

PR #1934, opened 2017-04-04
[WIP] Use un-copied subject directory in MRIsExpand

effigies:
`mris_expand` expects the entire subject directory to be available. This change abuses `_get_filecopy_info` in a different way from before, changing the output filename only in cases where mris_expand is to be run in a node.

Fixes issue in which absent `../mri/transforms/talairach.*` prevents xform from being written to the new surface, which in turn prevents surfaces converted to GIFTI from being loaded correctly by `wb_view`.


satra:
@effigies - is it worthwhile to check out which pieces are really touched by using reprozip on the call?

effigies:
Actually, I'm realizing the problem is in `mris_convert`.

Issue #1140, opened 2015-07-14
Documentation and testing for contrast functions in FSL 5.0.7+


poldrack:
Since ContrastMgr was deprecated in FSL 5.0.7+ (and won't work because the necessary corrections file is no longer generated by FILMgls), the ContrastMgr function should test for FSL version and handle this situation more gracefully.  This issue should also be documented more thoroughly in the documentation.

satra:
@poldrack - we should improve the documentation.

regarding the check, yes contrast manager should have a version check that raises a DeprecationError for FSL 5.0.7+

for an example of how this is used see the openfmri script, which uses this workflow:

https://github.com/nipy/nipype/blob/master/nipype/workflows/fmri/fsl/estimate.py#L65
https://github.com/nipy/nipype/blob/master/nipype/workflows/fmri/fsl/estimate.py#L135

PR #1940, opened 2017-04-05
doc: warn contrastmgr deprecation in fsl 5.0.7+

mgxd:
#1140


nipype/interfaces/fsl/model.py
```diff
@@ -1075,13 +1084,17 @@ class ContrastMgr(FSLCommand):
     """Use FSL contrast_mgr command to evaluate contrasts
 
     In interface mode this file assumes that all the required inputs are in the
-    same location.
+    same location. This has deprecated for FSL versions 5.0.7+ as the necessary
+    corrections file is no longer generated by FILMGLS.
     """
 
+    if LooseVersion(FSVersion) >= LooseVersion("5.0.7"):
```
satra: what if FSL is not installed?

nipype/interfaces/fsl/model.py
```diff
@@ -30,6 +31,14 @@
                     BaseInterfaceInputSpec)
 from .base import FSLCommand, FSLCommandInputSpec, Info
 
+FSVersion = "0"
```
satra: replace this section with the version property in FSL base class.

nipype/interfaces/fsl/model.py
mgxd: then FSVersion would be "0"

nipype/interfaces/fsl/model.py
```diff
@@ -1075,13 +1072,16 @@ class ContrastMgr(FSLCommand):
     """Use FSL contrast_mgr command to evaluate contrasts
 
     In interface mode this file assumes that all the required inputs are in the
-    same location.
+    same location. This has deprecated for FSL versions 5.0.7+ as the necessary
+    corrections file is no longer generated by FILMGLS.
     """
-
+    if Info.version() and LooseVersion(Info.version()) >= LooseVersion("5.0.7"):
+        warn("ContrastMgr is deprecated in FSL 5.0.7+", DeprecationWarning)
```
satra: you can use DeprecationWarning without import warn

PR #1941, opened 2017-04-05
ENH: Disable symlinks on CIFS filesystems

effigies:
There appear to be cases where hacks to make symlinks work on filesystems that don't directly support them can cause `FileNotFoundError`s. In particular, the CIFS network filesystem driver for Linux supports the `mfsymlinks` option, which enables symlinks at the VFS level, having the effect that symlinking can be subject to network timeouts or temporary inconsistencies.

CIFS+`mfsymlinks` is the Docker-supported way to mount filesystems into a container from a Windows host, which means this may be increasingly a problem for Nipype users, as containers are becoming a more popular environment for running workflows. An [error](https://github.com/poldracklab/fmriprep/pull/398#issuecomment-291723199) reported on poldracklab/fmriprep#398 ran into this issue when running an interface that requires 4 symlinks, which split into two `mapflow` directories (per dataset being analyzed), using the `MultiProc` plugin.

This PR permits users to disable symlinking in the config file with the `disable_symbolic_links` option.

Closes #1938.

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -238,7 +238,37 @@ def hash_timestamp(afile):
     return md5hex
 
 
-def _on_cifs(fname):
+def _generate_cifs_table():
+    """Construct a reverse-length ordered list of mount points that
+    fall under a CIFS mount.
+
+    This precomputation allows efficient checking for whether a given path
+    would be on a CIFS filesystem.
+
+    On systems without a ``mount`` command, or with no CIFS mounts, returns an
+    empty list.
+    """
+    exit_code, output = subprocess.getstatusoutput("mount")
+    # Not POSIX
+    if exit_code != 0:
+        return []
+
+    # (path, fstype) tuples, sorted by path length (longest first)
+    mount_info = sorted((line.split()[2:5:2] for line in output.splitlines()),
+                        key=lambda x: len(x[0]),
+                        reverse=True)
+    cifs_paths = [path for path, fstype in mount_info if fstype == 'cifs']
+    if cifs_paths == []:
+        return []
+
+    return [mount for mount in mount_info
+            if any(mount[0].startswith(path) for path in cifs_paths)]
+
+
+_cifs_table = _generate_cifs_table()
+
+
+def on_cifs(fname):
     """ Checks whether a PATH is on a CIFS filesystem mounted in a POSIX
     host (i.e., has the "mount" command).
 
@@ -246,17 +276,8 @@ def _on_cifs(fname):
     CIFS has partial support for symlinks that can break, causing misleading
     errors, so this test is used to disable them on such systems.
     """
-    exit_code, output = subprocess.getstatusoutput("mount")
-    # Not POSIX
-    if exit_code != 0:
-        return False
-
-    # (path, fstype) tuples, sorted by path length (longest first)
-    paths_types = sorted((line.split()[2:5:2] for line in output.splitlines()),
-                         key=lambda x: len(x[0]),
-                         reverse=True)
     # Only the first match counts
-    for fspath, fstype in paths_types:
+    for fspath, fstype in _cifs_table:
         if fname.startswith(fspath):
             return fstype == 'cifs'
     return False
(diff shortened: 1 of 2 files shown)


chrisfilo:
Determining the filesystem might be easier than we thought http://stackoverflow.com/a/25287521/616300

effigies:
That requires `psutil`, but if it seems worth it to add the dependency, we can check for CIFS. Could you install psutil on your docker container and get a working `is_cifs(path)` function?

chrisfilo:
Does not require psutil.
```
def is_cifs(path):
     exit_code, mount_str = subprocess.getstatusoutput("mount")
     if not exit_code:
         mount_points = [(line.split(" ")[2], line.split(" ")[4]) for line in mount_str.split("\n")]
         sorted_mount_points = sorted(mount_points, key=lambda x: len(x[0]), reverse=True)
         for k, v in sorted_mount_points:
             print(k,v)
             if path.startswith(k):
                 if v == "cifs":
                     return True
                 else:
                     return False
     else:
         return False
```

chrisfilo:
I also run the workflow with working directory inside the container (not mounted) and I did not get an error.

This means we should avoid symlinks only if the destination is on cifs.

effigies:
@chrisfilo I refactored your test a little. Want to grab 0d5359d from me and verify that this works?

```Python
from nipype.utils.filemanip import _on_cifs
assert (_on_cifs('/'), _on_cifs('/scratch')) == (False, True)
```

I think creating a CIFS mount in the docker image for testing might be excessive.

nipype/utils/filemanip.py
```diff
@@ -237,6 +238,30 @@ def hash_timestamp(afile):
     return md5hex
 
 
+def _on_cifs(fname):
```
chrisfilo: I would make it public - could be useful in other places

nipype/utils/filemanip.py
```diff
@@ -288,6 +313,10 @@ def copyfile(originalfile, newfile, copy=False, create_new=False,
     if hashmethod is None:
         hashmethod = config.get('execution', 'hash_method').lower()
 
+    # Don't try creating symlinks on CIFS
+    if _on_cifs(newfile):
```
chrisfilo: For performance reasons, I would only run this check if we are about to make a symlink.

effigies:
The performance issue is a good point. If this function's going to get called a lot, it might make sense to do something a little more persistent. We could do something like:

- At program start (or first call), call `mount`, and get `(mountpoint, fstype)` tuples.
- Prune all but superstrings of CIFS mounts
- Check target files against pruned list

effigies:
Latest commit may be breaking some style guidelines. Let me know if I should reorganize.

@chrisfilo could you verify this still works? Updated test:

```Python
from nipype.utils.filemanip import on_cifs
assert (on_cifs('/'), on_cifs('/scratch')) == (False, True)
```

chrisfilo:
I'm seeing huge performance hit when using MultiProc. It seems that mount is called each time a Node is being run (or checked in it needs to be rerun) - even if no files are being copied.

effigies:
On the most recent commit? The latest changes should make the mount call occur only once per run. Unless I misunderstand how things are being run...

chrisfilo:
Yeah I pulled to get the latest commit. I can double check, but I see the
slowdown even for checking precomputed inputs so I think I go the right
commit.

chrisfilo:
I'm doing more debugging. It seems that rechecking EPIToT1Transform (nothing gets rewritten) takes longer than usual (it's really slow) however the mri_expand is happening in the background so I don't know if the total running time will differ.

But you are right - the table is generated only once. I'll do more debugging.

effigies:
Re: `EPIToT1`, have you recently switched to content hashing, or using an EPI with more volumes? Because I've found that's always been slow when I used content hashing...

effigies:
Unable to replicate the slowdown on mine.

chrisfilo:
I think the slowdown is unrelated to the change in this PR. Everything looks good to me, but since this is a change at the core of nipype it would be nice to have @satra hava a look at it as well.

effigies:
Do you have `remove_unnecessary_outputs` disabled? If not, that might be your slowdown.

afni-dglen:
John Roger-Lee and I spent some time looking at the cifs parsing because it ended up blocking the installation on my local Mac. I'll create a pull request with my suggested changes in filemanip.py.

PR #1942, opened 2017-04-06
[ENH] Migrating resource handler to Node level

oesteban:
This PR:

- Sets the default `estimated_memory_gb` to 0.25 (most of the nodes of a workflow will never reach 1GB)
- Makes two new arguments available to creating Nodes (n_procs and mem_gb), that will set those two values in the inner interface (no need to set node.interface.num_threads anymore).
- Fixes a small error formatting one Exception (we had `__interface` instead of `_interface`).

TODO list:

- Remove estimated_memory_gb and num_threads from the BaseInterface
- Find a way to synchronize the Node level num_threads and the inner interface inputs (maybe a new metadata threads=True to identify these inputs?)
- Resource handling of MapNodes (check if these changes would affect those).

### DIFF
--- a/nipype/interfaces/base.py
+++ b/nipype/interfaces/base.py
@@ -774,7 +774,7 @@ class BaseInterface(Interface):
                             self.__class__.__name__)
 
         self.inputs = self.input_spec(**inputs)
-        self.estimated_memory_gb = 0
+        self.estimated_memory_gb = 0.25
         self.num_threads = 1
 
         if from_file is not None:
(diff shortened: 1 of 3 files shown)


oesteban:
@ccraddock may you have a look into these changes, and maybe give some feedback about the TODO list? Thanks!

ccraddock:
I will have a look over the next few days. Really busy at the moment. But at first, I would recommend to make the default memory something > 0, such as .5 or .25

oesteban:
Up! cc/ @ccraddock

Issue #1021, opened 2015-01-16
Workflow.write_graph chocks on workflow name starting with a digit


gerddie:
With the example program 

```
from nipype.pipeline.engine import Workflow
w = Workflow(name="1234a") 
w.write_graph()
```

_dot_ will fail, because the created dot file contains identifiers that start with a digit that _dot_ wants to interpret as number.

PR #1944, opened 2017-04-06
FIX: catch write_graph errors for 'hierarchical/colored

mgxd:
#1021

### DIFF
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -425,6 +425,9 @@ connected.
                 base_dir = os.getcwd()
         base_dir = make_output_dir(base_dir)
         if graph2use in ['hierarchical', 'colored']:
+            if self.name[:1].isdigit(): # these graphs break if int
+                raise ValueError('{} graph failed, workflow name cannot begin '
+                                 'with a number'.format(graph2use))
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/workflows.py
```diff
@@ -425,6 +425,9 @@ def write_graph(self, dotfilename='graph.dot', graph2use='hierarchical',
                 base_dir = os.getcwd()
         base_dir = make_output_dir(base_dir)
         if graph2use in ['hierarchical', 'colored']:
+            if self.name[:1].isdigit(): # these graphs break if int
+                raise ValueError('{} graph failed, workflow name cannot begin '
+                                 'with a number'.format(graph2use))
```
satra: if workflow name cannot begin with a number, shouldn't this be a check when a workflow/node is created/cloned? why is a workflow name check in graph generation?

mgxd: @satra the error only occurs when the workflow graph is generated if graph2use is specified as either hierarchical/colored - the other options generate graphs just fine. I'm fine with moving the check to the `Workflow` initialization if we want to go in that direction - was trying to affect the smallest range of legacy workflows

PR #1948, opened 2017-04-09
[ENH] Add MRIsCombine to FreeSurfer utils.

tsalo:
MRIsCombine uses mris_convert to combine two surface files.


nipype/interfaces/freesurfer/utils.py
```diff
@@ -954,6 +954,63 @@ def _gen_outfilename(self):
...
+    _cmd = 'mris_convert'
+    input_spec = MRIsCombineInputSpec
+    output_spec = MRIsCombineOutputSpec
+
+    def _list_outputs(self):
+        outputs = self.output_spec().get()
+        if any(self.inputs.out_file.startswith(pre) for pre in ['lh.', 'rh.']):
+            outputs['out_file'] = self.inputs.out_file
+        else:
+            outputs['out_file'] = 'lh.' + self.inputs.out_file
```
tsalo: From what I've seen:
- `mris_convert --combinesurfs lh.pial rh.pial out.stl` creates lh.out.stl
- `mris_convert --combinesurfs rh.pial lh.pial out.stl` creates lh.out.stl
- `mris_convert --combinesurfs rh.pial rh.smoothwm out.stl` creates lh.out.stl
- `mris_convert --combinesurfs lh.pial rh.pial rh.out.stl` creates rh.out.stl

So it doesn't seem to matter what order you pass in the files or even if both files are from the right hemisphere. As long as the output file doesn't start with `lh.` or `rh.`, it will prepend `lh.` to the output filename.

effigies: Haha. That's awful. Can you add a short comment explaining that this is a FreeSurfer behavior you're accommodating?

nipype/interfaces/freesurfer/utils.py
```diff
@@ -954,6 +954,63 @@ def _gen_outfilename(self):
...
+class MRIsCombineInputSpec(FSTraitedSpec):
+    """
+    Uses Freesurfer's mris_convert to combine two surface files into one.
+    """
+    in_file1 = File(exists=True, mandatory=True, position=1,
+                    argstr='--combinesurfs %s',
+                    desc='File to be combined with in_file2')
+    in_file2 = File(exists=True, mandatory=True, position=2,
+                    argstr='%s',
+                    desc='File to be combined with in_file1')
```
tsalo: I didn't realize `traits.List` allowed for `maxlen` and `minlen`. I definitely prefer it your way. Just tested it and it works, so I'll commit it in a second.

nipype/interfaces/freesurfer/utils.py
tsalo: Added an explanation to the docstring for `_list_outputs` in a7bafda. Should be clear now.

effigies:
Actually, I'm going to be obnoxious. That `lh.*` behavior is so wildly unexpected, it should be in the docstring for the entire interface so it shows up in the generated documentation.

Also, can you see if `out_file="./bh.pial"` gives different behavior? From my experience with FreeSurfer, I suspect they'll interpret a name with path information in it literally.

tsalo:
Putting in the path information worked. It wrote out the file given, with no changes. [MRIMarchingCubes](https://github.com/nipy/nipype/blob/8c4e6028243f3f44280e69b50ce8d07c91f908e4/nipype/interfaces/freesurfer/utils.py#L1077) deals with this by using `./%s` as the out_file argstr, but that seems to cause problems if you put in an out_file with an absolute path (like `/home/tsalo/out.stl`, because then it tries to combine them as `.//home/tsalo/out.stl` and then it fails.

Rather than putting the path in the argstr, is there a way to reformat it in something like `_list_outputs`, but for inputs? Is that a thing?

Whatever the fix is, it'll probably need to be implemented in other FreeSurfer functions- especially MRIMarchingCubes.

effigies:
Yeah, I wouldn't do it in the `argstr`, and I have no doubt that lots of these interfaces need clearing up. Doing something in `_list_outputs` is a little late, to my mind, especially if `mris_convert` likes, as other functions do, to put outputs in the same directory as the inputs, not in the current working directory. (If it outputs to the current working directory, you can pick out a better outfile name, rename and return the better name.)

`MRIsExpand` does a bit of hackery in [`_get_filecopy_info`](https://github.com/effigies/nipype/blob/39a1e3d3e7708f442f7f15e7617da9babdec3a72/nipype/interfaces/freesurfer/utils.py#L2995), which is only run when in a `Node` context, giving the advantage the the command is run as advertised outside a `Node`, and runs properly isolated inside. You could do something similar:

```Python
def _get_filecopy_info(self):
    super(MRIsCombine, self)._get_filecopy_info()
    if isdefined(self.inputs.out_file):
        self.inputs.out_file = os.path.abspath(self.inputs.out_file)
```

It's not elegant, but it is the [main pre-execution hook](https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/nodes.py#L594) that doesn't also get called when running an interface directly.

We should probably decide on a single hack, and apply it to all the interfaces, or better yet, update the `FreeSurferCommand` interface to automatically (possibly triggered by a trait attribute) accommodate these naming conventions when run in a `Node`.

tsalo:
I don't know how pervasive the renaming behavior is, but I think it probably only applies to functions that work on surface files. What about a new class inheriting from `FSCommand` for surface-related interfaces. Something like `FSSurfaceCommand`. I can add your version of `_get_filecopy_info` to that and have the interfaces we currently know are affected (`MRIsCombine`, `MRIsConvert`, and `MRIMarchingCubes`) inherit from that instead of `FSCommand`.

tsalo:
I took a look at the different ways different interfaces are handling the same issue.
- Some (e.g., `VolumeMask`, `Aparc2Aseg`, `Contrast`) output everything to the subject directory using a function called `copy_to_subjdir`. I'm not sure it's best to force a specific output directory.
- `MRIsExpand` checks the beginning of `out_file`. If the file doesn't have either prefix (lh. or rh.), it takes the prefix from `in_file` and uses that. It's sort of similar to what I did.
- Several others (e.g., `Apas2Aseg`, `RelabelHypointensities`, `MRIsCalc`, `Jacobian`) just set the outputs by applying `abspath` to the inputs in `_list_outputs`. I tested out `Jacobian`, and I don't think this works. It'll run, but shouldn't work in a `Node` context or if you're trying to use `outputs.out_file` at all. See below for my process.
- A few (e.g., `SmoothTessellation`, `MRIMarchingCubes`) do something similar to the simple `_list_outputs` approach, but have `_gen_filename` and `_gen_outfilename` to edit the filenames a bit more. If the simple `_list_outputs` way doesn't work, I don't think these can either. I might be missing something though.


To check the `_list_outputs` method, I just ran `Jacobian`:
```python
import nipype.interfaces.freesurfer as fs
jacobian = fs.Jacobian()
jacobian.inputs.in_origsurf = 'lh.pial'
jacobian.inputs.in_mappedsurf = 'lh.pial'
jacobian.inputs.out_file = 'outfile.pial'
res = jacobian.run()
```

Checking the inputs and outputs:
```python
jacobian.cmdline
'mris_jacobian lh.pial lh.pial outfile.pial'

res.inputs['out_file']
'outfile.pial'

res.outputs['out_file']
'/home/data/nbc/sub-01/surf/outfile.pial'
```

But the file that is actually created is still `lh.outfile.pial`.

effigies:
Cool. Thanks for rounding up these cases.

> Some (e.g., VolumeMask, Aparc2Aseg, Contrast) output everything to the subject directory using a function called copy_to_subjdir. I'm not sure it's best to force a specific output directory.

I believe these were added as part of the [`reconall_workflow`](https://nipype.readthedocs.io/en/latest/interfaces/generated/nipype.workflows.smri.freesurfer.recon.html#create-reconall-workflow) effort. And it actually seems to do something similar to `MRIsExpand`, which is, in a node context, copy a bunch of implicit dependencies to the working directory. (It can also be used to copy everything to a subject directory, which is probably how it's used in the workflow, which tries to replicate the `recon-all` script in detail.)

I agree that this probably isn't an appropriate solution to interfaces that aren't used in `recon-all` or other pipeline scripts.

> MRIsExpand checks the beginning of out_file. If the file doesn't have either prefix (lh. or rh.), it takes the prefix from in_file and uses that. It's sort of similar to what I did.

Yes, it's basically what you did, except the naming behavior of `mris_expand` isn't super surprising. It's really just partially replicating [`MRISbuildFileName()`](https://raw.githubusercontent.com/freesurfer/freesurfer/dev/utils/mrisurf.c). There may be naming edge cases that aren't being captured, since we're working from the filename and not its contents.

> Several others (e.g., Apas2Aseg, RelabelHypointensities, MRIsCalc, Jacobian) just set the outputs by applying abspath to the inputs in _list_outputs. I tested out Jacobian, and I don't think this works. It'll run, but shouldn't work in a Node context or if you're trying to use outputs.out_file at all. See below for my process.

Yeah, this is definitely a bug. It works under the assumption that nobody actually changes `out_file`. I suspect this one could use the `MRIsExpand._associated_file` pretty straightforwardly.

> A few (e.g., SmoothTessellation, MRIMarchingCubes) do something similar to the simple _list_outputs approach, but have _gen_filename and _gen_outfilename to edit the filenames a bit more. If the simple _list_outputs way doesn't work, I don't think these can either. I might be missing something though.

These happen before it runs, so I think it should actually be fine, especially if they're generating absolute paths (which I'm pretty sure they should). But again, this correct behavior is predicated on users not specifying `out_file` manually.

----

I like the `FSSurfaceCommand` option. Possibly moving `MRIsExpand._associated_file` into `FSSurfaceCommand` would be a good start. And we could try to separate renaming and similar options from `_get_filecopy_info` by adding an extra hook:

```Python
class FSSurfaceCommand(FSCommand):
    ...
    def _get_filecopy_info(self):
        self._normalize_filenames()
        return super(FSSurfaceCommand, self)._get_filecopy_info()

    def _normalize_filenames(self):
        """Filename normalization routine to perform only when run in Node context"""
        pass
```

Then I'd rewrite `MRIsExpand`:

```Python
class MRIsExpand(FSSurfaceCommand):
    _cmd = 'mris_expand'
    input_spec = MRIsExpandInputSpec
    output_spec = MRIsExpandOutputSpec

    def _list_outputs(self):
        outputs['out_file'] = self._associated_file(self.inputs.in_file,
                                                    self.inputs.out_name)
        return outputs

    def _normalize_filenames(self):
        in_file = self.inputs.in_file

        pial = self.inputs.pial
        if not isdefined(pial):
            pial = 'pial'
        self.inputs.pial = self._associated_file(in_file, pial)

        if isdefined(self.inputs.thickness) and self.inputs.thickness:
            thickness_name = self.inputs.thickness_name
            if not isdefined(thickness_name):
                thickness_name = 'thickness'
            self.inputs.thickness_name = self._associated_file(in_file,
                                                               thickness_name)

        self.inputs.sphere = self._associated_file(in_file, self.inputs.sphere)

        super(MRIsExpand, self)._normalize_filenames()
```

-----

WDYT?

tsalo:
Should `_normalize_filenames` just `pass` in `FSSurfaceCommand` or should it try to rename `out_file`, since that's the most common output? E.g.,

```python
class FSSurfaceCommand(FSCommand):
    ...
    def _get_filecopy_info(self):
        self._normalize_filenames()
        return super(FSSurfaceCommand, self)._get_filecopy_info()

    def _normalize_filenames(self):
        """Filename normalization routine to perform only when run in Node context"""
        if isdefined(self.inputs.out_file):
            self.inputs.out_file = os.path.abspath(self.inputs.out_file) 
```

If `_normalize_filenames` deals with `inputs.out_file`, then there's no need to do anything with `_list_outputs`, right?
Also, if `out_file` (or whatever other outputs) has the full path, then is `_associated_file` necessary? As long as FreeSurfer doesn't add the hemisphere prefix to any output file arguments that include the path, the interface shouldn't need to add the prefix from the input file to the output files, unless that's something it assumes users want.

effigies:
This comment gets into the weeds a bit, so tl;dr:

1) `interface.run()` should not change the inputs, but `Node(interface).run()` can.
2) `interface._list_outputs()` must account for both contexts, that is, it should describe the behavior of the wrapped program on *any* input.
3) I'd be hesitant to make too many assumptions about user's desires. If they want to pass an argument with path information, they can.
4) If you want to just document the surprising behavior and save refactoring `FSSurfaceCommand` for another PR, I'm totally fine with that. (Likewise, I'm fine with hashing out all of these issues here; but this does threaten to become a little bigger than perhaps you originally intended.)

-----
To address your comment in detail:

> Should _normalize_filenames just pass in FSSurfaceCommand or should it try to rename out_file, since that's the most common output?

I wouldn't do that, because there's no guarantee that all `FSSurfaceCommand`s will have an `out_file` parameter (even if all of them do now). `isdefined` only works on existing entries in the spec.

> If `_normalize_filenames` deals with `inputs.out_file`, then there's no need to do anything with `_list_outputs`, right?

No, because `_list_outputs` needs to be agnostic of whether it was run in a Node or not; `_normalize_filenames` only applies to Nodes.

> Also, if out_file (or whatever other outputs) has the full path, then is _associated_file necessary?

`_associated_file` is also useful for describing where to find inputs that FreeSurfer is expecting to find in the same directory. See `MRIsExpand._get_filecopy_info` where implicit inputs are defined so they can be copied over. If there's no immediate need to put it in `FSSurfaceCommand`, don't worry about it. That can go in a later refactor, if it ever becomes relevant.

> As long as FreeSurfer doesn't add the hemisphere prefix to any output file arguments that include the path, the interface shouldn't need to add the prefix from the input file to the output files, unless that's something it assumes users want.

But sometimes we (or at least I) want it to add the hemisphere prefix, and I'll signal that by using an `out_file` specification that FreeSurfer will add a prefix to.

Consider the following cases for `mris_expand` (where `in_file = <indir>/lh.smoothwm`):

| Input | Node output | Non-Node output |
|--------|------------------|-------------------------|
| `out_file='graymid'` | `<workdir>/lh.graymid` | `<indir>/lh.graymid` |
| `out_file='./graymid'` | `<workdir>/graymid` | `<workdir>/graymid` |

The right column is determined by how the underlying program treats the inputs. The bottom row is inflexible because the user gave path information. So our only real question is how to treat the top left corner. If we always do `out_file = os.path.abspath(out_file)` in Nodes, then we get:

| Input | Node output | Non-Node output |
|--------|------------------|-------------------------|
| `out_file='graymid'` | `<workdir>/graymid` | `<indir>/lh.graymid` |
| `out_file='./graymid'` | `<workdir>/graymid` | `<workdir>/graymid` |

Which, in this context, is probably not what the user wanted. Now in your case, I think you want to actually adjust the behavior to be less surprising, like:

| Input | Node output | Non-Node output |
|--------|------------------|-------------------------|
| `out_file='out.stl'` | `<workdir>/out.stl` | `<indir>/lh.out.stl` |
| `out_file='./out.stl'` | `<workdir>/out.stl` | `<workdir>/out.stl` |

This seems completely reasonable. The `FSSurfaceCommand` approach I proposed gives each interface the flexibility to decide on what the "expected" behavior is.

tsalo:
Thanks for the detailed reply. I'll make the minimal changes and leave the full refactoring for later/someone with more knowledge. I think I just have one question left. When you say that `interface.run()` shouldn't change the inputs, do you mean that running:
```python
import nipype.interfaces.freesurfer as fs
mris = fs.MRIsCombine()
mris.inputs.in_files = ['lh.pial', 'rh.pial']
mris.inputs.out_file = 'out.stl'
mris.cmdline
```
should give you `'mris_convert --combinesurfs lh.pial rh.pial out.stl'` but create `lh.out.stl`, so that your inputs are untouched in the command? Or, should `cmdline` give you `'mris_convert --combinesurfs lh.pial rh.pial lh.out.stl'` so it creates what your `cmdline` says it will create?

effigies:
The former. `mris.inputs.out_file = 'out.stl'` should result in `out.stl` being specified on the command line. But then `mris.run().outputs.out_file` should be the full path to `lh.out.stl`. So you should also run your `out_file` through `os.path.abspath` before setting it in `_list_outputs`.

nipype/interfaces/freesurfer/base.py
```diff
@@ -162,8 +162,45 @@ def version(self):
...
 
+class FSSurfaceCommand(FSCommand):
+    """Support for FreeSurfer surface-related functions.
+    For some functions, if the output file is not specified starting with
+    'lh.' or 'rh.', FreeSurfer prepends the prefix from the input file to the
+    output filename. Output out_file must be adjusted to accommodate this.
+    By including the full path in the filename, we can also avoid this behavior.
+    """
+    def __init__(self, **inputs):
+        super(FSSurfaceCommand, self).__init__(**inputs)
```
effigies: No need for an `__init__` if all it calls is `super`.

nipype/interfaces/freesurfer/base.py
```diff
@@ -162,8 +162,45 @@ def version(self):
...
+
+    def _get_filecopy_info(self):
+        """Filename normalization routine to perform only when run in Node
+        context
+        """
```
effigies: Comment belongs in `_normalize_filenames`.

nipype/interfaces/freesurfer/base.py
```diff
@@ -162,8 +162,45 @@ def version(self):
...
+    @staticmethod
+    def _associated_file(in_file, out_name):
+        """Based on MRIsBuildFileName in freesurfer/utils/mrisurf.c
+
+        Use in_file prefix to indicate hemisphere for out_name, rather than
+        inspecting the surface data structure.
+        """
+        path, base = os.path.split(out_name)
+        if path == '':
+            _, in_file = os.path.split(in_file)
```
effigies: This should be `path, in_file = ...`. The logic here is "if `out_file` doesn't have path information, use `in_file`'s path".

nipype/interfaces/freesurfer/utils.py
```diff
@@ -954,6 +954,73 @@ def _gen_outfilename(self):
...
+
+    @staticmethod
+    def _associated_file(in_file, out_name):
+        """Unlike the standard _associated_file, which uses the prefix from
+        in_file, in MRIsCombine, it uses 'lh.' as the prefix for the output
+        file no matter what the inputs are.
+        """
+        path, base = os.path.split(out_name)
+        if path == '':
+            _, in_file = os.path.split(in_file)
```
effigies: Using `_` here instead of replacing `path` means that `mris_convert` outputs to the working directory by default. Is this correct?

nipype/interfaces/freesurfer/utils.py
```diff
@@ -954,6 +954,73 @@ def _gen_outfilename(self):
...
+                                      'in_files.')
+
+
+class MRIsCombine(FSSurfaceCommand):
+    """
+    Uses Freesurfer's mris_convert to combine two surface files into one.
+
+    For complete details, see the `mris_convert Documentation.
+    <https://surfer.nmr.mgh.harvard.edu/fswiki/mris_convert>`_
+
```
effigies: Here might be a good place to add a comment about outputs (so it shows up in [the docs](https://nipype.readthedocs.io/en/latest/interfaces/generated/nipype.interfaces.freesurfer.utils.html)):

```
If given an out_file that does not begin with 'lh.' or 'rh.', mris_convert will prepend 'lh.'
to the file name.
To avoid this behavior, consider setting out_file = './<filename>', or leaving out_file blank.
```

effigies:
@satra would you mind having a look at this one? Here's another case where we need to do a little pre-run fiddling to get FreeSurfer commands to run nicely in a Node, so I've suggested injecting a hook for `FSSurfaceCommand`s that goes just before `_get_filecopy_info`.

nipype/interfaces/freesurfer/base.py
tsalo: Sorry that's my mistake. With `mris_convert --combinesurfs` (`MRIsCombine`) it outputs to the working directory, but I didn't realize that that wasn't how regular calls to `mris_convert` worked. Will fix.

nipype/interfaces/freesurfer/utils.py
tsalo: With `MRIsCombine`, it outputs to the working directory if path information isn't provided. So I think that's what we want.

nipype/interfaces/freesurfer/base.py
tsalo: So go straight to `_get_filecopy_info`?

nipype/interfaces/freesurfer/utils.py
effigies: Sounds good. You can just remove this line, then, since `in_file` isn't used at all otherwise. Actually, the whoe thing could be shortened to:

```Python
path, base = os.path.split(out_name)
if path == '' and base[:3] not in ('lh.', 'rh.'):
    base = 'lh.' + base
return os.path.abspath(os.path.join(path, base))
```

And if you prefer, you could just do this directly in `_list_outputs`, rather than override `_associated_file`. (That's up to you, though.)

nipype/interfaces/freesurfer/base.py
effigies: Yup.

effigies:
LGTM. Assuming tests pass, and there are no objections, I'll merge tonight or in the morning.

Thanks for your patience.

nipype/interfaces/freesurfer/utils.py
```diff
@@ -954,6 +954,77 @@ def _gen_outfilename(self):
...
+        file no matter what the inputs are.
+        """
+        path, base = os.path.split(out_name)
+        if path == '':
+            hemis = ('lh.', 'rh.')
+            if base[:3] not in hemis:
+                base = 'lh.' + base
+        return os.path.abspath(os.path.join(path, base))
+
+    def _normalize_filenames(self):
```
satra: it says this function is used in the Node context, but where is it used inside Node?

satra:
@effigies , @tsalo - in general it looks good - i left a question about `Node` context.

nipype/interfaces/freesurfer/utils.py
effigies: https://github.com/nipy/nipype/pull/1948/files#diff-36ff1313779ecc73abba56e55501f0b2R173

satra: thanks.

PR #1949, opened 2017-04-11
enh: trait for imaging files + implementation in SPM preproc

mgxd:
should be extended to the other spm interfaces - is there a better way to approach this?

also updated some missed auto-tests


nipype/interfaces/spm/preprocess.py
```diff
@@ -217,6 +217,15 @@ def _parse_inputs(self):
...
         einputs = super(Realign, self)._parse_inputs()
+        is_gz = lambda x: x.endswith('.gz')
+        if isdefined(self.inputs.in_files):
+            if isinstance(self.inputs.in_files, list):
+                for imgf in self.inputs.in_files:
+                    if is_gz(imgf):
+                        raise TypeError('Input files must be uncompressed')
+            else:
+                if is_gz(self.inputs.in_files):
+                    raise TypeError('Input files must be uncompressed')
```
effigies: Not sure if there's a better approach in general, but the test can be made much more concise:

```Python
if isdefined(self.inputs.in_files) and any(imgf.endswith('.gz')
                                           for imgf in filename_to_list(self.inputs.in_files)):
    raise TypeError('Input files must be uncompressed')
```

satra:
i think this is a good place to start writing a new traits `ImageFile(types=['nifti-pair', 'nifti1'], compressed=False)` and replace the `File` trait in the SPM tools for in files with this new trait.

also SPM can accept list of lists of 3D files so the check in this PR has to account for those scenarios. this is another reason to write a new Trait.

mgxd:
@satra maybe something like this? only in this case, file endings are checked instead of types - any recommendation for that?

```python
class ImageFile(File):
    """ Defines a trait of specific neuroimaging files """
    
    def __init__(self, value='', filter=None, auto_set=False, entries=0, 
                 exists=False, types=[], compressed=True, **metadata):
        """ Trait handles neuroimaging files.

        Parameters
        ----------
        types : list
            The accepted file-types
        compressed : boolean
            Indicates whether the file-type can compressed
        """
        self.types = types
        self.compressed = compressed
        super(ImageFile, self).__init__(value, filter, auto_set, entries, exists,
                                        types, compressed, **metadata)

    def _validate_file(self):
        # check value depending on possible types
        if self.value and self.types:
            if not any(self.value.endswith(x) for x in types):
                raise TraitsError(
                    args="{} is not included in allowed types: {}".format(
                        self.value, ','.join(self.types)))
```

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +218,38 @@ def __init__(self, value='', auto_set=False, entries=0,
...
+        super(ImageFile, self).__init__(value, filter, auto_set, entries,
+                                        exists, **metadata)
+
+    def validate(self, object, name, value):
+        """ Validates that a specified value is valid for this trait.
+        """
+        validated_value = super(ImageFile, self).validate(object, name, value)
+        if validated_value and self.types:
+            if self.compressed:
+                self.types.extend([x + '.gz' for x in self.types])
```
satra: how would this handle `mgz`?

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +218,38 @@ def __init__(self, value='', auto_set=False, entries=0,
...
+        self.types = types
+        self.compressed = compressed
+        super(ImageFile, self).__init__(value, filter, auto_set, entries,
+                                        exists, **metadata)
+
+    def validate(self, object, name, value):
+        """ Validates that a specified value is valid for this trait.
+        """
+        validated_value = super(ImageFile, self).validate(object, name, value)
+        if validated_value and self.types:
```
satra: can you check if `self.types` always returns `True` even if the metadata is not there? in the past traits would some weird things like that.

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +218,38 @@ def __init__(self, value='', auto_set=False, entries=0,
         super(Directory, self).__init__(value, auto_set, entries, exists,
                                         **metadata)
 
+class ImageFile(File):
+    """ Defines a trait of specific neuroimaging files """
+
+    def __init__(self, value='', filter=None, auto_set=False, entries=0,
+                 exists=False, types=['nii', 'hdr', 'img'], compressed=True,
```
satra: the reason i would suggest going with file format types instead of extensions is that programs that accept nifti-1 may not be able to run with nifti-2.

so here is an incomplete list: http://www.reproducibleimaging.org/module-dataprocessing/03-data/ (see mri data section)

and here is a description of nifti-1 vs nifti-2 check for future reference (i don't want to implement all validators right now).

nipype/interfaces/traits_extension.py
mgxd: yeah - if `types` is defined as an empty list, it'll return `False`

nipype/interfaces/traits_extension.py
mgxd: that case would have to added into types

nipype/interfaces/traits_extension.py
mgxd: link to nifti1 vs nifti2?

maybe we can work around this by adding a field to `ImageFile` that specifies `format_version`?

satra: https://brainder.org/2015/04/03/the-nifti-2-file-format/

oesteban:
Since you are working on this new trait, I have a thought since long ago about the File trait. This could be a good PR for it.

Can we please change the messages when the trait validation fails?

- if the problem is the type, keep the message.
- BUT if it didn't validate because the metadata `exists` is `True` but the file is not found, could we just say that the file was not found?. The _value XX is not str or unicode_ message is very confusing.

nipype/interfaces/traits_extension.py
satra: say one of the other interfaces (e.g., dcm2niix) were to restrict input files to dicoms. how would you determine that the input was valid for the MGH dicoms for example, which don't have extensions?

i'm also thinking of how we can use this interface to auto-(un)compress eventually instead of restricting it.

further compressed nifti pair would involve img/hdr.gz pairs. if we check only the header, then spm could still run into trouble.

so types can be format with details about compressed and uncompressed being something like this to start with:

```python
{'nifti1': [('.nii', '.nii.gz'), (('.hdr', '.img'), ('.hdr', '.img.gz'))],
 'mgh': [('.mgh', '.mgz')], 
  ...
}
```

at least for now, since this solution is focused on spm, let's use Nifti1 as the type and check for the various combinations listed above.

satra: @gllmflndn: does spm now support nifti-2?

gllmflndn: SPM12 should read NIfTI-2 files - I don't think it has been used much yet though.

nipype/interfaces/traits_extension.py
satra: @mgxd - are you planning to fix this?

mgxd: yes, i'll revisit it today

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +223,68 @@ def __init__(self, value='', auto_set=False, entries=0,
         super(Directory, self).__init__(value, auto_set, entries, exists,
                                         **metadata)
 
+# lists of tuples
+# each element consists of :
+# - uncompressed (tuple[0]) extension
+# - compressed (tuple[1]) extension
+img_fmt_types = {
```
mgxd: @satra I haven't included validation yet, but LMKWYT

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +223,68 @@ def __init__(self, value='', auto_set=False, entries=0,
...
+        self.exts = extensions
+        super(ImageFile, self).__init__(value, filter, auto_set, entries,
+                                        exists, **metadata)
+
+    def grab_exts(self, exts=[]):
+        for fmt in self.types:
+            if fmt in img_fmt_types:
+                exts.extend(sum([[u for u in y[0]] if isinstance(y[0], tuple)
+                            else [y[0]] for y in img_fmt_types[fmt]], []))
+                if self.compressed:
```
satra: `compressed=True` should allow for compressed and normal. perhaps we should change `compressed` to `allow_compressed`

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +223,68 @@ def __init__(self, value='', auto_set=False, entries=0,
...
+# - uncompressed (tuple[0]) extension
+# - compressed (tuple[1]) extension
+img_fmt_types = {
+        'nifti1': [('.nii', '.nii.gz'),
+                   (('.hdr', '.img'), ('.hdr', '.img.gz'))],
+        'mgh': [('.mgh', '.mgz'), ('.mgh', '.mgh.gz')],
+        'nifti2': [('.nii', '.nii.gz'),
+                   (('.hdr', '.img'), ('.hdr', '.img.gz'))],
+        'cifti2': [('.nii', '.nii.gz'),
+                   (('.hdr', '.img'), ('.hdr', '.img.gz'))],
```
satra: cifti doesn't allow hdr-img i think - can you check.

nipype/interfaces/traits_extension.py
mgxd: it does - it will extend on the normal

nipype/interfaces/traits_extension.py
mgxd: looks like not supported in nifti2 or cifti2. also should img/hdr be separated as it's own ANALYZE format?

satra: i don't think we should support analyze :)

satra:
@mgxd - looks good. the new traits need some tests.

nipype/interfaces/traits_extension.py
```diff
@@ -218,6 +222,69 @@ def __init__(self, value='', auto_set=False, entries=0,
...
+        compressed : boolean
+            Indicates whether the file format can compressed
+        """
+        self.types = types
+        self.allow_compressed = allow_compressed
+        super(ImageFile, self).__init__(value, filter, auto_set, entries,
+                                        exists, **metadata)
+
+    def grab_exts(self):
+        # TODO: file type validation
```
djarecka: @mgxd : are you still planning to add something more to this validation? or should I remove TODO?

mgxd: Right now we are only checking the extensions, but in the future it would be nice to check bit differences (like say between nifti1 and nifti2)

PR #1952, opened 2017-04-12
WIP - BF -  give sdc_fmb access to ants reg settings (esp  initial_moving_transform_com)

WillForan:
Running `sdc_fmb` where `in_file` and `bmap_mag` or `phase` are not close in coordinate space results in error running ants

> Description: itk::ERROR: MattesMutualInformationImageToImageMetricv4(0x2600ad0): Too many samples map outside moving image buffer. There are only 151 valid points out of 2473 total points. The images do not sufficiently overlap. They need to be initialized to have more overlap before this metric will work. For instance, you can align the image centers by translation.

Setting `initial_moving_transform_com = 1` resolves this error. 

This pull request creates a new function `antreg_fmm2b0` where ant settings can be changed or added.  `sdc_fmb` is parameterized to to use the new function.

### DIFF
--- a/nipype/workflows/dmri/fsl/artifacts.py
+++ b/nipype/workflows/dmri/fsl/artifacts.py
@@ -516,7 +516,8 @@ head-motion correction)
 
 
 def sdc_fmb(name='fmb_correction', interp='Linear',
-            fugue_params=dict(smooth3d=2.0)):
+            fugue_params=dict(smooth3d=2.0),
+            fmm2b0=antreg_fmm2b0() ):
     """
     SDC stands for susceptibility distortion correction. FMB stands for
     fieldmap-based.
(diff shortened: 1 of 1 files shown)


nipype/workflows/dmri/fsl/artifacts.py
```diff
@@ -913,3 +894,54 @@ def _get_zoom(in_file, enc_dir):
...
+       'sigma_units': ['vox'] * 2,
+       'shrink_factors': [[6], [1]],  # ,[1] ]
+       'use_estimate_learning_rate_once': [True] * 2,
+       'use_histogram_matching': [True] * 2,
+       'initial_moving_transform_com': 0,
+       'collapse_output_transforms': True,
+       'winsorize_upper_quantile': 0.995,
+    }
+    defaults.update(alternatives)
+    for k,v in defaults.iteritems():
```
satra: i believe this for loop can be replaced with: `fmb20.inputs.trait_set(**defaults)`

oesteban:
@WillForan, a couple of suggestions here:

- I think we should start shipping some default settings for ants with nipype and load them with the `from_file` initialization argument. The same way we are doing in niworkflows (e.g. [settings file](https://github.com/poldracklab/niworkflows/blob/d6612ceb1e44ca2df30b404ef853d136ede277bb/niworkflows/data/t1-mni_registration_000.json)). Then:
  ```
  fmm2b0 = pe.Node(ants.Registration(output_warped_image=True, 
                                     from_file=your_settings.json),
                   name="FMm_to_B0")
  ```

- The original problem (`itk::ERROR: MattesMutualInformationImageToImageMetricv4`) typically happens for the confluence of two reasons. First, as you said, a bad initialization. Second, the gradient scales/smoothing/etc. is too large.
  - For the bad initialization, we could/should use [this new interface](https://github.com/nipy/nipype/pull/1980) that will be integrated soon (I hope). This is what the authors of ANTs use in their incredibly solid `antsBrainExtraction.sh` script. It is rather straightforward to use (default parameters should work). The only caveat is to pass in masked fixed and moving images.
  - For the scaling of the optimizer step, I'd suggest revising the `transform_parameters` since 1.0 is probably too large.

oesteban:
@satra: WDYT about shipping some settings for ANTS?, so that users can start off reliable options and experts from the ITK/ANTs world can vet them.

satra:
@oesteban - i think any settings that are helpful would be nice. although ants folks thinks of ants as a framework and one may need to optimize settings for each use case. so i don't know whether we should include them or simply create a set of json files that people could use for various ants needs.

oesteban:
@satra I think settings like the ones you showed me once for T1w to MNI registration (which are pretty close -if not the same- to those they use in the `antsBrainExtraction.sh`) would be really useful.

Then, some other settings like this registration here, you are to release the workflow with a certain set of parameters anyway, so why don't you give them a better place and make them available to other workflows and interfaces?

PR #1958, opened 2017-04-19
Centralize FreeSurfer version parsing

effigies:
This change replaces `interfaces.freesurfer.preprocess.FSVersion` with `interfaces.freesurfer.Info.looseversion()`, and makes the corresponding updates throughout the code.

The goal is to get a well-behaved version string that is comparable across versions, including no installation. `LooseVersion` is used in anticipation of the most common use case.

`FSCommand().version` is updated to pull the version string out of `Info.looseversion()` to ensure that the same logic is applied whenever a parsed version string is used.

Release versions use simple version strings (no githash). Post-v6 `dev` versions follow format suggested by @satra in #1928: `6.0.0-dev.GITHASH`. `dev` versions without githashes are parsed as before.

Closes #1928.

### DIFF
--- a/nipype/interfaces/freesurfer/base.py
+++ b/nipype/interfaces/freesurfer/base.py
@@ -17,15 +17,16 @@ See the docstrings for the individual classes for 'working' examples.
 from __future__ import print_function, division, unicode_literals, absolute_import
 from builtins import open, object, str
 
-
 import os
 
+from ... import LooseVersion
 from ...utils.filemanip import fname_presuffix
 from ..base import (CommandLine, Directory,
                     CommandLineInputSpec, isdefined,
                     traits, TraitedSpec, File)
 
 __docformat__ = 'restructuredtext'
+
 
 class Info(object):
     """ Freesurfer subject directory and version information.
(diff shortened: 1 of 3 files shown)


satra:
@effigies - thanks for this. but i need to bring up one issue with my suggestion.

```python
In [11]: LooseVersion('6.0.0') > LooseVersion('6.0.0-rc1')
Out[11]: False

In [12]: LooseVersion('6.0.0') > LooseVersion('6.0.0-dev.g123456')
Out[12]: False
```

the first case often predates the release, while the second case post-dates the release. also github hashes are not comparable, which is why git describe looks like this `6.0.0-1234-g123456` the middle number ensures comparison except for comparing across branches.

i really don't know a good solution to this that goes across the software we support. we may have to do this on a case by case basis.

satra:
unless we do it like nipype dev version == `0.13.0-gcf2d093.dev` (we should fix that as well :) ), where after release we bump the dev version to the next release.

effigies:
Agreed that `-dev` and `-rc1` play havoc with `LooseVersion`, but I think our checks will be `>= LooseVersion('6.0.0')` or similar almost all the time. We *could* run up against a situation where somebody's running a release candidate after we make a branch dependent on the stable release, but then the fix is "run stable".

I'm also not *super* worried about ordering of dev releases based on githashes, but they do actually solve this problem by using `YYYYMMDD-GITHASH`, so we could do that. e.g. `6.0.0-dev-YYYMMDD-GITHASH`. It's kind of moot unless/until we're make a logical branch on "newer than this specific dev version", but there's no reason not to allow for that eventuality.

PR #1962, opened 2017-04-20
FIX: Restore FSVersion to string

effigies:
#1958 changed `FSVersion` from a string to a `LooseVersion` object. This change corrects this issue.


satra:
@effigies - was this breaking something, if so should there be a test for this to prevent future breakage?

effigies:
Yes, see poldracklab/niworkflows#151.

I'm actually not sure that anybody else would have depended on this, since it's something we added post 0.13.0-rc1, but I also don't want to pull the rug out from anybody who did. It didn't show up in our tests because we don't pin to master, but to specific merge commits. So it was found by a contributor who keeps his nipype bleeding edge.

Test passes locally.

PR #1963, opened 2017-04-21
RF: Derive MRIsExpand from FSSurfaceCommand

effigies:
This change takes advantage of work in #1948 to handle some common quirks in FreeSurfer commands that write surface files.

Some doc fixes and a minor update to `FSSurfaceCommand` to not push absolute paths at that stage (consistent with how `MRIsExpand._associated_file` worked).

@tsalo I made a couple changes to `MRIsCombine`, if you'd have a look. It's mostly docs, but I also moved the `out_file` description into `_list_outputs`, just because it didn't really need the machinery of `_associated_file`. I can revert these if you prefer.


effigies:
Added tests for `FSSurfaceCommand` and `MRIsExpand`. Not for `MRIsCombine`. In the process realized there was an option that shouldn't be enabled.

PR #1964, opened 2017-04-21
FIX: Return simple filename from export_graph if simple form is requested

effigies:
Currently the detailed graph is always returned, so code that depends on the return value of `export_graph`, such as the Sphinx workflow extension.

`export_graph()` currently always writes out a detailed graph and a (potentially) simple one. I didn't change this, because I'm not sure if anybody is expecting both graphs.

Related: poldracklab/fmriprep#463


effigies:
This resolves our issue. See https://github.com/poldracklab/fmriprep/pull/464#issuecomment-296264991.

PR #1966, opened 2017-04-23
ENH: ReconAll handle less common cases

effigies:
This PR is two-in-one. Can be split if needed.

Following on #1790/#1824, the first updates the "steps" checking to only check the steps implied by the directive.

 * `Undefined` removed the directive flag, and so also the steps checking
 * `autorecon2-*` are considered the same as `autorecon2` for these purposes. I can look more into if they should be done differently if that's desired.
* Non-`autorecon*` directives are given the same treatment as `all`, mainly because it doesn't change the behavior from before, and I haven't looked up what they actually do.

The primary goal is to enable `ReconAll(directive=Undefined)` to behave as expected when a subject's recon directory already exists. I don't believe that the others will have much effect but avoid throwing in a ton of `-no*` flags that aren't needed.

Related: poldracklab/fmriprep#429

-----

The second commit is an update to expert files, following on #1857. FreeSurfer saves its expert options files on the first introduction. If they are ever set again, they need to be passed with `-xopts-overwrite`, or else it throws an error. There are also `-xopts-use` (default) and `-xopts-clean`. The proposed change does the following:

* Checks the expert options file to be written against an existing file; if the contents are the same, `-xopts-use` is passed instead of writing a new expert options file.
* If an `-xopts-*` flag is passed by the user, this check is skipped, on the assumption the user knows what they're doing.

An FMRIPREP user [encountered an issue](https://neurostars.org/t/fmriprep-error-interface-reconallrpt-failed-to-run/331/10) with FreeSurfer complaining about needing `-xopts-overwrite` when due to memory reasons the node which set the expert options crashed and had to be rerun.

Related: poldracklab/fmriprep#425


satra:
@effigies - while we are at this should we add options like `-mprage`, `-bigventricles`?

effigies:
Yeah, I'm fine with that. There are so many options that I don't know what all is worth putting in. Do you have others you use regularly?

satra:
most of the time i just run with `-all` but a few have come up that are now relevant to some of the things we are doing - normally i just stick them in the generic `args`

a few more from the recent release:

```
 Segmentation of substructures of hippocampus and brainstem
  -hippocampal-subfields-T1 : segmentation of hippocampal subfields using input T1 scan
  -hippocampal-subfields-T2 file ID : segmentation using an additional scan (given by file);
                                      ID is a user-defined identifier for the analysis
  -hippocampal-subfields-T1T2 file ID : segmentation using additional scan (given by file) and input T1
                                        simultaneously; ID is a user-defined identifier for the analysis
  -brainstem-structures : segmentation of brainstem structures
```

these require the matlab common runtime and very specifically following these instructions:

https://surfer.nmr.mgh.harvard.edu/fswiki/MatlabRuntime

i'm hoping to build a container with everything and then reduce using reprozip for recon-all.

- run single recon with all options (T1 + T2 + brainstem + hippocampal subfields)
- run longitudinal version
- run custom template building

effigies:
`-brainstem` is easy enough. Do you know how to encode the hippocampal subfields options in a single trait?

Something like:

```Python
hippocampal_subfields = traits.Enum(
    'T1',
    traits.Tuple(traits.Enum('T2', 'T1T2'), File(exists=True), traits.Str),
    argstr='-hippocampal-subfields-%s ...')
```

But not actually that, because trying `ReconAll(hippocampal_subfields=('T2', '/dev/null', 'test'))` gives:

```
TraitError: The 'hippocampal_subfields' trait of a ReconAllInputSpec instance
must be u'T1' or <traits.trait_types.Tuple object at 0x7f6aed881510>, but a
value of ('T2', '/dev/null', 'test') <type 'tuple'> was specified.
```

Similarly

```Python
hippocampal_subfields = traits.Tuple(
    traits.Enum('T1', 'T2', 'T1T2'), File(exists=True), traits.Str(),
    argstr='-hippocampal-subfields-%s ...')
```
yields:
```
TraitError: The 'hippocampal_subfields' trait of a ReconAllInputSpec instance
must be a tuple of the form: ('T1' or 'T2' or 'T1T2', an existing file name, a
unicode string), but a value of ('T2', '/dev/null', 'test') <class 'tuple'> was specified.
```

It seems this should be doable, but I can just do an `xor` set, too.

nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -630,8 +630,26 @@ class ReconAllInputSpec(CommandLineInputSpec):
...
+    brainstem = traits.Bool(argstr='-brainstem-structures',
+                            desc='Segment brainstem structures')
+    hippocampal_subfields_T1 = traits.Bool(
+        argstr='-hippocampal-subfields-T1', min_ver='6.0.0',
+        desc='segment hippocampal subfields using input T1 scan')
+    hippocampal_subfields_T2 = traits.Tuple(
+        File(exists=True), traits.Str(),
+        argstr='-hippocampal-subfields-T2 %s %s', min_ver='6.0.0',
+        desc=('segment hippocampal subfields using T2 scan, identified by '
+              'ID (may be combined with hippocampal_subfields_T1)'))
```
effigies: @satra What do you think of this implementation? The tuple might be a little awkward; if you'd prefer, I could make a `hippocampal_subfields_id` that gets pulled in in `_format_arg`. If so, is there a reasonable default ID name?

satra: is the file here the same as the `T2 input`, in which case, should we simply say this field requires T2?

effigies: No. A separate filename is [parsed](https://github.com/freesurfer/freesurfer/blob/dev/scripts/recon-all#L7443-L7454). I don't know enough about these processes to know if you'd use the same T2 for hippocampal subfields as for pial surface refinement, but certainly it's not required by their code that they be the same file.

PR #1968, opened 2017-04-24
enh: multiple masks for compcor

mgxd:
allows multiple masks to be passed in - outputs are stacked in the same components file


nipype/algorithms/confounds.py
```diff
@@ -434,7 +441,7 @@ class TCompCorInputSpec(CompCorInputSpec):
 
 class TCompCorOutputSpec(CompCorInputSpec):
     # and all the fields in CompCorInputSpec
-    high_variance_mask = File(exists=True, desc="voxels excedding the variance threshold")
+    high_variance_mask = InputMultiPath(File(exists=True, desc="voxels excedding the variance threshold"))
```
satra: should this be OutputMultiPath?

nipype/algorithms/confounds.py
```diff
@@ -302,7 +302,7 @@ def _list_outputs(self):
 class CompCorInputSpec(BaseInterfaceInputSpec):
     realigned_file = File(exists=True, mandatory=True,
                           desc='already realigned brain image (4D)')
-    mask_file = File(exists=True, desc='mask file that determines ROI (3D)')
+    mask_file = InputMultiPath(File(exists=True, desc='mask file(s) that determines ROI (3D)'))
```
satra: how does this work for `TCompCor`? i see that only the first one is being used.

nipype/algorithms/confounds.py
```diff
@@ -464,7 +471,7 @@ def _run_interface(self, runtime):
                              .format(self.inputs.realigned_file, imgseries.ndim, imgseries.shape))
 
         if isdefined(self.inputs.mask_file):
-            in_mask_data = nb.load(self.inputs.mask_file, mmap=NUMPY_MMAP).get_data()
+            in_mask_data = nb.load(self.inputs.mask_file[0], mmap=NUMPY_MMAP).get_data()
```
satra: should we consider a union or intersection of masks, and a new Enum trait (for 'union', 'intersect') and xor-ed with an index trait if we want to index just one of the masks.

```
merge_method = traits.Enum('union', 'intersect', 'none', xor=['mask_index'], ...)
mask_index = traits.Range(0, low=0, xor=['merge_method'], ...)
```

`none` corresponds to per mask.

satra: if you can make these changes, we can get them in.

mgxd: @satra maybe you can take a look, there is a failing test I have to figure out first

mgxd:
Still have to add some basic testing but LMKWYT

nipype/algorithms/confounds.py
```diff
@@ -456,52 +500,98 @@ class TCompCor(CompCor):
...
+                threshold_std = np.percentile(tSTD, 100. *
+                                        (1. - self.inputs.percentile_threshold))
+                mask = tSTD >= threshold_std
+
+                mask_data = np.zeros_like(in_mask)
+                mask_data[in_mask != 0] = mask
+                # save mask
+                mask_file = os.path.abspath('mask{}.nii'.format(i))
+                nb.Nifti1Image(mask_data, nb.load(
+                    self.inputs.realigned_file).affine).to_filename(mask_file)
```
satra: this should also have the numpy MMAP

nipype/algorithms/confounds.py
```diff
@@ -456,52 +500,98 @@ class TCompCor(CompCor):
...
-                       nb.load(self.inputs.realigned_file).affine).to_filename(mask_file)
-        IFLOG.debug('tCompcor computed and saved mask of shape {} to mask_file {}'
-                   .format(mask.shape, mask_file))
-        self.inputs.mask_file = mask_file
-        self._set_header('tCompCor')
-
+            # save mask
+            mask_file = os.path.abspath('mask.nii')
+            nb.Nifti1Image(mask_data,
+                  nb.load(self.inputs.realigned_file).affine).to_filename(mask_file)
```
satra: and here

nipype/algorithms/confounds.py
```diff
@@ -301,23 +301,29 @@ def _list_outputs(self):
 
 class CompCorInputSpec(BaseInterfaceInputSpec):
     realigned_file = File(exists=True, mandatory=True,
-                          desc='already realigned brain image (4D)')
-    mask_file = File(exists=True, desc='mask file that determines ROI (3D)')
-    components_file = File('components_file.txt', exists=False,
-                           usedefault=True,
-                           desc='filename to store physiological components')
+        desc='already realigned brain image (4D)')
+    mask_files = InputMultiPath(File(exists=True,
```
satra: this should go through a deprecation cycle. so keep the old name and point to the new name. 

e.g.: https://github.com/nipy/nipype/blob/master/nipype/interfaces/spm/preprocess.py#L812

Issue #1965, opened 2017-04-22
fmri_spm.py  - No crash file, workflow just seems to stall


jnecus:
### Summary
I am attempting to use the example tutorial workflow :fmri_spm.py

Everything seems to run OK until it reaches rapidart (ra.ArtifactDetect()).

At this point the following warning is provided in the IPython terminal. No crash file is given and all processes just seem to kind of stop. A blank matplotlib figure also pops up.

**As an additional note:** ArtifactDetect will run  to completion (as no crash file is generated);however, in order to allow the workflow to progress I must keep on manually exiting (Ctrl+C) and restarting the workflow until every subject has surpassed this step.
 
### Warning

> because the backend has already been chosen; 
> matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
> or matplotlib.backends is imported for the first time.

### Platform details:
```
{'nibabel_version': '2.1.0', 'networkx_version': '1.11', 'numpy_version': '1.11.3', 'sys_platform': 'linux2', 'sys_version': '2.7.13 |Anaconda custom (64-bit)| (default, Dec 20 2016, 23:09:15) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]', 'commit_source': u'archive substitution', 'commit_hash': '%h', 'pkg_path': '/home/joe/anaconda2/lib/python2.7/site-packages/nipype', 'sys_executable': '/home/joe/anaconda2/bin/python', 'traits_version': '4.6.0', 'scipy_version': '0.18.1'}
0.13.0-dev
```

djarecka:
can you please point me to the specific tutorial you're using? are you just running  [fmri_spm.py](https://github.com/nipy/nipype/blob/master/examples/fmri_spm.py)?

Do you know what is a version of your `matplotlib`?

jnecus:
Link to tutorial: ` http://nipype.readthedocs.io/en/latest/users/examples/fmri_spm.html`

Matplotlib version:  `1.5.1`

Only thing that I have changed from this tutorial (other than imputs and contrasts) is:

Set default path to be SPM8 (didnt seem to work with SPM12): `mlab.MatlabCommand.set_default_paths('/software/matlab/spm8')`

djarecka:
[...]

PR #1969, opened 2017-04-24
changing backend in ArtifactDetect

djarecka:
not sure if this cause the #1965, but I believe that the backend should be set there anyway.


satra:
restart circle. i think this looks reasonable.

Issue #1971, opened 2017-04-25
Problem with importing NUMPY_MMAP


isukrit:
### Summary
I was trying to run pipeline for rs-fMRI pre processing (http://nipype.readthedocs.io/en/latest/users/examples/rsfmri_vol_surface_preprocessing_nipy.html). I get the following 2 errors:
from nipype.algorithms.compcor import ACompCor
ImportError: No module named compcor

if I comment this out I get:
from nipype.utils import NUMPY_MMAP
ImportError: cannot import name NUMPY_MMAP

### Platform details:
please paste the output of: `python -c "import nipype; print(nipype.get_info()); print(nipype.__version__)"`

{'nibabel_version': '2.1.0', 'networkx_version': '1.11', 'numpy_version': '1.11.0', 'sys_platform': 'linux2', 'sys_version': '2.7.12 (default, Nov 19 2016, 06:48:10) \n[GCC 5.4.0 20160609]', 'commit_source': '(none found)', 'commit_hash': '<not found>', 'pkg_path': '/usr/lib/python2.7/dist-packages/nipype', 'sys_executable': '/usr/bin/python', 'traits_version': '4.5.0', 'scipy_version': '0.17.0'}
0.13.0-dev

mgxd:
@isukrit these imports are outdated and the tutorial should be fixed. To fix this, use the following imports instead.

```python
from nipype.algorithms.confounds import aCompCor
from nipype.utils.config import NUMPY_MMAP
```

satra:
closed via #1973

isukrit:
Solved it for ACompCor. NUMPY_MMAP still gives the same error.

Traceback (most recent call last):
  File "rsfmri_vol_surface_preprocessing_nipy.py", line 78, in <module>
    from nipype.utils.config import NUMPY_MMAP
ImportError: cannot import name NUMPY_MMAP

ashgillman:
@isukrit I had same issue, info can be found here https://github.com/pypa/setuptools/issues/261

You need to upgrade setuptools >= 21.2.1

PR #1973, opened 2017-04-25
fix: incorrect import paths

mgxd:
fixes error in example from #1971 

[This page](http://nipype.readthedocs.io/en/latest/users/examples/rsfmri_vol_surface_preprocessing_nipy.html) also needs to be updated, but where is it?


satra:
that page gets generated automatically from the code.

PR #1977, opened 2017-04-28
update tests to py36, clean up documentation towards release

satra:


nipype/info.py
```diff
@@ -94,16 +95,16 @@ def get_nipype_gitversion():
 """
 
 # versions
-NIBABEL_MIN_VERSION = '2.0.1'
+NIBABEL_MIN_VERSION = '2.1.0'
 NETWORKX_MIN_VERSION = '1.7'
 NUMPY_MIN_VERSION = '1.6.2'
```
oesteban: We should probably raise to numpy>=1.11, I think the mmap issue was solved with that.

oesteban:
@satra I'm updating the `nipype/base` image to add `g++`. With that one installed, the new image should build correctly. Will send a modification to your PR when I check everything is ok (within a couple of hours).

nipype/info.py
satra: since we support the mmap issue, perhaps we can change this after the release.

PR #1978, opened 2017-04-28
FIX: Typo in private method name

effigies:
#1966 bug.

Sorry.


effigies:
Testing locally. Will post when I'm sure this fixes the problem.

effigies:
Fixes the issue. `ReconAll` runs fine with and without existing expert options.

PR #1979, opened 2017-04-28
small edit of the MCFLIRT tests

djarecka:


nipype/interfaces/fsl/tests/test_preprocess.py
```diff
@@ -344,6 +350,9 @@ def test_mcflirt(setup_flirt):
 
     for name, settings in list(opt_map.items()):
         fnt = fsl.MCFLIRT(in_file=infile, **{name: settings[1]})
+        outfile = os.path.join(os.getcwd(), nme)
+        outfile = fnt._gen_fname(outfile, suffix='_mcf')
```
effigies: Since these are always the same, they could go up under `_, nme = os.path.split(infile)`, rather than be in the loop body.

satra: the first line could move, but outfile variable name would need to change because the second outfile has to be generated within the loop.

effigies:
Good catch. Sorry about that. Certainly don't hold this up based on that
comment.

PR #1980, opened 2017-04-28
[ENH] Add interface to antsAffineInitializer

oesteban:
Adds a very simple interface to `antsAffineInitializer`, which is used
in the `antsBrainExtraction.sh` script to initialize the registration to
template.


nipype/interfaces/ants/utils.py
```diff
@@ -178,3 +179,49 @@ def _list_outputs(self):
...
+    >>> init.cmdline # doctest: +ALLOW_UNICODE
+    'antsAffineInitializer 3 fixed1.nii moving1.nii transform.mat 15.000000 0.100000 0 10'
+
+    """
+    _cmd = 'antsAffineInitializer'
+    input_spec = AffineInitializerInputSpec
+    output_spec = AffineInitializerOutputSpec
+
+    def _list_outputs(self):
+        return {'out_file': os.path.abspath(self.inputs.out_file)}
```
satra: i'll merge this, but does this not happen automatically from the specified `out_file` in inputs?

PR #1985, opened 2017-05-01
[ENH] Add "profiling" outputs to ants.Registration

oesteban:
Adds a `profiling` flag that enables parsing of the standard output from ANTS. This first version only reads two values: the total elapsed time and the final value of the metric.


nipype/interfaces/ants/registration.py
```diff
@@ -379,6 +379,8 @@ class RegistrationInputSpec(ANTSCommandInputSpec):
         low=0.0, high=1.0, value=0.0, argstr='%s', usedefault=True, desc="The Lower quantile to clip image ranges")
 
     verbose = traits.Bool(argstr='-v', default=False)
+    profiling = traits.Bool(False, usedefault=True,
+                            desc='generate profiling output fields')
```
chrisfilo: There does not seem to be any performance penalty on collecting this information - is there a need for this input?

nipype/interfaces/ants/registration.py
```diff
@@ -395,6 +397,8 @@ class RegistrationOutputSpec(TraitedSpec):
     warped_image = File(desc="Outputs warped image")
     inverse_warped_image = File(desc="Outputs the inverse of the warped image")
     save_state = File(desc="The saved registration state to be restored")
+    metric_value = traits.Float(desc='the final value of metric')
+    elapsed_time = traits.Float(desc='the total elapsed time as reported by ANTs')
```
chrisfilo: Is there a difference between this `elapsed_time` and `duration` stored in every `CommandLineInterface` `runtime`? If you need this as an output it might be better to enable this for all interfaces.

nipype/interfaces/ants/registration.py
oesteban: Not really.

My original idea is that we grow the number of outputs that are parsed from the log. When there are a bunch, there's a good chance that logs formatting change or we make mistakes. Being an added feature (and nonfunctional), it could come handy to have a switch to avoid parsing the stdout.

But I have no concern with removing the input.

nipype/interfaces/ants/registration.py
oesteban: Elapsed time is calculated by ANTs so it refers exclusively to the registration process. The duration of the runtime object adds the (little) overhead of nipype. In general, they will be equivalent, but this could be a good handle to catch problems.

Additionally, my plan would be to add the elapsed_time per registration stage at some point. Then, it will make sense to be consistent with ANTs calculations.

oesteban:
@chrisfilo good to go?

chrisfilo:
Please add a changelog entry and you can go ahead and merge it.

PR #1986, opened 2017-05-01
[FIX] Define ANTSPATH for ants.BrainExtraction automatically

### LINKED ISSUE #1901: [BUG] antsBrainExtraction interface fails when $ANTSPATH is not defined, but exit code is 0
### Summary
 antsBrainExtraction interface fails when $ANTSPATH is not defined, but exit code is 0

The problem is the following
```
$ antsBrainExtraction.sh -h
we cant find the N4 program -- does not seem to exist. please (re)define $ANTSPATH in your environment.
$ echo $?
0
```
when nipype gets to check the output, since the exit code is zero, it breaks. We could check, at least, for this output message and reraise the message.

cc @rwblair .

oesteban:
Only when it is not already defined. Also check if the interface failed for that reason nonetheless.

Fixes #1901

### DIFF
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -692,6 +692,33 @@ class BrainExtraction(ANTSCommand):
     output_spec = BrainExtractionOutputSpec
     _cmd = 'antsBrainExtraction.sh'
 
+    def _run_interface(self, runtime, correct_return_codes=(0,)):
+        out_environ = self._get_environ()
+        if out_environ.get('ANTSPATH') is None:
+            runtime.environ.update(out_environ)
+            executable_name = self.cmd.split()[0]
+            exist_val, cmd_path = _exists_in_path(executable_name,
+                                                  runtime.environ)
+            if not exist_val:
+                raise IOError("command '%s' could not be found on host %s" %
+                              (self.cmd.split()[0], runtime.hostname))
+
+            runtime.environ.update({'ANTSPATH': cmd_path})
+
+        runtime = super(BrainExtraction, self)._run_interface(runtime)
+
+        if 'we cant find the N4 program' in runtime.stdout:
+            errmsg = ('antsBrainExtraction.sh requires the environment variable '
+                      'ANTSPATH to be defined')
+            if runtime.stderr is None:
+                runtime.stderr = errmsg
+            else:
+                runtime.stderr += '\n' + errmsg
+            runtime.returncode = 1
+            self.raise_exception(runtime)
+
+        return runtime
+
     def _list_outputs(self):
         outputs = self._outputs().get()
         outputs['BrainExtractionMask'] = os.path.join(os.getcwd(),
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -710,9 +710,14 @@ class BrainExtraction(ANTSCommand):
         runtime = super(BrainExtraction, self)._run_interface(runtime)
 
         # Still, double-check if it didn't found N4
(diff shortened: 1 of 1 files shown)


nipype/interfaces/ants/segmentation.py
```diff
@@ -692,6 +693,35 @@ class BrainExtraction(ANTSCommand):
...
+        if out_environ.get('ANTSPATH') is None:
+            runtime.environ.update(out_environ)
+            executable_name = self.cmd.split()[0]
+            exist_val, cmd_path = _exists_in_path(executable_name, runtime.environ)
+            if not exist_val:
+                raise IOError("command '%s' could not be found on host %s" %
+                              (self.cmd.split()[0], runtime.hostname))
+
+            # Set the environment variable if found
+            runtime.environ.update({'ANTSPATH': os.path.dirname(cmd_path)})
```
satra: i believe `ANTSPATH` requires a trailing `/`. could you confirm this and make sure we adjust for it?

nipype/interfaces/ants/segmentation.py
```diff
@@ -692,6 +693,35 @@ class BrainExtraction(ANTSCommand):
...
+
+        runtime = super(BrainExtraction, self)._run_interface(runtime)
+
+        # Still, double-check if it didn't found N4
+        if 'we cant find the N4 program' in runtime.stdout:
```
satra: since the script can be in a different location than the N4 executable, can we not do this check before running in this particular case?

satra:
in general, for next release, i would like to decouple these checks as requirements of interfaces and have a generic checker mechanism that spans local paths and container id/paths.

nipype/interfaces/ants/segmentation.py
oesteban: Most of the scripts don't need the trailing `/`. Particularly, the `antsBrainExtraction.sh` doesn't need it.

But I can add it anyways if you want.

nipype/interfaces/ants/segmentation.py
oesteban: Sorry I'm lost here. Which check would you remove?

oesteban:
Sounds good. The problem here is a bit different, since it is the definition of `ANTSPATH` for the ANTs scripts. For next release we can have an `ANTSScript` to complement the `ANTSCommand` that takes care of the `ANTSPATH` with more generality.

blakedewey:
This looks good to me. It should cover the bases for when errors could occur. I agree that a more general solution in the future would be preferable, even recreating this script in Nipype would be a great addition as well.

oesteban:
+1 to adding a pure-nipype workflow for antsBrainExtraction

oesteban:
@satra I left a couple of questions, if you had a minute I can get this ready to merge :)

nipype/interfaces/ants/segmentation.py
satra: i meant to put the `N4` check before calling `_run_interface` instead of checking the output.

nipype/interfaces/ants/segmentation.py
satra: then not necessary.

nipype/interfaces/ants/segmentation.py
oesteban: Judging from this: https://github.com/stnava/ANTs/blob/646459bf297a50cdf608ea7d39cb79b44906ad22/Scripts/antsBrainExtraction.sh#L5-L20, I think we just parse the output for "we cant find" and raise the message for the missing tool.

For next release we can decouple or, as @blakedewey suggested, replicate antsBrainExtraction in pure nipype.

Issue #1959, opened 2017-04-20
PBS TypeError crash


gjcooper:
### Summary
With a script that was previously running successfully with Multiproc I am now attempting to run on our Uni grid running PBS. The first couple of jobs submitted crash and return a TypeError.

Issue first occurred on release 0.12.1, tried upgrading to 0.13.0rc1 as it seemed to have new code realting to python3.
### Actual behavior
Trace for the crashes appears as:

```bash
170420-12:05:51,31 workflow ERROR:
         ['Node selectfiles.a1 failed to run on host rcglogin.']
170420-12:05:51,31 workflow INFO:
         Saving crash info to /home/gjc216/crash-20170420-120551-gjc216-selectfiles.a1-f1cc94d5-dfe1-41ae-b203-a8d7385aa7f9.pklz
170420-12:05:51,31 workflow INFO:
         Traceback (most recent call last):
  File "/home/gjc216/.virtualenvs/nipype/lib/python3.5/site-packages/nipype/pipeline/plugins/base.py", line 245, in run
    result = self._get_result(taskid)
  File "/home/gjc216/.virtualenvs/nipype/lib/python3.5/site-packages/nipype/pipeline/plugins/base.py", line 518, in _get_result
    if self._is_pending(taskid):
  File "/home/gjc216/.virtualenvs/nipype/lib/python3.5/site-packages/nipype/pipeline/plugins/pbs.py", line 55, in _is_pending
    return errmsg not in e
TypeError: a bytes-like object is required, not 'str'
```
### Expected behavior
Selectfiles should run successfully
### How to replicate the behavior
Pretty sure it is related to the unicode/str change in python3, so I am **assuming** that any python3 run ising PBS will fail
### Platform details:
```
{'scipy_version': '0.19.0', 'commit_source': 'installation', 'sys_version': '3.5.3 (default, Mar 17 2017, 08:49:30) \n[GCC 4.8.5 20150623 (Red Hat 4.8.5-11)]', 'traits_version': '4.6.0', 'networkx_version': '1.11', 'numpy_version': '1.12.1', 'nibabel_version': '2.1.0', 'sys_platform': 'linux', 'pkg_path': '/home/gjc216/.virtualenvs/nipype/lib/python3.5/site-packages/nipype', 'sys_executable': '/home/gjc216/.virtualenvs/nipype/bin/python', 'commit_hash': '4f08c9a'}
0.13.0-rc1
```

[...]

PR #1987, opened 2017-05-01
FIX: TypeError bug for PBS process communication in python3

gjcooper:
When using python3 to call the PBS status command line script there is a TypeError crash (bytes-like object not str required). To fix this I have used the interfaces.base CommandLine class.

CommandLine already uses the default encoding to decode the
stdout/stderr of the command, so take advantage of that functionality
rather than re-implementing.

First try was to implement decoding pattern seen elsewhere - then I
changed my mind and used a CommandLine object from the interfaces.base
module. (It was already imported to the module anyway)

First Attempt
-------------

Decode the stderr using the default encoding

* Fixes TypeError bug for PBS jobs running on python3.x
* Fixes issue #1959

Second Attempt
--------------

Use the nipype CommandLine class instead of Popen

Remove unused modules

Ignore exceptions in qstat CommandLine call.

### DIFF
--- a/nipype/pipeline/plugins/pbs.py
+++ b/nipype/pipeline/plugins/pbs.py
@@ -6,7 +6,6 @@ from builtins import str, open
 
 import os
 from time import sleep
-import subprocess
 
 from ...interfaces.base import CommandLine
 from .base import (SGELikeBatchManagerBase, logger, iflogger, logging)
@@ -46,13 +45,12 @@ class PBSPlugin(SGELikeBatchManagerBase):
         super(PBSPlugin, self).__init__(template, **kwargs)
 
     def _is_pending(self, taskid):
-        #  subprocess.Popen requires taskid to be a string
-        proc = subprocess.Popen(["qstat", str(taskid)],
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        _, e = proc.communicate()
+        result = CommandLine('qstat {}'.format(taskid),
+                             environ=dict(os.environ),
+                             terminal_output='allatonce',
+                             ignore_exception=True).run()
         errmsg = 'Unknown Job Id'  # %s' % taskid
-        return errmsg not in e
+        return errmsg not in result.runtime.stderr
 
     def _submit_batchtask(self, scriptfile, node):
         cmd = CommandLine('qsub', environ=dict(os.environ),


gjcooper:
Additionally,

I have noticed other places in the codebase that perhaps would suffer from the same bug. I could attempt fixes for these Popen calls (via a similar method) however I wouldn't have the capability to test the other batch submission systems.

In particular I have noticed there there are possible similar problems in:
* `_is_pending` in the `oar` plugin module
* `_qacct_verified_complete` and `_run_qstat` in the `sge` plugin module

There are some other places, but I would be less confident adjusting those to use CommandLine.

nipype/pipeline/plugins/pbs.py
```diff
@@ -46,13 +45,12 @@ def __init__(self, **kwargs):
...
 
     def _is_pending(self, taskid):
-        #  subprocess.Popen requires taskid to be a string
-        proc = subprocess.Popen(["qstat", str(taskid)],
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        _, e = proc.communicate()
+        result = CommandLine('qstat {}'.format(taskid),
+                             environ=dict(os.environ),
+                             terminal_output='allatonce',
```
oesteban: I'm not sure `'allatonce'` generates both `stdout` and `stderr` in the runtime object. Have you checked?

nipype/pipeline/plugins/pbs.py
```diff
@@ -46,13 +45,12 @@ def __init__(self, **kwargs):
         super(PBSPlugin, self).__init__(template, **kwargs)
 
     def _is_pending(self, taskid):
-        #  subprocess.Popen requires taskid to be a string
-        proc = subprocess.Popen(["qstat", str(taskid)],
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        _, e = proc.communicate()
+        result = CommandLine('qstat {}'.format(taskid),
+                             environ=dict(os.environ),
```
oesteban: I think you can remove this line if you want to use the default environment

nipype/pipeline/plugins/pbs.py
satra: `allatonce` doesn't stream but still separates `stdout` and `stderr`

satra:
@gjcooper - i'll merge this, but it would be good to clean up the other places where you spot this.

PR #1991, opened 2017-05-02
ENH: Update multi-stage recon-all directives

effigies:
The goal of this PR is to give access to most* [multi-stage flags](https://github.com/freesurfer/freesurfer/blob/3e316ab/scripts/recon-all#L6641-L6938), in particular the `autorecon2-volonly` and `autorecon-hemi` flags.

At present, estimating thread consumption is difficult, and `recon-all -parallel -openmp N` can consume up to 2N threads at OpenMP-enabled sections of the surface processing sections of `autorecon2` and `autorecon3`. This change allows `autorecon2-volonly` to be tagged as consuming N threads, and (for instance) two invocations of `autorecon-hemi` to each be tagged as consuming N threads.

This also updates the docstring to document and test the more recent commandline altering changes.

--- 
\* The one that's missing is `gcatrain-iter`, which takes 2 arguments and would thus be fiddlier to deal with. (I'm not really sure why `gcareg-iters` and `gcareg-tol` are included in that section; but I wouldn't call them directives.) I can put it in for completeness sake, but given that I'm not even sure how to test it, that feels like an invitation to write a bug.

### DIFF
--- a/nipype/interfaces/freesurfer/preprocess.py
+++ b/nipype/interfaces/freesurfer/preprocess.py
@@ -611,10 +611,18 @@ class Resample(FSCommand):
 class ReconAllInputSpec(CommandLineInputSpec):
     subject_id = traits.Str("recon_all", argstr='-subjid %s',
                             desc='subject name', usedefault=True)
-    directive = traits.Enum('all', 'autorecon1', 'autorecon2', 'autorecon2-cp',
-                            'autorecon2-wm', 'autorecon2-inflate1',
-                            'autorecon2-perhemi', 'autorecon3', 'localGI',
-                            'qcache', argstr='-%s', desc='process directive',
+    directive = traits.Enum('all', 'autorecon1',
+                            # autorecon2 variants
+                            'autorecon2', 'autorecon2-volonly',
+                            'autorecon2-perhemi', 'autorecon2-inflate1',
+                            'autorecon2-cp', 'autorecon2-wm',
+                            # autorecon3 variants
+                            'autorecon3', 'autorecon3-T2pial',
+                            # Mix of autorecon2 and autorecon3 steps
+                            'autorecon-pial', 'autorecon-hemi',
+                            # Not "multi-stage flags"
+                            'localGI', 'qcache',
+                            argstr='-%s', desc='process directive',
                             usedefault=True, position=0)
     hemi = traits.Enum('lh', 'rh', desc='hemisphere to process',
                        argstr="-hemi %s")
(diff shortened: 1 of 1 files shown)


nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -950,14 +989,24 @@ def _is_resuming(self):
...
         if name == 'hippocampal_subfields_T1' and \
                 isdefined(self.inputs.hippocampal_subfields_T2):
-            return ''
+            return None
         if all((name == 'hippocampal_subfields_T2',
                 isdefined(self.inputs.hippocampal_subfields_T1) and
                 self.inputs.hippocampal_subfields_T1)):
-            trait_spec.argstr = trait_spec.argstr.replace('T2', 'T1T2')
+            argstr = trait_spec.argstr.replace('T2', 'T1T2')
+            return argstr % value
```
effigies: This part here was a (minor) bug, in that `trait_spec` is non-ephemeral, so changing `trait_spec.argstr` changes it for all future calls. Unlikely to affect anything except the doctests, but still worth cleaning up.

nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -950,14 +989,24 @@ def _is_resuming(self):
     def _format_arg(self, name, trait_spec, value):
         if name == 'T1_files':
             if self._is_resuming():
-                return ''
+                return None
```
effigies: Switched to returning `None` just because that avoids the extra spaces in the `cmdline` string, making the doctests less painful to debug.

effigies:
Marking this [WIP] until I verify that we don't need to rework the `_steps` nonsense to fit these new directives.

effigies:
Needed to update `_steps`.

effigies:
This is working as expected on my end.

nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -765,65 +805,51 @@ class ReconAll(CommandLine):
...
-                       'surf/lh.curv', 'surf/rh.curv',
-                       'surf/lh.area', 'surf/rh.area',
-                       'label/lh.cortex.label', 'label/rh.cortex.label'], []),
-            ('smooth2', ['surf/lh.smoothwm', 'surf/rh.smoothwm'], []),
-            ('inflate2', ['surf/lh.inflated', 'surf/rh.inflated',
-                          'surf/lh.sulc', 'surf/rh.sulc',
-                          'surf/lh.inflated.H', 'surf/rh.inflated.H',
-                          'surf/lh.inflated.K', 'surf/rh.inflated.K'], []),
+            ]
+        _autorecon2_lh_steps = [
```
satra: only a nitpick - is there a reason this is called `_lh_` as opposed to say `_hemi_`

otherwise looks reasonable to me.

effigies: Because I specify `lh` in the filenames. I then make a `rh` copy below, just swapping the filenames out.

PR #1993, opened 2017-05-03
WIP: Adding a python wrapper for ICA_AROMA

jdkent:
This is my attempt at adding ICA_AROMA's functionality to nipype, although it would probably be better just to incorporate the code from ICA_AROMA.py, I didn't want to take the code and then have it be maintained in two places, so I wrote a wrapper treating the python file as a commandline tool. I don't know the best place for commandline tools, but ICA_AROMA.py uses fsl tools so I placed it there.

Thanks for looking!


chrisfilo:
+1 for calling AROMA instead of copying the code

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+                      argstr='%.2f')
+    melodic_dir = Directory(exists=True,
+                            desc='path to MELODIC directory if MELODIC has already been ran',
+                            argstr='-meldir %s')
+    affmat = File(exists=True,
+                  desc='path/name of the mat-file describing the '
+                  'affine registration (e.g. FSL FLIRT) of the '
+                  'functional data to structural space (.mat file)',
+                  argstr='-affmat %s',mandatory=False,xor=['featDir'])
+    warp = File(exists=True,
```
effigies: I'd prefer `warp_file`. And possibly `fnirt_warp_file`, since each packages has its own warp format.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+                desc='File name of the warp-file describing '
+                'the non-linear registration (e.g. FSL FNIRT) '
+                'of the structural data to MNI152 space (.nii.gz)',
+                argstr='-warp %s',mandatory=False,xor=['featDir'])
+    mc = File(exists=True,
```
effigies: `mcparams` or `motion_parameters` might be more descriptive.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+              desc='motion parameters file',
+              argstr='-mc %s',mandatory=False,xor=['featDir'])
+    denoise_type = traits.Str(argstr='-den %s',
```
effigies: This should be an `Enum`, since we know all of the possible options.

```Python
denoise_type = traits.Enum('nonaggr', 'aggr', 'both', 'no', usedefault=True, desc='...')
```

I set `nonaggr` to the default here, since that seems consistent with the [source](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA.py#L42).

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+                              '-none: only classification, no denoising '
+                              '-nonaggr (default): non-aggresssive denoising, i.e. partial component regression '
+                              '-aggr: aggressive denoising, i.e. full component regression '
+                              '-both: both aggressive and non-aggressive denoising (two outputs)',
+                              mandatory=True)
+
+class ICA_AROMAOutputSpec(TraitedSpec):
+    out_file = OutputMultiPath(File(exists=True), 
+                               desc='if generated: 1-(non aggressive denoised volume),'
+                               '2-(aggressive denoised volume)')
```
effigies: This is a point where you can do the `OutputMultiPath` if you want, but I think having your original `aggr_denoised_file` and `nonaggr_denoised_file` (or similar) would be better, just because names are easier to deal with than indices.

For similar reasons, you'll also want anything that's in the `out_dir`. Looking through the source, I see:

* [`mask_file`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA.py#L164-L185)
* [`melodic_dir`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L35)
* [`melodic_ic_thr`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L119-L122) (I don't know enough about this file to have a more descriptive name)
* [`melodic_ic_mni`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA.py#L196) (same)
* [`feature_scores`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L482)
* [`classified_motion_components`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L485)
* [`classification_overview`](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L491)

You could also do `out_dir`, if that seems reasonable (I see [MELODIC does this](https://github.com/nipy/nipype/blob/e7e2c002133c99fac47aafa2d180bb524546c220/nipype/interfaces/fsl/model.py#L1552-L1554)).

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    >>> mc='/path/to/mcImg.par'
+    >>> mask='/path/to/mcImg_mask.nii.gz'
+    >>> denoise_type='both'
+    >>> AROMA_obj.inputs.infile=func
+    >>> AROMA_obj.inputs.affmat=affmat
+    >>> AROMA_obj.inputs.warp=warp
+    >>> AROMA_obj.inputs.mc=mc
+    >>> AROMA_obj.inputs.mask=mask
+    >>> AROMA_obj.inputs.denoise_type=denoise_type
+    >>> AROMA_obj.inputs.outdir=outDir
```
effigies: I wouldn't separate `AROMA_obj.inputs.x = y` from `y = 'somefile'`. It's extra lines without much value.

Also, these files need to exist. You could create a set of empty files in `nipype/testing/data/feat_dir` or something like that, and then use:

```Python
>>> AROMA_obj.inputs.in_file = 'feat_dir/mcImg_brain.nii.gz'
```

And so on.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    >>> AROMA_obj.cmdline
+    'ICA_AROMA.py -affmat /path/to/functoT1.mat -den both 
+    -i /path/to/mcImg_brain.nii.gz 
+    -m /path/to/mcImg_mask.nii.gz 
+    -mc /path/to/mcImg.par 
+    -o /path/to/ICA_AROMA_testout 
+    -warp /path/to/T1toMNI_warp.nii.gz'
+    >>> AROMA_obj.run()
```
effigies: Don't actually run this during the doctests. Do

```Python
>>> AROMA_obj.run()  # doctest: +SKIP
```

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+          outputs['out_file'] = os.path.join(outdir,'denoised_func_data_nonaggr.nii.gz')
+        elif denoising_strategy is "aggr":
+          outputs['out_file'] = os.path.join(outdir,'denoised_func_data_aggr.nii.gz')
+        elif denoising_strategy is "both":
+          outputs['out_file'] = (os.path.join(outdir,'denoised_func_data_nonaggr.nii.gz'), os.path.join(outdir,'denoised_func_data_aggr.nii.gz'))
+        elif denoising_strategy is "none":
+          print "No denoising selected"
+        else:
+          raise RuntimeError('denoise_type must be specified as one of'
+                             ' noaggr,aggr,both, or none')
```
effigies: This particular exception will be caught by the `Enum` trait, above.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+        denoising_strategy = input_spec.denoise_type
+
+        if denoising_strategy is "noaggr":
+          outputs['out_file'] = os.path.join(outdir,'denoised_func_data_nonaggr.nii.gz')
+        elif denoising_strategy is "aggr":
+          outputs['out_file'] = os.path.join(outdir,'denoised_func_data_aggr.nii.gz')
+        elif denoising_strategy is "both":
+          outputs['out_file'] = (os.path.join(outdir,'denoised_func_data_nonaggr.nii.gz'), os.path.join(outdir,'denoised_func_data_aggr.nii.gz'))
+        elif denoising_strategy is "none":
+          print "No denoising selected"
```
effigies: We support Python 3, so if you're going to print, use `print()`.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+        outputs = self.output_spec.get()
+        outdir = self.input_spec.outdir
+        denoising_strategy = input_spec.denoise_type
+
+        if denoising_strategy is "noaggr":
+          outputs['out_file'] = os.path.join(outdir,'denoised_func_data_nonaggr.nii.gz')
+        elif denoising_strategy is "aggr":
+          outputs['out_file'] = os.path.join(outdir,'denoised_func_data_aggr.nii.gz')
+        elif denoising_strategy is "both":
+          outputs['out_file'] = (os.path.join(outdir,'denoised_func_data_nonaggr.nii.gz'), os.path.join(outdir,'denoised_func_data_aggr.nii.gz'))
```
effigies: Assuming you split the output like I suggest above, you could do the following:

```Python
out_dir = os.path.abspath(self.inputs.out_dir)

if self.inputs.denoise_type in ('aggr', 'both'):
    outputs['aggr_denoised_file'] = os.path.join(out_dir, 'denoised_func_data_aggr.nii.gz')
if self.inputs.denoise_type in ('nonaggr', 'both'):
    outputs['nonaggr_denoised_file'] = os.path.join(out_dir, 'denoised_func_data_nonaggr.nii.gz')

File: nipype/interfaces/fsl/ICA_AROMA.py [hunk 10]
```diff
@@ -0,0 +1,115 @@
...
+    CommandLine,
+    File,
+    Directory,
+    traits,
+    OutputMultiPath
+)
+import os
+
+class ICA_AROMAInputSpec(CommandLineInputSpec):
+    featDir = Directory(exists=True,
```
jdkent: I haven't heard of flake8, that looks really useful, thanks!

argstr: D'oh, I made a couple of errors where I didn't specify the string correctly

I'll move 'desc' to the end and move 'mandatory' next to 'exists'. Question: if mandatory=False is the default behavior, should I still be explicit or  is that too wordy?

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+                      'the TR will be extracted from the header of the fMRI nifti file.',
+                      argstr='%.2f')
+    melodic_dir = Directory(exists=True,
+                            desc='path to MELODIC directory if MELODIC has already been ran',
+                            argstr='-meldir %s')
+    affmat = File(exists=True,
+                  desc='path/name of the mat-file describing the '
+                  'affine registration (e.g. FSL FLIRT) of the '
+                  'functional data to structural space (.mat file)',
+                  argstr='-affmat %s',mandatory=False,xor=['featDir'])
```
jdkent: I saw mat_file used in the fsl interface so I'll switch it to that

nipype/interfaces/fsl/ICA_AROMA.py
effigies: If it's non-mandatory, don't include it. But since you're in the situation where you have to either have `feat_dir` or `in_file`, then both should be mandatory, but `xor` one another. (The mandatory check is smart enough to figure out these required-but-mutually-exclusive cases.)

I think your motion correction params, affine transform and warp are all required (if not using `feat_dir`), so they should be set to mandatory as well, but the mask is generated by the program if not supplied, so it doesn't need to be set to mandatory.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    OutputMultiPath
+)
+import os
+
+class ICA_AROMAInputSpec(CommandLineInputSpec):
+    featDir = Directory(exists=True,
+                        desc='If a feat directory exists and temporal filtering '
+                        'has not been run yet, ICA_AROMA can use the files in '
+                        'this directory.',mandatory=False,xor=['infile','mask','affmat','warp','mc'])
+    infile = File(exists=True,
```
jdkent: adapted this change into updated script

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+
+class ICA_AROMAInputSpec(CommandLineInputSpec):
+    featDir = Directory(exists=True,
+                        desc='If a feat directory exists and temporal filtering '
+                        'has not been run yet, ICA_AROMA can use the files in '
+                        'this directory.',mandatory=False,xor=['infile','mask','affmat','warp','mc'])
+    infile = File(exists=True, 
+                  desc='volume to be denoised', 
+                  argstr='-i %s',mandatory=False,xor=['featDir'])
```
jdkent: done (I think).

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    outdir = Directory(desc='path to output directory', 
+                  argstr='-o %s',mandatory=True)
```
jdkent: your comment here helped me understand how nipype works a bit better, thanks!

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    mask = File(exists=True, 
+                desc='path/name volume mask', 
+                argstr='-m %s',mandatory=False,xor=['featDir'])
```
jdkent: done.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    dim = traits.Int(desc='Dimensionality reduction when running '
+                     'MELODIC (defualt is automatic estimation)', 
+                     argstr='-dim %d')
+    TR = traits.Float(desc='TR in seconds. If this is not specified '
+                      'the TR will be extracted from the header of the fMRI nifti file.',
+                      argstr='%.2f')
```
jdkent: I agree, I added another decimal to the float.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,115 @@
...
+    melodic_dir = Directory(exists=True,
+                            desc='path to MELODIC directory if MELODIC has already been ran',
```
jdkent: oops, thanks for catching that!

nipype/interfaces/fsl/ICA_AROMA.py
jdkent: chose `fnirt_warp_file`

nipype/interfaces/fsl/ICA_AROMA.py
jdkent: chose `motion_parameters`

jdkent:
@effigies : I believe I responded to all of your edits, thanks for the detailed feedback and suggestions, it made it that much easier to change the code.

effigies:
@jdkent Thanks, it looks much improved! I'll have a more detailed look after the tests run. (Possibly in the morning. There's a bit of a test backlog.)

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+    >>> AROMA_obj = ICA_AROMA.ICA_AROMA()
+    >>> AROMA_obj.inputs.in_file=example_data('functional.nii')
+    >>> AROMA_obj.inputs.mat_file=example_data('func_to_struct.mat')
+    >>> AROMA_obj.inputs.fnirt_warp_file=example_data('warpfield.nii')
+    >>> AROMA_obj.inputs.motion_parameters=example_data('functional.par')
+    >>> AROMA_obj.inputs.mask=example_data('mask.nii.gz')
+    >>> AROMA_obj.inputs.denoise_type='both'
+    >>> AROMA_obj.inputs.out_dir='ICA_testout'
+    >>> AROMA_obj.cmdline
+    u'ICA_AROMA.py -den both -warp /home/travis/build/nipy/nipype/nipype/testing/data/warpfield.nii -i /home/travis/build/nipy/nipype/nipype/testing/data/functional.nii -m /home/travis/build/nipy/nipype/nipype/testing/data/mask.nii.gz -affmat /home/travis/build/nipy/nipype/nipype/testing/data/func_to_struct.mat -mc /home/travis/build/nipy/nipype/nipype/testing/data/functional.par -o ICA_testout'
```
effigies: You don't need `example_data()`, which adds the full path, for doctests.

How about this:

```Python
>>> AROMA_obj.inputs.in_file = 'functional.nii'
>>> AROMA_obj.inputs.mat_file = 'func_to_struct.mat'
>>> AROMA_obj.inputs.fnirt_warp_file = 'warpfield.nii'
>>> AROMA_obj.inputs.motion_parameters = 'fsl_mcflirt_movpar.txt'
>>> AROMA_obj.inputs.mask = 'mask.nii.gz'
>>> AROMA_obj.inputs.denoise_type = 'both'
>>> AROMA_obj.inputs.out_dir='ICA_testout'
>>> AROMA_obj.cmdline # doctest: +ALLOW_UNICODE
'ICA_AROMA.py -den both -warp warpfield.nii -i functional.nii -m mask.nii.gz \
-affmat func_to_struct.mat -mc fsl_mcflirt_movpar.txt -o ICA_testout'
```

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+
+    """
+    _cmd = 'ICA_AROMA.py'
+    input_spec = ICA_AROMAInputSpec
+    output_spec = ICA_AROMAOutputSpec
+
+    def _list_outputs(self):
+
```
effigies: No blank line here.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+        out_dir = os.path.abspath(self.inputs.out_dir)
+        outputs['out_dir'] = out_dir
+        #outputs = self.output_spec.get()
+        #outdir = self.input_spec.outdir
+        #denoising_strategy = input_spec.denoise_type
```
effigies: Let's drop these comments.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if self.inputs.denoise_type in ('aggr', 'both'):
+            outputs['aggr_denoised_file'] = os.path.join(out_dir, 'denoised_func_data_aggr.nii.gz')
+        if self.inputs.denoise_type in ('nonaggr', 'both'):
+            outputs['nonaggr_denoised_file'] = os.path.join(out_dir, 'denoised_func_data_nonaggr.nii.gz')
+        
+        return outputs
```
effigies: Add a newline at the end of file.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+    nonaggr_denoised_file=File(exists=True,
+                              desc='if generated: non aggressively denoised volume' )
+    out_dir = Directory(exists=True, 
+                       desc='directory contains (in addition to the denoised files): '
+                       'melodic.ica + classified_motion_components + '
+                       'classification_overview + feature_scores + melodic_ic_mni)')
+   
+class ICA_AROMA(CommandLine):
+    """
+
```
effigies: Should describe what this interface does and link to external documentation. Have a look at some other interfaces for examples. You can largely copy this from any existing ICA_AROMA documentation.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+    File,
+    Directory,
+    traits,
+    OutputMultiPath
+)
+import os
+
+class ICA_AROMAInputSpec(CommandLineInputSpec):
+    feat_dir = Directory(exists=True, mandatory=True,
+                        argstr='-feat %s',
```
effigies: These should be shifted right by one space (aligned with the inside of the parentheses).

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+    CommandLine,
+    File,
+    Directory,
+    traits,
+    OutputMultiPath
+)
+import os
+
+class ICA_AROMAInputSpec(CommandLineInputSpec):
+    feat_dir = Directory(exists=True, mandatory=True,
```
chrisfilo: I don't think feat_dir is mandatory in ICA AROMA.

effigies: `feat_dir` or `in_file` and friends. `xor` handles competing `mandatory=True` traits.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,106 @@
...
+                     'the TR will be extracted from the '
+                     'header of the fMRI nifti file.')
+    melodic_dir = Directory(exists=True, argstr='-meldir %s',
+                           desc='path to MELODIC directory if MELODIC has already been run')
+    mat_file = File(exists=True, mandatory=True,
+                   argstr='-affmat %s', xor=['feat_dir'],
+                   desc='path/name of the mat-file describing the '
+                   'affine registration (e.g. FSL FLIRT) of the '
+                   'functional data to structural space (.mat file)')
+    fnirt_warp_file = File(exists=True, mandatory=True,
```
chrisfilo: This also does not seem to be mandatory - ICA AROMA also supports inputs that are in the MNI space.

nipype/interfaces/fsl/ICA_AROMA.py
chrisfilo: Oh ok - sorry!

nipype/interfaces/fsl/ICA_AROMA.py
effigies: This may be true. I honestly haven't dug into their documentation to be sure what all is actually mandatory.

Here is their argparse setup: https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA.py#L27-L37

jdkent: good catch @chrisfilo, fnirt_warp_file and mat_file don't appear to be mandatory

ICA-AROMA checking inputs: https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA.py#L109-L116

effigies: I would actually check inside this function, to see what it does: https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA.py#L197

jdkent: got it: https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L148-L163

nipype/interfaces/fsl/ICA_AROMA.py
jdkent: thanks for this!

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,112 @@
...
+    >>> from nipype.testing import example_data
+    >>> AROMA_obj = ICA_AROMA.ICA_AROMA()
+    >>> AROMA_obj.inputs.in_file = 'functional.nii'
+    >>> AROMA_obj.inputs.mat_file = 'func_to_struct.mat'
+    >>> AROMA_obj.inputs.fnirt_warp_file = 'warpfield.nii'
+    >>> AROMA_obj.inputs.motion_parameters = 'functional.par'
+    >>> AROMA_obj.inputs.mask = 'mask.nii.gz'
+    >>> AROMA_obj.inputs.denoise_type = 'both'
+    >>> AROMA_obj.inputs.out_dir = 'ICA_testout'
+    >>> AROMA_obj.cmdline # doctest: +ALLOW_UNICODE
```
jdkent: How do I get rid of the u'string' output? (and I need to to change the motion parameter file)

effigies: The `u''` part is handled by the `+ALLOW_UNICODE` flag. The problem here is actually that there's a tab after the newline. You can either remove the spaces before `-affmat` or just remove the newline entirely and just have the whole string on one line.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,112 @@
...
+                               '-none: only classification, no denoising '
+                               '-nonaggr (default): non-aggresssive denoising, i.e. partial component regression '
+                               '-aggr: aggressive denoising, i.e. full component regression '
+                               '-both: both aggressive and non-aggressive denoising (two outputs)')
+
+class ICA_AROMAOutputSpec(TraitedSpec):
+    aggr_denoised_file=File(exists=True, 
+                           desc='if generated: aggressively denoised volume')
+    nonaggr_denoised_file=File(exists=True,
+                              desc='if generated: non aggressively denoised volume' )
```
effigies: For both of these traits, add spaces around `=`, and align your `desc` with `exists`.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,112 @@
...
+    out_dir = Directory(exists=True, 
+                        desc='directory contains (in addition to the denoised files): '
+                        'melodic.ica + classified_motion_components + '
+                        'classification_overview + feature_scores + melodic_ic_mni)')
+   
+class ICA_AROMA(CommandLine):
+    """
+    Interface for the ICA_AROMA.py script (v0.3 beta).
```
effigies: I'd remove the version number. Unless the interface changes, we don't really need to tag versions, since it'll cause confusion about whether the interface is still valid for future versions. (If it does, then we'll figure out how to make it work with multiple versions.)

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,112 @@
...
+    a data-driven method to identify and remove motion-related independent
+    components from fMRI data. To that end it exploits a small, but robust
+    set of theoretically motivated features, preventing the need for classifier
+    re-training and therefore providing direct and easy applicability. 
+
+
+    See link for further documentation: https://github.com/rhr-pruim/ICA-AROMA
+
+
+
```
effigies: Limit blank lines (here and above the `See link...` line) to one.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,108 @@
...
+    motion_parameters = File(exists=True, mandatory=True,
+                             argstr='-mc %s', xor=['feat_dir'],
+                             desc='motion parameters file')
+    denoise_type = traits.Enum('nonaggr', 'aggr', 'both', 'no', usedefault=True,
+                               mandatory=True, argstr='-den %s',
+                               desc='Type of denoising strategy: '
+                               '-none: only classification, no denoising '
+                               '-nonaggr (default): non-aggresssive denoising, i.e. partial component regression '
+                               '-aggr: aggressive denoising, i.e. full component regression '
+                               '-both: both aggressive and non-aggressive denoising (two outputs)')
```
effigies: This `desc` string is:

```
Type of denoising strategy: -none: only classification, no denoising -nonaggr (default): non-aggresssive denoising, i.e. partial component regression -aggr: aggressive denoising, i.e. full component regression -both: both aggressive and non-aggressive denoising (two outputs)
```

Is this correct, or do you want it to be:

```
Type of denoising strategy:
-none: only classification, no denoising
-nonaggr (default): non-aggresssive denoising, i.e. partial component regression
-aggr: aggressive denoising, i.e. full component regression
-both: both aggressive and non-aggressive denoising (two outputs)
```

If the latter, replace the space at the end of each line with `\n`. If the former is what you intended, no change necessary.

jdkent: I think I want the latter, thanks! I wasn't sure if newlines were interpreted within single quotes or if I had to go to double quotes (because that makes a difference in bash, I didn't know in python).

effigies: In Python, they're the same, you just use whichever looks nicer to you. The only difference is that with double-quotes you have to escape double-quotes and with single-quotes you have to escape single-quotes. (My preference is to use single quotes unless the string contains a single quote, but there is no enforced rule in nipype.)

effigies:
LGTM. Any final comments, @chrisfilo?

jdkent:
I'm afraid I may be in over my head interpreting what I need to change to make this code pass the ci/circleci test. Does this have to do with modifying the docker files?

effigies:
I think it's a random failure. I'll restart the job when I can. Might be in
the morning.

effigies:
@jdkent Can you merge from master? The tests should be fixed.

effigies:
This looks like it should have passed. @satra did something change with the doctests? Circle is complaining about a missing `functional.nii`, but [nipype/testing/data/functional.nii](https://github.com/nipy/nipype/blob/master/nipype/testing/data/functional.nii) exists.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,108 @@
...
+    set of theoretically motivated features, preventing the need for classifier
+    re-training and therefore providing direct and easy applicability. 
+
+    See link for further documentation: https://github.com/rhr-pruim/ICA-AROMA
+
+    Example
+    -------
+
+    >>> from nipype.interfaces.fsl import ICA_AROMA 
+    >>> from nipype.testing import example_data
```
effigies: You can remove the `example_data` import. Shouldn't have any effect on the tests, though.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -0,0 +1,108 @@
+#ICA_AROMA pulled from: https://github.com/rhr-pruim/ICA-AROMA
+#This assumes ICA_AROMA.py is already installed and callable via $PATH
```
satra: @effigies, @jdkent - we are missing the standard header that imports the testing paths for doctest. see any other interface file in nipype.

jdkent: Thanks! Sorry, I'm new to all this.

jdkent:
The last line having trouble in the Travis CI build says this:
`nipype/pipeline/engine/tests/test_utils.py::test_mapnode_crash3 /home/travis/.travis/job_stages: line 53: 12967 Segmentation fault      (core dumped) py.test -v --doctest-modules nipype`

Would this be anything related to how I have ICA_AROMA formatted?

mgxd:
@jdkent that's a very specific and not always reproducible bug we've been encountering in our test builds, unrelated to ICA_AROMA. I've restarted those specific builds

effigies:
@mgxd I've restarted them multiple times. The bug may be getting more reliable...

effigies:
Not going to hold this one up for the segfault bug. Merging.

PR #1997, opened 2017-05-05
doc: fix rst formatting

satra:


effigies:
I take it this is meant to resolve the issue that's causing a ll Circle builds to fail?

satra:
@effigies - yes. so we should merge this once tests pass.

however, i snuck in one change. `requires` no longer raises warnings.

effigies:
Cancelled the previous build to move this one up in the queue (and because that one was going to fail anyway). If you want to merge master/rebase now, then we'll get test results on the latest merges, too.

satra:
current master is merged in here.

effigies:
So it is.

PR #1999, opened 2017-05-06
remove bug to have a temporal derivative for multiple variables

jokedurnez:
`ev_parameters.pop('derivs')` removes `'derivs'` from the `ev_parameters`.  As a consequence, only for the first regressor, a temporal derivative is added.  This fixes the issue.

### DIFF
--- a/nipype/interfaces/fsl/model.py
+++ b/nipype/interfaces/fsl/model.py
@@ -209,9 +209,9 @@ class Level1Design(BaseInterface):
                             raise Exception('FSL environment variables not set')
                         else:
                             ev_parameters['fsldir'] = '/usr/share/fsl'
-                    try:
-                        ev_parameters['temporalderiv'] = int(bool(ev_parameters.pop('derivs')))
-                    except KeyError:
+                    if 'derivs' in ev_parameters.keys():
+                        ev_parameters['temporalderiv'] = int(bool(ev_parameters['derivs']))
+                    else:
                         ev_parameters['temporalderiv'] = False
                     if ev_parameters['temporalderiv']:
                         evname.append(name + 'TD')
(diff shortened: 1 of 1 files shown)


satra:
@jokedurnez - looks reasonable 

@TheChymera - could you please take a look?

nipype/interfaces/fsl/model.py
```diff
@@ -209,9 +209,9 @@ def _create_ev_files(
                             raise Exception('FSL environment variables not set')
                         else:
                             ev_parameters['fsldir'] = '/usr/share/fsl'
-                    try:
-                        ev_parameters['temporalderiv'] = int(bool(ev_parameters.pop('derivs')))
-                    except KeyError:
+                    if 'derivs' in ev_parameters.keys():
```
oesteban: I think this would be more reliable.
```
ev_parameters['temporalderiv'] = bool(ev_parameters.get('derivs', False))
```

BTW, I don't understand why the int casting if `ev_parameters['derivs']` is `True` and then the boolean is left for false.

nipype/interfaces/fsl/model.py
```diff
@@ -393,9 +393,9 @@ def _list_outputs(self):
                         cwd, 'ev_%s_%d_%d.txt' % (name, runno,
                                                   len(evname)))
                     if field == 'cond':
-                        try:
-                            ev_parameters['temporalderiv'] = int(bool(ev_parameters.pop('derivs')))
-                        except KeyError:
+                        if 'derivs' in ev_parameters.keys():
```
oesteban:
```
ev_parameters['temporalderiv'] = int(bool(ev_parameters.get('derivs', False)))
```

TheChymera:
I agree that `ev_parameters['temporalderiv'] = bool(ev_parameters.get('derivs', False))` is the best way to go.

Also, I recommend you fix the `.tcl` file in a separate PR. Perhaps also linking some documentation for the `fmri(inputtype)` variable. I'm really unsure what switching it effects.

jokedurnez:
Oh sorry about the tcl file.  I actually have submitted another PR for this.
I'm not going to write documentation.  Switching this variable to folders searches for lower level feat directories.  Given that this is the template for the first level, it makes no sense to make it into a variable.

jokedurnez:
Ok, I've removed the tcl file from the pull request and added the suggested changes.

Issue #2004, opened 2017-05-10
niftyreg interfaces failing because of base and OMP_NUM_THREADS


byvernault:
### Summary

I tried to install the master version of nipype and run our workflows using niftyreg and niftyseg. It looks like the node using niftyreg/niftyseg are not working because of the code to set the OMP_NUM_THREADS in the _run_interface.

Right now, the code is the following:

if not isdefined(self.inputs.environ['OMP_NUM_THREADS']):
            self.inputs.environ['OMP_NUM_THREADS'] = self.num_threads

But the self.inputs.environ is not populated from the class CommandLine giving us the error  below.
 
I could fix it by changing to this:

if 'OMP_NUM_THREADS' not in os.environ:
            os.environ['OMP_NUM_THREADS'] = str(self.num_threads)

or do you recommend something else? Where should self.inputs.environ be set?

### Actual behavior

Error when running a niftyreg node:

170510-11:56:37,203 workflow INFO:
[2017-05-10 11:56:37] 	 Traceback (most recent call last):
[2017-05-10 11:56:37]   File "/Users/byvernault/home-local/code/anaconda/lib/python2.7/site-packages/nipype/pipeline/plugins/linear.py", line 43, in run
[2017-05-10 11:56:37]     node.run(updatehash=updatehash)
[2017-05-10 11:56:37]   File "/Users/byvernault/home-local/code/anaconda/lib/python2.7/site-packages/nipype/pipeline/engine/nodes.py", line 372, in run
[2017-05-10 11:56:37]     self._run_interface()
[2017-05-10 11:56:37]   File "/Users/byvernault/home-local/code/anaconda/lib/python2.7/site-packages/nipype/pipeline/engine/nodes.py", line 1287, in _run_interface
[2017-05-10 11:56:37]     updatehash=updatehash))
[2017-05-10 11:56:37]   File "/Users/byvernault/home-local/code/anaconda/lib/python2.7/site-packages/nipype/pipeline/engine/nodes.py", line 1193, in _collate_results
[2017-05-10 11:56:37]     (self.name, '\n'.join(msg)))
[2017-05-10 11:56:37] Exception: Subnodes of node: register2GAD failed:
[2017-05-10 11:56:37] Subnode 0 failed
[2017-05-10 11:56:37] Error:
[2017-05-10 11:56:37] u'OMP_NUM_THREADS\nInterface RegAladin failed to run. '

### How to replicate the behavior

Run an interface using niftyreg like follow:

[...]

PR #2005, opened 2017-05-10
Fixing the base interface _run_interface function for NiftyRegCommand

byvernault:
Dear Nipype Team,

In our previous release of the niftyreg interfaces, int the base interface, we added some line of code to set at least by default the number of thread use by niftyreg tools.

I just realized this morning that the code I used isn't correct since I am not checking if the key exists in the dictionary. I edited with @effigies advice from the issue #2004 . 

Basically, I am setting the self.inputs.environ['OMP_NUM_THREADS'] to the environ variable if it exists, else to the num_threads if not found (default 1).

I ran a node using the interface RegAladin and it worked. I tested as well one interface for NiftySeg and it worked as well.

Let me know if you think I should change something from those few lines.

Kind Regards,

### DIFF
--- a/nipype/interfaces/niftyreg/base.py
+++ b/nipype/interfaces/niftyreg/base.py
@@ -26,7 +26,7 @@ import shutil
 import subprocess
 from warnings import warn
 
-from ..base import CommandLine, isdefined, CommandLineInputSpec, traits
+from ..base import CommandLine, CommandLineInputSpec, traits, Undefined
 from ...utils.filemanip import split_filename
 
 
@@ -47,8 +47,9 @@ def no_nifty_package(cmd='reg_f3d'):
 class NiftyRegCommandInputSpec(CommandLineInputSpec):
     """Input Spec for niftyreg interfaces."""
     # Set the number of omp thread to use
-    omp_core_val = traits.Int(desc='Number of openmp thread to use',
-                              argstr='-omp %i')
+    omp_core_val = traits.Int(int(os.environ.get('OMP_NUM_THREADS', '1')),
+                              desc='Number of openmp thread to use',
+                              argstr='-omp %i', usedefault=True)
 
 
 class NiftyRegCommand(CommandLine):
@@ -58,7 +59,10 @@ class NiftyRegCommand(CommandLine):
     _suffix = '_nr'
     _min_version = '1.5.30'
 
+    input_spec = NiftyRegCommandInputSpec
+
     def __init__(self, required_version=None, **inputs):
+        self.num_threads = 1
         super(NiftyRegCommand, self).__init__(**inputs)
         self.required_version = required_version
         _version = self.get_version()
@@ -73,6 +77,29 @@ class NiftyRegCommand(CommandLine):
                     msg = 'The version of NiftyReg differs from the required'
                     msg += '(%s != %s)'
                     warn(msg % (_version, self.required_version))
+        self.inputs.on_trait_change(self._omp_update, 'omp_core_val')
+        self.inputs.on_trait_change(self._environ_update, 'environ')
+        self._omp_update()
+
+    def _omp_update(self):
+        if self.inputs.omp_core_val:
+            self.inputs.environ['OMP_NUM_THREADS'] = \
+                str(self.inputs.omp_core_val)
+            self.num_threads = self.inputs.omp_core_val
+        else:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                del self.inputs.environ['OMP_NUM_THREADS']
+            self.num_threads = 1
+
+    def _environ_update(self):
+        if self.inputs.environ:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                self.inputs.omp_core_val = \
+                    int(self.inputs.environ['OMP_NUM_THREADS'])
+            else:
+                self.inputs.omp_core_val = Undefined
+        else:
+            self.inputs.omp_core_val = Undefined
 
     def check_version(self):
         _version = self.get_version()
@@ -101,16 +128,6 @@ class NiftyRegCommand(CommandLine):
 
     def exists(self):
         return self.get_version() is not None
-
-    def _run_interface(self, runtime):
-        # Update num threads estimate from OMP_NUM_THREADS env var
-        # Default to 1 if not set
-        omp_key = 'OMP_NUM_THREADS'
-        if omp_key in os.environ:
-            self.inputs.environ[omp_key] = os.environ[omp_key]
-        else:
-            self.inputs.environ[omp_key] = str(self.num_threads)
-        return super(NiftyRegCommand, self)._run_interface(runtime)
 
     def _format_arg(self, name, spec, value):
         if name == 'omp_core_val':
--- a/nipype/interfaces/niftyreg/base.py
+++ b/nipype/interfaces/niftyreg/base.py
@@ -105,8 +105,11 @@ class NiftyRegCommand(CommandLine):
     def _run_interface(self, runtime):
         # Update num threads estimate from OMP_NUM_THREADS env var
         # Default to 1 if not set
-        if not isdefined(self.inputs.environ['OMP_NUM_THREADS']):
-            self.inputs.environ['OMP_NUM_THREADS'] = self.num_threads
+        omp_key = 'OMP_NUM_THREADS'
+        if omp_key in os.environ:
+            self.inputs.environ[omp_key] = os.environ[omp_key]
+        else:
+            self.inputs.environ[omp_key] = str(self.num_threads)
         return super(NiftyRegCommand, self)._run_interface(runtime)
 
     def _format_arg(self, name, spec, value):
--- a/nipype/interfaces/niftyreg/tests/test_auto_NiftyRegCommand.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_NiftyRegCommand.py
@@ -12,6 +12,9 @@ def test_NiftyRegCommand_inputs():
     ignore_exception=dict(nohash=True,
     usedefault=True,
     ),
+    omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
+    ),
     terminal_output=dict(nohash=True,
     ),
     )
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegAladin.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegAladin.py
@@ -46,6 +46,7 @@ def test_RegAladin_inputs():
     nosym_flag=dict(argstr='-noSym',
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     platform_val=dict(argstr='-platf %i',
     ),
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegAverage.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegAverage.py
@@ -43,6 +43,7 @@ def test_RegAverage_inputs():
     usedefault=True,
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     out_file=dict(argstr='%s',
     genfile=True,
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegF3D.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegF3D.py
@@ -76,6 +76,7 @@ def test_RegF3D_inputs():
     noz_flag=dict(argstr='-noz',
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     pad_val=dict(argstr='-pad %f',
     ),
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegJacobian.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegJacobian.py
@@ -13,6 +13,7 @@ def test_RegJacobian_inputs():
     usedefault=True,
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     out_file=dict(argstr='%s',
     name_source=['trans_file'],
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegMeasure.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegMeasure.py
@@ -19,6 +19,7 @@ def test_RegMeasure_inputs():
     mandatory=True,
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     out_file=dict(argstr='-out %s',
     name_source=['flo_file'],
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegResample.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegResample.py
@@ -18,6 +18,7 @@ def test_RegResample_inputs():
     inter_val=dict(argstr='-inter %d',
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     out_file=dict(argstr='%s',
     name_source=['flo_file'],
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegTools.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegTools.py
@@ -34,6 +34,7 @@ def test_RegTools_inputs():
     noscl_flag=dict(argstr='-noscl',
     ),
     omp_core_val=dict(argstr='-omp %i',
+    usedefault=True,
     ),
     out_file=dict(argstr='-out %s',
     name_source=['in_file'],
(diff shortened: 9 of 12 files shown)


nipype/interfaces/niftyreg/base.py
```diff
@@ -105,8 +105,11 @@ def exists(self):
     def _run_interface(self, runtime):
         # Update num threads estimate from OMP_NUM_THREADS env var
         # Default to 1 if not set
-        if not isdefined(self.inputs.environ['OMP_NUM_THREADS']):
-            self.inputs.environ['OMP_NUM_THREADS'] = self.num_threads
+        omp_key = 'OMP_NUM_THREADS'
+        if omp_key in os.environ:
```
oesteban: I'd suggest this instead the `if .. else`:
`self.inputs.environ[omp_key] = os.getenv(omp_key, '%d' % self.num_threads)`

chrisfilo:
This PR is probably not the right place for it, but the `OMP_NUM_THREADS` variable also controls threads in AFNI commands. We should abstract this away as a separate class (say `OpenMPInterface` and `OpenMPInterfaceInputSpec`) and use multi inheritance.

oesteban:
I was coming back to this and, correct me if I'm wrong, but this PR does not actually fix the issue:

If `OMP_NUM_THREADS` is defined in the environment (e.g. you run `export OMP_NUM_THREADS=4` before nipype), then nipype will automatically load that in the interface environment.

So it is not really necessary to make sure of setting it again, it will be already in the environment of the call to subprocess.

Then, I've noticed several issues in the code (sorry I reviewed the PR too quickly before):

- `self.numthreads` vs. `self.num_threads` (https://github.com/mmodat/nipype/blob/fcae4f71a5f208ad51215c322ca5e6ac3af99fd3/nipype/interfaces/niftyreg/base.py#L117 and https://github.com/mmodat/nipype/blob/fcae4f71a5f208ad51215c322ca5e6ac3af99fd3/nipype/interfaces/niftyreg/base.py#L112). Additionally, since `self.num[_]?threads` is not set in the `__init__`, the use of this member is very risky. Please use `setattr`/`getattr`.

- `self.numthreads` is not necessary. Why not just:
```Python
    if isdefined(self.inputs.omp_core_val):
        self.inputs.environ['OMP_NUM_THREADS'] = '%d' % self.input.omp_core_val
    return super(NiftyRegCommand, self)._run_interface(runtime)
```
Then, the format_arg function would go away.

If OMP_NUM_THREADS is not specified (not defined in the environment and omp_core_val is not defined), then I think nifty* should handle that situation. (meaning, I don't think it is a good idea to define a default value here).

oesteban:
@effigies just reminded me that BaseInterface has a `num_threads` property (which will be used in MultiProc to see if there are free cpus for an upcoming node). You can set it using `self.num_threads` safely. Something like:

```Python
    if isdefined(self.inputs.omp_core_val):
        self.inputs.environ['OMP_NUM_THREADS'] = '%d' % self.input.omp_core_val

    if self.inputs.environ.get('OMP_NUM_THREADS') is not None:
        self.num_threads =  int(self.inputs.environ.get('OMP_NUM_THREADS'))
    return super(NiftyRegCommand, self)._run_interface(runtime)
```

satra:
@oesteban - i don't think we should modify `self.inputs.environ` inside `_run_interface`. inside that function `runtime.environ` should be used. 

also self.num_threads should be set whenever `self.input.omp_core_val` is set, which is why i suggested adopting things in afni.

when the multiproc plugin executes, it will want to know how many threads to use, `_run_interface` is too late to set that value.

essentially here are the scenarios:
- nothing set in interface, nothing in environment, everything defaults to 1 thread
- nothing in interface, `OMP_NUM_THREADS` in environment, interface to use `OMP_NUM_THREADS`
- `omp_core_val` is set, should override `runtime.environ`. since the nifti tools can invoke this internally, the safest option would be to unset `OMP_NUM_THREADS` if it exists in environment.

byvernault:
hmm so how do you want me to proceed?

I can also remove the option from the interface and let nipype take care of this and the nifty executables are going to use the OMP_NUM_THREADS value from the env variables. 

WDYT?

oesteban:
As @chrisfilo said and @satra confirmed, I think we should give this a general solution.

satra:
@oesteban - i really would like to release 0.13 before the general solution. can you see if the solution below works?

@byvernault - so for these interfaces i would suggest the following, and remove overloading `_run_interface`:

```
diff --git a/nipype/interfaces/niftyreg/base.py b/nipype/interfaces/niftyreg/base.py
index a7672166f..0c7e22d60 100644
--- a/nipype/interfaces/niftyreg/base.py
+++ b/nipype/interfaces/niftyreg/base.py
@@ -47,8 +47,8 @@ def no_nifty_package(cmd='reg_f3d'):
 class NiftyRegCommandInputSpec(CommandLineInputSpec):
     """Input Spec for niftyreg interfaces."""
     # Set the number of omp thread to use
-    omp_core_val = traits.Int(desc='Number of openmp thread to use',
-                              argstr='-omp %i')
+    omp_core_val = traits.Int(1, desc='Number of openmp thread to use',
+                              argstr='-omp %i', use_default=True)
 
 
 class NiftyRegCommand(CommandLine):
@@ -73,6 +73,15 @@ class NiftyRegCommand(CommandLine):
                     msg = 'The version of NiftyReg differs from the required'
                     msg += '(%s != %s)'
                     warn(msg % (_version, self.required_version))
+        self.num_threads = 1
+        self.inputs.on_trait_change(self._omp_update, 'omp_core_val')
+
+    def _omp_update(self):
+        if self.inputs.omp_core_val:
+            self.inputs.environ = {'OMP_NUM_THREADS': self.inputs.omp_core_val}
+            self.num_threads = self.inputs.omp_core_val
+        else:
+            self.num_threads = 1
```

satra:
updated the `usedefault` keyword.

effigies:
@satra I could be wrong, but I think your solution will always set `num_threads = 1` unless it's modified after instantiation (so `Interface(omp_core_val=2)` would not actually set `num_threads` to 2).

```Diff
diff --git a/nipype/interfaces/niftyreg/base.py b/nipype/interfaces/niftyreg/base.py
index a7672166f..0c7e22d60 100644
--- a/nipype/interfaces/niftyreg/base.py
+++ b/nipype/interfaces/niftyreg/base.py
@@ -47,8 +47,8 @@ def no_nifty_package(cmd='reg_f3d'):
 class NiftyRegCommandInputSpec(CommandLineInputSpec):
     """Input Spec for niftyreg interfaces."""
     # Set the number of omp thread to use
-    omp_core_val = traits.Int(desc='Number of openmp thread to use',
-                              argstr='-omp %i')
+    omp_core_val = traits.Int(1, desc='Number of openmp thread to use',
+                              argstr='-omp %i', usedefault=True)
 
 
 class NiftyRegCommand(CommandLine):
@@ -73,6 +73,15 @@ class NiftyRegCommand(CommandLine):
                     msg = 'The version of NiftyReg differs from the required'
                     msg += '(%s != %s)'
                     warn(msg % (_version, self.required_version))
+        self._omp_update()
+        self.inputs.on_trait_change(self._omp_update, 'omp_core_val')
+
+    def _omp_update(self):
+        if self.inputs.omp_core_val:
+            self.inputs.environ['OMP_NUM_THREADS'] = self.inputs.omp_core_val
+            self.num_threads = self.inputs.omp_core_val
+        else:
+            self.num_threads = 1
```

Also modified your environment update line, as it currently clobbers any user-set `self.inputs.environ`.

WDYT? (And what should happen to `environ` in the `else` case?)

byvernault:
Sure, I will edit the code for this PR. 

I remember you wanted to do a release so I was trying to fix this so those interfaces are working well. 

Thank you for your help guys.

satra:
@effigies  and @byvernault - here is an update. if this passes muster, @byvernault - you should be able to patch this in:

```python
diff --git a/nipype/interfaces/niftyreg/base.py b/nipype/interfaces/niftyreg/base.py
index a7672166f..fca2a1b97 100644
--- a/nipype/interfaces/niftyreg/base.py
+++ b/nipype/interfaces/niftyreg/base.py
@@ -26,7 +26,8 @@ import shutil
 import subprocess
 from warnings import warn
 
-from ..base import CommandLine, isdefined, CommandLineInputSpec, traits
+from ..base import (CommandLine, isdefined, CommandLineInputSpec, traits,
+                    Undefined)
 from ...utils.filemanip import split_filename
 
 
@@ -47,8 +48,8 @@ def no_nifty_package(cmd='reg_f3d'):
 class NiftyRegCommandInputSpec(CommandLineInputSpec):
     """Input Spec for niftyreg interfaces."""
     # Set the number of omp thread to use
-    omp_core_val = traits.Int(desc='Number of openmp thread to use',
-                              argstr='-omp %i')
+    omp_core_val = traits.Int(1, desc='Number of openmp thread to use',
+                              argstr='-omp %i', usedefault=True)
 
 
 class NiftyRegCommand(CommandLine):
@@ -58,7 +59,10 @@ class NiftyRegCommand(CommandLine):
     _suffix = '_nr'
     _min_version = '1.5.30'
 
+    input_spec = NiftyRegCommandInputSpec
+
     def __init__(self, required_version=None, **inputs):
+        self.num_threads = 1
         super(NiftyRegCommand, self).__init__(**inputs)
         self.required_version = required_version
         _version = self.get_version()
@@ -73,6 +77,26 @@ class NiftyRegCommand(CommandLine):
                     msg = 'The version of NiftyReg differs from the required'
                     msg += '(%s != %s)'
                     warn(msg % (_version, self.required_version))
+        self.inputs.on_trait_change(self._omp_update, 'omp_core_val')
+        self.inputs.on_trait_change(self._environ_update, 'environ')
+
+    def _omp_update(self):
+        if self.inputs.omp_core_val:
+            self.inputs.environ['OMP_NUM_THREADS'] = \
+                str(self.inputs.omp_core_val)
+            self.num_threads = self.inputs.omp_core_val
+        else:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                del self.inputs.environ['OMP_NUM_THREADS']
+            self.num_threads = 1
+
+    def _environ_update(self):
+        if self.inputs.environ:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                self.inputs.omp_core_val = \
+                    int(self.inputs.environ['OMP_NUM_THREADS'])
+        else:
+            self.inputs.omp_core_val = Undefined
 
     def check_version(self):
         _version = self.get_version()

```

satra:
one extra else:

```python
+    def _environ_update(self):
+        if self.inputs.environ:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                self.inputs.omp_core_val = \
+                    int(self.inputs.environ['OMP_NUM_THREADS'])
+            else:
+                self.inputs.omp_core_val = Undefined
+        else:
+            self.inputs.omp_core_val = Undefined
```

satra:
here is the full thing (also added an initial sync with a call to `self._omp_update()`):

```python
diff --git a/nipype/interfaces/niftyreg/base.py b/nipype/interfaces/niftyreg/base.py
index a7672166f..ca0e2e634 100644
--- a/nipype/interfaces/niftyreg/base.py
+++ b/nipype/interfaces/niftyreg/base.py
@@ -26,7 +26,8 @@ import shutil
 import subprocess
 from warnings import warn
 
-from ..base import CommandLine, isdefined, CommandLineInputSpec, traits
+from ..base import (CommandLine, isdefined, CommandLineInputSpec, traits,
+                    Undefined)
 from ...utils.filemanip import split_filename
 
 
@@ -47,8 +48,8 @@ def no_nifty_package(cmd='reg_f3d'):
 class NiftyRegCommandInputSpec(CommandLineInputSpec):
     """Input Spec for niftyreg interfaces."""
     # Set the number of omp thread to use
-    omp_core_val = traits.Int(desc='Number of openmp thread to use',
-                              argstr='-omp %i')
+    omp_core_val = traits.Int(1, desc='Number of openmp thread to use',
+                              argstr='-omp %i', usedefault=True)
 
 
 class NiftyRegCommand(CommandLine):
@@ -58,7 +59,10 @@ class NiftyRegCommand(CommandLine):
     _suffix = '_nr'
     _min_version = '1.5.30'
 
+    input_spec = NiftyRegCommandInputSpec
+
     def __init__(self, required_version=None, **inputs):
+        self.num_threads = 1
         super(NiftyRegCommand, self).__init__(**inputs)
         self.required_version = required_version
         _version = self.get_version()
@@ -73,6 +77,29 @@ class NiftyRegCommand(CommandLine):
                     msg = 'The version of NiftyReg differs from the required'
                     msg += '(%s != %s)'
                     warn(msg % (_version, self.required_version))
+        self.inputs.on_trait_change(self._omp_update, 'omp_core_val')
+        self.inputs.on_trait_change(self._environ_update, 'environ')
+        self._omp_update()
+
+    def _omp_update(self):
+        if self.inputs.omp_core_val:
+            self.inputs.environ['OMP_NUM_THREADS'] = \
+                str(self.inputs.omp_core_val)
+            self.num_threads = self.inputs.omp_core_val
+        else:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                del self.inputs.environ['OMP_NUM_THREADS']
+            self.num_threads = 1
+
+    def _environ_update(self):
+        if self.inputs.environ:
+            if 'OMP_NUM_THREADS' in self.inputs.environ:
+                self.inputs.omp_core_val = \
+                    int(self.inputs.environ['OMP_NUM_THREADS'])
+            else:
+                self.inputs.omp_core_val = Undefined
+        else:
+            self.inputs.omp_core_val = Undefined
 
     def check_version(self):
         _version = self.get_version()
```

satra:
this tries to keep `omp_core_val` in sync with `OMP_NUM_THREADS`

effigies:
@satra @byvernault I want to preface this by saying that I think that patch is fine as it stands.

Two comments, though.

1) It appears that if `omp_core_val` is `0`, `None` or `Undefined` then nipype will count it as one thread, and niftyreg will get no `OMP_NUM_THREADS` environment variable. Is it the case that `niftyreg` will only use one thread, or is this a way of telling nipype not to track that resource? If we want `num_threads` to track actual usage, why not just set `OMP_NUM_THREADS` to 1?

2) I want to be sure we're all on the same page about the intended behavior, because my impression (from admittedly intermittent scanning of the threads) is there's been some amount of talking past each other regarding what the people expect the default behavior with regard to that environment variable to be.

The default behavior in this patch is to set `omp_core_val` and `inputs.environ['OMP_NUM_THREADS']` to `1`, regardless of the `OMP_NUM_THREADS` environment variable passed to nipype. Now, I think this is reasonable and that, if someone wants to use the environment variable, they should do so explicitly as:

```Python
x = pe.Node(Interface(), name='x')
x.inputs.omp_core_val = int(os.get_env('OMP_NUM_THREADS', 1))
```

Just want to make sure nobody's surprised.

satra:
@effigies - i set the default for `omp_core_val` to 1. so unless the user changes something, it will default to 1 independent of what the environment variable is. 

soon after this release we should generalize the thread pieces across interfaces and clearly indicate preference order.

byvernault:
My intent was that if the user wants to increase the omp_core_val for this node, he can do it by using the options omp_core_val (it's an option in the niftyseg executables). 

If it's not set, then I want the interface to use OMP_NUM_THREADS. If Nothing is specify or set, then we use 1.

byvernault:
What I can do is all the changes from @satra but set the default value of omp_core_val like below:

```
omp_core_val = traits.Int(int(os.environ.get('OMP_NUM_THREADS', '1')),
                                          desc='Number of openmp thread to use',
                                          argstr='-omp %i', usedefault=True)
```

WDYT?

byvernault:
I added the changes from @satra with my last message for the default value. Is that ok? I can edit again if needed.

I also changed the NiftySegCommand to be a children class of NiftyFitCommand instead of NiftyRegCommand since there are no omp_core_val.

Cheers,

nipype/interfaces/niftyreg/base.py
```diff
@@ -47,8 +47,9 @@ def no_nifty_package(cmd='reg_f3d'):
 class NiftyRegCommandInputSpec(CommandLineInputSpec):
     """Input Spec for niftyreg interfaces."""
     # Set the number of omp thread to use
-    omp_core_val = traits.Int(desc='Number of openmp thread to use',
-                              argstr='-omp %i')
+    omp_core_val = traits.Int(int(os.environ.get('OMP_NUM_THREADS', '1')),
```
effigies: I personally don't like this because it makes nipype's behavior more dependent on hidden state, which means that different people testing and using the code may see different behavior.

But I'll leave the final call to @satra.

nipype/interfaces/niftyreg/base.py
```diff
@@ -102,13 +129,6 @@ def version(self):
...
         return self.get_version() is not None
 
-    def _run_interface(self, runtime):
-        # Update num threads estimate from OMP_NUM_THREADS env var
-        # Default to 1 if not set
-        if not isdefined(self.inputs.environ['OMP_NUM_THREADS']):
-            self.inputs.environ['OMP_NUM_THREADS'] = self.num_threads
-        return super(NiftyRegCommand, self)._run_interface(runtime)
-
     def _format_arg(self, name, spec, value):
```
effigies: You should also get rid of `_format_arg` in this PR.

PR #2008, opened 2017-05-10
Add int to temporal derivatives

jokedurnez:
Related to : https://github.com/nipy/nipype/pull/1999

It has to be int(bool()): fsl reads 0's or 1's.


oesteban:
thanks a lot @jokedurnez, can I ask you to do the same with line [393](https://github.com/jokedurnez/nipype/blob/7ad5d1966773a3a4cd415ca9a8767f100472135d/nipype/interfaces/fsl/model.py#L393)?

PR #2009, opened 2017-05-11
added migp options

TheChymera:
https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;e95b4651.1501
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4289914/


satra:
@TheChymera - need to run `make specs` to update the auto-test for this interface.

PR #2018, opened 2017-05-13
Making zenodo instructions more prominent

chrisfilo:


satra:
@chrisfilo - you should also take a look at the PULL_REQUEST_TEMPLATE that i added in .github

chrisfilo:
It seems like mention of .zenodo.json is already there.

satra:
yes - i kind of committed it without review ;) - that's why it would be good if that suffices

chrisfilo:
Up to you.

PR #2020, opened 2017-05-14
FIX: ?h.white is actually created in the -pial step

effigies:
In this installment of my never-ending series of minor tweaks to `ReconAll`, I've stumbled across the most annoyingly undocumented (outside of source code) change in process in FreeSurfer v6.0, which is that `?h.white` are created in the `-pial` step, to take advantage of [post-aparc refinements](https://github.com/freesurfer/freesurfer/blob/dev/scripts/recon-all#L3517-L3521). If you happen to kill a process running `-pial` after the other outputs are produced but before `?h.white` is produced, rerunning will incorrectly skip over this step.

I can't guarantee this is the last PR of the series, but I sincerely hope so.

Related: #1991 #1824 #1790


satra:
@effigies - this is the reprozip trace of recon-all.

https://www.dropbox.com/s/1w7zjz22v1tuyrm/config.yml?dl=0

using this you should be able to create the graph of recon-all with reprozip (it has a graph generation tool). in the future we hope to create a reprozip to nipype workflow tool, but we need to understand the graph a bit better first.

effigies:
Ha. I was going to say, I'm nearly ready to start hacking on the reprozip workflow generator just to be done with all of these. Maybe next Thursday?

satra:
should this be added in for 0.13.1?

effigies:
If you're comfortable with it. We can patch around it (or depend on a post
release commit), so either way.

PR #2017, opened 2017-05-13
patches for release 0.13.1

satra:
Changes proposed in this pull request
- allows conda forge builds to work across python versions
- updates mailmap for contributors
- adds more contributors

TODO:

- [x] finalize contributors after monday
- [x] update base docker builds by monday night. (@oesteban - i'll need your help for this).

in the future we need to test conda-forge builds before we release.

PR #2022, opened 2017-05-15
[ENH] Revise CircleCI settings

oesteban:
- Try to build nipype/base in circle build, and cache it properly.


oesteban:
Submitted rebuild to see if caching starts working

satra:
@oesteban - should this be merged?

oesteban:
@satra I'm trying to see the caching working. But no luck so far. I tried splitting the cached files, so that they are smaller. Maybe that way circle would swallow them. But I don't see any changes.

Let me trigger another build. I'll enable ssh and see if the files are there.

satra:
@oesteban - sounds good. you could also check the total size of the cache after it's created, and we can talk to the circle folks to see if the size is the limitation.

satra:
@oesteban - did this work?

oesteban:
Yes, the largest tarred image is 8.5 GB. I think we can gzip it. Splitting the images was not a good idea, I just expands the tar file for each single image.

satra:
ok i'm testing the gzipped version in the other PR.

oesteban:
Altogether, the tar (no compression) makes 12GB

oesteban:
I close this, follow up in #2017

PR #2028, opened 2017-05-17
ENH: Permit external packages to deprecate interfaces

effigies:
External projects, such as [poldracklab/niworkflows](https://github.com/poldracklab/niworkflows), may maintain a set of interfaces, that over time require deprecating portions of their input and output specs. At present, the deprecation machinery is intrinsically tied to nipype's version number.

This PR adds a `package_version` attribute to each interface, which defaults to nipype's, but allows other packages easily to add their own versioning information to specs with a deprecation schedule.

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -45,7 +45,7 @@ def get_nipype_gitversion():
 if __version__.endswith('-dev'):
     gitversion = get_nipype_gitversion()
     if gitversion:
-        __version__ = __version__.replace('-dev', '-' + gitversion + '.dev')
+        __version__ = '{}+{}'.format(__version__, gitversion)
 
 CLASSIFIERS = ['Development Status :: 5 - Production/Stable',
                'Environment :: Console',
@@ -141,7 +141,8 @@ REQUIRES = [
     'funcsigs',
     'pytest>=%s' % PYTEST_MIN_VERSION,
     'mock',
-    'pydotplus'
+    'pydotplus',
+    'packaging',
 ]
 
 if sys.version_info <= (3, 4):
--- a/nipype/interfaces/base.py
+++ b/nipype/interfaces/base.py
@@ -31,6 +31,7 @@ from textwrap import wrap
 from warnings import warn
 import simplejson as json
 from dateutil.parser import parse as parseutc
+from packaging.version import Version
 
 from .. import config, logging, LooseVersion, __version__
 from ..utils.provenance import write_provenance
@@ -43,7 +44,7 @@ from .traits_extension import (
 from ..external.due import due
 
 runtime_profile = str2bool(config.get('execution', 'profile_runtime'))
-nipype_version = LooseVersion(__version__)
+nipype_version = Version(__version__)
 iflogger = logging.getLogger('interface')
 
 FLOAT_FORMAT = '{:.10f}'.format
@@ -445,7 +446,7 @@ class BaseTraitedSpec(traits.HasTraits):
             else:
                 msg3 = ''
             msg = ' '.join((msg1, msg2, msg3))
-            if LooseVersion(str(trait_spec.deprecated)) < self.package_version:
+            if Version(str(trait_spec.deprecated)) < self.package_version:
                 raise TraitError(msg)
             else:
                 if trait_spec.new_name:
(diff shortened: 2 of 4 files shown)


effigies:
I'm moving to [`packaging.version.Version`](https://packaging.pypa.io/en/latest/version/) instead of `LooseVersion` for spec deprecations. It's [PEP440](https://www.python.org/dev/peps/pep-0440/)-compliant, which is the standard that pip uses to resolve versions, while `LooseVersion` is just a pain wrt pre-release version strings.

This did require me to switch the gitversion scheme to `1.0.0-dev+githash` as opposed to `1.0.0-githash.dev`. Will this cause problems? I assume it's mostly for readability; neither is more `cmp`able than the other.

Obviously it isn't appropriate for things like `min_ver`, where we have no control over external versioning schemes.

cc @satra @chrisfilo

nipype/interfaces/base.py
```diff
@@ -31,6 +31,7 @@
 from warnings import warn
 import simplejson as json
 from dateutil.parser import parse as parseutc
+from packaging.version import Version
 
 from .. import config, logging, LooseVersion, __version__
```
satra: remove LooseVersion from everywhere?

nipype/info.py
```diff
@@ -141,7 +141,8 @@ def get_nipype_gitversion():
     'configparser',
     'pytest>=%s' % PYTEST_MIN_VERSION,
     'mock',
-    'pydotplus'
+    'pydotplus',
+    'packaging',
```
satra: `packing` should be added to requirements.txt and rtd_requirements as well?

satra:
can this wait for the next release? or should it be included now?

effigies:
It can wait.

nipype/interfaces/base.py
effigies: It's used for versioning on external programs, e.g. FSVersion < 6.0 or FSLVersion > 507. Since these don't follow PEP440, I intended to leave them as they are.

effigies:
@satra This is ready for another look, now the release is out.

satra:
@effigies - could you please merge with master?

effigies:
It's rebased off master.

satra:
@effigies - that's weird, i haven't seen that error in all the builds i have run in the last few days? also i don't see a commit message above that merges off of current master. (circle doesn't merge with master when a PR is run).

effigies:
Yeah, I noticed that. I'm guessing the issue is caused by [removing](https://github.com/nipy/nipype/commit/8946bcab9d0e2f24e5364e42d4a7766e00237cb8#diff-29944324a3cbf9f4bd0162dfe3975d88L3) the updated version of docker, but why it's waited until now to exhibit this failure is unknown.

First commit: b700261
Parent: ff02ed0

Sorry if I'm being dense, but I'm not sure what you're expecting to see.

effigies:
So it seems like restoring docker loses caching...

satra:
@effigies - since we are testing this out, could you please now remove those two lines?

effigies:
Removed that commit.

effigies:
Passing. Doesn't look like we benefited from caching, but the Circle caching mechanism is opaque to me. Perhaps it takes a couple attempts to get something usable to a future build.

nipype/info.py
satra: i think, this would also need to be added to `SETUP_REQUIRES`. you can check by trying to install nipype in a clean environment using `python setup.py install`.

effigies: Building the docker image (I never figured out how to do virtualenvs) from scratch. Is that an equivalent check?

satra: you can do this inside a basic continuumio/miniconda docker instance - think of that as a virtualenv.

```
docker run -it --rm -v /path/to/nipype:/nipype-src continuumio/miniconda bash
> cd /nipype-src
> python setup.py bdist_wheel
```

satra: and more generally download and install miniconda and then do: conda create -n envname python=2/3 for creating virtual envs

effigies: Builds and installs without issue. (Had to apt-get install build-essential, conda install scipy, lxml and dependencies, but otherwise no problems.)

satra: so that latter are runtime dependencies not setup dependencies. there is difference between setup requires and run/test requires. setup requires are things that the setup.py file imports.

satra: e.g. see: https://github.com/conda-forge/nipype-feedstock/blob/master/recipe/meta.yaml#L17

effigies: Right. I don't think `packaging` is required by that. Running `python setup.py bdist_wheel` worked fine. I then also ran `python setup.py install` which required the other dependencies.

Sorry if I'm missing something.

PR #2034, opened 2017-05-18
[ENH] Add copy_header option to N4BiasFieldCorrection

oesteban:
This N4 algorithm does not propagate the nifti headers after run.
This PR allows setting the option ``copy_header=True`` and that will
overwrite the N4 output with the corrected data and the original headers.

Default value of `copy_header` is `False`


nipype/interfaces/ants/segmentation.py
```diff
@@ -384,6 +387,23 @@ def _list_outputs(self):
...
+
+        if self.inputs.copy_header:
+            import nibabel as nb
+            refnii = nb.load(self.inputs.input_image)
+            hdr = refnii.header.copy()
+            out_file = self._gen_filename('output_image')
+            nii = nb.load(out_file)
+            hdr.set_data_dtype(nii.header.get_data_dtype())
+            nb.Nifti1Image(nii.get_data(), refnii.affine, hdr).to_filename(
+                out_file)
```
effigies: Does `N4BiasFieldCorrection` always use NIfTI images, or are we making this assumption? To be more general, this could be:

```Python
if self.inputs.copy_header and runtime.returncode in correct_return_codes:
    import nibabel as nb
    in_img = nb.load(self.inputs.input_image)
    out_file = self._gen_filename('output_image')
    out_img = nb.load(out_file, mmap=False)
    new_img = out_img.__class__(out_img.get_data(), in_img.affine, in_img.header)
    new_img.set_data_dtype(out_img.get_data_dtype())
    new_img.to_filename(out_file)
```

A couple additional notes:

* Should make sure we ran successfully before reading/overwriting the output
* `mmap=False` on `nb.load()` - See nipy/nibabel#492
* No need to copy header (that's done in [image creation](https://github.com/nipy/nibabel/blob/master/nibabel/filebasedimages.py#L192))
* I prefer to do `get/set_data_dtype` on the image itself, rather than the header, even though they're equivalent. Since an image contains a reference to the header and not the other way around, if some image class needs (in a future nibabel version) to modify the data portion of the image during this call, it can. Your call on this one.

oesteban: `N4BiasFieldCorrection` virtually accepts any [of the file formats supported by ITK](https://itk.org/Wiki/ITK/File_Formats). So I guess it is worth checking the image type (Analyze or Nifti) before copying the header. @satra any thoughts?

effigies:
I imagine we should do the same for `bias_image`, if it's being created?

satra:
@oesteban - what aspects of the header does it not copy? as in, is the output file an invalid Nifti file? and is this an issue that should be fixed in ANTs?

oesteban:
@satra we started here https://github.com/poldracklab/fmriprep/issues/497 . I don't know exactly what are the problems, I could go deeper into it.

@effigies that would make sense for consistency

satra:
@oesteban - i looked at the issue, the voxel size bit seems related to floating point representation and is fine. the shape mismatch seems weird. so two questions

1. does nibabel complain when the output image is loaded that there is a mismatch in the header?
2. is there really a mismatch in shape?

i'm worried that we shouldn't be patching bugs in other software, if they are indeed bugs. has this been reported on ants? and if we patch, we need to be very clear to un-patch it when it's fixed.

@kaczmarj, @mgxd - could we create a test across different versions of ANTs for N4 and test this issue?

satra:
@oesteban - any thoughts on my comment?

oesteban:
@satra https://github.com/stnava/ANTs/issues/452

We can go ahead with this patch in our settings (fmriprep and Co.) and close this PR if you prefer to wait for ANTs to solve the issue. On the other hand, they don't release very often, so this patch could be an incentive for using nipype in the meantime.

Since the major problem is the change of the qform code, I think nipype could very well use this patch.

oesteban:
Update: since the qform issue derives from ITK, it is not likely to change (and we probably should address this in all ITK-related interfaces, to keep workflows consistent).

Then the type-casting is the user's responsibility. I can see nipype helping on that.

Issue #1970, opened 2017-04-25
Nipype installation test failure 


rattar1989:
### Summary
Hi all,
I installed nipype step by step based on the nipype guide without any error in installation process, but when I tested the system by nipype.test() I encounered: FAILED (SKIP=77, errors=5, failures=3)

### Actual behavior
======================================================================
ERROR: Failure: InterpolationSyntaxError ('%' must be followed by '%' or '(', found: '%h')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nose/loader.py", line 251, in generate
    for test in g():
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nipype/pipeline/engine/tests/test_utils.py", line 342, in test_provenance
    psg = write_workflow_prov(eg, prov_base, format='all')
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py", line 1270, in write_workflow_prov
    prov_doc = ProvStore().add_results(result)
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nipype/utils/provenance.py", line 376, in add_results
    for key, value in list(get_info().items()):
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nipype/__init__.py", line 60, in <lambda>
    get_info = lambda: _get_pkg_info(os.path.dirname(__file__))
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nipype/pkg_info.py", line 82, in get_pkg_info
    src, hsh = pkg_commit_hash(pkg_path)
  File "/home/rahman/anaconda3/lib/python3.6/site-packages/nipype/pkg_info.py", line 52, in pkg_commit_hash
    archive_subst = cfg_parser.get('commit hash', 'archive_subst_hash')
  File "/home/rahman/anaconda3/lib/python3.6/configparser.py", line 797, in get
    d)
  File "/home/rahman/anaconda3/lib/python3.6/configparser.py", line 393, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/home/rahman/anaconda3/lib/python3.6/configparser.py", line 443, in _interpolate_some
    "found: %r" % (rest,))
[...]

PR #2042, opened 2017-05-23
BF: Do not double % the field for hash in COMMIT_INFO.txt

yarikoptic:
Addresses an issue listed in #1970

Before:
```shell
$> git archive HEAD nipype/COMMIT_INFO.txt | grep -a archive_su 
archive_subst_hash=%h
```

After:
```shell
$> git archive HEAD nipype/COMMIT_INFO.txt | grep -a archive_su                             
archive_subst_hash=df035ac81
```
but imho correct fix to code should also allow for that line still stay with original value `archive_subst_hash=$Format:%h$`

### DIFF
--- a/nipype/COMMIT_INFO.txt
+++ b/nipype/COMMIT_INFO.txt
@@ -1,6 +1,6 @@
 # This is an ini file that may contain information about the code state
 [commit hash]
 # The line below may contain a valid hash if it has been substituted during 'git archive'
-archive_subst_hash=$Format:%%h$
+archive_subst_hash=$Format:%h$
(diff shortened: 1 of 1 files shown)


yarikoptic:
yeap -- not a full fix since now parser would fail to parse it whenever it is run within git sources so that '%h' still present as is ;)

mgxd:
@yarikoptic is this stale?

djarecka:
@yarikoptic - are you still working on it? the issue is closed

effigies:
Yarik, please reopen if you intend to work more on this.

PR #2048, opened 2017-05-23
Replace deprecated HasTraits.set with trait_set

effigies:
Tests are currently failing, I think at least partially because of a bunch of deprecation warnings about `HasTraits.set`.

This fixes that (I think). Then hopefully the logs will be clear enough to identify other issues.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,6 @@
 machine:
+  pre:
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
   environment:
     OSF_NIPYPE_URL: "https://files.osf.io/v1/resources/nefdp/providers/osfstorage"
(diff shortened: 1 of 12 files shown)


effigies:
Cleaned up some simple nibabel-related warnings.

circle.yml
```diff
@@ -1,4 +1,6 @@
 machine:
+  pre:
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
```
oesteban: This would break the cache (if it worked with 1.9, we didn't get it to work, right @satra?). Caching does not work with 1.10.

satra: sth weird happened with docker. i thought i had removed 1.9 and caching worked, but not after it got merged. we still have to revisit the cache situation. from a time perspective restoring and saving the cache takes almost as time as building it.

oesteban: yes, we should probably roll back to the docker pull solution, and manually update nipype/base

oesteban: (I don't have hopes for get the cache to work. Definitely not with 1.10 but 1.9 neither)

PR #2053, opened 2017-05-25
WIP: Traitlets 

djarecka:
Moving from `traits` to `traitlets`. 
- this is WIP, still need a lot of cleaning etc., see my notes in the code
- i was running only tests  from `interfaces/tests/test_base.py` for now
- it works with dev version of `traitlets`
- more tests should be added, some classes have no single test

main problems/differences:
- couldn't find anything for `_ = traits.Disallow`
- `Undefined` is not the same as in `traits` (using `default_value=None`)
- for every trait you have some default value, i.e. if you ask for `a = traitlets.Int()` the default value will be `0` and `isdefined(a)` gives `True`. You have to define `a` as `a = traitlets.Int(default_value=None, allow_none=True)`.
- if you ask for a list, i.e. `a = traitlets.List(default_value=None, allow_none=True)` the default value is `[]` not `None` (!), so I changed the `isdefined` method, so it gives `False` for `isdefined([])`.


nipype/interfaces/base.py
```diff
@@ -72,7 +72,7 @@ class Str(traitlets.Unicode):
 #traits.Str = Str
 
 
-# dj TOTHINK: do i need it?
+# dj TOASK: nipype almost doest use it, should we remove?
```
oesteban: On the contrary, I would suggest using more customized errors.

djarecka: sure, but someone has to use it ;-) I was simply marking things when I saw that they are (almost) not used.

However I agree, that customized errors are nice and they also should be tested better. I found a few places that the raised exception was probably not the one that was expected, but the test was passing.

effigies:
Some comments on the post. I haven't had a chance to look at the code in detail.

> * couldn't find anything for _ = traits.Disallow

If I remember, this had the effect of not permitting new attributes to be added to a `HasTraits` object?

What about moving back to `BaseTraitedSpec`, and then:

```Python
class TraitedSpec(BaseTraitedSpec):
    def __setattr__(self, name, value):
        if name not in self.__dict__:
            raise NameError  # or something
        super(TraitedSpec, self).__setattr(name, value)
```

> * for every trait you have some default value, i.e. if you ask for a = traitlets.Int() the default value will be 0 and isdefined(a) gives True. You have to define a as a = traitlets.Int(default_value=None, allow_none=True).

Currently the default is "Undefined" unless "usedefault=True" is set... I think it's probably going to make sense to subclass the common traits so that `default_value=None, allow_none=True` are set. Or else it's going to be a pretty major task to add this boilerplate to *most* traits in nipype.

> * if you ask for a list, i.e. a = traitlets.List(default_value=None, allow_none=True) the default value is [] not None (!), so I changed the isdefined method, so it gives False for isdefined([]).

As discussed on gitter, I think we're going to want `[]` to be a valid, defined value. I could be wrong; I don't have a specific case in mind where we'd absolutely need to distinguish between an empty list and no inputs, but it strikes me as unexpected, and having a strong potential to bite somebody.

djarecka:
@effigies -  
* re `Dissalow`:
    I tried to write my own `__setattr__`, the simple approach doesn't work but I was able to write something that seemed to work. However, at the end I was told that it's very likely that my subclass will not be compatible with the future `traitlets`, so I gave up. You can see this [issue](https://github.com/ipython/traitlets/issues/397).
    Later, I've learnt that what is allow and what raises error is a bit more complicated, and I'm not sure at the end what will be the future strategy for `traitlets`. You can read this [conversation](https://github.com/ipython/traitlets/issues/400), I'm waiting for an answer. 

* re `default_value=None`:
    Yes, I understand your point and was thinking about it. That was a @satra suggestion to not write extra subclasses, assuming that the changes for all interfaces will be done (semi-)automatically. I'm open for suggestions, can change this.

nipype/interfaces/base.py
```diff
@@ -715,7 +715,7 @@ def _get_filecopy_info(self):
 
 
 class BaseInterfaceInputSpec(TraitedSpec):
-    ignore_exception = Bool(help="Print an error message instead \
+    ignore_exception = Bool(default_value=False, help="Print an error message instead \
```
oesteban: We probably want to make ignore_exception a BaseInterfaceSpec class attribute, this could be a good point to make this deprecation.

djarecka: ok, i added it to my todo. want to finish other things first

djarecka:
@satra , @oesteban :
I'm changing `nilearn.py` and I have problems with [tests](https://github.com/djarecka/nipype/blob/7048722ebd243d22369b7b8057ab6626227c691c/nipype/interfaces/tests/test_nilearn.py). I already rearrange them, because I was having problems that the tests were "not independent", but I realized I should change more because values of global variables, e.g. `Labels` are changing.  I can change tests, so they all pass (e.g. change to `Labels.copy()` [here](https://github.com/djarecka/nipype/blob/7048722ebd243d22369b7b8057ab6626227c691c/nipype/interfaces/tests/test_nilearn.py#L90)), but I don't like the fact, that in previous version with `traits` the value of `Labels` is not changing even if `iface.SignalExtraction.inputs.class_labels` is changing). It could be the source of big differences between versions for traits that are lists (or other mutable). Any thoughts?

In the current version, I run only two tests, and the second, [test_signal_extr_4d_global_no_shared](https://github.com/djarecka/nipype/blob/7048722ebd243d22369b7b8057ab6626227c691c/nipype/interfaces/tests/test_nilearn.py#L99) is failing because `Labels` has already changed.

djarecka:
@satra - coming back to our discussion about `DynamicTraitedSpec` and `traits.Disallow`. I have problem with a [test](https://github.com/nipy/nipype/pull/2053/files#diff-9c3d38dfe6ae95376cb0a6dbb3a68a06R444) (taken from doctests) that it's actually related to this issue. Reading my assert statements you can see how `param1` and `param3` are set. You can access them directly but `get` doesn't return them, since they are not treated as traits and [trait_values](https://github.com/djarecka/nipype/blob/e727fd612c379db388ba84404d64fc5d7da08866/nipype/interfaces/base.py#L502) doesn't return them.

PR #2056, opened 2017-05-27
ENH: ICA_AROMA: make current working directory default output directory

jdkent:
- Make the default output directory the current working directory so you don't have to specify an out_dir, which I believe makes ICA_AROMA more Node friendly.


jdkent:
I was a little hasty getting this up. Please ignore until I fix it.

satra:
@jdkent - could you please resolve conflicts with master? relative imports should always be used in nipype - we stopped being careful during review.

jdkent:
@satra: you've probably answered this a thousand times, but why use relative versus absolute imports? Not that I abide by PEP for everything, but [PEP 328](https://docs.python.org/2.5/whatsnew/pep-328.html) wants to make absolute imports standard. Does it have to do with making sure we are importing the right packages if we have multiple versions of nipype installed? Sorry for badgering, but I am curious.

effigies:
@jdkent I think you may be misreading PEP 328, which is about interpretation of import statements. Pre-PEP328, `import x` would look for `x` in the local package, *i.e.* all imports were assumed relative. PEP328 moved to implicit absolute imports and explicit relative imports.

But yes, using relative imports does ensure that we use a package-local version, rather than one found in the PYTHONPATH. (It also helps with [including nipype as a submodule](https://github.com/nipy/nipype/pull/2059) in [other packages](https://github.com/poldracklab/niworkflows/pull/174).)

Incidentally, @oesteban it looks like `from __future__ import absolute_import` was only necessary pre-2.6? Nipype doesn't support before 2.7, so that's probably why all the missing `absolute_import` requirements haven't had any effect.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -103,12 +104,20 @@ class ICA_AROMA(CommandLine):
 
     def _list_outputs(self):
         outputs = self.output_spec().get()
-        out_dir = os.path.abspath(self.inputs.out_dir)
-        outputs['out_dir'] = out_dir
-
+        if isdefined(self.inputs.out_dir):
```
jdkent: I think I need the isdefined module for this line to run, please let me know if I'm mistaken.

effigies: Good catch. Please add `isdefined` to the `from ..base import` line.

PR #2057, opened 2017-05-28
ENH: added interface for FSL's dual_regression

TheChymera:


satra:
add the auto test file from `make specs`

PR #2059, opened 2017-06-01
ENH: Reduce within-package absolute imports

effigies:
With our (FMRIPREP/MRIQC) mismatched release cycles, we're running into the situation where correctly managing dependencies puts an undue burden on users (at least those installing natively, rather than running the Docker/Singularity images). We're attempting to move to include nipype as a submodule in niworkflows. Our projects will then import from `niworkflows.nipype`, which will avoid clashing with other installs of nipype.

Absolute imports assume that the root module is named `nipype`, where relative imports will permit (relatively) easy use as a submodule.

This should not affect the actual behavior of nipype at all, but will require that future PRs refrain from using absolute imports.

Changes proposed in this pull request:
- Move to relative imports in non-testing code

### DIFF
--- a/nipype/interfaces/fsl/ICA_AROMA.py
+++ b/nipype/interfaces/fsl/ICA_AROMA.py
@@ -10,6 +10,7 @@
     ...                            '../../testing/data'))
     >>> os.chdir(datadir)
 """
+from __future__ import print_function, division, unicode_literals, absolute_import
 from ..base import (TraitedSpec, CommandLineInputSpec, CommandLine,
                     File, Directory, traits)
 import os
--- a/nipype/interfaces/fsl/ICA_AROMA.py
+++ b/nipype/interfaces/fsl/ICA_AROMA.py
@@ -10,16 +10,10 @@
     ...                            '../../testing/data'))
     >>> os.chdir(datadir)
 """
-from nipype.interfaces.base import (
-    TraitedSpec,
-    CommandLineInputSpec,
-    CommandLine,
-    File,
-    Directory,
-    traits,
-    OutputMultiPath
-)
+from ..base import (TraitedSpec, CommandLineInputSpec, CommandLine,
+                    File, Directory, traits)
 import os
+
 
 class ICA_AROMAInputSpec(CommandLineInputSpec):
     feat_dir = Directory(exists=True, mandatory=True,
--- a/nipype/interfaces/fsl/fix.py
+++ b/nipype/interfaces/fsl/fix.py
@@ -54,7 +54,7 @@ outgraph = fix_pipeline.run()
 
 """
 
-from nipype.interfaces.base import (
+from ..base import (
     TraitedSpec,
     CommandLineInputSpec,
     CommandLine,
@@ -64,11 +64,8 @@ from nipype.interfaces.base import (
     BaseInterfaceInputSpec,
     traits
 )
-from nipype.interfaces.traits_extension import (
-    Directory,
-    File,
-    isdefined
-)
+from ..traits_extension import Directory, File, isdefined
+
 import os
 
 class TrainingSetCreatorInputSpec(BaseInterfaceInputSpec):
--- a/nipype/interfaces/fsl/fix.py
+++ b/nipype/interfaces/fsl/fix.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """The fix module provides classes for interfacing with the `FSL FIX
@@ -53,6 +54,7 @@ fix_pipeline.write_graph()
 outgraph = fix_pipeline.run()
 
 """
+from __future__ import print_function, division, unicode_literals, absolute_import
 
 from ..base import (
     TraitedSpec,
@@ -65,8 +67,8 @@ from ..base import (
     traits
 )
 from ..traits_extension import Directory, File, isdefined
-
 import os
+
 
 class TrainingSetCreatorInputSpec(BaseInterfaceInputSpec):
     mel_icas_in = InputMultiPath(Directory(exists=True), copyfile=False,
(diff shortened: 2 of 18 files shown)


chrisfilo:
pycharm is showing me many more occurences of "from nipype" in non-testing code:
image

effigies:
The slicer modules are auto-generated. It looks like the [generating script](https://github.com/nipy/nipype/blob/master/nipype/interfaces/slicer/generate_classes.py#L41-L43) has been updated, I just didn't take the time to figure out how to run it.

chrisfilo:
I think there are still some imports apart from the autogenerated classes (in dipy and petpvc for example).

effigies:
I think I got all of them now.

nipype/interfaces/fsl/ICA_AROMA.py
```diff
@@ -10,17 +10,11 @@
...
-from nipype.interfaces.base import (
-    TraitedSpec,
-    CommandLineInputSpec,
-    CommandLine,
-    File,
-    Directory,
-    traits,
-    OutputMultiPath
-)
+from ..base import (TraitedSpec, CommandLineInputSpec, CommandLine,
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/interfaces/fsl/fix.py
```diff
@@ -54,7 +54,7 @@
 
 """
 
-from nipype.interfaces.base import (
+from ..base import (
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/interfaces/niftyseg/base.py
```diff
@@ -16,8 +16,8 @@
 See the docstrings of the individual classes for examples.
 """
 
-from nipype.interfaces.niftyreg.base import no_nifty_package
-from nipype.interfaces.niftyfit.base import NiftyFitCommand
+from ..niftyreg.base import no_nifty_package
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/utils/__init__.py
```diff
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-from __future__ import absolute_import
```
oesteban: Why removing this?

nipype/workflows/dmri/fsl/artifacts.py
```diff
@@ -4,13 +4,13 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 from __future__ import division
 
-from nipype.utils import NUMPY_MMAP
-
 from ....interfaces.io import JSONFileGrabber
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/workflows/smri/freesurfer/autorecon1.py
```diff
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 
-from nipype.utils import NUMPY_MMAP
-from nipype.interfaces.utility import Function,IdentityInterface
-import nipype.pipeline.engine as pe  # pypeline engine
-from nipype.interfaces.freesurfer import *
+from ....utils import NUMPY_MMAP
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/workflows/smri/freesurfer/autorecon2.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
-from nipype.interfaces.utility import Function, IdentityInterface, Merge
-import nipype.pipeline.engine as pe  # pypeline engine
-from nipype.interfaces.freesurfer import *
+from ....interfaces.utility import Function, IdentityInterface, Merge
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/workflows/smri/freesurfer/autorecon3.py
```diff
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
-from nipype.interfaces.utility import IdentityInterface, Merge, Function
-import nipype.pipeline.engine as pe  # pypeline engine
-from nipype.interfaces.freesurfer import *
+from ....interfaces.utility import IdentityInterface, Merge, Function
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/workflows/smri/freesurfer/ba_maps.py
```diff
@@ -1,11 +1,10 @@
 # -*- coding: utf-8 -*-
 import os
-import nipype
-from nipype.interfaces.utility import Function,IdentityInterface
-import nipype.pipeline.engine as pe  # pypeline engine
-from nipype.interfaces.freesurfer import *
-from nipype.interfaces.io import DataGrabber
-from nipype.interfaces.utility import Merge
+from ....interfaces.utility import Function, IdentityInterface
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

nipype/workflows/smri/niftyreg/groupwise.py
```diff
@@ -7,9 +7,9 @@
 """
 
 from builtins import str, range
-import nipype.interfaces.utility as niu
-import nipype.interfaces.niftyreg as niftyreg
-import nipype.pipeline.engine as pe
+from ....interfaces import utility as niu
```
oesteban: for backwards compatibility, make sure to add ```from __future__ import absolute_import``` before this

effigies:
I thought `from __future__ import absolute_import` was for if you wanted to import within the package using absolute import paths. If we're eliminating them, it seemed unnecessary.

Instead it [seems](https://stackoverflow.com/questions/33743880/what-does-from-future-import-absolute-import-actually-do#33744115) that it does the opposite, always preferring an installed package to the local package. Which makes me wonder why we were using absolute imports for within-package imports at all.

effigies:
@oesteban Updated.

PR #2063, opened 2017-06-04
[FIX] fix cluster threshold in randomise + change default prefix

jokedurnez:
- The base name tbss in randomise is very confusing, so I changed it to randomise (and removed the _, as FSL adds it too).
- Also, the cluster thresholds with 2 digits are not read.  Not sure about the other thresholds, so I didn't touch those.


mgxd:
@jokedurnez can you run `make specs` and add the updated auto-test?

satra:
@jokedurnez - it would be great if you can merge master and add the updated auto-test.

jokedurnez:
I believe the tests failed exactly because of one of the changes I made, so I'm not sure what to do.  Probably I should change the test as well, but that would require me to take some time to dive into the testing stuff.  I'll try, but it won't be tomorrow... :\

oesteban:
Hi @jokedurnez, sorry for joining late to the party.

I think that doing the following will resolve this error:

  1. At your nipype's repo root directory, with your `patch-8` branch checked out, run  `make specs`

  2. Then `git add nipype/interfaces/fsl/tests/test_auto_Randomise.py`

  3. `git commit -m 'updated Randomise specs' && git push`

oesteban:
Please follow-up in #2369. I kept your commits (@jokedurnez's and @mgxd's) there.

PR #2070, opened 2017-06-08
ENH: Enable per-stage masking in ants.Registration

effigies:
antsRegistration permits per-stage masking, and accepts invalid filenames to indicate that no mask should be applied, providing [`NULL` as the example string of choice](https://github.com/stnava/ANTs/blob/master/Scripts/antsRegistrationSyN.sh#L85).

- Permit `fixed_image_mask` and `moving_image_mask` to be a (single or list of) `'NULL'` string(s) or existing file(s)

This should change the command lines, but not semantics, of existing interface calls.


nipype/interfaces/ants/registration.py
```diff
@@ -219,12 +219,14 @@ class RegistrationInputSpec(ANTSCommandInputSpec):
                             usedefault=True, desc='image dimension (2 or 3)')
     fixed_image = InputMultiPath(File(exists=True), mandatory=True,
                                  desc='image to apply transformation to (generally a coregistered functional)')
-    fixed_image_mask = File(argstr='%s', exists=True,
-                            desc='mask used to limit metric sampling region of the fixed image')
+    fixed_image_mask = InputMultiPath(
+        traits.Either('NULL', File(exists=True)),
+        desc='mask used to limit metric sampling region of the fixed image')
```
chrisfilo: Please describe the use of special string 'NULL' in the desc.

effigies:
Thanks. I'll wait for the first round of tests to finish, but won't waste time with a full Circle build on the latest.

PR #2074, opened 2017-06-14
[REF] Change dynamic loading of plugin to rely on relative import

oesteban:
Fixes https://github.com/poldracklab/niworkflows/issues/179.

- Replaces the old `__import__` (which, btw is apparently not very Python 3 friendly) by a default import of the plugins module.


effigies:
This runs up against circular imports. I propose a kind of hacky solution (which I think is the best we'll get until we drop Py2) in oesteban/nipype#5.

Most of the changes are a reversion. It reduces the diff to [one line](https://github.com/nipy/nipype/compare/e02cb4a...effigies:fix/Plugin-absolute-import).

PR #2076, opened 2017-06-14
fix: do not create a nipype folder in the home directory by default

satra:
- Nipype by default should be stateless and not create any change outside of the directory it is called from.

### DIFF
--- a/nipype/utils/config.py
+++ b/nipype/utils/config.py
@@ -88,26 +88,11 @@ class NipypeConfig(object):
     def __init__(self, *args, **kwargs):
         self._config = configparser.ConfigParser()
         config_dir = os.path.expanduser('~/.nipype')
-        mkdir_p(config_dir)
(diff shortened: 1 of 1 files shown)


oesteban:
We probably should also provide a system-wide way to configure settings, _a la_ matplotlib and do not use the `~/.nipype` anymore. WDYT?

satra:
people can still use that location, just that we don't create it by default. the default config is the one that comes from the library.

oesteban:
Yes -- the idea is to limit the config sources:
- Installation wide (like matplotlib, under nipype installation folder in pythonpath)
- Run-wide: at the working directory level.

satra:
@oesteban - we have always had that.  three locations: (in the nipype code, ~/.nipype, cwd). 

however, after this PR, it will not write a file by default in ~/.nipype. but in all cases, local always supersedes the other two.

PR #2078, opened 2017-06-15
ENH: Enable antsRegistration 2.2.0+ masking

effigies:
- Add `fixed`/`moving_image_masks` input spec entries, mutually exclusive with the singular versions
  - New entries are `min_ver=2.2.0`
  - Plural reflects that each mask can be set per-transform stage
- Old entries set to `max_ver=2.1.0`; they produce different command-lines
- `AntsCommand` is given a usable `version` property, based on running `antsRegistration --version`

---------
#### Old PR text, for context.

- Add missing import

Bug introduced in #2070.


effigies:
Turns out I failed to hook up these inputs, so the downstream testing I was doing was worthless. Added a doctest, which found another bug.

Ready for review.

effigies:
It turns out this is a post 2.1.0 development... Reverting #2070 and will update the input spec accordingly.

effigies:
Status: I can run the desired command in ANTs 2.1.0 (though it crashes; it seems they changed how they did masking). I'm upgrading our setup to ANTs 2.2.0, so I'll be able to test it today.

effigies:
Status: I'm able to use this to pass a mask in ANTs 2.1.0; I'll test with 2.2.0 today.

satra:
@effigies - for nipype containers, we should move over to neurodebian install for ants to minimize footprint, now that the latest version is available there.

effigies:
@satra Done.

PR #2084, opened 2017-06-21
ENH: Include realignment matrices in TOPUP outputs

mwaskom:
FSL's `topup` binary has two undocumented options: `--dfout` and `--rbmout` (see [here](https://www.mail-archive.com/hcp-users@humanconnectome.org/msg00670.html)). The Nipype interface exposes the former but not the latter, which writes out transformation matrices that encode the rigid realignment from each frame to the first frame of the input file. These matrices are necessary to properly apply the warpfields to the input images. This PR extends the currently existing logic for `--dfout` to `--rbmout`.


effigies:
Looks reasonable. Assume you've actually used it and the file names work out correctly?

mwaskom:
Yep I have a workflow that needed this functionality and things look good.

Issue #2091, opened 2017-06-23
Error raised from DataSink on Python 3


mwaskom:
I'm transitioning some workflow code to Python3 and have run into an issue in nipype that I think is a Python3 conversion problem. Here is the traceback I see:

```
Traceback (most recent call last):
  File "/home/mwaskom/anaconda2/envs/py36/bin/run_fmri.py", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/mwaskom/code/lyman/scripts/run_fmri.py", line 525, in <module>
    main(sys.argv[1:])
  File "/home/mwaskom/code/lyman/scripts/run_fmri.py", line 355, in main
    lyman.run_workflow(ffx, "ffx", args)
  File "/home/mwaskom/code/lyman/lyman/frontend.py", line 196, in run_workflow
    wf.run(plugin, plugin_args)
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/site-packages/nipype/pipeline/engine/workflows.py", line 590, in run
    runner.run(execgraph, updatehash=updatehash, config=self.config)
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/site-packages/nipype/pipeline/plugins/base.py", line 266, in run
    result=result))
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/site-packages/nipype/pipeline/plugins/base.py", line 306, in _clean_queue
    raise RuntimeError("".join(result['traceback']))
RuntimeError: Traceback (most recent call last):
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/site-packages/nipype/pipeline/plugins/base.py", line 255, in run
    result=result))
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/site-packages/nipype/pipeline/plugins/base.py", line 306, in _clean_queue
    raise RuntimeError("".join(result['traceback']))
RuntimeError: Traceback (most recent call last):
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/site-packages/nipype/interfaces/io.py", line 72, in copytree
    os.makedirs(dst)
  File "/home/mwaskom/anaconda2/envs/py36/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/home/mwaskom/studies/fluid/analysis/nback-block_load/gf52/ffx/mni/smoothed/n0'

[...]

PR #2092, opened 2017-06-23
Fix: String-based error inspection on Python 3

mwaskom:
See https://github.com/nipy/nipype/issues/2091

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -71,7 +71,7 @@ def copytree(src, dst, use_hardlink=False):
     try:
         os.makedirs(dst)
     except OSError as why:
-        if 'File exists' in why:
+        if 'File exists' in str(why):
             pass
         else:
             raise why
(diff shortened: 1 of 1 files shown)


effigies:
Since these are `OSError`s, can you do `in why.strerror` and `in inst.strerror`? While `OSError.__str__()` will probably always include `strerror`, it doesn't seem to be part of the documented API, so it seems better to me to check the `strerror` attribute.

mwaskom:
I couldn't find anything in the docs one way or the other about the preferred API but I am trying it your way.

nipype/interfaces/io.py
```diff
@@ -86,7 +86,7 @@ def copytree(src, dst, use_hardlink=False):
                 copyfile(srcname, dstname, True, hashmethod='content',
                          use_hardlink=use_hardlink)
         except (IOError, os.error) as why:
-            errors.append((srcname, dstname, str(why)))
+            errors.append((srcname, dstname, why.strerror))
```
mwaskom: Done.

PR #2105, opened 2017-06-28
FIX: adding ICA_AROMA to fsl __init__.py

jdkent:
Fixes: being able to import ICA_AROMA like other fsl packages

- adds ICA_AROMA to fsl/interfaces/\_\_init\_\_.py


effigies:
Unfortunately, this is changing the API, since `fsl.interfaces.ICA_AROMA` is the module, not the class, so changing this has the potential to induce errors for anybody using this.

If it's worth making the change, we should make the name of the class and the module different, along with importing the class into `__init__.py`. I'd suggest changing the file name to `ica_aroma.py` or just `aroma.py`, and I'm inclined to leave the class name the same.

@satra I'd appreciate your input on a potential API change.

satra:
>I'd suggest changing the file name to ica_aroma.py or just aroma.py, and I'm inclined to leave the class name the same.

i like that `from .aroma import ICA_AROMA`

let's make the file name change and leave the class intact.

PR #2107, opened 2017-06-29
ENH: Add cosine-basis high-pass-filter to CompCor

effigies:
- Adds a high-pass filter using the discrete cosine transform
- Enabled with `high_pass_filter`
- TR can be specified as an input or derived from the NIfTI header
- Cosine regressors can be saved

Todo:

* [x] `make specs`
* [x] Downstream testing (poldracklab/fmriprep#577

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -325,7 +325,7 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
     components_file = traits.Str('components_file.txt', usedefault=True,
                                  desc='Filename to store physiological components')
     num_components = traits.Int(6, usedefault=True) # 6 for BOLD, 4 for ASL
-    use_regress_poly = traits.Bool(True, usedefault=True,
+    use_regress_poly = traits.Bool(True, usedefault=True, xor=['high_pass_filter'],
                                    desc=('use polynomial regression '
                                          'pre-component extraction'))
     regress_poly_degree = traits.Range(low=1, default=1, usedefault=True,
@@ -334,7 +334,7 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
                                      'file (one column). If undefined, will '
                                      'default to "CompCor"'))
     high_pass_filter = traits.Bool(
-        False, usedefault=True,
+        False, xor=['use_regress_poly'],
         desc='Use cosine basis to remove low-frequency trends pre-component '
              'extraction')
     high_pass_cutoff = traits.Float(
@@ -437,7 +437,8 @@ class CompCor(BaseInterface):
 
         components, hpf_basis = compute_noise_components(
             imgseries.get_data(), mask_images, degree,
-            self.inputs.num_components, self.inputs.high_pass_filter, TR)
+            self.inputs.num_components, self.inputs.high_pass_filter,
+            self.inputs.high_pass_cutoff, TR)
 
         components_file = os.path.join(os.getcwd(), self.inputs.components_file)
         np.savetxt(components_file, components, fmt=b"%.10f", delimiter='\t',
@@ -838,7 +839,7 @@ def is_outlier(points, thresh=3.5):
     return timepoints_to_discard
 
 
-def cosine_filter(data, timestep, period_cut, remove_mean=False, axis=-1):
+def cosine_filter(data, timestep, period_cut, remove_mean=True, axis=-1):
     datashape = data.shape
     timepoints = datashape[axis]
 
@@ -952,7 +953,8 @@ def combine_mask_files(mask_files, mask_method=None, mask_index=None):
 
 
 def compute_noise_components(imgseries, mask_images, degree, num_components,
-                             high_pass_filter=False, repetition_time=0):
+                             high_pass_filter=False, period_cut=128,
+                             repetition_time=0):
     """Compute the noise components from the imgseries for each mask
 
     imgseries: a nibabel img
@@ -966,6 +968,7 @@ def compute_noise_components(imgseries, mask_images, degree, num_components,
     returns:
 
     components: a numpy array
+    hpf_basis: a numpy array, if high_pass_filter is True, else None
 
     """
     components = None
@@ -983,16 +986,14 @@ def compute_noise_components(imgseries, mask_images, degree, num_components,
         # Zero-out any bad values
         voxel_timecourses[np.isnan(np.sum(voxel_timecourses, axis=1)), :] = 0
 
+        # Use either cosine or Legendre-polynomial detrending
         if high_pass_filter:
-            # If degree == 0, remove mean in same pass
             voxel_timecourses, hpf_basis = cosine_filter(
-                voxel_timecourses, repetition_time,
-                self.inputs.high_pass_cutoff, remove_mean=(degree == 0))
-
-        # from paper:
-        # "The constant and linear trends of the columns in the matrix M were
-        # removed [prior to ...]"
-        if degree > 0 or not high_pass_filter:
+                voxel_timecourses, repetition_time, period_cut)
+        else:
+            # from paper:
+            # "The constant and linear trends of the columns in the matrix M were
+            # removed [prior to ...]"
             voxel_timecourses = regress_poly(degree, voxel_timecourses)
 
         # "Voxel time series from the noise ROI (either anatomical or tSTD) were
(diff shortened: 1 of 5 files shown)


effigies:
This is ready to go.

satra:
@effigies - the code looks reasonable.

why are we adding high pass filtering to compcor? i can't recollect this being in the behzadi paper. further the polynomial regression achieves a similar effect. it's not a cosine basis, but the functionality is the same.

nipype/algorithms/confounds.py
```diff
@@ -794,6 +835,27 @@ def is_outlier(points, thresh=3.5):
...
 
 
+def cosine_filter(data, timestep, remove_mean=False, axis=-1):
+    datashape = data.shape
+    timepoints = datashape[axis]
+
+    data = data.reshape((-1, timepoints))
+
+    frametimes = timestep * np.arange(timepoints)
+    X = _full_rank(_cosine_drift(128, frametimes))[0]
```
satra: this number appears to be coming from the spm default high-pass filter. if this is indeed necessary shouldn't the high-pass filter cutoff be a variable?

effigies:
From Behzadi et al.

> These low-frequency components can appear in the CompCor spectra because we remove only constant and linear trend terms prior to the principal component analysis. Note that, as stated in the Methods section, the use of the discrete cosine transform (DCT) low-frequency nuisance terms is reserved for the statistical assessments performed with the GLM. As a result, the space spanned by the principal components may partially overlap with the space spanned by the discrete cosine transform nuisance terms.

I'm definitely not an expert on the theory here, and certainly the Legendre basis set seems equally valid. One distinction I would draw between these approaches is in selecting the degree; perhaps there's a way to select the polynomial order based on frequency cutoff, but it's a very natural approach with a DCT basis. At present, the default order is 1, and I think I've seen "order 2 polynomial detrending" in papers to account for nonlinear drift.

This is my first foray into CompCor territory, so I don't have strong opinions here. Arguments are welcome.

Also, I did write this to be "side-by-side" compatible with polynomial detrending, but if it makes more sense to be an `xor` situation, let me know.

nipype/algorithms/confounds.py
effigies: Yeah, I'll add that.

nipype/algorithms/confounds.py
```diff
@@ -945,3 +1022,69 @@ def _compute_tSTD(M, x, axis=0):
     stdM[stdM == 0] = x
     stdM[np.isnan(stdM)] = x
     return stdM
+
+
+# _cosine_drift and _full_rank copied from nipy/modalities/fmri/design_matrix
```
chrisfilo: Is there a a big disadvantage of depending on nipy here instead of copying the code?

chrisfilo:
Hi @satra - highpass filtering has been requested as a more explicit form of detrending by an FMRIPREP user. The idea is to capture nontrivial variation with the CompCor components.

Providing a highpass cutoff rather than polynomial order seem to be a more interpretable way to do this kind of detrending. The functionality is optional and should not interfere with the traditional uses.

nipype/algorithms/confounds.py
effigies: Originally I did it simply to avoid adding a dependency. I ended up [augmenting](https://github.com/nipy/nipype/pull/2107/commits/aed10c8f843c14773b1ceaf4845d23f68fc83cc4) it to set a minimum order of 1.

satra:
@effigies - it would be good to make regress_poly and high_pass_filter be xor - since they are supposed to achieve similar end goals. if we are going to do high-pass-filter does it make sense to simply do bandpass filter? i don't know why it should be high-pass in particular?

@chrisfilo - on the user side, the goal of compcor is to remove nuisance projectors. to some extent this overlaps with the notion of temporal filtering. the reason a basic polynomial regressor was used was to clearly take out linear/quadratic trends as per the paper. by allowing for more general filtering, users will have to be careful as to what they are really meaning by nuisance and what impact such filtering has on subsequent steps. finally, by bringing more general filtering into the picture, everything about filter design comes into play. something often ignored in the algorithms that implement this.

chrisfilo:
+1 for xor

- bandpass might be a bit harder to implement (I'm not sure it can be done
just with cosines) - I would leave it for another PR

- I do agree this might seem more complex. The idea is that users wanting
to limit CompCor just to higher frequencies should include the high pass
cosine regressors returned by this interface. This way you can get the full
set of regressors used in CompCor for denoising (high pass cosines and PCA
components). From what I am hearing from users this is a valuable option to
have.




On Jul 5, 2017 2:05 PM, "Satrajit Ghosh" <notifications@github.com> wrote:

@effigies <https://github.com/effigies> - it would be good to make
regress_poly and high_pass_filter be xor - since they are supposed to
achieve similar end goals. if we are going to do high-pass-filter does it
make sense to simply do bandpass filter? i don't know why it should be
high-pass in particular?

@chrisfilo <https://github.com/chrisfilo> - on the user side, the goal of
compcor is to remove nuisance projectors. to some extent this overlaps with
the notion of temporal filtering. the reason a basic polynomial regressor
was used was to clearly take out linear/quadratic trends as per the paper.
by allowing for more general filtering, users will have to be careful as to
what they are really meaning by nuisance and what impact such filtering has
on subsequent steps. finally, by bringing more general filtering into the
picture, everything about filter design comes into play. something often
ignored in the algorithms that implement this.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/nipy/nipype/pull/2107#issuecomment-313181252>, or mute
the thread

nipype/algorithms/confounds.py
```diff
@@ -323,21 +323,35 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
                               desc=('Position of mask in `mask_files` to use - '
                                     'first is the default.'))
     components_file = traits.Str('components_file.txt', usedefault=True,
-                           desc='Filename to store physiological components')
+                                 desc='Filename to store physiological components')
     num_components = traits.Int(6, usedefault=True) # 6 for BOLD, 4 for ASL
-    use_regress_poly = traits.Bool(True, usedefault=True,
+    use_regress_poly = traits.Bool(True, usedefault=True, xor=['high_pass_filter'],
```
effigies: After adding the `xor` entries, in order to use `high_pass_filter`, we need to turn off `usedefault` here, which means we could be breaking existing scripts in one way or another.

Given that, maybe the better way to do this is to switch to an `Enum`, such as

```Python
pre_filter=traits.Enum('polynomial', 'high_pass', False, usedefault=True,
                       desc='Detrend time series prior to component extraction')
```

WDYT?

chrisfilo:
Sounds like a great idea to me, but we have to make sure it's backwards
compatible.

effigies:
Okay. I've moved to an `Enum`-based filter selection, and deprecated `use_regress_poly`. It is no longer set `True` by default, but the `pre_filter=='polynomial'` default is equivalent. If `use_regress_poly==True`, `pre_filter` is overridden. I'm counting on nipype's deprecation warnings to let people know to move off of it.

I'm also generally making the filtering regressors available. There's no cost to keeping them around if people want.

nipype/algorithms/confounds.py
```diff
@@ -323,21 +323,34 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
...
     components_file = traits.Str('components_file.txt', usedefault=True,
-                           desc='Filename to store physiological components')
+                                 desc='Filename to store physiological components')
     num_components = traits.Int(6, usedefault=True) # 6 for BOLD, 4 for ASL
-    use_regress_poly = traits.Bool(True, usedefault=True,
+    pre_filter = traits.Enum('polynomial', 'cosine', False, usedefault=True,
+                             desc='Detrend time series prior to component '
+                                  'extraction')
+    use_regress_poly = traits.Bool(True,
+                                   deprecated='0.15.0', new_name='pre_filter',
```
effigies: @satra I set this to deprecate at 0.15.0, two minor versions in the future. Should that be 3.0.0, instead? It's pretty unclear what a deprecation schedule should look like.

effigies:
@satra I think this is ready to go, but want your input on the deprecation.

satra:
@effigies - looks fine to me. at this point it's possible it will get deprecated with the planned 1.0 release in fall. i think that's ok.

PR #2113, opened 2017-07-15
WIP/ENH: Dask plugin

effigies:
This is a new plugin using [dask](http://dask.pydata.org/en/latest/), a parallel distribution engine that handles a huge amount of the details of intra- and inter-node IPC with a common interface. Its default invocation is effectively synonymous with `MultiProc`, although it's a little heavier since it uses Tornado under the hood.

The basic goal is to push a lot of the scheduling down the stack. For instance, we're not limited to the vagaries of a topological sort; dask does heuristic timing of nodes and can predict (some) node runtimes and adjust the run order. Further, it allows us to pass new jobs from within workers (which is how I've implemented `MapNode` distribution). This may be a useful avenue for implementing conditional or other control-flow nodes.

Finally, there are built-in profiling and job-completion report features, which would be very nice to have available in large nipype pipelines.

@oesteban is working on a launcher-like wrapper that spins up workers across nodes in a TACC job. By changing the plugin parameters, nipype could distribute jobs across nodes with the same plugin.

This preliminary work was built at the SciPy 2017 sprint.

TODO:

* [ ] Engage in resource management
* [ ] Possibly enable using `dask.multiprocessing` instead of `dask.distributed` to allow users to avoid Tornado overhead on local runs
* [ ] Remove debug code

### DIFF
--- a/nipype/pipeline/plugins/__init__.py
+++ b/nipype/pipeline/plugins/__init__.py
@@ -18,6 +18,7 @@ from .sgegraph import SGEGraphPlugin
 from .lsf import LSFPlugin
 from .slurm import SLURMPlugin
 from .slurmgraph import SLURMGraphPlugin
+from .dask import DaskPlugin
 
 from .callback_log import log_nodes_cb
 from . import  semaphore_singleton
--- a/nipype/pipeline/plugins/dask.py
+++ b/nipype/pipeline/plugins/dask.py
@@ -41,9 +41,12 @@ def run_node(node, updatehash, *args):
         dask_graph = {
             snode.fullname: (partial(run_node, snode, updatehash), [])
             for snode in subnodes}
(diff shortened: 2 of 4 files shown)


oesteban:
> Possibly enable using `dask.multiprocessing` instead of `dask.distributed` to allow users to avoid Tornado overhead on local runs

AFAIK the downside of `dask.multiprocessing` is the GIL, so in practice `distributed` is faster.

effigies:
I'm pretty sure there was a `get` that depended on Python's `multiprocessing` module, which isn't subject to the GIL. But it might not have been `dask.multiprocessing`. Either way, worth looking into.

effigies:

effigies:
Reopened to add a failing script to the diff. Someone else is still welcome to take the lead on reviving this one.

PR #2117, opened 2017-07-18
fix: retrieve aseg and wmparc stats properly

satra:
freesurfersource did not receive aseg and wmparc stats properly. this address the issue.


effigies:
Yup. Looks good.

PR #2122, opened 2017-07-19
ENH: Allow CompCor to skip initial volumes

effigies:
Follow-up to #2107.

- Permit CompCor to ignore initial volumes, intended to take the output of the `NonSteadyStateDetector` interface
- Adds a "SteadyState" column to the `filter_basis`
- `filter_basis` and `components` are calculated on the truncated series, then zero-padded to fit original series

Todo:

* [x] Downstream testing


satra:
@effigies - zero-padding will change the filter characteristics. if necessary this should be applied only to the pca components. on the other hand you can add N columns with a 1 at the location for every volume that is discarded (similar to ArtifactDetect).

nipype/algorithms/confounds.py
```diff
@@ -451,6 +470,12 @@ def _run_interface(self, runtime):
                      'cosine': 'cos'}[self.inputs.pre_filter]
             ncols = filter_basis.shape[1] if filter_basis.size > 0 else 0
             header = ['{}{:02d}'.format(ftype, i) for i in range(ncols)]
+            if nvols:
+                ss_col = np.zeros((components.shape[0], 1),
+                                  dtype=filter_basis.dtype)
+                ss_col[:nvols] = 1
```
chrisfilo: I think we should model each volume with a separate regressor. Like artifacts in ArtDetect

effigies:
@satra Ah, you're right. The "steady state" column should be N columns with a single 1 in each.

But to be clear, the data is truncated, the filter basis is constructed and applied before extracting the components. After that, the filter basis is zero-padded, to include in the GLM along with the components. Does that resolve the issue for you, or is there a problem I'm missing?

effigies:
@satra I believe the CircleCI issues are unrelated to this PR.

I also added docs to explain the behavior of this change and the high-pass filter. Let me know what you think.

effigies:
Sorry. Realized a bug was being picked up downstream. Will check the box when ready.

effigies:
@satra Sorry to bug you, but do you have any issues with this as it stands?

satra:
lgtm

Issue #2124, opened 2017-07-24
Missing coma in inputs traits definition of FSL FNIRT


mmromero:
### Summary
Definition of [`FNIRT.inputs.intensity_mapping_model`](https://github.com/nipy/nipype/blob/0b3a116997eda4c364fea2b0e4e3dc902b1ede4e/nipype/interfaces/fsl/preprocess.py#L880) is missing a coma between 'global_non_linear' and 'local_linear' and thus both options are unusable. 

### Script/Workflow details

[`FNIRT.inputs.intensity_mapping_model`](https://github.com/nipy/nipype/blob/0b3a116997eda4c364fea2b0e4e3dc902b1ede4e/nipype/interfaces/fsl/preprocess.py#L880)

### Platform details:

{'sys_platform': 'darwin', 'scipy_version': '0.19.1', 'networkx_version': '1.11', 'commit_source': '(none found)', 'commit_hash': '<not found>', 'sys_version': '3.5.3 | packaged by conda-forge | (default, May 12 2017, 15:35:12) \n[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]', 'nipype_version': '1.0.0-dev', 'sys_executable': '/Users/miguel/anaconda/bin/python', 'numpy_version': '1.12.1', 'traits_version': '4.6.0', 'pkg_path': '/Users/miguel/anaconda/lib/python3.5/site-packages/nipype', 'nibabel_version': '2.1.0'}
1.0.0-dev

### Execution environment

Choose one
- Container [Tag: ???]
- My python environment inside container [Base Tag: ???]
- My python environment outside container

mgxd:
closed in #2125

PR #2125, opened 2017-07-24
fix to #2124

djarecka:
adding the missing coma in fnirt


satra:
lgtm - merge when travis passes - there are some weird issues with circle: timeouts, provenance errors, etc.,. seems to show up differently on different cases. need to check on docker builds and such.

PR #2134, opened 2017-07-27
enh: selectfiles select(only)dirs too

### LINKED ISSUE #2132: How to use SelectFiles to read all directories inside a folder?
Here is my data_organization function. This is used in the first node of my workflow. 
```
def data_organization(in_path):
    do_something...
    return out_path
```
The function receives `in_path` as an input (a directory) which is a path of one subject folder. Then it organizes all the sub-folders in a desired way and it returns the path of the organized folder. Folders' name have a pattern. As `SelectFiles` finds **files** with specific patterns, is it possible to use it to find specific **directories** and pass them other nodes (e.g. to the above node)?

mgxd:
Fixes #2132  .

- ENH: exclusively select directories


satra:
could you please add a test with some fake directories and ensure that the test covers scenarios where there is and is not an os.sep at the end of the template

quick question: why couldn't the user at the separator at the end? is this addition necessary?

satra:
on 20s of further thought, i'm not sure this is change is essential, as a user can easily add the separator, no?

mgxd:
@satra the separator will be removed since [`abspath` is called before the `glob` search](https://github.com/nipy/nipype/blob/master/nipype/interfaces/io.py#L1300-L1304)

effigies:
@mgxd Maybe just check `template[-1] == os.sep` before applying `abspath`, and append it afterwards if it was there?

mgxd:
@effigies I like that better than my initial solution

nipype/interfaces/io.py
```diff
@@ -1296,13 +1296,21 @@ def _list_outputs(self):
 
         for field, template in list(self._templates.items()):
 
+            find_dirs = False
+            if template[-1] == os.sep:
+                find_dirs = True
```
effigies:
```Python
find_dirs = template[-1] == os.sep
```

nipype/interfaces/io.py
```diff
@@ -1186,17 +1186,17 @@ def _list_outputs(self):
...
-                                default=False, usedefault=True,
-                                desc=("Whether to return outputs as a list even when only one file "
-                                      "matches the template. Either a boolean that applies to all "
-                                      "output fields or a list of output field names to coerce to "
-                                      " a list"))
+        default=False, usedefault=True,
+        desc=("Whether to return outputs as a list even when only one file "
+              "matches the template. Either a boolean that applies to all "
+              "output fields or a list of output field names to coerce to "
+              " a list"))
```
effigies: I think these changes should be reverted, or the arguments on the first line of each trait should also be dropped to a newline and un-indented.

mgxd:
GTK-thanks!

PR #2137, opened 2017-08-01
FIX: Set prov==1.5.0 dependency

effigies:
This PR explores the issues with the currently failing builds.

Currently checking whether the prov 1.5.1 release is the issue.


effigies:
prov 1.5.1 seems to be the issue.

@satra do we want to go ahead and merge this for now, and then start a new branch for relaxing the constraint and fixing the underlying issues?

PR #2139, opened 2017-08-01
fix #2115: changing range into list in MultiPath class

### LINKED ISSUE #2115: iterfield in MapNode can't be range (py3)
### Summary
In py3 `range` is not a list, and if iterfield in `MapNode` is defined as `range` than entire range is send to the interface at once.  

### Script/Workflow details

```
from nipype import Function, MapNode

def square_func(x):
    print(x, type(x))
    return x ** 2

square = Function(["x"], ["f_x"], square_func)

square_node = MapNode(square, name="square", iterfield=["x"])

square_node.inputs.x = range(4)
res  = square_node.run()
```

### Actual behavior
print statement from `square_func` gives `range(0, 4) <class 'range'>`

### Expected behavior
print statement from `square_func` should give  `0 <class 'int'>`, etc.

djarecka:
Checking if type of value is `range` (py3). If this is a case the value is changed to a list.

It fixes the problem in `MapNode`, now if `iterfield` is a `range` it is treated as a single element list - `[range()]`.


effigies:
Is there any reason not to make this work for any `collections.Iterable` (excepting `str`)?

djarecka:
Not sure... Can try to change, but exception would be needed at least for `str` and `dict` I guess. Anything  else you can think of that shouldn't be included?

effigies:
Good point. Maybe `collections.Sequence` would be better than `Iterable`, to exclude mappings.

But... this also might be over-engineering. Maybe carving an exception to make `range` behave nicely is simpler and covers actual use cases better.

satra:
@djarecka - can you make sure to merge master.

djarecka:
@effigies : ok, i change to a sequence, but noticed that string is also a sequence...
I actually moved the if statement to the very beginning of `validate`,  so sequences are treated exactly as lists. Let me know if those changes make sense now.

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -456,6 +456,22 @@ def test_mapnode_iterfield_check():
     with pytest.raises(ValueError): mod1._check_iterfield()
 
 
+@pytest.mark.parametrize("x_inp, f_exp", [
+        (3, [9]), ([2, 3], [4, 9]), ((2,3), [4,9]), (range(3), [0, 1, 4])
+        ])
+def test_mapnode_iterfield_type(x_inp, f_exp):
+    from nipype import MapNode, Function
+    def square_func(x):
+        return x ** 2
```
effigies: What about make it:

```Python
def double_func(x):
    return x * 2
double = Function(["x"], ["f_x"], double_func)
```

Then you can test that strings are treated as expected, as well. (`'x' * 2 == 'xx'`)

djarecka: done

effigies:
LGTM.

PR #2155, opened 2017-08-17
ENH: Allow input weight for afni's volreg.

williford:
This PR should be discussed before accepting.

I've added the following to `VolregInputSpec`:
```
    in_weight_file = File(
        desc='File for input weighting volume',
        argstr="-weight '%s[0]'",
        exists=True)
```
Apparently, `3dvolreg` requires the volume number for the weight. I don't know of a good way to do this with NiPype, so I make it just take the first volume (I think typically someone would only pass a single volume image). Is this sufficient?

Also there doesn't seem to be a standard way for naming the input weights. I went with a more verbose version, but feedback is welcome.

I tried to run `make check-before-commit` and received the error "ImportError: No module named packaging.version". I'm running the code in virtualenv with Python3, using pip to install.

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -119,7 +119,7 @@ class AlignEpiAnatPyOutputSpec(TraitedSpec):
         desc="matrix to volume register and align epi"
              "to anatomy and put into standard space")
     epi_vr_motion = File(
-        desc="motion parameters from EPI time-series" 
+        desc="motion parameters from EPI time-series"
              "registration (tsh included in name if slice"
              "timing correction is also included).")
     skullstrip = File(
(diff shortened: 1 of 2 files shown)


effigies:
One option is to use a tuple trait, which would require the user to specify the volume.

```Python
in_weight_volume = traits.Tuple(File(exists=True), traits.Int, argstr="-weight '%s[%d]'", ...)
```

However, using the 0th volume seems like a reasonable default, so what about the following:

```Python
in_weight_volume = traits.Either(traits.Tuple(File(exists=True), traits.Int),
                                 File(exists=True), argstr="-weight '%s[%d]'", ...)
```

Then, in the interface class, override `_format_arg`:

```Python
    if name == 'in_weight_volume' and not isinstance(value, Tuple):
        value = (value, 0)
    return super(Volreg, self)._format_arg(name, trait_spec, value)
```

Thus, if you pass `volreg.inputs.in_weight_volume = 'functional.nii'`, you get `-weight 'functional.nii[0]'`, but if you pass `volreg.inputs.in_weight_volume = ('functional.nii', 10)`, you get `-weight 'functional.nii[10]'`. (I'd add these to the docstring to make sure that it works as I expect.)

satra:
@williford - could you please merge with current master and update this PR w.r.t @effigies comments?

effigies:
Hi @williford. Do you have a few minutes to finish this? It would be nice to have it in for 1.0.

If you don't have the time, I can push my suggested changes to your branch.

williford:
@effigies Sorry for my delay! I just implemented your suggestions, updated the minimal testing, and enabled maintainers to modify my PR. Feel free to make the changes you want.

I did get one error when running tests, but it was unrelated to volreg. The volreg tests passed.

williford:
I failed to fetch the current changes from nipype master branch...

williford:
I'm closing this PR, since I created a clean version that is up-to-date with Master here:
#2396

PR #2163, opened 2017-08-22
FIX: Detecting and apropriately warning about unconnected duplicatlely nodes

TheChymera:
If the workflow contained an unconnected node with a duplicate name, the check failed with a nondescript `IndexError`, which required the user to either manually re-check their entire connections list for any errors at all, or hack the nipype code with print calls to find out which node is triggering the exception.

Now, the check gives the correct name duplication error and the name of the node, even if the node has no lineage.


oesteban:
Could you write a unit test for this? That'd be very much appreciated :)

TheChymera:
@oesteban  added :)

TheChymera:
@oesteban @satr ping?

PR #2167, opened 2017-08-23
enh: calcmedian interface

mgxd:
- ENH: `CalculateMedian` interface that takes one or more 4D realigned timeseries and returns the average across each timeseries' medians.


nipype/algorithms/misc.py
```diff
@@ -1380,6 +1380,51 @@ def merge_rois(in_files, in_idxs, in_ref,
     return out_file
 
 
+class CalculateMedianInputSpec(BaseInterfaceInputSpec):
+    in_file = InputMultiPath(File(exists=True, mandatory=True,
```
satra: `in_files` since multiple files are acceptable.

nipype/algorithms/misc.py
```diff
@@ -1380,6 +1380,51 @@ def merge_rois(in_files, in_idxs, in_ref,
...
+      desc="One or more realigned Nifti 4D timeseries"))
+    median_file = traits.Str('median.nii.gz', usedefault=True,
+      desc="Filename to store median image")
```
satra: perhaps add a boolean flag to indicate whether the median should be computed individually or across files. and generate the appropriate outputs based on the flag.

nipype/algorithms/misc.py
```diff
@@ -1380,6 +1380,51 @@ def merge_rois(in_files, in_idxs, in_ref,
     return out_file
 
 
+class CalculateMedianInputSpec(BaseInterfaceInputSpec):
+    in_file = InputMultiPath(File(exists=True, mandatory=True,
+      desc="One or more realigned Nifti 4D timeseries"))
+    median_file = traits.Str('median.nii.gz', usedefault=True,
```
satra: use name_source and template

mgxd: I defined a `_gen_fname` instead that should handle multiple `in_files` accordingly

PR #2173, opened 2017-08-30
fix: allow translation thresholding

### LINKED ISSUE #2165: BUG: Cannot use translation_threshold in ArtifactDetect
### Summary

ArtifactDetect currently doesn't allow `translation_threshold` to be used, because it always thinks `norm_threshold` is set.

### Actual behavior

Error is thrown: "ArtifactDetect requires a value for input 'use_norm' because one of norm_threshold is set. For a list of required inputs, see ArtifactDetect.help()"

### Expected behavior

ArtifactDetect should use `translation_threshold` or (in the reduced example below) it should throw an error because the files are empty.

Create an empty file called "empty.file" (`touch empty.file`), then run the following file (python2 will probably give the same result).

### Script/Workflow details

```
#!/usr/bin/env python3

import nipype.algorithms.rapidart as ra

art = ra.ArtifactDetect(
    mask_type='file',
    translation_threshold=1.0,  # translation in mm
    rotation_threshold=0.02,  # rotation in radians
    zintensity_threshold=3.0,  # z-score
    use_norm=False,
    parameter_source='AFNI',
    save_plot=True,
    realignment_parameters='empty.file',
    realigned_files='empty.file',
    mask_file='empty.file',
)
art.run()
```

### Platform details:

Please paste the output of: `python -c "import nipype; print(nipype.get_info()); print(nipype.__version__)"`

> {'traits_version': '4.6.0', 'commit_hash': 'ab14a0e', 'pkg_path': '/home/jonathan/.virtualenvs/mri-py3/lib/python3.4/site-packages/nipype', 'nibabel_version': '2.1.0', 'sys_version': '3.4.2 (default, Oct  8 2014, 10:45:20) \n[GCC 4.9.1]', 'nipype_version': '1.0.0-dev', 'scipy_version': '0.19.1', 'commit_source': 'installation', 'sys_executable': '/home/jonathan/.virtualenvs/mri-py3/bin/python', 'numpy_version': '1.13.1', 'networkx_version': '1.11', 'sys_platform': 'linux'}
> 1.0.0-dev

### Execution environment

I'm running the code in a virtualenv.

mgxd:
Fixes #2165.

- FIX: disable `use_norm` providing default value, effectively deprecating the input. If `norm_threshold` is defined, `ArtifactDetect` interface will assume `use_norm` is True. If not defined, will assume it's False.


kastman:
I think you may actually want to remove the use_norm bool altogether - setting `use_norm = False` still requires a norm_threshold even with the latest fix. I think this would be an appropriate time to remove it from the interface completely, and just select based on which set of mutually exclusive thresholds are provided?

mgxd:
@kastman yes, I agree - setting it to False will still require an input for `norm_threshold`. My only hesitation is backwards compatibility of previous scripts, as the old method of preforming calculations relied on whether the `use_norm` input was defined.

kastman:
I've been out of the loop - any chance nipype is going to 1.0 soon? ;) I think this is normally handled with a deprecation warning, right?

I suppose that future scripts going forward could know not to set `use_norm`, but it was still surprising to see the pathology. Guess that's the least-bad solution as long as it's documented.

satra:
@mgxd and @kastman - the original reason why `use_norm` became the default is that it doesn't make much sense to use separate thresholds (seemed quite arbitrary, what people were doing), while `use_norm` seemed a little more principled around computing the motion of the head from volume to volume.

Therefore to go back to individual thresholds, one would have to do: `....inputs.use_norm = Undefined` and we had hoped that may be a sufficient bias towards using `use_norm` :)

so i'm -1 on this particular change, but would be happy to see the documentation improved.

satra:
@mgxd and @kastman -  a ping here on status. i think a documentation improvement would be sufficient. what do you think?

kastman:
Thanks @mgxd - that's a clear and straightforward compromise. Looks good from a user perspective.

williford:
So the unit of `norm_threshold` is mm and it roughly estimates the maximum displacement of any point on the surface of the brain (by using a polyhedron representation)?

satra:
@williford - the unit is mm and reflects the max displacement of the center of the faces of a cube placed around the head and transformed by the same motion as the head.

effigies:
@satra I assume this is just a quick heuristic, but has anybody by chance worked out the relation of this measure to the max displacement of a given voxel in the brain? e.g. For supremum of voxel displacements `svd` and some factor `f` reasonably close to 1, it could be shown `norm in (svd, f * svd]`?

satra:
@effigies - the code for doing some of that is already in that module. this was based on a set of papers by chao gan on looking at voxel displacement maps. i think these can be useful in other situations. one could potential regress out these 4d maps instead of the motion parameters. they may provide more sensitivity at given voxel. they can also be used to estimate potential signal confounds in specific regions of the brain.

to answer your first question, i don't think anyone has looked at the relative contributions/differences of the two measures. we have simply used the composite norm as a confound measure and for determining outliers and it is highly correlated with framewise displacement.

PR #2171, opened 2017-08-27
[ENH] update for networkx dev version (should fix #2159)

djarecka:
The new code should work with both versions -- ver 1.x  and  2.x (dev).

summary of changes:
 - `*_iter` has been removed, changing `nodes_iter` to `nodes`, etc. `nodes` gives dict-like `NodeView` (iterate or change to list first to access a specific element)
 - `.node` has been removed, changing to `nodes[node]`
 - `add_node` and `add_edge` don't have `attr_dict` anymore, have to change `**dict`
 
 - `subgraph` gives a view and `.copy()` has different depth, created two versions of [this part](https://github.com/djarecka/nipype/blob/18a29890c12e25d70915fb14d0219dd75dd7e3ce/nipype/pipeline/engine/utils.py#L786) . Suggestions how to simplify are welcome, don't understand all details of those various copies.

**Note**, I changed also `interfaces.cmtk`, but don't have any tests, so don't know if it works.

Edit (by @effigies):

Closes #2159 
Closes #2194

PR #2174, opened 2017-09-01
ENH: BIDS Data Grabber

adelavega:
- Added a BIDSDataGrabber Interface. This interface has a similar API to DataGrabber, and wraps pybids grabbit to query files in a BIDS project.
  - Can filter results using pybids entities with dynamic, user-defined `infields`
  - Can define dynamic `outfields` to execute multiple queries, such as grabbing both func and anat files for a subject in a given run.
- Added a BIDS dataset (from pybids test data) to test the interface, and added tests using doctest
   - In the future this should be a git submodule


adelavega:
cc: @oesteban @satra @tyarkoni @chrisfilo

tyarkoni:
Do we need to duplicate the entire BIDS dataset from pybids? Since pybids is a dependency for this module, you should be able to grab the test data from the original source. I think if you just get bids.__file__ after `import bids` you should be ready to go.

adelavega:
Oh yeah, good call.

effigies:
Personally, I think it's reasonable to add pybids as a dependency. It's pretty lightweight.

tyarkoni:
Right now it is, but we will likely include pandas (and hence numpy etc.) as a dependency of pybids once we start adding transformation and models module. Not sure how people feel about that...

adelavega:
@satra comments 9/4/17: Consider adding optional and required outfields, such that an error is thrown is nothing is found for required outfields, and Undefined is passed if nothing found for optional outputs.

I'll work on getting travis & circle to work and ping you for a review.

nipype/interfaces/bids.py
```diff
@@ -0,0 +1,171 @@
...
+    from bids.grabbids import BIDSLayout
+except ImportError:
+    have_pybids = False
+else:
+    have_pybids = True
+
+from warnings import warn
+
+class BIDSDataGrabberInputSpec(DynamicTraitedSpec):
+    base_dir = traits.Directory(exists=True,
```
effigies: `Directory` is defined in [`interfaces.traits_extension.py`](https://github.com/nipy/nipype/blob/master/nipype/interfaces/traits_extension.py#L192-L223). Import from `interfaces.base`.

nipype/interfaces/bids.py
```diff
@@ -150,7 +154,18 @@ def _list_outputs(self):
...
+            args.update(filters)
+            filelist = layout.get(return_type='file',
+                                      **args)
+            if len(filelist) == 0:
+                msg = 'Output key: %s returned no files' % (
+                    key)
+                if self.inputs.raise_on_empty:
+                    raise IOError(msg)
+                else:
+                    warn(msg)
```
oesteban: I would advocate here for using nipype's logging:

At the top:

```
from .. import logging
(...)

LOGGER = logging.getLogger('workflows')
```

And here:

```
LOGGER.warning('Output key: %s returned no files', key)
```

nipype/interfaces/bids.py
```diff
@@ -0,0 +1,171 @@
...
+
+from warnings import warn
+
+class BIDSDataGrabberInputSpec(DynamicTraitedSpec):
+    base_dir = traits.Directory(exists=True,
+                                desc='Path to BIDS Directory.',
+                                mandatory=True)
+    output_query = traits.Dict(key_trait=Str,
+                               value_trait=traits.Dict,
+                               desc='Queries for outfield outputs')
```
chrisfilo: I wonder if we should give this field a more sensible default - for example {"func": {"modality": "func"}, "anat": {"modality": "anat"}}. This would save time figuring out the syntax for 80% of users.

chrisfilo:
Looks great - thanks for putting this together Alejandro.

nipype/interfaces/bids.py
adelavega: I'm a fan. The current default of outputting all items is probably almost never usable.

adelavega: Here's an other idea for sensible defaults: If no infields are defined, infields for all BIDS entities in project are automatically created, and `Undefined` by default. If user connects an input to them, great, if not, they are not passed to pybids. It saves having to define infields.

chrisfilo: Sounds like a great idea!

adelavega:
I'm having trouble getting the tests to run because pybids's test data is not getting copied over.

In travis (for example), NIPYPE_EXTRAS are installed using -e, like this:
    
    travis_retry pip install -e .[$NIPYPE_EXTRAS]

Installing pybids with -e *should* make the data folder available (which is not installed normally as its not a "package", on purpose). However, it looks like -e is supposed to be used with local files or VCS urls, and when used with pypi it just seems to do a normal install, without coping over the data. 

Any clues on how to get pip to download the whole package and then install in editable mode, so that non-package folders are available?

oesteban:
Hi @adelavega, that way you are installing nipype in editable mode, but NOT pybids. To do so, just add some lines after this https://github.com/nipy/nipype/blob/4a0bd3d928f522e16a97526f88b0b9c608db9d34/.travis.yml#L39:

```
git clone https://github.com/INCF/pybids.git
pushd pybids
pip install -e .
popd
```
It will get installed then in editable mode.

adelavega:
Sweet thanks. I was trying to avoid installing directly from github but if you don't mind, then that works for me!

adelavega:
Allright, I've incorporated all of your changes and got Travis CI working with the new tests. Any more feedback on "smarter" defaults is appreciated, and any hints on where to inject pybids pip -e installation into Circle.

adelavega:
Note: I changed the name of the file to `bids_utils` because otherwise python 2.7 imported that file instead of pybids in `import bids`

satra:
@adelavega  - let's make sure we remove any commit in this tree that adds data to the repo.

adelavega:
Okay @satra I re-wrote history.

adelavega:
Yay, all tests passing on Circle and Travis. @satra, feel free to make a final review whenever you're ready.

oesteban:
Sorry @adelavega, yes, that's the right place to insert the installation of pybids :+1:

nipype/interfaces/bids_utils.py
```diff
@@ -0,0 +1,169 @@
...
+            outfields = ['func', 'anat']
+            self.inputs.output_query = {
+                "func": {"modality": "func"},
+                "anat": {"modality": "anat"}}
+        else:
+            self.inputs.output_query = {}
+
+        # If infields is None, use all BIDS entities
+        if infields is None:
+            bids_config = join(dirname(gb.__file__), 'config', 'bids.json')
```
satra: as a future improvement to a GrabbitIT interface that could serve as a base class, could this be passed along as a parameter?

this would also allow arbitrary and newer layouts to be used.

satra:
perhaps this should be added to `io` as a reference, since people are used to pulling out any io related things. it can stay in this file.

`from nipype.interfaces.io import BIDS...`

if you are up for reorganizing `io` into a directory with a base class and separate files for the different pieces. let's at least do the `io` reference for this one and then re-organize in a separate PR.

otherwise this LGTM

nipype/interfaces/bids_utils.py
adelavega: Yes. The issue here is that I cannot get the entities of a project (or BIDS entities in general), without initializing a BIDSLayout with a specific dataset. So this was the only way I could think of to access the entities definition.

adelavega:
Great! I've added the reference @satra so feel free to merge.

oesteban:
Tests will require #2171 to be merged in first.

satra:
@adelavega - i think this is good to go, but just to be sure could you please merge with master? we had to fix a few things.

adelavega:
@satra done! travis passed so i assume circle will too soon

PR #2184, opened 2017-09-13
Removed exists True for the mask file when not used.

sulantha2006:
This causes an error if the mask is not specified. 'none' is not a file, so fails with File not found.

Fixes # .
This causes an error if the mask is not specified. 'none' is not a file, so fails with File not found.
-
-

### DIFF
--- a/nipype/interfaces/ants/visualization.py
+++ b/nipype/interfaces/ants/visualization.py
@@ -23,7 +23,7 @@ class ConvertScalarImageToRGBInputSpec(ANTSCommandInputSpec):
                        position=1)
     output_image = traits.Str('rgb.nii.gz', argstr='%s', usedefault=True,
                               desc='rgb output image', position=2)
-    mask_image = File('none', argstr='%s', exists=True,
+    mask_image = File('none', argstr='%s',
                       desc='mask image', position=3, usedefault=True)
     colormap = traits.Str(argstr='%s', usedefault=True,
                           desc=('Possible colormaps: grey, red, green, '


nipype/interfaces/ants/visualization.py
```diff
@@ -23,7 +23,7 @@ class ConvertScalarImageToRGBInputSpec(ANTSCommandInputSpec):
                        position=1)
     output_image = traits.Str('rgb.nii.gz', argstr='%s', usedefault=True,
                               desc='rgb output image', position=2)
-    mask_image = File('none', argstr='%s', exists=True,
+    mask_image = File('none', argstr='%s',
```
effigies: What about:

```Python
mask_image = traits.Either('none', File(exists=True), argstr='%s', ...)
```

That way we get the existence check while allowing 'none'.

effigies:
Also, run `make specs` and add the updated test for this interface.

satra:
@sulantha2006 - could you please update this with @effigies suggestion, and merge with current master? also the `_format_arg` function seems to be doing nothing in this interface and `list_outputs` can be supplanted by `name_source`.

nipype/interfaces/ants/visualization.py
oesteban: @sulantha2006, do you think you can modify this in order for us to get it merged? Thanks a lot for the contribution

effigies:
Hi @sulantha2006, are you going to have time to finish this up, soon?

effigies:

PR #2185, opened 2017-09-14
WIP: Dtitk

kesshijordan:
- Added DTITK interfaces (http://dti-tk.sourceforge.net/pmwiki/pmwiki.php)

Note: make check-before-commit is throwing an error (see below) I am having trouble figuring out. If anyone has suggestions, please let me know.

Thanks!
- Kesshi

____ ERROR collecting nipype/interfaces/dtitk/tests/test_auto_AffineTask.py ____
nipype/interfaces/dtitk/tests/test_auto_AffineTask.py:3: in <module>
    from ..registration import AffineTask
E   ValueError: attempted relative import beyond top-level package
____ ERROR collecting


mgxd:
@kesshijordan 
- Your travis build is failing because the doctests are actually being run - however dtitk isn't present in the environment. You can just add `# doctest: +SKIP` to the line in your doctest that has either `node.run()` or `node.cmdline`

- As for circle, could you try merging with current nipype master?

kesshijordan:
@mgxd I'm not sure how to debug this; if you have any suggestions they are most welcome. 

I merged with the current nipype master, but am getting the error, below, in the circleci.
+ echo 'These tests were designed to be run at 4x parallelism.'
These tests were designed to be run at 4x parallelism.

Are you familiar with the error, below? It's the same one I get when I try make check-before-commit and was having trouble figuring out.

Thanks!
Kesshi

____ ERROR collecting nipype/interfaces/dtitk/tests/test_auto_AffineTask.py ____
nipype/interfaces/dtitk/tests/test_auto_AffineTask.py:3: in <module>
    from ..registration import AffineTask
E   ValueError: Attempted relative import in non-package

mgxd:
@kesshijordan 

regarding travis - you're missing an `__init__.py` in your `dtitk/tests` directory.

for circle - did you start working on this branch awhile ago? It seems your circle build is not running on 4x parallelism, you may have to manually change that within your circleci build settings. Perhaps @oesteban can chime in here

oesteban:
Yes - for circle just go to https://circleci.com/gh/kesshijordan/nipype/edit#parallel-builds and set `4x` for your builds.

mgxd:
@kesshijordan you are close to passing the checks - you'll just have to run `make specs` and add the updated autotests

kesshijordan:
thank you @mgxd and @oesteban!

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,22 @@
+{
```
mgxd: this file isn't necessary

kesshijordan: Two quick questions:
1) There isn't anything in the directories from .cache on: Is there any reason not to remove everything including .cache? (or is it better to just delete lastfailed?)
2) I see the comment about lastfailed not being necessary: is that the only comment, or is there somewhere else I should be looking for other comments?

Thanks again, @mgxd!
Kesshi

satra: @kesshijordan - `lastfailed` is the only one that's included in the commit. probably got added by mistake. but in general nothing in `.cache` should be in the commit.

satra:
@kesshijordan and @mgxd - could this be reviewed/updated soon? i think it's very close.

.cache/v/cache/lastfailed
kesshijordan: Thanks, @satra. I removed the .cache file from git and have pushed the change. @mgxd is there anything else I need to address?

nipype/interfaces/dtitk/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""The dtitk module provides classes for interfacing with the `Diffusion
+Tensor Imaging Toolkit (DTI-TK)
+<http://dti-tk.sourceforge.net/pmwiki/pmwiki.php>`_ command line tools.
+
+Top-level namespace for dti-tk.
+"""
+
+# from .base import ()
+from ..dtitk.registration import (RigidTask, AffineTask, DiffeoTask,
```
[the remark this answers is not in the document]
kesshijordan: fixed

nipype/interfaces/dtitk/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+                                  ComposeXfmTask, diffeoSymTensor3DVolTask,
+                                  affSymTensor3DVolTask, affScalarVolTask,
+                                  diffeoScalarVolTask)
+from ..dtitk.utils import (TVAdjustOriginTask, TVAdjustVoxSpTask,
```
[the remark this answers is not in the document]
kesshijordan: fixed

satra:
is this still a wip, or is this ready for merge?

kesshijordan:
Still a WIP... I need to test the way I addressed the last few comments to make sure I don't mess anything up naming the outputs.

satra:
@kesshijordan - any chance we can get this in by the oct 31 deadline? we are going to cut a new release.

kesshijordan:
👍

kesshijordan:
To update/ask for help: @satra and @mgxd I've been running into some issues testing these modules, and I want to make sure they're all working before they're released so if one of you has a quick second to suggest something for item 1 that would be greatly appreciated (I just figured out what the issue was with item 2, but I don't know if this is a wider issue that should be addressed or just a problem for me, so I wanted to note it).

1) I suspect this is a rather basic question but if someone can suggest a better way it would greatly facilitate testing. Every time I make a change, I have to do python setup.py --force to be able to import the new dtitk module (otherwise, the module can't be found). This seems to only work in ipython... I would prefer to do this in a notebook or script so I can test more efficiently and I'm guessing my approach is wrong... what procedure should I be using to reinstall nipype with the changes so that I can import the module like I can for the others? (e.g. I can import nipype.interfaces.fsl but nipype.interfaces.dtitk fails)

2) I just figured out that instead of throwing the error from the command-line, I get a utf-8 decoding error back (see, below) if I put in inputs that cause DTI-TK to throw an error.

/Users/kesshijordan/repos/nipype/nipype/interfaces/base.py in _read(self, drain)
   1286         "Read from the file descriptor"
   1287         fd = self.fileno()
-> 1288         buf = os.read(fd, 4096).decode(self.default_encoding)
   1289         if not buf and not self._buf:
   1290             return None

UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 13: invalid start byte

mgxd:
@kesshijordan 

Question 1: installing in developer mode will really make your life easier, as you can make changes and see them in action as soon as you save. 

To do this, I would first recommend uninstalling all versions of nipype, then `cd` to your cloned repository and installing with `pip install -e .`

Question 2: what input caused this error? Py 2/3?

satra:
@kesshijordan - a recent PR from @oesteban takes care of the unicode issue. could you please make sure you are merged with current master?

oesteban:
@satra I think that PR did not affect the `'stream'` type of `terminal_output`.

@kesshijordan, in order to facilitate your debugging, could you set `terminal_output = 'file_split'` on your interface? That would avoid the problem you are getting (after you merge current master as Satra suggested).

I have opened an issue (link is above) for the UnicodeDecodeError problem.

kesshijordan:
Thanks, @oesteban! I set it when I use the interface (like the example below), and it will run without the UTF-8 error. This doesn't interfere with my debugging, but FYI: if I run it such that the command line would throw and error, the node runs until manually killed with control C. When you kill it and open up the stderr.nipype, the node appears to be running over and over with jumbled letters replacing the filenames. 

https://github.com/nipy/nipype/blob/1a7401328baf4aca0f163ca18d3a88b913412eb3/doc/users/saving_workflows.rst

kesshijordan:
Thanks for the pointers, @mgxd. I tried your suggestion for installing and it worked... this is much easier! Regarding question #2: the error occurred because I hadn't set some of the inputs the command line needed; I am using python 3 (though when I first ran this, I think I was using 2.7). I got the rigid registration interface to work and will modify the subsequent interfaces in the same way so they take two niftis and a set of parameters instead of a nifti and a text file containing a nifti (different functions in dti-tk). I'm having trouble getting the generalized file naming to work based on other examples (I keep getting errors that it isn't a "file name" when I manipulate the string so right now I'm just leaving it at only supporting .nii.gz so I can try to get the rest done before the deadline tomorrow.)

stderr.nipype
```diff
@@ -0,0 +1,3 @@
+/Users/kesshijordan/repos/dtitk-2.3.3-Darwin-x86_64/scripts/dti_diffeomorphic_reg: line 191: [: 1.0: integer expression expected
```
mgxd: you can remove this

stdout.nipype
```diff
@@ -0,0 +1,3 @@
+registering /Users/kesshijordan/ref_data/interscan_dys/Trio/IC004-1_dti_b2000_EC_tensorx1000_aff.nii.gz to /Users/kesshijordan/ref_data/interscan_dys/Prisma/IC004-4_dti_b2500_SC_EC_tensorx1000.nii.gz ...
```
mgxd: Also remove

nipype/interfaces/tests/test_auto_SimpleInterface.py
```diff
@@ -0,0 +1,16 @@
+# AUTO-GENERATED by tools/checkspecs.py - DO NOT EDIT
```
mgxd: remove this too

effigies:
Is this something we want in for 1.0? I merged master and updated the auto tests.

effigies:
Tests are passing. Is this good to merge? @mgxd @kesshijordan

PR #2187, opened 2017-09-15
[ENH] Support for multiple initial-moving-transforms for ants.Registration and ComposeMultiTransforms

Gilles86:
In my workflows I use multiple registration steps that I want to inspect and build upon each other. For that I think it would be useful if you can use multiple transforms to initialize ANTSRegistration and/or compose new transforms out of a list of transforms.

I tried to implement that in this pull request. It would be good if someone would review this code. The antsRegistration script and its interface are quite messy.

In a related note: the antsRegistration-interface currently outputs transforms in the order of which you would like to apply them. However, if you want to input them into ans.ApplyTransforms or, now, as initial_moving_transform for ants.Registration you should give them in the reverse order. I think that is quite unclear and not user-friendly. However, I can see that it's too late to reverse this now. Should ants.Registration maybe get an input flag which can inverse the transforms already for you?


satra:
@Gilles86 - regarding the reversing of transforms, there is already an output (`reverse_transforms`) that does this. also one can use the `composite_transform/inverse_composite_transform` if there is a no need to combine with other external transforms.

satra:
@Gilles86 - just checking in on the status of this. 

also a merge with current master would be advised.

Gilles86:
Hey @satra, 

Thanks for the tip about the reverse_transforms. I misunderstood that output field for a list of inverse transforms! Maybe we should extend the documentation a bit? What do you think?

I merged this branch with origin/master and made some new commits that make this branch pass all tests. Do we need anything else?

oesteban:
@Gilles86 - if you could extend the documentation as you proposed that'd be awesome. Since you have been looking into this the latest, you are probably the best indicated candidate to do so :)

Gilles86:
I updated the documentation with some information I think is useful.

nipype/interfaces/ants/utils.py
```diff
@@ -225,3 +225,46 @@ class AffineInitializer(ANTSCommand):
...
+    transforms = InputMultiPath(File(exists=True), argstr='%s', mandatory=True,
+                                position=3, desc='transforms to average')
+
+
+class ComposeMultiTransformOutputSpec(TraitedSpec):
+    output_transform = File(exists=True, desc='Composed transform file')
+
+
+class ComposeMultiTransform(ANTSCommand):
+    """
```
oesteban: Please add a one-line description of what ComposeMultiTransform does.

nipype/interfaces/ants/utils.py
```diff
@@ -225,3 +225,46 @@ class AffineInitializer(ANTSCommand):
...
+    >>> compose_transform.inputs.transforms = ['struct_to_template.mat', 'func_to_struct.mat']
+    >>> compose_transform.inputs.output_transform = 'func_to_template.mat'
+    >>> compose_transform.cmdline # doctest: +ALLOW_UNICODE
+    'ComposeMultiTransform 3 func_to_template.mat struct_to_template.mat func_to_struct.mat'
+    """
+    _cmd = 'ComposeMultiTransform'
+    input_spec = ComposeMultiTransformInputSpec
+    output_spec = ComposeMultiTransformOutputSpec
+
+    def _format_arg(self, opt, spec, val):
```
oesteban: This should not be necessary, please remove.

nipype/interfaces/ants/utils.py
```diff
@@ -225,3 +225,46 @@ class AffineInitializer(ANTSCommand):
 
     def _list_outputs(self):
         return {'out_file': os.path.abspath(self.inputs.out_file)}
+
+
+class ComposeMultiTransformInputSpec(ANTSCommandInputSpec):
+    dimension = traits.Enum(3, 2, argstr='%d', usedefault=True, mandatory=True,
+                            position=0, desc='image dimension (2 or 3)')
+    output_transform = File(argstr='%s', mandatory=True, position=1,
```
oesteban: Please use `name_source`:

```Python
        output_transform = File(argstr='%s', mandatory=True, position=1,
                                   name_source=['transforms'], name_template='%s_composed',
                                   desc='the name of the resulting transform.')
```

And remove member `_list_outputs` that will not be necessary anymore.

nipype/interfaces/ants/utils.py
Gilles86: If I don't include this I get an error:

```traits.trait_errors.TraitError: Cannot set the undefined 'dimension' attribute of a 'ANTSCommandInputSpec' object.```
```/src/nipype/nipype/interfaces/ants/utils.py:256: UnexpectedException```

Maybe this functionality is not yet implemented for ```ANTSCommand```?

oesteban: I think that'd be fixed by removing `mandatory=True` from the `dimension` input trait.

Gilles86:
Thanks for the feedback, @oesteban!

I think I implemented all of it. leaving out the input/outputspec-definition in the Interface-class doesn't work for me. Maybe you can have a look why this is.

nipype/interfaces/ants/utils.py
Gilles86: No :(.

oesteban: Yay! (it works on my settings at least :+1: )

Gilles86: Cool! This works for me too. Maybe it was too late yesterday :).

oesteban:
Sorry I forgot to update the specs. Hopefully this is the one passing.

PR #2190, opened 2017-09-19
create_featreg_preproc: reference run modification

oliver-contier:
ENH

Added possibility to specify which functional run the reference volume should be picked from to the function 'create_featreg_preproc'. The reference run can be selected by integer indexing or with a string ('first', 'middle', 'last') similar to how the reference volume is selected. Also added simple test functions respectively.

We ran an analysis with these changes on nipype version 0.11.0 and it worked fine.


nipype/workflows/fmri/fsl/preprocess.py
```diff
@@ -18,6 +18,25 @@ def getthreshop(thresh):
...
+    filemap = {'first': 0, 'last': -1, 'middle': len(files) // 2}
+
+    if isinstance(whichrun, str):
+        if whichrun not in filemap.keys():
+            raise(KeyError, 'Sorry, whichrun must be either integer index'
+                            'or string in form of "first", "last" or "middle')
+        else:
+            return files[filemap[whichrun]]
+
+    if isinstance(files, list):
```
satra: perhaps this if clause should be at the beginning to account for the potential that files is not a list, but a single filename.

oliver-contier:
Should now also work in case a single file (i.e. functional run) is given.

PR #2193, opened 2017-09-21
[WIP,ENH] Revision to the resource profiler

oesteban:
This PR revises the resource profiler with the following objectives:

  - Increase robustness (and making sure it does not crash nipype)
  - Extend profiling to all interfaces (including pure python)

The increase of robustness will be expected from:

  1. Trying to reduce (or remove at all if possible) the logger callback
     to register the estimations of memory and cpus. This could be achieved
     by making interfaces responsible or keeping track of their resources
     to then collect all results after execution of the node.

  2. Centralize profiler imports, like the config or logger object
     so that the applicability of the profiler is checked only once.

This first commit just creates one new module nipype.utils.profiler, and
moves the related functions in there.

**Important**: this PRs removes the old `filemanip` logger and replaces it by a more generic `utils`. Documentation has been updated accordingly.


oesteban:
Hi @satra, @effigies,

Before I get further with this PR, I'd love your feedback on these issues:

  - Renaming logger `filemanip` -> `utils` and attach events from `nipype.utils.filemanip`, `nipype.utils.provenance` and `nipype.utils.profiler` to it.
  - Make interfaces responsible of monitoring themselves, and do it in a separate process so nipype is robust against errors in monitoring (see [tear-up](https://github.com/oesteban/nipype/blob/7d953cc74020767d2e8f155d9d1f1b3746cb6486/nipype/interfaces/base.py#L1084-L1096) and [tear-down](https://github.com/oesteban/nipype/blob/7d953cc74020767d2e8f155d9d1f1b3746cb6486/nipype/interfaces/base.py#L1127-L1131)). Additionally: all interfaces can measure their performance now (not only commandline interfaces and niu.Function).
  - Since the interface logs itself and they are meant to work in shared filesystems, a new .prof file is generated in the interface base directory containing the traces of memory and cpus. That will be retrieved and finally the peaks saved in the runtime object, allowing for removal of the profiler callback and log. WDYT?.
  - I had to [simplify the run interface logic](https://github.com/oesteban/nipype/blob/7d953cc74020767d2e8f155d9d1f1b3746cb6486/nipype/interfaces/base.py#L1098-L1125), trying to have the least lines possible within `try` ... `except`.

effigies:
This all seems reasonable on its face. I haven't really looked into the profiler before, so I don't have strong opinions of cases to consider. I'll try to have a more detailed look and might have more to say...

setup.py
```diff
@@ -148,6 +148,7 @@ def main():
         entry_points='''
            [console_scripts]
            nipypecli=nipype.scripts.cli:cli
+           nipype_mprof=nipype.utils.profiler:main
```
satra: can we put this under `nipypecli`? it would be nice to have a single cli.

satra:
@oesteban - in principle this looks reasonable, here are my questions:

1. is the profiler stable across os x and linux? i see that duration is now being returned from the profiler, so that would mean that we need to have the profiler on to get this. i'm fine with that, but that would mean that the profiler should be on by default and be robust across platforms (i.e. not require sudo on some).

2. has this been tested with linux cgroups limiting threads? does this have an impact on performance, since this is running a separate process. as opposed to the previous case where it was a nested subprocess.

oesteban:
Thanks for the comments :)

I am currently finishing this and will start testing all those things soon.

Regarding the duration: I am still working on making the run function robuster, but in principle the storing the duration will be the same way it was even before the resource profiler was included.

I'm more concerned about the cgroups, as you say this Popen will fork a new subprocess, but I guess this is new to the eyes of the scheduler (and thus will take one thread out). I'll find out how cgroups work here.

satra:
@oesteban - it may still be listed under the parent process, but the question is how much of the resources does that thread take up. possibly not much, but just worth checking.

satra:
@oesteban - you can use pstree to check.

setup.py
oesteban: I think we can even remove this executable.

oesteban:
This is nearing a review-able status. Main changes:

- Switched monitoring to a thread, which is started here https://github.com/oesteban/nipype/blob/3f34711e33129415a108b0d3e93c8e9b2e8ac66e/nipype/interfaces/base.py#L1090-L1091 and stopped here: https://github.com/oesteban/nipype/blob/3f34711e33129415a108b0d3e93c8e9b2e8ac66e/nipype/interfaces/base.py#L1132. The Monitor is implemented here: https://github.com/oesteban/nipype/blob/3f34711e33129415a108b0d3e93c8e9b2e8ac66e/nipype/utils/profiler.py#L30-L54.

- Revised documentation and options in depth, trying to make everything more consistent.

- Now MultiProc does not block if one asks for excessive resources. A warning is issued, unless a new option `plugin_args['raise_insufficient']` is `True` (in that case an error is raised). This has been documented.

- TODO:
  - Remove former monitor completely.
  - Test this individually

I would really appreciate comments from @satra about the new approach using `Event` and a general look with fresh eyes (@effigies?) will be very useful as well.

nipype/interfaces/base.py
```diff
@@ -1513,13 +1359,11 @@ def _process(drain=0):
...
                 for stream in res[0]:
                     stream.read(drain)
+
         while proc.returncode is None:
-            if runtime_profile:
-                mem_mb, num_threads = \
-                    get_max_resources_used(proc.pid, mem_mb, num_threads)
             proc.poll()
             _process()
-            time.sleep(interval)
```
effigies: This loop seems kind of busy. Maybe add `time.sleep(0)` to at least yield control? Or is there something in `_process()` that will yield?

nipype/interfaces/spm/base.py
```diff
@@ -151,18 +151,18 @@ def version(matlab_cmd=None, paths=None, use_mcr=None):
...
-                matlab_cmd = 'matlab -nodesktop -nosplash'
+
+        # Test if matlab is installed, exit quickly if not.
+        clout = CommandLine('which matlab', ignore_exception=True,
+                            terminal_output='allatonce').run()
+        if clout.runtime.returncode is not 0:
+            return None
+
+        use_mcr = use_mcr or 'FORCE_SPMMCR' in os.environ
+        matlab_cmd = (os.getenv('SPMMCRCMD') or
```
effigies: To be equivalent to what was replaced:

```Python
matlab_cmd = ((use_mcr and os.getenv('SPMMCRCMD')) or
```

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -697,9 +696,9 @@ def func1(in1):
     assert n1.num_subnodes() == len(n1.inputs.in1)
 
     # test running the workflow on default conditions
-    w1.run(plugin='MultiProc')
+    # w1.run(plugin='MultiProc')
```
effigies: Is this meant to still be commented out?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,53 +111,37 @@ def __init__(self, plugin_args=None):
...
-            if self._config['execution']['poll_sleep_duration']:
-                self._timeout = float(self._config['execution']['poll_sleep_duration'])
-            sig_received=self._event.wait(self._timeout)
-            if not sig_received:
-                logger.debug('MultiProcPlugin timeout before signal received. Deadlock averted??')
-            self._event.clear()
+        logger.debug('MultiProcPlugin starting in "%sdaemon" mode (n_procs=%d, mem_gb=%0.2f)',
+                     'non' if non_daemon else '', self.processors, self.memory_gb)
+        self.pool = (NonDaemonPool(processes=self.processors)
+                     if non_daemon else Pool(processes=self.processors))
```
effigies: Possibly slightly cleaner, if it can fit on one line:

```Python
self.pool = (NonDaemonPool if non_daemon else Pool)(processes=self.processors)
```

I don't know. Might not be an improvement.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -217,32 +178,38 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
         executing_now = []
 
         # Check to see if a job is available
-        currently_running_jobids = np.flatnonzero((self.proc_pending == True) & \
-                                (self.depidx.sum(axis=0) == 0).__array__())
+        currently_running_jobids = np.flatnonzero(
+            self.proc_pending & (self.depidx.sum(axis=0) == 0).__array__())
```
effigies: What work is `__array__` doing here?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -217,32 +178,38 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
...
+
+            busy_memory_gb += min(est_mem_gb, self.memory_gb)
+            busy_processors += min(est_num_th, self.processors)
 
         free_memory_gb = self.memory_gb - busy_memory_gb
         free_processors = self.processors - busy_processors
 
         # Check all jobs without dependency not run
-        jobids = np.flatnonzero((self.proc_done == False) & \
+        jobids = np.flatnonzero((self.proc_done == False) &
```
effigies: `~self.proc_done`?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -300,9 +268,8 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
                         hash_exists, _, _, _ = self.procs[
                             jobid].hash_exists()
                         logger.debug('Hash exists %s' % str(hash_exists))
-                        if (hash_exists and (self.procs[jobid].overwrite == False or
-                                             (self.procs[jobid].overwrite == None and
-                                              not self.procs[jobid]._interface.always_run))):
+                        if hash_exists and not self.procs[jobid].overwrite and \
+                           not self.procs[jobid]._interface.always_run:
```
effigies: This isn't equivalent to what it replaced. Is that intentional?

nipype/utils/draw_gantt_chart.py
```diff
@@ -102,61 +102,14 @@ def log_to_dict(logfile):
...
-                node = json.loads(l)
-            except ValueError:
-                pass
-
-            if not node:
-                continue
-
-            #if it is a start node, add to unifinished nodes
-            if 'start' in node:
-                node['start'] = parser.parse(node['start'])
```
effigies: You can remove the `parser` import from this file.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -251,14 +218,15 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
                         key=lambda item: (self.procs[item]._interface.estimated_memory_gb,
                                           self.procs[item]._interface.num_threads))
 
-        if str2bool(config.get('execution', 'profile_runtime')):
+        resource_monitor = str2bool(config.get('execution', 'resource_monitor', 'false'))
```
effigies: May want to check for `profile_runtime`, use value and raise deprecation warning.

nipype/utils/profiler.py
```diff
@@ -0,0 +1,309 @@
...
+except ImportError as exc:
+    psutil = None
+
+from .. import config, logging
+from .misc import str2bool
+from builtins import open
+
+proflogger = logging.getLogger('utils')
+
+resource_monitor = str2bool(config.get('execution', 'resource_monitor'))
```
effigies: Or raise deprecation warning here...

nipype/pipeline/plugins/multiproc.py
oesteban: I didn't want to mess much around these checks since they work. But yeah, I don't get it either. I'll clean up.

nipype/pipeline/plugins/multiproc.py
oesteban: `self.proc_done` is not a numpy array

nipype/pipeline/plugins/multiproc.py
oesteban: Oh yeah

nipype/interfaces/base.py
```diff
@@ -44,37 +44,33 @@
...
-        import psutil
-    except ImportError as exc:
-        iflogger.info('Unable to import packages needed for runtime profiling. '\
-                    'Turning off runtime profiler. Reason: %s' % exc)
-        runtime_profile = False
-
 __docformat__ = 'restructuredtext'
 
+if sys.version_info < (3, 3):
+    setattr(sp, 'DEVNULL', os.devnull)
```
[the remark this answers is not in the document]
oesteban: Leftover from the previous implementation with a process.

nipype/interfaces/base.py
oesteban: I wanted to bring it back to the original https://github.com/nipy/nipype/blame/c0d148aec7505e9c0df439c5b53d452ae478f352/nipype/interfaces/base.py#L1259 but probably @satra can enlighten us a bit here. I'd say the `time.sleep(0)` seems like a good idea.

nipype/pipeline/plugins/multiproc.py
oesteban: Added a deprecation on the config itself

nipype/utils/draw_gantt_chart.py
oesteban: done

nipype/utils/profiler.py
oesteban: get now raises warning (it is not `DeprecationWarning` bc they are filtered by default)

nipype/utils/logger.py
```diff
@@ -69,8 +69,8 @@ def update_logging(self, config):
         self.disable_file_logging()
         self._logger.setLevel(logging.getLevelName(config.get('logging',
                                                               'workflow_level')))
-        self._fmlogger.setLevel(logging.getLevelName(config.get('logging',
-                                                                'filemanip_level')))
+        self._utlogger.setLevel(logging.getLevelName(config.get('logging',
+                                                                'utils_level')))
```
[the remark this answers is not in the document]
oesteban: Will work on this

nipype/interfaces/base.py
```diff
@@ -903,7 +896,7 @@ def _outputs_help(cls):
         """
         helpstr = ['Outputs::', '']
         if cls.output_spec:
-            outputs = cls.output_spec()  #pylint: disable=E1102
+            outputs = cls.output_spec()  # pylint: disable=E1102
```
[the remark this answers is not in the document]
oesteban: Seems not https://stackoverflow.com/a/20789397/6820620

nipype/pipeline/plugins/multiproc.py
oesteban: I was wrong. Changing to `~self.proc_done`

PR #2197, opened 2017-09-24
CI: Add pip install --pre entry to Travis

effigies:
This gives our dependencies a chance to break a Travis build before breaking everything downstream. I would treat failures of `--pre` builds as non-blocking for most PRs.


satra:
we should figure out a way to do this such that it's not a failure when something breaks here. also we should also do this on circle.

effigies:
Is the concern PRs showing failure when there is none? Or master showing failure?

satra:
@effigies - both.

satra:
@effigies - alright let's go with this for now and revisit when things break.

satra:
just a quick note that the flag only works when there is a pre-release on pypi.

PR #2198, opened 2017-09-25
RF: Separate mcparam -> affine translation from norm calculation in ArtifactDetect

effigies:
`algorithms.rapidart._calc_norm` uses an affine-based algorithm to calculate norms and displacements, translating from a list of motion parameters at each time point. This PR separates the two processes, creating a `_calc_norm_affine` function that performs the core functionality, which `_calc_norm` passes a list of affine matrices translated from the motion parameters.

- New function: `_calc_norm_affine`
- Update: `_calc_norm` translates motion parameters, calls `_calc_norm_affine`
- Removed `_nanmean`, since the minimum numpy (1.9) includes `np.nanmean`

### DIFF
--- a/nipype/algorithms/rapidart.py
+++ b/nipype/algorithms/rapidart.py
@@ -99,6 +99,29 @@ def _calc_norm(mc, use_differences, source, brain_pts=None):
 
     """
 
+    affines = [_get_affine_matrix(mc[i, :], source)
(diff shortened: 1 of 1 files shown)


satra:
LGTM. is numpy 1.9 now our minimum?

effigies:
Since #2182.

PR #2200, opened 2017-09-27
[ENH] New ResourceMonitor (replaces resource profiler)

oesteban:
Supersedes #2193 (I removed the specs to make code review easier)

This PR revises the resource profiler (with a new name: **resource monitor**):

  - [x] Created new `nipype.utils.profiler` where all the code related to resource monitoring is gathered together. Tests are also brought into the `nipype.utils` module.
  - [x] Removed monitoring responsibility from the runner into the interfaces. Consequences:
    1. Resources can be monitored when running bare interfaces
    2. All kinds of interfaces can be monitored (it was reduced to only `CommandLine`-based and `Function` interfaces before).
    3. There was a bit of a mixture of resource-management and resource-monitoring in the node execution code. That has been cleaned up.
    4. Interfaces save the traces directly to a file, so the log callback is not needed anymore (it has been left here just in case users want to generate the old gantt chart).
    5. Interfaces place peak usage in their runtime object.
  - [x] Previously, even though resource consumption with a certain frequency, only peak usage was finally reported in the Gantt chart. Now the traces are kept in the `proc_dict` member of the `runtime` object.
  - [x] When the workflow is finished, the execution graph is post-processed and a new `runtime_monitor.json` file is generated, with all the traces and necessary information to reconstruct the performance profile with the resolution of the new `runtime_monitor_frequency` option (default: every 1s).
  - [x] The resource manager of MultiProc blocked if a node was defined with `n_procs` > MultiProc's setting `n_procs` (similarly with the estimated memory). Now, either a warning or an error are raised (depending on the new plugin arg `raise_insufficient`).

**Important**: this PRs removes the old `filemanip` logger and replaces it by a more generic `utils`. Documentation has been updated accordingly and Deprecation warnings are generated consequently.


doc/users/config_file.rst
```diff
@@ -14,93 +14,97 @@ Logging
...
+    How detailed the logs regarding nipype utils, like file operations
+    (for example overwriting warning) or the resource profiler, should be
+    (possible values: ``INFO`` and ``DEBUG``; default value:
+    ``INFO``)
 *interface_level*
-	How detailed the logs regarding interface execution should be (possible
-	values: ``INFO`` and ``DEBUG``; default value: ``INFO``)
+    How detailed the logs regarding interface execution should be (possible
+    values: ``INFO`` and ``DEBUG``; default value: ``INFO``)
+*filemanip_level* (deprecated as of 0.13.2)
```
satra: since we are calling the next release 1.0, should this be deprecated as of `1.0` or `1.1`

doc/users/config_file.rst
```diff
@@ -146,18 +150,26 @@ Execution
     crashfiles allow portability across machines and shorter load time.
     (possible values: ``pklz`` and ``txt``; default value: ``pklz``)
 
+*resource_monitor*
+    Enables monitoring the resources occupation (possible values: ``true`` and
+    ``false``; default value: ``false``)
+
+*resource_monitor_frequency*
+    Sampling period (in seconds) between measurements of resources (memory, cpus)
+    being used by an interface. Requires ``resource_monitor`` to be ``true``.
```
satra: is there a default when `resource_monitor=='true'`?

docker/files/run_examples.sh
```diff
@@ -12,10 +12,16 @@ mkdir -p ${HOME}/.nipype ${WORKDIR}/logs/example_${example_id} ${WORKDIR}/tests
 echo "[logging]" > ${HOME}/.nipype/nipype.cfg
 echo "workflow_level = DEBUG" >> ${HOME}/.nipype/nipype.cfg
 echo "interface_level = DEBUG" >> ${HOME}/.nipype/nipype.cfg
-echo "filemanip_level = DEBUG" >> ${HOME}/.nipype/nipype.cfg
+echo "utils_level = DEBUG" >> ${HOME}/.nipype/nipype.cfg
 echo "log_to_file = true" >> ${HOME}/.nipype/nipype.cfg
 echo "log_directory = ${WORKDIR}/logs/example_${example_id}" >> ${HOME}/.nipype/nipype.cfg
 
+if [[ "${NIPYPE_RESOURCE_MONITOR:-0}" == "1" ]]; then
+    echo '[execution]' >> ${HOME}/.nipype/nipype.cfg
```
satra: just a note that `.nipype` is no longer created by default. so nipype doesn't touch anything when run.

oesteban:
@satra, shouldn't this 
https://github.com/oesteban/nipype/blob/0b00a2064e93c0ea8e2f57cb70df9209504140b7/nipype/interfaces/base.py#L1371

sleep for `poll_sleep_duration` seconds?

nipype/interfaces/base.py
```diff
@@ -738,9 +729,12 @@ def _get_filecopy_info(self):
...
 
 class BaseInterfaceInputSpec(TraitedSpec):
-    ignore_exception = traits.Bool(False, desc="Print an error message instead \
-of throwing an exception in case the interface fails to run", usedefault=True,
-                                   nohash=True)
+    ignore_exception = traits.Bool(False, usedefault=True, nohash=True,
+                                   desc='Print an error message instead of throwing an exception '
+                                        'in case the interface fails to run')
+    resource_monitor = traits.Bool(True, usedefault=True, nohash=True,
+                                   desc='Disable the resource monitor for this interface '
```
satra: is this really saying disable when set to `True`?

also should `resource_monitor` be an input or a class attribute like `_redirect_x`?

nipype/interfaces/base.py
```diff
@@ -1063,13 +1055,19 @@ def run(self, **inputs):
         results :  an InterfaceResult object containing a copy of the instance
         that was executed, provenance information and, if successful, results
         """
+        from ..utils.profiler import resource_monitor, ResourceMonitor
+
+        enable_rm = resource_monitor and getattr(self.inputs, 'resource_monitor', True)
```
satra: should this be `getattr(self, 'resource_monitor', True)`?

assuming it should be a class attribute rather than an input.

nipype/interfaces/base.py
```diff
@@ -1080,66 +1078,74 @@ def run(self, **inputs):
...
+
+            # Add provenance (if required)
+            if store_provenance:
+                # Provenance will only throw a warning if something went wrong
+                results.provenance = write_provenance(results)
+
+            # Make sure runtime profiler is shut down
+            if enable_rm:
+                import numpy as np
+                mon_sp.stop()
```
satra: assuming that this is a blocking call for now.

nipype/interfaces/base.py
```diff
@@ -1170,7 +1176,7 @@ def aggregate_outputs(self, runtime=None, needed_outputs=None):
                                                       self.__class__.__name__))
                 try:
                     setattr(outputs, key, val)
-                    _ = getattr(outputs, key)
+                    getattr(outputs, key)
```
satra: this was a weird traits error with dynamic traits, where this wouldn't work without the `_` . if things pass perhaps this is no longer necessary. but let's hold this thought in our heads in case people start reporting weird errors.

nipype/interfaces/base.py
```diff
@@ -1513,13 +1364,12 @@ def _process(drain=0):
...
                     stream.read(drain)
+
         while proc.returncode is None:
-            if runtime_profile:
-                mem_mb, num_threads = \
-                    get_max_resources_used(proc.pid, mem_mb, num_threads)
             proc.poll()
             _process()
-            time.sleep(interval)
+            time.sleep(0)
```
satra: is `0` what we want here?

nipype/interfaces/afni/base.py
```diff
@@ -45,6 +45,7 @@ def version():
         """
         try:
             clout = CommandLine(command='afni --version',
+                                resource_monitor=False,
```
satra: would something bad happen if `resource_monitor` is left out from here? i.e., does nested resource monitoring work? if not, perhaps it would be good to add this info somewhere.

nipype/pipeline/engine/utils.py
```diff
@@ -1294,6 +1296,51 @@ def write_workflow_prov(graph, filename=None, format='all'):
     return ps.g
 
 
+def write_workflow_resources(graph, filename=None):
+    import simplejson as json
+    if not filename:
+        filename = os.path.join(os.getcwd(), 'resource_monitor.json')
+
+    big_dict = {
+        'time': [],
```
satra: is `time` different from `duration`?

nipype/utils/config.py
```diff
@@ -64,7 +67,6 @@
 parameterize_dirs = true
 poll_sleep_duration = 2
 xvfb_max_wait = 10
-profile_runtime = false
```
satra: shouldn't we add the resource_monitor options here?

nipype/interfaces/base.py
satra: @oesteban - i made this comment here - that's how long it has been since i wrote that code. now i think it may be there to create a no-op. 

but it shouldn't be `poll_sleep_duration` since this is trying to get stdout/error as quickly as possible and report on the screen.

nipype/pipeline/engine/utils.py
oesteban: yes, time is the result of `time.time()`. Duration is `time.time()` at start - `time.time()` at end.

nipype/interfaces/base.py
effigies: 0 was my suggestion to just yield control rather than busy wait, but 0.1 is probably fast enough.

docker/files/run_examples.sh
oesteban: mmm what you mean? this is currently working out...

satra: if it's working great - just that in the past nipype would create this directory by default, now it doesn't.

nipype/interfaces/base.py
oesteban: Yes, that `desc` is wrong.

Your question is pretty good. Why `ignore_exception` is an input? shouldn't it be a class attribute? But sure, we can make it a class attribute.

nipype/interfaces/base.py
oesteban: what do you mean?

nipype/interfaces/base.py
oesteban: It wasn't there before the ResourceMultiProc was merged, so I wonder if it is really necessary.

nipype/interfaces/base.py
satra: i agree - several things have bothered me about being inputs/class attributes. 

`terminal_output`, `ignore_exception` are particular offenders. but for those we would need to go through a deprecation cycle. for this, i think we can start correctly :)

nipype/interfaces/base.py
satra: as in it does not return till the process is stopped and all information gathered. i.e. it's not an asynchronous process. i left this comment before i got to resource_profiler. it is indeed a blocking call.

nipype/interfaces/afni/base.py
oesteban: Yes, that's why I had to add all of those `resource_monitor`. Basically, both inner and outer interfaces send traces to the same file (with all the derived effects you might expect from races)

nipype/utils/config.py
oesteban: I have added a `default` option to `nipype.config.get` *ala* `dict.get`

nipype/interfaces/base.py
satra: perhaps we can try without it and see what happens.

doc/users/config_file.rst
oesteban: added default value to the documentation

docker/files/run_examples.sh
oesteban: if it still reads it, then it works :)

nipype/interfaces/base.py
oesteban: yep, it waits for the monitor to shut down.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,207 +114,190 @@ def __init__(self, plugin_args=None):
...
+                     len(self.pending_tasks), len(jobids),
+                     free_memory_gb, self.memory_gb, free_processors, self.processors)
 
-        free_memory_gb = self.memory_gb - busy_memory_gb
-        free_processors = self.processors - busy_processors
 
-        # Check all jobs without dependency not run
-        jobids = np.flatnonzero((self.proc_done == False) & \
-                                (self.depidx.sum(axis=0) == 0).__array__())
+        if (len(jobids) + len(self.pending_tasks)) == 0:
```
effigies: Drop parentheses.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,207 +114,190 @@ def __init__(self, plugin_args=None):
...
-        jobids = sorted(jobids,
-                        key=lambda item: (self.procs[item]._interface.estimated_memory_gb,
-                                          self.procs[item]._interface.num_threads))
-
-        if str2bool(config.get('execution', 'profile_runtime')):
-            logger.debug('Free memory (GB): %d, Free processors: %d',
-                         free_memory_gb, free_processors)
+        # jobids = sorted(jobids,
+        #                 key=lambda item: (self.procs[item]._get_mem_gb(),
+        #                                   self.procs[item]._get_n_procs()))
```
effigies: Should this be uncommented or removed?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,207 +114,190 @@ def __init__(self, plugin_args=None):
...
+                continue
+
+            free_memory_gb -= next_job_gb
+            free_processors -= next_job_th
+            logger.info('Allocating %s ID=%d (%0.2fGB, %d threads). Free: %0.2fGB, %d threads.',
+                        self.procs[jobid]._id, jobid, next_job_gb, next_job_th,
+                        free_memory_gb, free_processors)
+
+            # change job status in appropriate queues
+            self.proc_done[jobid] = True
```
effigies: Done? I see that's how this was before; just want to make sure this makes sense.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,207 +114,190 @@ def __init__(self, plugin_args=None):
...
+            self.proc_pending[jobid] = True
+
+            if str2bool(self.procs[jobid].config['execution']['local_hash_check']):
+                logger.debug('checking hash locally')
+                try:
+                    hash_exists, _, _, _ = self.procs[jobid].hash_exists()
+                    overwrite = self.procs[jobid].overwrite
+                    always_run = self.procs[jobid]._interface.always_run
+                    if (hash_exists and (overwrite is False or
+                                         (overwrite is None and not always_run))):
```
effigies: Can drop two sets of parentheses.

```Python
if hash_exists and (overwrite is False or
                    overwrite is None and not always_run):
```

nipype/pipeline/plugins/oar.py
```diff
@@ -10,9 +10,10 @@
 import subprocess
 import simplejson as json
 
-from .base import (SGELikeBatchManagerBase, logger, iflogger, logging)
+from .base import (SGELikeBatchManagerBase, logger, logging)
+from ... import logging
```
effigies: Multi-importing `logging`.

nipype/pipeline/plugins/pbs.py
```diff
@@ -8,8 +8,10 @@
 from time import sleep
 
 from ...interfaces.base import CommandLine
-from .base import (SGELikeBatchManagerBase, logger, iflogger, logging)
+from ... import logging
+from .base import (SGELikeBatchManagerBase, logger, logging)
```
effigies: Multi-importing `logging`.

nipype/pipeline/plugins/sge.py
```diff
@@ -15,9 +15,10 @@
 
 import random
 
+from ... import logging
 from ...interfaces.base import CommandLine
-from .base import (SGELikeBatchManagerBase, logger, iflogger, logging)
-
+from .base import (SGELikeBatchManagerBase, logger, logging)
```
effigies: Multi-importing `logging`.

nipype/pipeline/plugins/slurm.py
```diff
@@ -12,8 +12,11 @@
 import re
 from time import sleep
 
+from ... import logging
 from ...interfaces.base import CommandLine
-from .base import (SGELikeBatchManagerBase, logger, iflogger, logging)
+from .base import (SGELikeBatchManagerBase, logger, logging)
```
effigies: Multi-importing `logging`.

nipype/pipeline/plugins/slurm.py
```diff
@@ -12,8 +12,11 @@
...
+
+iflogger = logging.getLogger('interface')
```
effigies: Too many blank lines.

nipype/utils/config.py
```diff
@@ -11,21 +11,24 @@
...
+from simplejson import load, dump
+from ..external import portalocker
 from future import standard_library
 standard_library.install_aliases()
 
-import configparser
-from ..external import portalocker
 
+CONFIG_DEPRECATIONS = {
+    'profile_runtime': ('resource_monitor', '1.13.2'),
```
effigies:
```Diff
- 'profile_runtime': ('resource_monitor', '1.13.2'),
+ 'profile_runtime': ('resource_monitor', '1.0'),
+ 'filemanip_level': ('utils_level', '1.0'),
```

nipype/utils/config.py
```diff
@@ -114,13 +116,29 @@ def set_log_dir(self, log_dir):
...
 
-    def get(self, section, option):
-        return self._config.get(section, option)
+    def get(self, section, option, default=None):
+        if option in CONFIG_DEPRECATIONS:
+            msg = ('Config option "%s" has been deprecated as of nipype %s. Please use '
+                   '"%s" instead.') % (option, CONFIG_DEPRECATIONS[option][1],
+                                       CONFIG_DEPRECATIONS[option][0])
+            warn(msg)
+            option = CONFIG_DEPRECATIONS[option][0]
```
effigies: This is a message directed at users, but it's the nipype dev who's going to request the old versions of these. I think we need a warning for users that their config file contains a deprecated option.

Perhaps in `__init__`, you can do at the end:

```Python
for option in CONFIG_DEPRECATIONS:
    if self._config.has_option(option):
        # Warn
        new_option = CONFIG_DEPRECATIONS[option][0]
        if not self._config.has_option(new_option):
            self._config.set(new_option, self._config.get(option))
```

Or similar.

nipype/utils/logger.py
```diff
@@ -53,30 +55,35 @@ def enable_file_logging(self):
...
+                                                                'utils_level')))
         self._iflogger.setLevel(logging.getLevelName(config.get('logging',
                                                                 'interface_level')))
         if str2bool(config.get('logging', 'log_to_file')):
             self.enable_file_logging()
 
     def getLogger(self, name):
+        if name == 'filemanip':
+            warn('The "filemanip" logger has been deprecated and replaced by '
+                 'the "utils" logger as of nipype 1.13.2')
```
effigies: 1.0?

nipype/utils/spm_docs.py
```diff
@@ -27,7 +27,7 @@ def grab_doc(task_name):
 
     """
 
-    cmd = matlab.MatlabCommandLine()
+    cmd = matlab.MatlabCommandLine(resource_monitor=False)
```
effigies: `MatlabCommandLine` doesn't exist. Perhaps `MatlabCommand`?

nipype/utils/tests/test_profiler.py
```diff
@@ -1,33 +1,20 @@
...
 from __future__ import print_function, division, unicode_literals, absolute_import
 from builtins import open, str
 
 # Import packages
-from nipype.interfaces.base import (traits, CommandLine, CommandLineInputSpec,
-                                    runtime_profile)
 import pytest
-import sys
+from nipype.utils.profiler import resource_monitor as run_profile
+from nipype.interfaces.base import (traits, CommandLine, CommandLineInputSpec)
```
effigies: No need for parentheses.

oesteban:
Thanks a lot for all the comments! I'll start fixing ASAP

nipype/interfaces/afni/base.py
oesteban: Do you think this will bite someone? This problem only happens if there are nested interfaces. I think it is not too much to ask somebody with the skills to do that knowingly that they add the `resource_monitor=True`. I think it is and edge case, isn't it?

oesteban: We could check if the profiler file exists not to run an extra ResourceMonitor...

nipype/interfaces/base.py
```diff
@@ -766,23 +757,24 @@ class BaseInterface(Interface):
     _additional_metadata = []
     _redirect_x = False
     references_ = []
+    resource_monitor = True
```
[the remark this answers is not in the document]
oesteban: Good idea!

nipype/pipeline/engine/nodes.py
```diff
@@ -79,7 +78,7 @@ class Node(EngineBase):
 
     def __init__(self, interface, name, iterables=None, itersource=None,
                  synchronize=False, overwrite=None, needed_outputs=None,
-                 run_without_submitting=False, n_procs=1, mem_gb=None,
+                 run_without_submitting=False, n_procs=1, mem_gb=0.25,
```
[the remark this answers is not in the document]
oesteban: I discussed this with @ccraddock. At that point it seemed reasonable.

On my experience, multiproc processes are taking around 170MB when running utility.Merge nodes and stuff like that. We can go to 200MB but I think we are here in a reasonable range already.

nipype/pipeline/plugins/base.py
```diff
@@ -9,189 +9,46 @@
...
+    Execution plugin API
+    ====================
 
+    Current status::
 
-class PluginBase(object):
-    """Base class for plugins"""
+        class plugin_runner(PluginBase):
+
+            def run(graph, config, updatehash)
```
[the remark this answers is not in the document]
oesteban: @satra what was `updatehash` meant for?

nipype/interfaces/afni/base.py
effigies: The problem is less going forward (though assuming that everybody's going to be up on the latest details of the optional parameters to `BaseInterface` seems dubious), but the already existing interfaces. Obviously we'll know to fix them in niworkflows/fmriprep/mriqc, but how many others exist?

If we can do such a check to not double-run, that'd be good, IMO.

effigies: Also, this `Info.version` is a great example of a use that's not obviously nested. It's not nested when I program it, and if I'm writing an interface, it might not occur to me to check whether `Info().version` will itself be running an interface.

nipype/pipeline/plugins/multiproc.py
oesteban: Thanks for pointing out. This was actually something I want to discuss with you, @satra and it probably may attract @chrisfilo's interest.

Do we really want to sort nodes by size? I'd say we don't (and I'd remove this commented code)

nipype/pipeline/plugins/multiproc.py
oesteban: This also struck me at first, but this is how it works :D. The name of this array is certainly confusing. It should be something like `proc_submitted` but that would change all plugins internally.

nipype/pipeline/plugins/multiproc.py
effigies: I think the idea is to fill as many slots as possible, and let the small nodes fill in the holes. Which makes sense; if I have something that's going to use half the cores and take a while, and a bunch of small jobs that can run more-or-less serially in the remaining cores, we'll use the processor pretty efficiently. (And actually would mean we should be sorting in the opposite order; threads then memory.)

One metric I've thought about is the number of nodes that will be able to be queued when a given node is completed. The priority then would be to remove bottlenecks first, and fill in around them.

Scheduling priority metric might be a sensible plugin argument. And the default could be `tsort` (graph as it comes in) or `mem_thread` (current strategy), and we could easily build alternatives like un-bottleneck...

ccraddock:
Agree with Markiewicz on this and is why we implemented it this way. There are many different scheduling strays out there - agree we should support a few.

nipype/pipeline/plugins/multiproc.py
satra: @oesteban, @effigies, and @ccraddock - my first thought is that we shouldn't get into the scheduling business - as that is a hairy one, even on a multi-core machine. that being said i'm ok with some basic scheduling strategies. since we don't know how long something is going to run (at least yet), the best we can do is prioritize cores, so favor small jobs over long jobs. however, since we have a graph, the other piece would be how many downstream jobs are dependent on the current node. as i write this, i realize why my first thought was first.

effigies: I mean, we're already in the scheduling business, at least insofar as there's topological sorting for all other plugins, and memory-then-threads for `MultiProc`.

The interface could be dead simple:

```Python
def mem_thread_scheduler(jobids, plugin):
    return sorted(jobids, key=lambda item: (plugin.procs[item]._get_mem_gb(),
                                            plugin.procs[item]._get_n_procs()))

def tsort(jobids, plugin):
    return jobids
```

nipype/utils/config.py
oesteban: Done

nipype/pipeline/plugins/multiproc.py
satra: @effigies - we are only partially in the scheduling business :) we hand of things from the graph. while the simple interface would work, it's a starting point, as this doesn't take the weight of dependencies or their run profiles into account.

this is almost like playing chess :) peeking a few more levels may give us an A* algorithm.

but i would be ok with this as a start. however, more generally this could be applied to any of the scheduler plugins (except the graph ones).

oesteban: Since it was already in the codebase, I thought that we could allow users to try this "mem_thread" scheduler. To do so, `MultiProc` allows now a `plugin_args['scheduler'] = 'mem_threads'` argument. However, I do agree with @satra that scheduling is a pretty hairy problem (and worse when you don't have a wallclock time).

oesteban:
Ok, I finally ended up with a pretty massive PR. Sorry about that. The reason for it is that the two concepts of resource management and resource monitoring are confused at many points of the current implementation. In this PR I try to disentangle both.

An example of the confusion is here:
https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/multiproc.py#L254-L256
Here we check if the profiler (monitoring) is ON to write a debug trace... on the two important variables that will decide if a job is submitted or not a bit further. That debug trace belongs to resource management but is protected with an if checking on the monitoring.

# Changes affecting the resource manager of MultiProc
MultiProc, following the former implementation is still doing some resource management based on the estimations that are passed along with the two `Node` arguments `mem_gb` and `n_procs`.

The old `Interface.estimated_memory_gb` and `Interface.num_threads` have been **deprecated**. Interfaces should not be aware of resource management, so I moved that responsibility completely to the plugin. I've tried to spit enough warnings when trying to use the old API.

MultiProc has two new `plugin_args`. First, we have `plugin_args['scheduler'] = ('tsort'|'mem_threads')` and second is `plugin_args['raise_insufficient'] = (True|False)` which defaults to `True`. When `raise_insufficient` is `True`, then MultiProc will not allow jobs that exceed the nominal maximums of RAM and CPUs, raising an exception. If `False`, then the job will execute when no other job is running (occupying all available resources). The `scheduler` arg is discussed above and left there to avoid removing a pre-existing feature.

Also, with the great comments from @effigies, I think this PR cleans up a lot MultiProc's code.

# Changes affecting the resource monitoring
The resource monitoring had its tests disabled, and there are some issues reported about it. My main concern about it is that the implementation was intertwined with the MultiProc plugin, and would only work for `Function` and `CommandLine` interfaces wrapped in a `Node`.

In this PR, the responsibility of monitoring is moved to the Interfaces. Interfaces are the object that nipype ultimately runs in the target infrastructure (any plugin). So moving this to the interfaces will enable monitoring the resources consumption of ANY plugin. Even if you want to run an interface in an ipython terminal.

It is done with the new `ResourceMonitor` class of `nipype.utils.profiler`. This new object uses `Event`s to sample the resources with a given frequency. These traces are written to a file in the base directory of the interface.

When a workflow is run with the resource_monitor enabled, the info of this file is collected and stored in the `runtime` object. Finally, the same way we collect provenance, the resource profile is collected together in a json file that can be used to plot the gantt chart.

I haven't removed the callback infrastructure, since someone could be interested in keeping it. But, if find the time to generate the equivalent to the gantt chart based on the new traces, I think we will be able to remove it at that point.

# Known issues:
- The estimation of threads we are getting is incorrect. I'll need to revise this and enable the parameterized tests of monitoring (https://github.com/oesteban/nipype/blob/8a5e7a3c3d51f5e46d03f665cfaa3cc9b02d9065/nipype/interfaces/tests/test_resource_monitor.py#L45-L83)

nipype/utils/config.py
```diff
@@ -94,6 +97,15 @@ def __init__(self, *args, **kwargs):
...
             self._config.read([config_file, 'nipype.cfg'])
 
+        for option in CONFIG_DEPRECATIONS:
+            for section in ['execution', 'logging']:
+                if self._config.has_option(section, option):
+                    new_option = CONFIG_DEPRECATIONS[option][0]
+                    if not self._config.has_option(section, new_option):
+                        # Warn implicit in get
+                        self._config.set(section, new_option,
+                                         self._config.get(option))
```
effigies: If you want this implicit in `get`, what you probably want the local `get`:

```Python
self.set(section, new_option, self.get(section, option))
```

Also, you're missing the section in this `get`, in either case.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,207 +118,174 @@ def __init__(self, plugin_args=None):
...
+        """
 
-        # Check to see if a job is available
-        currently_running_jobids = np.flatnonzero((self.proc_pending == True) & \
-                                (self.depidx.sum(axis=0) == 0).__array__())
+        # Check to see if a job is available (jobs without dependencies not run)
+        # See https://github.com/nipy/nipype/pull/2200#discussion_r141605722
+        jobids = np.nonzero(~self.proc_done & (self.depidx.sum(0) == 0))[1]
+        jobids = np.flatnonzero(
+            ~self.proc_done & (self.depidx.sum(axis=0) == 0).__array__())
```
satra: is this intended to be chosen from? (like a or b)?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -134,207 +118,174 @@ def __init__(self, plugin_args=None):
...
-                                          self.procs[item]._interface.num_threads))
-
-        if str2bool(config.get('execution', 'profile_runtime')):
-            logger.debug('Free memory (GB): %d, Free processors: %d',
-                         free_memory_gb, free_processors)
-
-        # While have enough memory and processors for first job
-        # Submit first job on the list
+        if len(jobids) + len(self.pending_tasks) == 0:
+            logger.debug('No tasks are being run, and no jobs can '
```
satra: shouldn't this raise an Exception?

nipype/pipeline/plugins/multiproc.py
satra: the reason for this is the linear plugin, where it truly meant `proc_done`, but it was indeed short for proc_submission_done.

nipype/utils/config.py
satra: but shouldn't it also be here in the `default_cfg`?

nipype/utils/profiler.py
```diff
@@ -0,0 +1,333 @@
...
+            raise RuntimeError('Frequency (%0.2fs) cannot be lower than zero' % freq)
+
+        if fname is None:
+            fname = '.nipype.prof'
+
+        self._pid = pid
+        self._fname = fname
+        self._freq = freq
+
+        self._log = open(self._fname, 'w')
```
satra: what if we put this in memory or added a timestamp to _fname? wouldn't that take care of nested profilers? i would really try and avoid `resource_profile=False` flag if possible.

nipype/pipeline/plugins/multiproc.py
oesteban: Good catch. No, the second line shouldn't be there

nipype/pipeline/plugins/multiproc.py
oesteban: I was still trying to understand why the execution goes through this, but it actually happens (see https://654-17021985-gh.circle-artifacts.com/2/home/ubuntu/work/logs/example_fmri_spm_nested_MultiProc__data_examples__l2pipeline/pypeline.log and https://654-17021985-gh.circle-artifacts.com/2/home/ubuntu/work/logs/example_fmri_spm_nested_MultiProc__data_examples__level1/pypeline.log).

So this happens, and it seems to recover itself without getting trapped. I think we could keep the debug trace (meaning, do not elevate it to warning) and add a counter. If the counter reaches a threshold, then raise.

WDYT?

nipype/utils/profiler.py
oesteban: I see it independently. I think the `resource_profile=False` flag is a necessary tool that ensures you an interface will not trigger the profiler at all, regardless nipype's config.

Then we have the issue of accidentally having two monitors logging to the same file. We have identified the case of nested interfaces because it quickly showed up, but there might be other corner cases in complex workflows that we haven't seen yet. For this reason, I agree we need to work more on this.

@satra, is there a way of writing this to memory, as you say, but making sure that everything that was traced gets finally written to a file no matter what?

oesteban: Also, with writing to memory we will miss a new feature that would allow you keep track of a workflow being run by looking at these profiler files.

nipype/utils/config.py
oesteban: not necessarily

nipype/utils/profiler.py
satra: how about doing: `self._fname = fname + pid + time()`

wouldn't that allow nested profilers?

nipype/pipeline/plugins/multiproc.py
oesteban: seems to me that `self.proc_pending` and `self.pending_task` are not in sync.

satra:
@oesteban - i sent a PR to your branch.

nipype/interfaces/base.py
```diff
@@ -760,29 +750,43 @@ class BaseInterface(Interface):
...
 
-    def __init__(self, from_file=None, **inputs):
+    def __init__(self, from_file=None, resource_monitor=None, **inputs):
         if not self.input_spec:
             raise Exception('No input_spec in class: %s' %
                             self.__class__.__name__)
 
         self.inputs = self.input_spec(**inputs)
-        self.estimated_memory_gb = 0.25
-        self.num_threads = 1
```
effigies: Removing this breaks AFNI interfaces. We'll need to rework these.

https://github.com/nipy/nipype/blob/6a8318f632d3b3243cf7d69d1ae387aaa26afff8/nipype/interfaces/afni/base.py#L187

nipype/utils/profiler.py
```diff
@@ -0,0 +1,367 @@
...
+                mem += self._process.memory_info().rss
+        except psutil.NoSuchProcess:
+            pass
+
+        # parent_mem = mem
+        # Iterate through child processes and get number of their threads
+        for child in self._process.children(recursive=True):
+            try:
+                with child.oneshot():
+                    cpu += child.cpu_percent()
```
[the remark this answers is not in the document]
oesteban: interval is blocking, so I avoided it for every child to make the measurement faster. Consequence is that first sample will underestimate usage.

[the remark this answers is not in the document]
oesteban: interval is blocking, so I avoided it for every child to make the measurement faster. Consequence is that first sample will underestimate usage.

nipype/interfaces/base.py
oesteban: Correct

nipype/interfaces/base.py
```diff
@@ -1080,66 +1089,72 @@ def run(self, **inputs):
...
+
+                runtime.mem_peak_gb = None
+                runtime.cpu_percent = None
+
+                # Read .prof file in and set runtime values
+                vals = np.loadtxt(mon_sp.fname, delimiter=',')
+                if vals.size:
+                    vals = np.atleast_2d(vals)
+                    runtime.mem_peak_gb = float(vals[:, 1].max() / 1024)
+                    runtime.cpu_percent = float(vals[:, 2].max())
```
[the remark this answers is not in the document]
oesteban: Probably not necessary

nipype/interfaces/base.py
oesteban: Ok, I've been looking at this. Regarding `num_threads` we have two sets of interfaces:
  - "ANTs-like" (ANTs, FreeSurfer, FSL, elastix): have an `inputs.num_threads` field.
  - AFNI, where `num_threads` is a class attribute.

We should unify this to only one option. And also make a decision about how information should flow between interfaces and nodes: should nodes set `num_threads` on the interface depending on their `n_procs` attribute. The alternative is that `n_procs` is read from the interface's `inputs` before running the node.

WDYT @satra, @effigies ?

effigies: My thought:

- If I set it in the `Node`, that should override and set it in the input spec.
- If I don't specify it in the `Node`, but it is defined in the `Interface`, then the node should pass that through.

Something like the following (I haven't looked up current naming conventions and API):

```Python
class Node:

    def __init__(self, n_procs=None):
        super().__init__()
        self._n_procs = n_procs
        if hasattr(self.inputs, 'num_threads') and self._n_procs is not None:
                self.inputs.num_threads = self._n_procs

    @property
    def n_procs(self):
        if self._n_procs is not None:
            return self._n_procs
        elif hasattr(self.inputs, 'num_threads') and isdefined(self.inputs.num_threads):
            return self.inputs.num_threads
        else:
            return 1
```

oesteban:
This PR now has two main issues on the table:

  - [This discussion](https://github.com/nipy/nipype/pull/2200#discussion_r142147570) regarding where the information of number of threads should reside (Interface or Node) and how it flows. My opinion is that Nodes keep the information and set it on the interfaces before running.
  - I'm having several issues with the estimation of resources, and the corresponding tests:
    - When testing `CommandLine` interfaces (https://github.com/oesteban/nipype/blob/ef097a6390b72b1550d60145d23b23597a48208c/nipype/interfaces/tests/test_resource_monitor.py#L46-L60), it works for my in my desktop 95% of the times (the 5% correspond to runs where the cpu_usage went a bit over the threshold producing peaks over the prescribed number of cores of the test). It always fails in CircleCI
    - When testing `Function` interfaces (https://github.com/oesteban/nipype/blob/ef097a6390b72b1550d60145d23b23597a48208c/nipype/interfaces/tests/test_resource_monitor.py#L65-L81) in my desktop, only the cpu count will work. Apparently there is some optimization in python that removes the memory allocation from the test, or we are measuring memory wrongly [here](https://github.com/oesteban/nipype/blob/ef097a6390b72b1550d60145d23b23597a48208c/nipype/utils/profiler.py#L76-L98), even though that code works for `CommandLine`s. The number of cpus is estimated correctly (with some glitches as for command lines) in my desktop. Memory and cpu estimations fail both in Circle.

effigies:
WRT `num_threads`, I would agree if the number of threads could always be known at the time of workflow construction, but there are [cases](https://github.com/poldracklab/fmriprep/blob/b16fe1cbc73959e7cf6c43fa2c88da762ebf36a4/fmriprep/workflows/anatomical.py#L260) where you want to set it dynamically based on inputs, for which the `input_spec` is the logical place to specify.

I think it sense to have two modes:

- Set statically in `Node`
- Set dynamically in `Interface().inputs`

Perhaps it makes sense to throw an error if a user tries to invoke both modes. So, if I have a `Node` such that `node.n_procs is not None` then a `RuntimeError` (or `ValueError`) should be thrown if I try `node.inputs.num_threads = <any value except node.n_procs>` (I think we can do this with `on_trait_change`), or to make a connection `wf.connect(in_node, 'in_port', node, 'num_threads')` (even easier).

nipype/interfaces/base.py
satra: for `MapNode` would this be set for every instance of the interface? my thought would be yes, since the number of instances are dynamically created. and the worst case scenario is to run every instance in serial mode, with best case scenario controlled by the total number of procs available.

i like the code, defaults to 1, overrides with `Node`, and sets it for each instance. we need to make sure this carries over to `MapNode` and we have a sync between the interface `num_threads` and `_n_procs`.

re: the case for inputs vs class attributes, this one is fuzzy in my mind. on one hand, bad parallel implementations may provide different output, which means should rerun if number of threads change, but we don't do this right now (`hash_files = False`). i think this should eventually be a class attribute that can be overridden for any instance. or we should be strict and remove `hash_files` and make the code rerun if this number changes.

oesteban: I'll answer here, including [this Chris' comment](https://github.com/nipy/nipype/pull/2200#issuecomment-333618018) that I think it was meant to be here.

I do fundamentally agree with both, and my idea was along the lines @effigies has described (keep the info statically on `Node.n_procs`, but have a dynamic option with `Interface.inputs.num_threads`). Thanks for this much clearer description.

I also agree with Satra on the following. In our experience, `antsRegistration`'s `nohash` (or `hash_files` if talking about the specs) should be `False`. Unless you compiled with very specific and unoptimized flags, you'd get different results if you change `num_threads`. But also if you don't change it and `num_threads` > 1. That would justify an input, in addition to @effigies example.

For nipype 1.0 we can keep the `inputs.num_threads` (meaning, I'll normalize AFNI to behave as the others) and we can reconsider removing the input for nipype 2.0, when we have more experience on how this turned out. Sounds good?

effigies: That sounds fine to me.

satra: sounds good to me as well.

oesteban:
Seems to be working now on fmriprep. Conversely, it was falling in a deadlock for MRIQC because it touches nipype's logging, and `run_node` was trying to write a trace (that wasn't actually sent to the logger ever). So removing that trace fixed the issue.

I'm re-running both. Final work will address:
  - Make sure that `MapNode`s work as discussed above.
  - Make a decision about merging without the profiler tests (as it was before) since it is unclear that these tests will ever work on CircleCI (and also the limitations of the resource monitor that I would address in a different PR).

nipype/pipeline/plugins/multiproc.py
satra: in that case, if no jobs are available in line 197 should we return similar to your no resources are available? after all this function's goal is to send jobs to workers.

oesteban: Will do. I'm not seeing this in any of the runs, only few cases in the very last loop.

oesteban:
@satra Am I correct thinking that this: https://github.com/oesteban/nipype/blob/d2599e2f86df754bfa0fa41b8f6789e2408fb7bd/nipype/pipeline/engine/nodes.py#L1150-L1151 would ensure that `MapNode`s actually inherit the memory/cpu constraints?

effigies:
@oesteban I think you actually want to pass `n_procs=self._n_procs`, so that when it's `None`, it attempts to read `self._interface.inputs.num_threads`.

oesteban:
Oh, good catch!

satra:
@oesteban - correct, mapnodes inherit constraints. solid work - really appreciate the refactoring that was much needed.

assuming tests pass, this looks good to me.

oesteban:
We'll need to revisit the estimation of resources, but what we have with this PR is one step further from where we were before and it is enough to get going.

nipype/interfaces/base.py
oesteban: BTW, I didn't comment on this, but back in #1371 I already played around these making them traits. And I wrote it was a good idea.

We may consider having a traited `interface.settings` object (or similar).

mgxd:
@oesteban what's up with `test_resource_monitor.py`? I see we're skipping it in CI tests - but running it locally will also lead to a fail

```python
    @pytest.mark.skipif(os.getenv('CI_SKIP_TEST', False), reason='disabled in CI tests')
    @pytest.mark.parametrize("mem_gb,n_procs", [(0.5, 3), (2.2, 8), (0.8, 4), (1.5, 1)])
    def test_cmdline_profiling(tmpdir, mem_gb, n_procs):
        """
        Test runtime profiler correctly records workflow RAM/CPUs consumption
        of a CommandLine-derived interface
        """
        from nipype import config
        config.set('execution', 'resource_monitor_frequency', '0.2')  # Force sampling fast
    
        tmpdir.chdir()
        iface = UseResources(mem_gb=mem_gb, n_procs=n_procs)
        result = iface.run()
    
        assert abs(mem_gb - result.runtime.mem_peak_gb) < 0.3, 'estimated memory error above .3GB'
>       assert int(result.runtime.cpu_percent / 100 + 0.2) == n_procs, 'wrong number of threads estimated'
E       AssertionError: wrong number of threads estimated
E       assert 3 == 1
E        +  where 3 = int(((366.80000000000001 / 100) + 0.2))
E        +    where 366.80000000000001 = Bunch(cmdline='/code/nipype/nipype/utils/tests/use_resources -g 1.500000 -p 1', command_path='/code/nipype/nipype/util...5.92338, 1508877125.970061]}, returncode=0, startTime='2017-10-24T20:32:03.693559', stderr='', stdout='', version=None).cpu_percent
E        +      where Bunch(cmdline='/code/nipype/nipype/utils/tests/use_resources -g 1.500000 -p 1', command_path='/code/nipype/nipype/util...5.92338, 1508877125.970061]}, returncode=0, startTime='2017-10-24T20:32:03.693559', stderr='', stdout='', version=None) = <nipype.interfaces.base.InterfaceResult object at 0x7f25b8754278>.runtime
```

oesteban:
Hi, as I [mentioned before](https://github.com/nipy/nipype/pull/2200#issuecomment-333612463) it seems that the measurements we get are not very accurate and those tests fail, especially when running in circleci (I was more successful on my desktop).

Since those tests were already disabled before refactoring, that code is left there for when we have the time and will of making sure that the cpu and memory estimates are accurate. I noted that we should revisit these tests in my [last comment](https://github.com/nipy/nipype/pull/2200#issuecomment-334047223)

mgxd:
@oesteban sounds good - perhaps we should just always skip it until we can assure the estimates are accurate, WDYT?

PR #2201, opened 2017-09-29
Afni 3dCM

salma1601:
add AFNI 3dCM to utils


nipype/interfaces/afni/utils.py
```diff
@@ -642,6 +642,105 @@ def _format_arg(self, name, spec, value):
...
+
+class CenterMassInputSpec(CommandLineInputSpec):
+    in_file = File(
+        desc='input file to 3dCM',
+        argstr='%s',
+        position=-2,
+        mandatory=True,
+        exists=True,
+        copyfile=True)
+    cm_file = File(
```
satra: this should have `hash_files=False` so that this value doesn't get hashed as a file.

salma1601: Ok !

Issue #1400, opened 2016-03-07
xvfbwrapper error


satra:
@oesteban - i'm getting this on my osx machine which has Xvfb

```
======================================================================
ERROR: Failure: RuntimeError (Xvfb did not start
Interface ComputeMeshWarp failed to run. )
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/software/anaconda/envs/devpype/lib/python2.7/site-packages/nose/loader.py", line 251, in generate
    for test in g():
  File "/software/nipy-repo/nipype/nipype/algorithms/tests/test_mesh_ops.py", line 29, in test_ident_distances
    res = dist_ident.run()
  File "/software/nipy-repo/nipype/nipype/interfaces/base.py", line 1032, in run
    runtime = self._run_wrapper(runtime)
  File "/software/nipy-repo/nipype/nipype/interfaces/base.py", line 980, in _run_wrapper
    vdisp.start()
  File "/software/anaconda/envs/devpype/lib/python2.7/site-packages/xvfbwrapper-0.2.8-py2.7.egg/xvfbwrapper.py", line 63, in start
    raise RuntimeError('Xvfb did not start')
RuntimeError: Xvfb did not start
Interface ComputeMeshWarp failed to run. 

======================================================================
ERROR: Failure: RuntimeError (Xvfb did not start
Interface ComputeMeshWarp failed to run. )
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/software/anaconda/envs/devpype/lib/python2.7/site-packages/nose/loader.py", line 251, in generate
    for test in g():
  File "/software/nipy-repo/nipype/nipype/algorithms/tests/test_mesh_ops.py", line 68, in test_trans_distances
    res = dist.run()
  File "/software/nipy-repo/nipype/nipype/interfaces/base.py", line 1032, in run
    runtime = self._run_wrapper(runtime)
  File "/software/nipy-repo/nipype/nipype/interfaces/base.py", line 980, in _run_wrapper
    vdisp.start()
  File "/software/anaconda/envs/devpype/lib/python2.7/site-packages/xvfbwrapper-0.2.8-py2.7.egg/xvfbwrapper.py", line 63, in start
    raise RuntimeError('Xvfb did not start')
RuntimeError: Xvfb did not start
Interface ComputeMeshWarp failed to run. 

----------------------------------------------------------------------
```

more generally this is what Xvfb does in user mode:

```
$ Xvfb 
[...]

PR #2203, opened 2017-10-03
ENH: Centralize virtual/physical $DISPLAYs

### LINKED ISSUE #1403: xvfb-run not ending gracefully for CommandLine interfaces

### LINKED ISSUE #2055: DISPLAY variable should be optional
### Summary

DISPLAY variable is always set to :1 when running a command line interface. There is currently now way to unset this variable

### Actual behavior

DISPLAY=:1 by default

### Expected behavior
Option to remove DISPLAY or any other environment variable.

### Platform details:

{'pkg_path': '/software/nipy-repo/nipype/nipype', 'commit_source': 'repository', 'commit_hash': 'acb37ce18', 'nipype_version': '1.0.0-gacb37ce18.dev', 'sys_version': '3.6.0 |Continuum Analytics, Inc.| (default, Dec 23 2016, 13:19:00) \n[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)]', 'sys_executable': '/software/miniconda3/envs/dev36pype/bin/python', 'sys_platform': 'darwin', 'numpy_version': '1.12.0', 'scipy_version': '0.18.1', 'networkx_version': '1.11', 'nibabel_version': '2.1.0', 'traits_version': '4.6.0'}
1.0.0-gacb37ce18.dev

### Execution environment

Choose one
- My python environment inside container [Base Tag: nipy/mindboggle]

### LINKED ISSUE #2206: AFNI's SkullStrip needs a $DISPLAY again after version 16.2.07
### Summary
It seems that AFNI's `3dSkullStrip` is again linking libraries that need a functional X server.

### Actual behavior
```
2017-09-25 20:17:53,524 workflow:INFO Running: 3dSkullStrip -input /scratch/work/workflow_enumerator/anatMRIQCT1w/AFNISkullStripWorkflow/_in_file_..data..circle-tests..sub-50137..anat..sub-50137_T1w.nii.gz/skullstrip/sub-50137_T1w_conformed_corrected.nii.gz -prefix sub-50137_T1w_conformed_corrected_skullstrip.nii.gz
2017-09-25 20:17:53,613 workflow:ERROR ['Node skullstrip.a0 failed to run on host 96ff9afc656c.']
2017-09-25 20:17:53,614 workflow:INFO Saving crash info to /scratch/out/logs/crash-20170925-201753-root-skullstrip.a0-c8da4299-b3c4-4afd-bda5-14d2a723f97d.txt
2017-09-25 20:17:53,614 workflow:INFO Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/plugins/multiproc.py", line 52, in run_node
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 372, in run
    self._run_interface()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 482, in _run_interface
    self._result = self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 613, in _run_command
    result = self._interface.run()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 1084, in run
    runtime = self._run_wrapper(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 1758, in _run_wrapper
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/afni/base.py", line 188, in _run_interface
    return super(AFNICommand, self)._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/afni/base.py", line 126, in _run_interface
    return super(AFNICommandBase, self)._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 1792, in _run_interface
    self.raise_exception(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 1716, in raise_exception
    **runtime.dictcopy()))
RuntimeError: Command:
3dSkullStrip -input /scratch/work/workflow_enumerator/anatMRIQCT1w/AFNISkullStripWorkflow/_in_file_..data..circle-tests..sub-50137..anat..sub-50137_T1w.nii.gz/skullstrip/sub-50137_T1w_conformed_corrected.nii.gz -prefix sub-50137_T1w_conformed_corrected_skullstrip.nii.gz
Standard output:

Standard error:
freeglut (3dSkullStrip): failed to open display ':1'

Return code: 1
Interface SkullStrip failed to run.
```

### Expected behavior
The interface should run, opening an Xvfb if no X is available.

# Other
  - How to replicate the behavior
  - Script/Workflow details
  - Platform details:
  - Execution environment

All of that on MRIQC's tests: https://circleci.com/gh/poldracklab/mriqc/1935

oesteban:
This PR addresses several (related) problems:

  - Makes `$DISPLAY` and the config `display_variable` optional (close #2055)
  - Should fix #1403 since xvfb-run is not used anymore.
  - Relates to #1400:
    - Will reduce its impact because now Xvfb is called only once and only if it is absolutely necessary
    - May make it worse in some cases when Xvfb fails to create a listener (root needed?).
  - (EDIT: added later) Fixes #2206.

This PR adds a `config.get_display()` which identifies what display should be used, and creates a virtual one if necessary. Also adds a few unit tests to the config object to make sure precedence is fulfilled. If the virtual framebuffer is created, it is created only once for the whole workflow (solves the problem of many virtual fbs on a mapnode with `_redirect_x = True`.

One open question is: do we want to double check that an X server is running if `$DISPLAY` or `display_variable` are `:0`?
https://github.com/oesteban/nipype/blob/e8e18f039ce6ab7c7bd19c045e083bf02275c7c0/nipype/utils/config.py#L180-L185


effigies:
> One open question is: do we want to double check that an X server is running if $DISPLAY or display_variable are :0?

This seems to me like outside the scope of nipype. If I set `DISPLAY=":0"` and don't have X running, programs that need X will crash. If an interface crashes and doesn't give a "No display found on $DISPLAY" message, I think it's within scope for that interface to give a better message.

oesteban:
Thanks a lot for the comment.

> This strikes me as being outside the scope of nipype. If I set `DISPLAY=":0"` and don't have X running, programs that need X will crash. If an interface crashes and doesn't give a "No display found on $DISPLAY" message, I think it's within scope for that interface to give a better message.

Totally agree, and that's the reason why I commented that part out, just to have the code handy if needed. It was just a suggestion that could make nipype easier for the user with better error messages (instead of relying on the third party libraries for these infrastructure things).

However, there is a hidden complexity: if you set the `$DISPLAY=':1'` because you have a VNC server running, then the way to identify the process is different. And even more complicated if the display is remote (`$DISPLAY='<ip-of-different-host>:6001'`).

I'll leave it there for the future, if we want to check for it only when `$DISPLAY='localhost:0'`.

nipype/utils/config.py
```diff
@@ -172,3 +173,52 @@ def update_matplotlib(self):
     def enable_provenance(self):
         self._config.set('execution', 'write_provenance', 'true')
         self._config.set('execution', 'hash_method', 'content')
+
+    def get_display(self):
```
effigies: If I understand it, if this has not been called before workflow runtime, then each node will get [a copy of `config`](https://github.com/nipy/nipype/blob/ea62ca6ea9a788d6f98c3a5a57b582e2b383280f/nipype/pipeline/engine/nodes.py#L274-L277) and thus start up its own `Xvfb` object.

And I don't see anywhere where it's stopped.

oesteban: The first time you call it, one display will be locked. As a matter of fact, when I first wrote the tests (https://github.com/oesteban/nipype/blob/fix/VirtualDisplay/nipype/utils/tests/test_config.py#L9-L49) I realized I had to manally reset the config object all the times because once it is called it is fixed to that display. Therefore, all those `config._display = None` and those re-setting/removing the nipype config setting.

Does this reply to your question?

effigies: Okay. I meant to put at the end of the first sentence: "Is this the desired behavior?" Sounds like it is.

Similarly, the fact that [`self._display.stop()`](https://github.com/nipy/nipype/pull/2203/files#diff-1cce190a2f31ba127aae1067c3860abfL1035) is no longer called anywhere is intentional?

oesteban: Yes, that was actually the problem: nipype would spawn an endless number of displays.

Well, it is probably nice to add a `self._display.stop()` at the end when running a workflow. I'm not sure though about how to do that when running bare interfaces.

effigies:
What about in __del__, or whatever python's cleanup method is? Or if
there's a post run hook.

On Oct 3, 2017 5:38 PM, "Oscar Esteban" <notifications@github.com> wrote:

*@oesteban* commented on this pull request.
------------------------------

In nipype/utils/config.py
<https://github.com/nipy/nipype/pull/2203#discussion_r142531022>:

> @@ -172,3 +173,52 @@ def update_matplotlib(self):
     def enable_provenance(self):
         self._config.set('execution', 'write_provenance', 'true')
         self._config.set('execution', 'hash_method', 'content')
+
+    def get_display(self):

Yes, that was actually the problem: nipype would spawn an endless number of
displays.

Well, it is probably nice to add a self._display.stop() at the end when
running a workflow. I'm not sure though about how to do that when running
bare interfaces.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/nipy/nipype/pull/2203#discussion_r142531022>, or mute
the thread

nipype/utils/config.py
oesteban: Ok, added a callback when python exists the session :D.

oesteban:
This PR fixed the problem in MRIQC, and the whole workflow run correctly: https://circleci.com/gh/oesteban/mriqc/1368

nipype/interfaces/afni/preprocess.py
```diff
@@ -2146,11 +2146,12 @@ class SkullStrip(AFNICommand):
...
     def __init__(self, **inputs):
         super(SkullStrip, self).__init__(**inputs)
+
         if not no_afni():
             v = Info.version()
 
-            # As of AFNI 16.0.00, redirect_x is not needed
-            if v[0] > 2015:
+            # Between AFNI 16.0.00 and 16.2.07, redirect_x is not needed
+            if v >= (2016, 0, 0) and v < (2016, 2, 7):
```
satra: woah  - is this really true - crazy!

oesteban: oh yeah, tell me about it. I'll report this to AFNI, to make sure this is not a mistake linking.

oesteban:
@satra: added the error plus a unit test for it

@effigies and Satra, this is ready for a final review.

doc/users/config_file.rst
```diff
@@ -73,13 +73,18 @@ Execution
...
+	``_redirect_x = True``. This option is very useful when the system has
+	an X server listening in the default port 6000, but ``$DISPLAY`` is
+	not defined. In that case, set ``display_variable = :0``. Similarly,
+	it can be used to point X-based interfaces to other servers, like VNC,
+	`xnest <http://www.x.org/archive/X11R7.5/doc/man/man1/Xnest.1.html>`_
+	or `Xvfb <http://www.x.org/archive/X11R6.8.1/doc/Xvfb.1.html>`_
+	and you would like to redirect all spawned windows to
+	it. If not set, nipype will try to configure a new virtual server using
+	Xvfb. (possible values: any X server address; default value: not
+	set)
```
effigies: This is a little unclear. If I understand it correctly, we use the input `$DISPLAY` variable unless this is set? Under that assumption, here's a proposed rewrite:

```
Override the ``$DISPLAY`` environment variable for interfaces that require an X server.
This option is useful if there is a running X server, but ``$DISPLAY`` was not defined
in nipype's environment. For example, if an X server is listening on the default port of
6000, set ``display_variable = :0`` to enable nipype interfaces to use it. It may also
point to displays provided by VNC, `xnest
<http://www.x.org/archive/X11R7.5/doc/man/man1/Xnest.1.html>`_ or `Xvfb
<http://www.x.org/archive/X11R6.8.1/doc/Xvfb.1.html>`_. If neither ``display_variable``
nor the ``$DISPLAY`` environment variable are set, nipype will try to configure a new
virtual server using Xvfb. (possible values: ...)
```

nipype/utils/tests/test_config.py
```diff
@@ -0,0 +1,122 @@
...
+    monkeypatch.setitem(os.environ, 'DISPLAY', dispstr)
+    assert config.get_display() == dispstr
+
+
+def test_display_config_and_system(monkeypatch):
+    """Check that when only both config and $DISPLAY are defined, the config takes precedence"""
+    config._display = None
+    dispstr = ':10'
+    config.set('execution', 'display_variable', dispstr)
+    monkeypatch.setitem(os.environ, 'DISPLAY', dispstr)
```
effigies: Should this not be something other than `dispstr`?

nipype/utils/tests/test_config.py
```diff
@@ -0,0 +1,122 @@
...
+    config._display = None
+    if config.has_option('execution', 'display_variable'):
+        config._config.remove_option('execution', 'display_variable')
+    monkeypatch.delitem(os.environ, 'DISPLAY', raising=False)
+    monkeypatch.setitem(sys.modules, 'xvfbwrapper', xvfbpatch)
+    assert config.get_display() == ":2010"
+
+
+def test_display_empty_patched(monkeypatch):
+    """Check that when no display is specified, a virtual Xvfb is used"""
```
effigies: Docstrings after this one are all the same.

nipype/utils/config.py
```diff
@@ -242,4 +252,77 @@ def resource_monitor(self, value):
...
+            self._display.stop()
+
+
+@atexit.register
+def free_display():
+    """Stop virtual display (if it is up)"""
+    from nipype import config
+    from nipype import logging
+    config.stop_display()
+    logging.getLogger('interface').info('Closing display (if virtual)')
```
effigies: If nipype is going to spawn many `Xvfb`s, shouldn't the stop be associated with the interface, rather than just closing one when Python quits?

I think I must be missing something here.

oesteban: Actually the opposite. You want nipype to run only one `Xvfb` if possible. Let's say you start an ipython session, run an interface with `_redirect_x = True` and then another interface than again requires X. With the current implementation only one `Xvfb` is started. Only when you exit ipython the Xvfb service is killed.

The rationale is precisely stop spawning many `Xvfb`s which was the case before

doc/users/config_file.rst
oesteban: You are right. `display_variable` should override `$DISPLAY`. I'll change the code and use your proposed text for the documentation.

satra:
@oesteban -  a couple of follow up questions to @effigies' comments:

i'm enumerating some use cases:

1. one machine, one user - single multiproc process
2. one machine, one user - multiple processes from different scripts/slurm/etc
3. one machine, many users

in my head, 1 and 2 would reuse the same xvfb, however this will not work unless there is some reference counting in place and the one that shuts it down is the last one to finish. this could also result in race conditions of some kind.

3 would launch different xvfb's per user. but within each user's space considerations for 1 and 2 would apply.

oesteban:
> 1. one machine, one user - single multiproc process
Will reuse same `Xvfb`.

> 2. one machine, one user - multiple processes from different scripts/slurm/etc
Will open one new `Xvfb` per process. This is not ideal, but after all we are reducing the overload of `Xvfb` by a lot.

> 3. one machine, many users
Will open as many `Xvfb`s as users x processes.

For use cases 2 and 3 we should generally recommend setting up VNC (for clusters) or a unique Xvfb, and setting `display_variable` in all cases.

There is a compromise here between how much we want to make it seamless to the user and basics of infrastructure. The former implementation was extremely inefficient (spawning literally hundreds of `Xvfb`, we've seen that in the early days of MRIQC) so this is a great advance. We also get rid of `xvfb-run` for `CommandLine` interfaces, also very advisable and not always installed in clusters.

nipype/utils/config.py
effigies: I see. And you're testing to make sure you're not producing multiple by running multiple interfaces? Because my [first comment](https://github.com/nipy/nipype/pull/2203#discussion_r142490521) noted that it looks like each Node makes a copy of `config` before spawning an `xvfb`, so it's not clear they would actually share it.

Again, this is from a very cursory look, so it's possible I'm missing things.

satra:
i think we still need to support 2.  imagine submitting slurm jobs on a cluster - it gets assigned to random nodes. if every job has to start a vnc server i don't think that's a good outcome. i'd rather have every job starting an xvfb session.

satra:
@oesteban - nevermind my previous comment on case 2 - i think we are still ok - case 2 will still launch a different xvfb per process.

oesteban:
@satra correct :). The vnc idea is when you have one node providing that, otherwise it's better to fall back to the default (ergo `Xvfb`).

oesteban:
An important addendum to https://github.com/nipy/nipype/pull/2203#discussion_r142783272 is that `Xvfb().stop()` ensures that the `$DISPLAY` variable is set to the original value. The only side effect that I could see if that `$DISPLAY == ''`, for which we unset `$DISPLAY` and it is not recovered. I thought this is not extremely relevant, though.

oesteban:
@satra, @effigies, this is ready for final review

nipype/utils/config.py
satra: @effigies - config makes copies of sections, not the entire config object. so as long as the config.get_display is called, nipype maintains a global config/state.

satra:
@oesteban - lgtm

PR #2204, opened 2017-10-03
allow float as weight in afni's Allineate

salma1601:
AFNI's `3dAllineate` accepts either float or dataset as weight

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -353,6 +353,13 @@ class AllineateInputSpec(AFNICommandInputSpec):
         argstr='-nomask',
         desc='Don\'t compute the autoweight/mask; if -weight is not '
              'also used, then every voxel will be counted equally.')
+    weight_file = File(
+        argstr='-weight %s',
+        exists=True,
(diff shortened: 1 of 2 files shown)


nipype/interfaces/afni/preprocess.py
```diff
@@ -353,12 +353,13 @@ class AllineateInputSpec(AFNICommandInputSpec):
         argstr='-nomask',
         desc='Don\'t compute the autoweight/mask; if -weight is not '
              'also used, then every voxel will be counted equally.')
-    weight_file = File(
```
oesteban: AFNI Allineate has been there for a long while. Would it make sense to leave the `weight_file` input here with the deprecated meta set?
```Python
    weight_file = File(
        argstr='-weight %s',
        exists=True,
        deprecated='1.0.0', new_name='weight',
        desc='Set the weighting for each voxel in the base dataset; '
             'larger weights mean that voxel count more in the cost function. '
             'Must be defined on the same grid as the base dataset')
```

salma1601: >Would it make sense to leave the weight_file input here with the deprecated meta set?

Ok, sorry.

PR #2205, opened 2017-10-03
 FIX: Create out_reg_file correctly in RobustRegister

effigies:
RobustRegister does not correctly return the full path to manually-specified `out_reg_file` names, and, if unspecified, creates them in the original folder.

- Make `out_reg_file` a normal `Either(True, File)` trait, with default value `True`
- `out_reg_file` is created in working directory by default

Also did some cleanup. See 5be5de94801bb9b07404a8c35d469f4d472fe241 for significant changes.


effigies:
I'll go ahead and merge, then.

PR #2209, opened 2017-10-04
Sort out terminal_output

### LINKED ISSUE #1407: [ENH] terminal_output as output
While addressing #1406 and #1398, I've found that the `terminal_output` is not mapped to any output when set to `'file'`.

I don't know if this was intended as a feature, but I see it more like an error. Moreover, we have the `StdOutCommandLine` that workarounds this behaviour.

What is the rationale behind maintaining these two ways to do the same thing? Would joining them together make sense? Or maybe `StdOutCommandLine` should disappear in favor of a `CommandLine` that provides one output (currently named `out_file` for the `StdOutCommandLine`) when `terminal_output = 'file'`?

I would implement whatever you see it is the best solution to this issue because it blocks me in finishing the two above-mentioned issues.

oesteban:
This PR revises the terminal_output feature, making it a class attribute and deprecating the old InputSpec trait.

It also implements new ways of storing the output of a CommandLine, as discussed in #1407. Thus, closes #1407.

Work in progress:
  - [x] Update documentation
  - [x] Before merging, I `make specs` should be run. I'm holding off to ease code review.


oesteban:
@satra, @effigies, this is ready for final review

satra:
@oesteban - this will need to resolve a few things post the display PR merge.

effigies:
@satra @oesteban The conflicts looked simple to resolve to me, so I went ahead and did it to get the tests queued up. Please check the diff to make sure you agree with the changes.

oesteban:
Testing one conflict that wasn't correctly resolved. If it passes locally I'll push very soon. Thanks a lot @effigies for the merge with upstream :)

nipype/interfaces/base.py
```diff
@@ -1616,11 +1638,9 @@ def _format_arg(self, name, trait_spec, value):
             # Depending on whether we stick with traitlets, and whether or
             # not we beef up traitlets.List, we may want to put some
             # type-checking code here as well
-            sep = trait_spec.sep
-            if sep is None:
-                sep = ' '
-            if argstr.endswith('...'):
+            sep = trait_spec.sep or ' '
```
effigies: This will replace `''` with `' '`. I'm not sure that's what people will expect. Actually equivalent:

```Python
sep = trait_spec.sep if trait_spec.sep is not None else ' '
```

nipype/interfaces/base.py
```diff
@@ -1761,10 +1781,7 @@ class StdOutCommandLine(CommandLine):
     input_spec = StdOutCommandLineInputSpec
 
     def _gen_filename(self, name):
-        if name == 'out_file':
-            return self._gen_outfilename()
-        else:
-            return None
+        return self._gen_outfilename() if name == 'out_file' else None
```
effigies: Functions return `None` implicitly. This could just drop the `else` condition.

```Python
def _getn_filename(self, name):
    if name == 'out_file':
        return self._gen_outfilename()
```

I'm not sure which one is easier to understand. Your choice.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -137,8 +137,11 @@ def _clear_task(self, taskid):
 
     def _submit_job(self, node, updatehash=False):
         self._taskid += 1
-        if getattr(node.inputs, 'terminal_output', '') == 'stream':
-            node.inputs.terminal_output = 'allatonce'
+
+        # Don't allow streaming outputs
+        if hasattr(node.interface, 'terminal_output') and \
+                node.interface.terminal_output == 'stream':
+            node.interface.terminal_output = 'allatonce'
```
effigies: This might be something I messed up with my merge, but I think the `getattr` approach is cleaner.

```Python
# Don't allow streaming outputs
if getattr(node.interface, 'terminal_output', '') == 'stream':
    node.interface.terminal_output = 'allatonce'
```

nipype/interfaces/base.py
```diff
@@ -1457,7 +1471,7 @@ class must be instantiated with a command argument
     {'args': '-al',
      'environ': {'DISPLAY': ':1'},
      'ignore_exception': False,
-     'terminal_output': 'stream'}
+     'terminal_output': <undefined>}
```
satra: this is syntactically off.

nipype/interfaces/base.py
```diff
@@ -1652,7 +1672,7 @@ def _filename_from_source(self, name, chain=None):
             ns = trait_spec.name_source
             while isinstance(ns, (list, tuple)):
                 if len(ns) > 1:
-                    iflogger.warn('Only one name_source per trait is allowed')
+                    iflogger.warning('Only one name_source per trait is allowed')
```
satra: i think in a next refactor, we should move this up to the base class.

nipype/interfaces/afni/tests/test_auto_ROIStats.py
```diff
@@ -25,9 +25,8 @@ def test_ROIStats_inputs():
     quiet=dict(argstr='-quiet',
     position=1,
     ),
-    terminal_output=dict(mandatory=True,
+    terminal_output=dict(deprecated='1.0.0',
```
satra: can we set version temporarily to `1.0.0` and make sure nothing breaks :) doesn't have to be a in a test. could be your local test.

doc/devel/interface_specs.rst
```diff
@@ -159,6 +159,70 @@ generated depending on inputs, by the tool.  OutputSpecs inherit from
...
+In all cases, except for the ``'none'`` setting of ``terminal_output``,
+the ``run()`` method will return a "runtime" object that will contain
+the streams in the corresponding properties (``runtime.stdout``
+for the standard output, ``runtime.stderr`` for the standard error, and
+``runtime.merged`` for both when streams are mixed, eg. when using the
+*file* option). ::
+
+  import nipype.interfaces.fsl as fsl
+  mybet = fsl.BET(from_file='bet-settings.json')
+  mybet.terminal_output = 'file_split'
```
satra: does this suggest that we should consider two types of fields in the json files? things that set node/interface attributes, like mem, proc, terminal_output, etc.,. and then a section that's exclusively the inputs?

oesteban: good point. We could think of a JSON representation of the interface state for nipype 2.0, don't you think?

oesteban:
I still have to add in your comments and `make specs`. Please hold off merging till then.

nipype/interfaces/base.py
oesteban: what about:
```Python
    # Use get_traitsfree() to check all inputs set
    >>> pprint.pprint(cli.inputs.get_traitsfree())  # doctest: +NORMALIZE_WHITESPACE +ALLOW_UNICODE
    {'args': '-al',
     'environ': {'DISPLAY': ':1'},
     'ignore_exception': False}
```

nipype/interfaces/base.py
oesteban: I generally prefer explicit over implicit ...

nipype/interfaces/afni/tests/test_auto_ROIStats.py
oesteban: Running locally with version `1.0.1`. Will also run some examples.

oesteban: Confirmed it runs correctly.

PR #2211, opened 2017-10-05
[ENH] Minor Xvfb and MultiProcPlugin cleanups

effigies:
Quick follow-up to #2200, #2203, #2208.

- Use relative import to ensure the correct nipype's config is being imported
- Remember `MultiProcPlugin._send_proc_to_workers` stats between calls to reduce noise


mgxd:
I was just about to comment about `_send_procs_to_workers` but you beat me to the punch :)

effigies:
Added a bunch of `make specs` updates that got missed in #2200.

@mgxd Let me know if there are any other cleanups that should go in with this.

mgxd:
LGTM - one little thing that could be useful is supplementing the information when running a `Function` interface [here](https://github.com/nipy/nipype/blob/6e95b3c25ea8fe461d3612cc3b1643cb0dae1fa8/nipype/pipeline/engine/nodes.py#L636-L646).

Currently, this is output in the log
```
171005-16:29:28,86 workflow INFO:
	 Running a "Function" interface.
171005-16:29:28,113 workflow INFO:
	 Running a "Function" interface.
```
which is nice, but maybe we can add the node/function name in this case to get a better understanding?

This could also be addressed in an add-on PR

effigies:
What do you think of 5b367f2?

oesteban:
Hopefully this will polish those little things here and there. Thanks a lot @effigies

effigies:
Reviewed Oscar's changes. I'm +1 to merge if tests pass.

PR #2214, opened 2017-10-05
fixes afni CenterMass bug

### LINKED ISSUE #2213: [Bug] Type error in afni's CenterMass
```Python
from nipype.interfaces import afni
a = afni.CenterMass()
a.inputs.in_file = 'anat.nii.gz'
a.inputs.cm_file = 'cm.txt'
a.run()
```
throws an error because single line file `cm.txt` is not loaded in an array of length one
```
TypeError: 'numpy.float64' object is not iterable
```

salma1601:
fixes #2213


effigies:
Does this have any effect on this logic?

https://github.com/nipy/nipype/blob/5526eeda7454c7799f0dc4b105bee1a2f7241206/nipype/interfaces/afni/utils.py#L738-L741

In particular I'm wondering whether the second option should now be `tuple(sout[0])`. I don't have much context here, so I could be entirely wrong.

salma1601:
>In particular I'm wondering whether the second option should now be tuple(sout[0]). I don't have much context here, so I could be entirely wrong.

You are entirely right. In the basic case, only one centre of mass is computed, so there is only one line in the `cm_file` and the output should be one coordinates tuple. If multiple centres of mass (for instance different ROIs), the output would be a list of tuples.
I am correcting, sorry for that.

salma1601:
>The fact that this is the first time this is coming up suggests nobody's been depending on this single-line case up to now, so it would be a minor breakage.

Well actually `afni.CenterMass` has been merged 6 days ago so may be code simplification is more important than sticking to previous API ?

effigies:
Cool. Looks good to me.

PR #2215, opened 2017-10-06
ENH: Flesh out ConcatenateLTA

effigies:
- Adds most options to `mri_concatenate_lta` (skipping `-rmsdiff` because it's a weird one and I don't need it)
- Update existing fields to match tool specs (see especially `in_lta2`)


effigies:
A quick review of this one would be greatly appreciated.

nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -2397,23 +2422,44 @@ class ConcatenateLTAOutputSpec(TraitedSpec):
...
     _cmd = 'mri_concatenate_lta'
     input_spec = ConcatenateLTAInputSpec
     output_spec = ConcatenateLTAOutputSpec
 
+    def _format_arg(self, name, spec, value):
+        if name == 'out_type':
+            value = {'VOX2VOX': 0, 'RAS2RAS': 1}[value]
+        return super(ConcatenateLTA, self)._format_arg(name, spec, value)
+
     def _list_outputs(self):
```
satra: perhaps this can be removed?

effigies: I'm guessing something else would be needed to change? Removing `_list_outputs` gets me with an `<undefined>` `outputs.out_file`.

satra: `name_source` in outputspec

satra:
@effigies - just merged oscar's terminal output PR, so make specs would need to be redone.

nipype/interfaces/freesurfer/preprocess.py
effigies: I'm not able to figure out how to use `name_source`. Has no effect I can see in output spec, and trying to follow the examples in the [docs](https://nipype.readthedocs.io/en/latest/devel/interface_specs.html) results in failing to generate a string for `inputs.out_file` and the interface fails.

satra: nevermind - the logic i had in mind is only implemented if `name_source` is in the input trait. i think the next refactor needs to take care of filename generation

effigies:
@satra Any other comments? Good to merge?

PR #2219, opened 2017-10-08
interface for AFNI's 3dNwarpCat

salma1601:
Trying to interface `3dNwarpCat`


salma1601:
@nadkarni-na Are there other functions to add in https://github.com/salma1601/nipype/blob/74deb3849e18f5e5202bf8a40208b017dc10c345/nipype/interfaces/afni/utils.py#L1596 ?

nadkarni-na:
yes but they're weird, this will do for now

salma1601:
Ok thanks !

satra:
@salma1601 - what's the difference between this and #2254? should this be merged before that? or is this redundant with it?

salma1601:
No difference, sorry I completely forgot I already opened a PR for NwarpCat.
There is a little fix I need to do here, then I remove the WIP label

PR #2083, opened 2017-06-20
WIP: Allow dicts as Interface input and output specs

mwaskom:
See #2081 for motivation. Here is a basic proof of principle:

```python
import os
import nibabel as nib
from nipype.interfaces.base import BaseInterface, traits


class MeanImage(BaseInterface):

    input_spec = dict(in_file=traits.File())
    output_spec = dict(mean_val=traits.Float())

    def _list_outputs(self):

        outputs = self._outputs().get()
        outputs["mean_val"] = self._mean_val
        return outputs

    def _run_interface(self, runtime):

        data = nib.load(self.inputs.in_file).get_data()
        self._mean_val = data.mean()
        return runtime


if __name__ == "__main__":

    fname = "/home/mwaskom/code/nipype/nipype/testing/data/tpm_00.nii.gz"
    interface = MeanImage(in_file=fname)
    print(interface.run().outputs)
```

I did run into some issues in type() with `__future__.unicode_literals` and then what I am guessing is whatever cross 2/3 magic nipype does to `str()` that will need to be sorted out...

PR #2220, opened 2017-10-09
ENH: Add SimpleInterface

effigies:
This adds `niworkflows.interfaces.base.SimpleInterface` to `nipype.interfaces.base`.

`SimpleInterface` is an interface that gives dictionary access to the output spec. We've found it a very useful boilerplate-reducer for making interfaces by simply defining input/output specs and `_results`.

Merging this would reduce friction for a lot of other poldracklab (niworkflows, mriqc, fmriprep) interfaces that people may want to see merged into nipype proper. I'm open to cleaning it up to be slightly less simple, as long as it achieves the same reduction in boilerplate in subclasses.

Slightly related: #2083

@oesteban EDITED: changed `_runtime` for `_results`.

### DIFF
--- a/nipype/interfaces/base.py
+++ b/nipype/interfaces/base.py
@@ -1241,6 +1241,11 @@ class SimpleInterface(BaseInterface):
     ...     def _run_interface(self, runtime):
     ...          self._results['doubled'] = double(self.inputs.x)
     ...          return runtime
+
+    >>> dbl = Double()
+    >>> dbl.inputs.x = 2
+    >>> dbl.run().outputs.doubled
+    4.0
     """
     def __init__(self, from_file=None, resource_monitor=None, **inputs):
         super(SimpleInterface, self).__init__(
--- a/nipype/interfaces/base.py
+++ b/nipype/interfaces/base.py
@@ -1212,6 +1212,41 @@ class BaseInterface(Interface):
             json.dump(inputs, fhandle, indent=4, ensure_ascii=False)
 
 
+class SimpleInterface(BaseInterface):
+    """ An interface pattern that allows outputs to be set in a dictionary
+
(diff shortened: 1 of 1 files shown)


nipype/interfaces/base.py
```diff
@@ -1212,6 +1212,41 @@ def save_inputs_to_json(self, json_file):
             json.dump(inputs, fhandle, indent=4, ensure_ascii=False)
 
 
+class SimpleInterface(BaseInterface):
+    """ An interface pattern that allows outputs to be set in a dictionary
```
oesteban: I'd suggest:

```Python
"""
An interface pattern that allows outputs to be set in a dictionary called ``_results``
that is automatically interpreted by ``_list_outputs()`` to find the outputs.

...
"""
```

nipype/interfaces/base.py
```diff
@@ -1212,6 +1212,41 @@ def save_inputs_to_json(self, json_file):
...
+
+    When implementing `_run_interface`, set outputs with::
```
oesteban: Double quotes for the docstring: 
```
When implementing ``_run_interface``, set outputs with::
```

nipype/interfaces/base.py
```diff
@@ -1212,6 +1212,41 @@ def save_inputs_to_json(self, json_file):
...
+
+        self._results[out_name] = out_value
+
+    This can be a way to upgrade a ``Function`` interface to do type checking:
```
oesteban: Missing:

```
Example
-------
```

That is a standard in the docstring of nipype interfaces

satra:
LGTM - i can see how this is useful and keeps data validation.

effigies:
@oesteban any more comments?

oesteban:
Nope :)

PR #2222, opened 2017-10-09
Afni qwarp output paths

### LINKED ISSUE #2221: AFNI Qwarp output paths are incorrect
### Summary
output paths are not correct: current directory is always used as parent directory and `+tlrc` is appended even to nifti outputs

``` Python
from nipype.interfaces import afni
a = afni.Qwarp()
a.inputs.in_file = '/tmp/in_file.nii.gz'
a.inputs.base_file = '/tmp/base_file.nii.gz'
a.inputs.out_file = '/tmp/warped.nii.gz'
res = a.run()
print(res.outputs.warped_source)
```
prints a wrong path
``` Python
In [20]:  /home/salma/CODE/nipype/warped14+tlrc.nii.gz
```

salma1601:
try fix #2221

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -15,7 +15,8 @@ from builtins import open
 import os
 import os.path as op
 
-from ...utils.filemanip import (load_json, save_json, split_filename)
+from ...utils.filemanip import (load_json, save_json, split_filename,
+                                fname_presuffix)
 from ..base import (
     CommandLineInputSpec, CommandLine, TraitedSpec,
     traits, isdefined, File, InputMultiPath, Undefined, Str)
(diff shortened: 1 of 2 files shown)


nipype/interfaces/afni/preprocess.py
```diff
@@ -3475,29 +3476,38 @@ def _list_outputs(self):
         if not isdefined(self.inputs.out_file):
```
oesteban: All of this should not be here:

1. The `out_file` input spec can only define one of `name_source` or `genfile`. Since we are trying to deprecate `genfile`, `name_source` should be preferred.

2. In theory, `name_source` should fix this issue.

satra: @oesteban - i'm leaving the afni stuff to @Shotgunosine and others to figure out. right now, because of the way afni works with various kinds of extensions, the straight up namesource can run into issues.

PR #2224, opened 2017-10-12
[FIX] 3dUndump interface

jdkent:
- fixes the interface for 3dUndump for some of its options
- adds documentation from the afni command 3dUndump


effigies:
Was `-srad` removed at a specific version and `-orient` added? You may want to have both, setting `max_ver` and `min_ver` arguments to these traits, rather than replacing one with the other.

jdkent:
@effigies: In the code, the option `-srad` was repeated twice and I think it they meant to include `-orient`, but it was an oversight. `-srad` [is still a valid option](https://github.com/jdkent/nipype/blob/f1a76cb950a9cc29b9da335f8cccf960b04fb0a6/nipype/interfaces/afni/utils.py#L2095)

effigies:
Ah, I didn't read enough context. LGTM.

PR #2226, opened 2017-10-13
rearranging nilearn test, so they are independent from each other

djarecka:
I had problem when testing traitlets that tests were not completely independent from each other, so decided to remove the class and rearrange, so they run now in different directories etc.


djarecka:
I can add it to my traitlets PR

Issue #2207, opened 2017-10-04
Raise exception if BIDSDataGrabber is imported using Python 2


adelavega:
Pybids no longer supports Python 2. Thus, BIDSDataGrabber should only be importable with Python 3, and raise an exception otherwise.

adelavega:
I'd assign myself but I can't ;)

oesteban:
I'd also include here setting the appropriate skips on tests:
  - [ ] Because Python 2
  - [ ] Because pybids was not installed in editable mode (very likely in user settings).

adelavega:
Cool. Looks like Python 2 support might continue enough to the point that perhaps raise an Exception is a bit heavy, but it probably makes sense to skip on tests either way unless pybids CI adds back official python 2 support.

effigies:
Is this actually a problem? Does BIDSDataGrabber depend on behavior since pybids dropped Python 2 support?

If not, can we set the Python 2 docker image to use the last Py2-compatible release tag?

So this:

https://github.com/nipy/nipype/blob/master/Dockerfile#L90-L91

Becomes

```Dockerfile
RUN curl -sSL https://github.com/INCF/pybids/archive/$TAG.tar.gz | tar xvz &&
    cd pybids && python setup.py develop
```

And you can set TAG to `0.3.0` for Py2, or `master` for current behavior.

PR #2228, opened 2017-10-13
ENH: More flexible BIDSDataGrabber creation

effigies:
- Remove `outfields` parameter - redundant with `output_query` input
- Allows inputs to be specified in `kwargs`
- auto tests


adelavega:
Cool. Although I think I had the `outfields` parameter because that gets called on `init` whereas `output_query` input is not called until run time. I figured the former would allow checking a workflow has the correct inputs/ouputs, but maybe that's not necessary?

adelavega:
Any desire to take care #2207 while you're at it? Although I'm not sure it was confirmed whether or not Python 2 support will be entirely dropped out of pybids, so maybe just add a skiptest if pybids was not installed in editable mode?

oesteban:
@adelavega, Chris' solution is equivalent to having `outfields`. Inputs can be passed through ``__init__`` so you could always have done:

```Python
bids = BIDSDataGrabber(output_query={ ... })
```
and that would have the same effect as:

```Python
bids = BIDSDataGrabber(outfields={ ... })
bids.inputs.output_query = { ... }
```

PR #2229, opened 2017-10-13
changing _format_arg in WarpImageMultiTransform, should fix #2192 

### LINKED ISSUE #2192: ANTS interface: incorrect parameter passing in WarpTimeSeriesImageMultiTransform
### Summary
when using WarpTimeSeriesImageMultiTransform.inputs.invert_affine = [1], the parameter passed to the corresponding ANTS command is ['-i'], and should instead be '-i' (as in WarpImageMutiTransfrom) in resampling.py

### Actual behavior
The command fails to run or to write with .cmdline
(TypeError: sequence item 0: expected string or Unicode, list found)

### Expected behavior
In [8]: wimt.cmdline
Out[8]: u'WarpTimeSeriesImageMultiTransform 4 my_fixed_time_series.nii.gz my_fixed_time_series_warped.nii.gz -R my_moving_image.nii.gz -i synAffine.txt synInverseWarp.nii.gz'

just run WarpTimeSeriesImageMultiTransform with .inputs.invert_affine = [1]

### Script/Workflow details
l.89:                         series += ['-i'], -> series += '-i',

### Platform details:

{'nibabel_version': '2.1.0', 'sys_executable': '/usr/bin/python', 'networkx_version': '1.11', 'numpy_version': '1.11.1rc1', 'sys_platform': 'linux2', 'sys_version': '2.7.12+ (default, Sep 17 2016, 12:08:02) \n[GCC 6.2.0 20160914]', 'commit_source': u'repository', 'commit_hash': '3f8c8b1', 'pkg_path': '/mnt/drive/pilou/Code/github/src/nipype/nipype', 'nipype_version': u'1.0.0-dev+g3f8c8b1', 'traits_version': '4.6.0', 'scipy_version': '0.17.1'}
1.0.0-dev+g3f8c8b1

### Execution environment

Choose one
My python environment outside container

djarecka:
I added a set of simple tests, mostly based on doctest, but just testing a few different scenarios.
Have 2 xfail test for wrong (i.e. out of range) `invert_affine`. Should this raise exception?

### DIFF
--- a/nipype/interfaces/ants/tests/test_resampling.py
+++ b/nipype/interfaces/ants/tests/test_resampling.py
@@ -19,75 +19,67 @@ def change_dir(request):
    request.addfinalizer(move2orig)
 
 
-
-def test_WarpImageMultiTransform(change_dir):
+@pytest.fixture()
+def create_wimt():
    wimt = WarpImageMultiTransform()
    wimt.inputs.input_image = 'diffusion_weighted.nii'
    wimt.inputs.reference_image = 'functional.nii'
    wimt.inputs.transformation_series = ['func2anat_coreg_Affine.txt','func2anat_InverseWarp.nii.gz', \
                                            'dwi2anat_Warp.nii.gz','dwi2anat_coreg_Affine.txt']
+   return wimt
+
+def test_WarpImageMultiTransform(change_dir, create_wimt):
+   wimt = create_wimt
    assert wimt.cmdline == 'WarpImageMultiTransform 3 diffusion_weighted.nii diffusion_weighted_wimt.nii -R functional.nii \
 func2anat_coreg_Affine.txt func2anat_InverseWarp.nii.gz dwi2anat_Warp.nii.gz dwi2anat_coreg_Affine.txt'
 
 
-def test_WarpImageMultiTransform_invaffine_1(change_dir):
-   wimt = WarpImageMultiTransform() 
-   wimt.inputs.input_image = 'diffusion_weighted.nii'
-   wimt.inputs.reference_image = 'functional.nii'
-   wimt.inputs.transformation_series = ['func2anat_coreg_Affine.txt','func2anat_InverseWarp.nii.gz', \
-                                           'dwi2anat_Warp.nii.gz','dwi2anat_coreg_Affine.txt']  
+def test_WarpImageMultiTransform_invaffine_1(change_dir, create_wimt):
+   wimt = create_wimt
    wimt.inputs.invert_affine = [1]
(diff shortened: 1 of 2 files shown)


nipype/interfaces/ants/tests/test_resampling.py
```diff
@@ -0,0 +1,93 @@
...
+   wimt = WarpImageMultiTransform()
+   wimt.inputs.input_image = 'diffusion_weighted.nii'
+   wimt.inputs.reference_image = 'functional.nii'
+   wimt.inputs.transformation_series = ['func2anat_coreg_Affine.txt','func2anat_InverseWarp.nii.gz', \
+                                           'dwi2anat_Warp.nii.gz','dwi2anat_coreg_Affine.txt']
+   wimt.inputs.invert_affine = [2]
+   assert wimt.cmdline == 'WarpImageMultiTransform 3 diffusion_weighted.nii diffusion_weighted_wimt.nii -R functional.nii func2anat_coreg_Affine.txt func2anat_InverseWarp.nii.gz dwi2anat_Warp.nii.gz -i dwi2anat_coreg_Affine.txt'
+
+
+@pytest.mark.xfail(reason="dj: should it fail?")
```
satra: this should not fail.

satra: how does it fail?

djarecka: @satra - fails, i.e. it doesn't raise any exception.

I found it slightly confusing that i can provide wrong index in invert_affine, e.g. 0 or 3 (if you have only 2 affine files), and i don't get any errors. Was wondering if it would be useful to add an extra check.

satra: i think the issue is that the description is not very clear.

technically, looking at the code, each affine is given a index. seems like this can cause some mishaps. so yes, raising exceptions when the values of index exceed total number of affines is good.

djarecka: I think that description of `invert_affine` trait is quite good. However IMO it's very hard to guess without reading description or the code that affine files are counted, they are getting another indices, and the new indices start from 1. 

Unfortunately changing this concept would make old script fail, so was thinking that at least adding some error with meaningful message might save people some time in case of problems.

satra: that sounds good, but i think it may be good to add an example to describe this situation clearly in the docstring examples.

djarecka: I add a comment to an existing doctest and add a new one, please see if the new one really returns cmd that makes sense

PR #2236, opened 2017-10-18
Remove NaNs from data before components calculation

oesteban:
Closes poldracklab/fmriprep#776

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -1067,7 +1067,7 @@ def compute_noise_components(imgseries, mask_images, num_components,
 
         # "The covariance matrix C = MMT was constructed and decomposed into its
         # principal components using a singular value decomposition."
-        u, _, _ = linalg.svd(M, full_matrices=False)
+        u, _, _ = linalg.svd(M.dropna(inplace=True), full_matrices=False)
         if components is None:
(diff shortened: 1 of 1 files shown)


effigies:
Are you sure there are NaNs, and it's not just singular?

oesteban:
Not sure, but svd won't take NaNs anyways.

nipype/algorithms/confounds.py
```diff
@@ -1067,7 +1067,7 @@ def compute_noise_components(imgseries, mask_images, num_components,
 
         # "The covariance matrix C = MMT was constructed and decomposed into its
         # principal components using a singular value decomposition."
-        u, _, _ = linalg.svd(M, full_matrices=False)
+        u, _, _ = linalg.svd(M[~np.isnan(M)], full_matrices=False)
```
satra: i don't see how anything in `M` would be a `nan` given line 1047 and the `_compute_tSTD` function which should return a 1 if the variance is 0.

is there a particular use case we can look at where you found this?

oesteban: I'm going to look closer into this. You're right. And this change would reshape `M` so it basically breaks the interface.

oesteban:
Confirmed this problem is not within Nipype's realm

Issue #1916, opened 2017-03-30
Test Errors on a readonly file system


satra:
### Summary
When running nipype tests in a singularity container it fails because the tests try to create files in a read only file system.

### Actual behavior

```
============================================= FAILURES ==============================================
____________________________ [doctest] nipype.interfaces.dcm2nii.Dcm2nii ____________________________
069 
070     Examples
071     ========
072 
073     >>> from nipype.interfaces.dcm2nii import Dcm2nii
074     >>> converter = Dcm2nii()
075     >>> converter.inputs.source_names = ['functional_1.dcm', 'functional_2.dcm']
076     >>> converter.inputs.gzip_output = True
077     >>> converter.inputs.output_dir = '.'
078     >>> converter.cmdline # doctest: +ALLOW_UNICODE
UNEXPECTED EXCEPTION: OSError(30, 'Read-only file system')
Traceback (most recent call last):

  File "/opt/conda/lib/python3.5/doctest.py", line 1321, in __run
    compileflags, 1), test.globs)

  File "<doctest nipype.interfaces.dcm2nii.Dcm2nii[5]>", line 1, in <module>

  File "/opt/conda/lib/python3.5/site-packages/nipype/interfaces/base.py", line 1678, in cmdline
    allargs = self._parse_inputs()

  File "/opt/conda/lib/python3.5/site-packages/nipype/interfaces/base.py", line 1911, in _parse_inputs
    value = self._gen_filename(name)

  File "/opt/conda/lib/python3.5/site-packages/nipype/interfaces/dcm2nii.py", line 198, in _gen_filename
    with open(config_file, "w") as f:

OSError: [Errno 30] Read-only file system: 'config.ini'

/opt/conda/lib/python3.5/site-packages/nipype/interfaces/dcm2nii.py:78: UnexpectedException
_________________________________________ test_level1design _________________________________________

    def test_level1design():
        l = Level1Design()
        runinfo = dict(cond=[{'name': 'test_condition', 'onset': [0, 10],
                              'duration':[10, 10]}],regress=[])
        runidx = 0
        contrasts = Undefined
        do_tempfilter = False
        orthogonalization = {}
        basic_ev_parameters = {'temporalderiv':False}
        convolution_variants = [
            ('custom', 7, {'temporalderiv':False, 'bfcustompath':'/some/path'}),
            ('hrf', 3, basic_ev_parameters),
            ('dgamma', 3, basic_ev_parameters),
[...]

PR #2252, opened 2017-10-24
pytest cleaning

djarecka:
changing str(tmpdir) to tmpdir.strpath in tests (see the discussion in #1916)

**update**
using more of the tmpdir methods in tests
updating tests (mostly doctests) so the work on read-only system (see #1916)
adding setting for doctests in pytest.ini (and removing some options from doctests)
adding some names to doctests_namespace in conftest.py


nipype/algorithms/tests/test_confounds.py
```diff
@@ -43,7 +43,7 @@ def test_dvars(tmpdir):
                          in_mask=example_data('ds003_sub-01_mc_brainmask.nii.gz'),
                          save_all=True,
                          intensity_normalization=0)
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
satra: tmpdir is a `py.path.local` object which has a lot of methods:

http://py.readthedocs.io/en/latest/path.html#py-path-local-local-file-system-path

nipype/algorithms/tests/test_modelgen.py
```diff
@@ -17,7 +17,7 @@
 
 
 def test_modelgen1(tmpdir):
-    tempdir = str(tmpdir)
+    tempdir = tmpdir.strpath
     filename1 = os.path.join(tempdir, 'test1.nii')
```
satra: so this could be simply `tmpdir.join('test.nii')`

djarecka: ok, can try to find those places automatically

djarecka: but `py.path.local` won't work everywhere, I'm recalling that I did have some problems with it when I was converting tests to pytest, and there were places where string worked better.   I can see that that at least for `nibabel.save` I'll have to keep strings.

satra: i meant places where the methods of `py.path.local` make sense, such as join. it's not a direct replacement for a string in all cases.

djarecka: my only point was that this test will not work if I change join part to `tmpdir.join('test.nii')`, it has to be string. 
Can always do `tmpdir.join('test.nii').strpath` if this is preferred.

oesteban:
For the doctests maybe a `conftest.py` file at the top folder with the following:

https://github.com/poldracklab/niworkflows/blob/e77c3c103b76f6eddce94625ba6171c13248d7d5/niworkflows/tests/conftest.py#L12-L14

would work out. Worth trying.

djarecka:
@oesteban that might be the best solution. Was trying to find something for doctests only, but haven't found yet. 
I mean, you can always add 
```
>>> tmp = getfixture('tmpdir')                                                                                            
>>> old = tmp.chdir()
```
but would prefer keep the docstrings as short as possible.

btw.  should we use `conftest.py` to import modules like `numpy`, `nibabel`, so we don't have to include imports in doctests?

satra:
could you please merge with master and resolve conflicts.

djarecka:
@oesteban what do you think about this approach?
 https://stackoverflow.com/questions/46962007/how-to-automatically-change-to-pytest-temporary-directory-for-all-doctests/46991331#46991331

It uses `tmpdir` and can be done either for doctests only or all tests

djarecka:
There is one extra problem with changing working directory for all doctests (or all tests). Many files with doctests start with [this part](https://github.com/nipy/nipype/blob/master/nipype/interfaces/ants/registration.py#L4) and the following tests assume that they are in a directory with data and will fail if I change to a temporary directory.

In `conftest.py` I can define `datadir` and add to every single test full path, e.g. `ants.inputs.fixed_image = [os.path.join(datadir, 'T1.nii')]`, but this change will be needed in many places.

@oesteban, @satra let me know if you have any opinion about it. Not sure how important to you is to run everything in readonly systems and is it ok to add this boilerplate to doctests, that should be probably as simple as possible. 

Or maybe we want to add some extra code only in tests you're having problem with?

satra:
@djarecka - i agree that we should not be making the examples more complex. let's just focus on the unit-test cleanup first, and we can tackle the doctest issue later. 

we can perhaps change the doctests for the specific interfaces that write files during doctests. that's a much smaller list.

@oesteban - regarding read only systems, why are we thinking of running tests on these systems? can't we test before such a container is made read only?

djarecka:
@satra yes, you can add those line within specific doctests if you want:
```
>>> tmp = getfixture('tmpdir')                                                                                            
>>> old = tmp.chdir()
```

do you want me to add anything specific to this PR?

satra:
@djarecka - knowing which interfaces fail doctests for read only systems would be good. if it's a small list, we can fix those rather than everything.

mgxd:
I think I covered most of the culprits [here](https://github.com/nipy/nipype/issues/1916#issuecomment-337969975)

djarecka:
@mgxd ok, I will review those tests

djarecka:
@satra I'm also reviewing other tests, I remember that I left `tempfile` in a few places, since they offer methods that I couldn't find with `tmpdir`, e.g. `mkstepm` that creates new file with specific suffix and prefix. 
Let me know if the goal is totally remove `tempfile`, there is always a way to do it with `tmpdir` differently.

satra:
@djarecka - the goal is to be semantically uniform such that developers can rely on expectations and to use the tmpdir methods when possible.

djarecka:
@mgxd can you check if those tests are working now in the readonly setting you were using?

nipype/algorithms/tests/test_confounds.py
oesteban: true, you can do `tmpdir.chdir()` here :+1:

nipype/algorithms/tests/test_overlap.py
```diff
@@ -20,7 +20,7 @@ def check_close(val1, val2):
     in1 = example_data('segmentation0.nii.gz')
     in2 = example_data('segmentation1.nii.gz')
 
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()` ?

nipype/algorithms/tests/test_splitmerge.py
```diff
@@ -14,13 +14,13 @@ def test_split_and_merge(tmpdir):
...
     in_mask = example_data('tpms_msk.nii.gz')
-    dwfile = op.join(str(tmpdir), 'dwi.nii.gz')
+    dwfile = op.join(tmpdir.strpath, 'dwi.nii.gz')
     mskdata = nb.load(in_mask, mmap=NUMPY_MMAP).get_data()
     aff = nb.load(in_mask, mmap=NUMPY_MMAP).affine
 
     dwshape = (mskdata.shape[0], mskdata.shape[1], mskdata.shape[2], 6)
     dwdata = np.random.normal(size=dwshape)
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()` ?

nipype/interfaces/afni/utils.py
```diff
@@ -74,7 +74,7 @@ class ABoverlap(AFNICommand):
     >>> aboverlap.inputs.in_file_a = 'functional.nii'
     >>> aboverlap.inputs.in_file_b = 'structural.nii'
     >>> aboverlap.inputs.out_file =  'out.mask_ae_overlap.txt'
-    >>> aboverlap.cmdline  # doctest: +ALLOW_UNICODE
```
oesteban: Have you tried this on Python 2 and it works?

nipype/interfaces/dcm2nii.py
```diff
@@ -70,13 +70,16 @@ class Dcm2nii(CommandLine):
     Examples
     ========
 
+    >>> tmp = getfixture('tmpdir')
```
oesteban: Please consider the ``.. testsetup::`` directive to hide these two lines in the documentation.

http://www.sphinx-doc.org/en/stable/ext/doctest.html#directive-testsetup

nipype/interfaces/fsl/tests/test_model.py
```diff
@@ -13,7 +13,7 @@
 
 @pytest.mark.skipif(no_fsl(), reason="fsl is not installed")
 def test_MultipleRegressDesign(tmpdir):
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/interfaces/io.py
```diff
@@ -2476,19 +2476,20 @@ class JSONFileGrabber(IOBase):
 
     Example
     -------
-
+    >>> tmp = getfixture('tmpdir')
```
oesteban: `.. testsetup::` ?

nipype/interfaces/tests/test_base.py
```diff
@@ -70,7 +70,7 @@ def test_bunch_hash():
 
 @pytest.fixture(scope="module")
 def setup_file(request, tmpdir_factory):
-    tmp_dir = str(tmpdir_factory.mktemp('files'))
+    tmp_dir = tmpdir_factory.mktemp('files').strpath
```
oesteban: Would this work?

```Python
@pytest.fixture(scope="module")
def setup_file(tmpdir):
    tmpdir.chdir()
    tmp_file = tmpdir.join('foo.txt')
    tmp_file.write('123456789\n')
    tmp_file.close()
    return tmp_file
```

nipype/interfaces/utility/base.py
```diff
@@ -2,15 +2,9 @@
...
-Various utilities
-
-    Change directory to provide relative paths for doctests
-    >>> import os
-    >>> filepath = os.path.dirname(os.path.realpath(__file__))
-    >>> datadir = os.path.realpath(os.path.join(filepath,
-    ...                            '../../testing/data'))
-    >>> os.chdir(datadir)
-
+    # changing to temporary directories
```
oesteban: `.. testsetup::`

nipype/interfaces/utility/tests/test_wrappers.py
```diff
@@ -16,7 +16,7 @@ def concat_sort(in_arrays):
 """
 
 def test_function(tmpdir):
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/interfaces/utility/tests/test_wrappers.py
```diff
@@ -59,11 +59,11 @@ def should_fail(tmpdir):
 
 def test_should_fail(tmpdir):
     with pytest.raises(NameError):
-        should_fail(str(tmpdir))
+        should_fail(tmpdir.strpath)
 
 
 def test_function_with_imports(tmpdir):
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/interfaces/utility/tests/test_wrappers.py
```diff
@@ -79,7 +79,7 @@ def test_aux_connect_function(tmpdir):
     """ This tests excution nodes with multiple inputs and auxiliary
     function inside the Workflow connect function.
     """
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/interfaces/utility/wrappers.py
```diff
@@ -1,17 +1,12 @@
...
-    Change directory to provide relative paths for doctests
-    >>> import os
-    >>> filepath = os.path.dirname(os.path.realpath(__file__))
-    >>> datadir = os.path.realpath(os.path.join(filepath,
-    ...                            '../../testing/data'))
-    >>> os.chdir(datadir)
-
-
 """
+# changing to temporary directories
```
oesteban:
```rst
.. testsetup::
   
    # changing to temporary directories
    tmp = getfixture('tmpdir')
    old = tmp.chdir()

File: nipype/pipeline/engine/tests/test_join.py [hunk 15]
```diff
@@ -149,7 +149,7 @@ def _list_outputs(self):
 
 
 def test_join_expansion(tmpdir):
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -196,7 +196,7 @@ def test_join_expansion(tmpdir):
 
 def test_node_joinsource(tmpdir):
     """Test setting the joinsource to a Node."""
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -214,7 +214,7 @@ def test_node_joinsource(tmpdir):
 
 def test_set_join_node(tmpdir):
     """Test collecting join inputs to a set."""
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -240,7 +240,7 @@ def test_unique_join_node(tmpdir):
     """Test join with the ``unique`` flag set to True."""
     global _sum_operands
     _sum_operands = []
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -265,7 +265,7 @@ def test_multiple_join_nodes(tmpdir):
     """Test two join nodes, one downstream of the other."""
     global _products
     _products = []
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -318,7 +318,7 @@ def test_identity_join_node(tmpdir):
     """Test an IdentityInterface join."""
     global _sum_operands
     _sum_operands = []
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -353,7 +353,7 @@ def test_multifield_join_node(tmpdir):
     """Test join on several fields."""
     global _products
     _products = []
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -393,7 +393,7 @@ def test_synchronize_join_node(tmpdir):
     """Test join on an input node which has the ``synchronize`` flag set to True."""
     global _products
     _products = []
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -430,7 +430,7 @@ def test_synchronize_join_node(tmpdir):
 
 def test_itersource_join_source_node(tmpdir):
     """Test join on an input node which has an ``itersource``."""
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -484,7 +484,7 @@ def test_itersource_join_source_node(tmpdir):
 def test_itersource_two_join_nodes(tmpdir):
     """Test join with a midstream ``itersource`` and an upstream
     iterable."""
-    os.chdir(str(tmpdir))
+    os.chdir(tmpdir.strpath)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -524,7 +524,7 @@ def test_itersource_two_join_nodes(tmpdir):
 
 def test_set_join_node_file_input(tmpdir):
     """Test collecting join inputs to a set."""
-    wd = str(tmpdir)
+    wd = tmpdir.strpath
     os.chdir(wd)
```
oesteban: `tmpdir.chdir()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -524,7 +524,7 @@ def test_itersource_two_join_nodes(tmpdir):
...
     open('test.nii', 'w+').close()
```
oesteban: `tmpdir.join('test.nii').close()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -524,7 +524,7 @@ def test_itersource_two_join_nodes(tmpdir):
...
     open('test2.nii', 'w+').close()
```
oesteban: `tmpdir.join('test2.nii').close()`

nipype/pipeline/engine/tests/test_join.py
```diff
@@ -547,7 +547,7 @@ def test_set_join_node_file_input(tmpdir):
 
 def test_nested_workflow_join(tmpdir):
     """Test collecting join inputs within a nested workflow"""
-    wd = str(tmpdir)
+    wd = tmpdir.strpath
     os.chdir(wd)
```
oesteban: `tmpdir.chdir()`

pytest.ini
```diff
@@ -1,3 +1,4 @@
 [pytest]
 norecursedirs = .git build dist doc nipype/external tools examples src
-addopts = --doctest-modules
\ No newline at end of file
+addopts = --doctest-modules
+doctest_optionflags = ALLOW_UNICODE NORMALIZE_WHITESPACE
```
oesteban: Gotcha!

djarecka:
@oesteban thanks for comments. I'm still reviewing those test, and will try to implement all you points. At the beginning I change some parts automatically, but it didn't actually work as good, so will review chdir parts etc.

djarecka:
I'm reviewing tests, and tried to remove all remainders of `tempfile`. There are a few tests that were xfailing or skipping that I was afraid to change, one test that really had no idea why I'm getting errors (pipeline/plugins/tests/test_multiproc_nondaemon.py), and two tests that I still have questions:

- you will see one test in `nipype/interfaces/tests/test_io.py` that fails because of too long name, but I'm simply not sure if this is really right that `DataSink` creates so long path (see the error message)

- `nipype/interfaces/tests/test_nilearn.py` this has tests within a class and after introducing temporary file to `setup_class` `teardown_class` stopped working (doesn't see self.orig_dir). I simply commented right now, but if anyone understands it, I'd be thankful for letting me know.

nipype/interfaces/dcm2nii.py
djarecka: @oesteban : do you have an example how to use it? I'm adding `.. testsetup::` before those lines and my tests are working, but everything is still visible if you ask for `help()`

mgxd:
@djarecka Here's the error log I ran into (I pulled from commit 8baac42 so this may be outdated) - all i'm really doing is `chmod -R a-w .` within my editable nipype clone and then running `make test-code`

------
```
====================================================================== FAILURES =======================================================================
__________________________________________________ [doctest] nipype.interfaces.base.SimpleInterface ___________________________________________________
1238     ...     input_spec = DoubleInputSpec
1239     ...     output_spec = DoubleOutputSpec
1240     ...
1241     ...     def _run_interface(self, runtime):
1242     ...          self._results['doubled'] = double(self.inputs.x)
1243     ...          return runtime
1244 
1245     >>> dbl = Double()
1246     >>> dbl.inputs.x = 2
1247     >>> dbl.run().outputs.doubled
UNEXPECTED EXCEPTION: PermissionError(13, 'Permission denied')

  File "/home/mathias/miniconda2/envs/dev3/lib/python3.6/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)

  File "<doctest nipype.interfaces.base.SimpleInterface[6]>", line 1, in <module>

  File "/code/nipype/nipype/interfaces/base.py", line 1080, in run
    mon_sp = ResourceMonitor(proc_pid, freq=mon_freq)

  File "/code/nipype/nipype/utils/profiler.py", line 44, in __init__
    self._logfile = open(self._fname, 'w')

PermissionError: [Errno 13] Permission denied: '.proc-7600_time-1509464139.0193746_freq-1.00'

/code/nipype/nipype/interfaces/base.py:1247: UnexpectedException
__________________________________________________________________ test_importerror ___________________________________________________________________

creating_graphs = ['/tmp/pytest-of-mathias/pytest-2/test_importerror0name0.pck', '/tmp/pytest-of-mathias/pytest-2/test_importerror0name1...p/pytest-of-mathias/pytest-2/test_importerror0name4.pck', '/tmp/pytest-of-mathias/pytest-2/test_importerror0name5.pck']

    @pytest.mark.skipif(have_cv, reason="tests for import error, cviewer available")
    def test_importerror(creating_graphs):
        graphlist = creating_graphs
        group1 = graphlist[:3]
        group2 = graphlist[3:]
    
        nbs = NetworkBasedStatistic()
        nbs.inputs.in_group1 = group1
        nbs.inputs.in_group2 = group2
        nbs.inputs.edge_key = "weight"
    
        with pytest.raises(ImportError) as e:
>           nbs.run()

../../interfaces/cmtk/tests/test_nbs.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../interfaces/base.py:1080: in run
    mon_sp = ResourceMonitor(proc_pid, freq=mon_freq)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AssertionError("Thread.__init__() was not called") raised in repr()] ResourceMonitor object at 0x7f7febedd7b8>, pid = 7600, freq = 1.0
fname = '.proc-7600_time-1509464141.1586614_freq-1.00', python = True

    def __init__(self, pid, freq=5, fname=None, python=True):
        # Make sure psutil is imported
        import psutil
    
        if freq < 0.2:
            raise RuntimeError('Frequency (%0.2fs) cannot be lower than 0.2s' % freq)
    
        if fname is None:
            fname = '.proc-%d_time-%s_freq-%0.2f' % (pid, time(), freq)
        self._fname = fname
>       self._logfile = open(self._fname, 'w')
E       PermissionError: [Errno 13] Permission denied: '.proc-7600_time-1509464141.1586614_freq-1.00'

../../utils/profiler.py:44: PermissionError
_______________________________________________________________ test_coherence_analysis _______________________________________________________________

    @pytest.mark.skipif(no_nitime, reason="nitime is not installed")
    def test_coherence_analysis():
        """Test that the coherence analyzer works """
        import nitime.analysis as nta
        import nitime.timeseries as ts
    
        # This is the nipype interface analysis:
        CA = nitime.CoherenceAnalyzer()
        CA.inputs.TR = 1.89
        CA.inputs.in_file = example_data('fmri_timeseries.csv')
        if display_available:
            tmp_png = tempfile.mkstemp(suffix='.png')[1]
            CA.inputs.output_figure_file = tmp_png
        tmp_csv = tempfile.mkstemp(suffix='.csv')[1]
        CA.inputs.output_csv_file = tmp_csv
    
>       o = CA.run()

../../interfaces/nitime/tests/test_nitime.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../interfaces/base.py:1080: in run
    mon_sp = ResourceMonitor(proc_pid, freq=mon_freq)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AssertionError("Thread.__init__() was not called") raised in repr()] ResourceMonitor object at 0x7f7febcb32b0>, pid = 7600, freq = 1.0
fname = '.proc-7600_time-1509464143.6017606_freq-1.00', python = True

    def __init__(self, pid, freq=5, fname=None, python=True):
        # Make sure psutil is imported
        import psutil
    
        if freq < 0.2:
            raise RuntimeError('Frequency (%0.2fs) cannot be lower than 0.2s' % freq)
    
        if fname is None:
            fname = '.proc-%d_time-%s_freq-%0.2f' % (pid, time(), freq)
        self._fname = fname
>       self._logfile = open(self._fname, 'w')
E       PermissionError: [Errno 13] Permission denied: '.proc-7600_time-1509464143.6017606_freq-1.00'

../../utils/profiler.py:44: PermissionError
_________________________________________________________________ test_mapnode_crash3 _________________________________________________________________

self = <nipype.pipeline.plugins.linear.LinearPlugin object at 0x7f7fe991e240>, graph = <networkx.classes.digraph.DiGraph object at 0x7f7fe991e978>
config = {'check': {'interval': '1209600'}, 'execution': {'crashdump_dir': '/code/nipype', 'crashfile_format': 'pklz', 'create_...ogging': {'interface_level': 'INFO', 'log_directory': '/home/mathias', 'log_rotate': '4', 'log_size': '16384000', ...}}
updatehash = False

    def run(self, graph, config, updatehash=False):
        """Executes a pre-defined pipeline in a serial order.
    
            Parameters
            ----------
    
            graph : networkx digraph
                defines order of execution
            """
    
        if not isinstance(graph, nx.DiGraph):
            raise ValueError('Input must be a networkx digraph object')
        logger.info("Running serially.")
        old_wd = os.getcwd()
        notrun = []
        donotrun = []
        nodes, _ = topological_sort(graph)
        for node in nodes:
            try:
                if node in donotrun:
                    continue
                if self._status_callback:
                    self._status_callback(node, 'start')
>               node.run(updatehash=updatehash)

/code/nipype/nipype/pipeline/plugins/linear.py:43: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = testmapnodecrash.myfunc, updatehash = False

    def run(self, updatehash=False):
        """Execute the node in its directory.
    
            Parameters
            ----------
    
            updatehash: boolean
                Update the hash stored in the output directory
            """
        # check to see if output directory and hash exist
    
        if self.config is None:
            self.config = deepcopy(config._sections)
        else:
            self.config = merge_dict(deepcopy(config._sections), self.config)
        if not self._got_inputs:
            self._get_inputs()
            self._got_inputs = True
        outdir = self.output_dir()
        logger.info("Executing node %s in dir: %s", self.fullname, outdir)
        if op.exists(outdir):
            logger.debug('Output dir: %s', to_str(os.listdir(outdir)))
        hash_info = self.hash_exists(updatehash=updatehash)
        hash_exists, hashvalue, hashfile, hashed_inputs = hash_info
        logger.debug(
            'updatehash=%s, overwrite=%s, always_run=%s, hash_exists=%s',
            updatehash, self.overwrite, self._interface.always_run, hash_exists)
        if (not updatehash and (((self.overwrite is None and
                                  self._interface.always_run) or
                                 self.overwrite) or not
                                hash_exists)):
            logger.debug("Node hash: %s", hashvalue)
    
            # by rerunning we mean only nodes that did finish to run previously
            json_pat = op.join(outdir, '_0x*.json')
            json_unfinished_pat = op.join(outdir, '_0x*_unfinished.json')
            need_rerun = (op.exists(outdir) and not
                          isinstance(self, MapNode) and
                          len(glob(json_pat)) != 0 and
                          len(glob(json_unfinished_pat)) == 0)
            if need_rerun:
                logger.debug(
                    "Rerunning node:\n"
                    "updatehash = %s, self.overwrite = %s, self._interface.always_run = %s, "
                    "os.path.exists(%s) = %s, hash_method = %s", updatehash, self.overwrite,
                    self._interface.always_run, hashfile, op.exists(hashfile),
                    self.config['execution']['hash_method'].lower())
                log_debug = config.get('logging', 'workflow_level') == 'DEBUG'
                if log_debug and not op.exists(hashfile):
                    exp_hash_paths = glob(json_pat)
                    if len(exp_hash_paths) == 1:
                        split_out = split_filename(exp_hash_paths[0])
                        exp_hash_file_base = split_out[1]
                        exp_hash = exp_hash_file_base[len('_0x'):]
                        logger.debug("Previous node hash = %s", exp_hash)
                        try:
                            prev_inputs = load_json(exp_hash_paths[0])
                        except:
                            pass
                        else:
                            logging.logdebug_dict_differences(prev_inputs,
                                                              hashed_inputs)
                cannot_rerun = (str2bool(
                    self.config['execution']['stop_on_first_rerun']) and not
                    (self.overwrite is None and self._interface.always_run))
                if cannot_rerun:
                    raise Exception(("Cannot rerun when 'stop_on_first_rerun' "
                                     "is set to True"))
            hashfile_unfinished = op.join(outdir,
                                          '_0x%s_unfinished.json' %
                                          hashvalue)
            if op.exists(hashfile):
                os.remove(hashfile)
            rm_outdir = (op.exists(outdir) and not
                         (op.exists(hashfile_unfinished) and
                             self._interface.can_resume) and not
                         isinstance(self, MapNode))
            if rm_outdir:
                logger.debug("Removing old %s and its contents", outdir)
                try:
                    rmtree(outdir)
                except OSError as ex:
                    outdircont = os.listdir(outdir)
                    if ((ex.errno == errno.ENOTEMPTY) and (len(outdircont) == 0)):
                        logger.warn(
                            'An exception was raised trying to remove old %s, but the path '
                            'seems empty. Is it an NFS mount?. Passing the exception.', outdir)
                    elif ((ex.errno == errno.ENOTEMPTY) and (len(outdircont) != 0)):
                        logger.debug(
                            'Folder contents (%d items): %s', len(outdircont), outdircont)
                        raise ex
                    else:
                        raise ex
    
            else:
                logger.debug(
                    "%s found and can_resume is True or Node is a MapNode - resuming execution",
                    hashfile_unfinished)
                if isinstance(self, MapNode):
                    # remove old json files
                    for filename in glob(op.join(outdir, '_0x*.json')):
                        os.unlink(filename)
            outdir = make_output_dir(outdir)
            self._save_hashfile(hashfile_unfinished, hashed_inputs)
            self.write_report(report_type='preexec', cwd=outdir)
            savepkl(op.join(outdir, '_node.pklz'), self)
            savepkl(op.join(outdir, '_inputs.pklz'),
                    self.inputs.get_traitsfree())
            try:
>               self._run_interface()

/code/nipype/nipype/pipeline/engine/nodes.py:407: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = testmapnodecrash.myfunc, execute = True, updatehash = False

    def _run_interface(self, execute=True, updatehash=False):
        """Run the mapnode interface
    
            This is primarily intended for serial execution of mapnode. A parallel
            execution requires creation of new nodes that can be spawned
            """
        old_cwd = os.getcwd()
        cwd = self.output_dir()
        os.chdir(cwd)
        self._check_iterfield()
        if execute:
            if self.nested:
                nitems = len(filename_to_list(flatten(getattr(self.inputs,
                                                              self.iterfield[0]))))
            else:
                nitems = len(filename_to_list(getattr(self.inputs,
                                                      self.iterfield[0])))
            nodenames = ['_' + self.name + str(i) for i in range(nitems)]
            self._collate_results(self._node_runner(self._make_nodes(cwd),
>                                                   updatehash=updatehash))

/code/nipype/nipype/pipeline/engine/nodes.py:1326: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = testmapnodecrash.myfunc, nodes = <generator object MapNode._node_runner at 0x7f7fe987a258>

    def _collate_results(self, nodes):
        self._result = InterfaceResult(interface=[], runtime=[],
                                       provenance=[], inputs=[],
                                       outputs=self.outputs)
        returncode = []
        for i, node, err in nodes:
            self._result.runtime.insert(i, None)
            if node.result:
                if hasattr(node.result, 'runtime'):
                    self._result.interface.insert(i, node.result.interface)
                    self._result.inputs.insert(i, node.result.inputs)
                    self._result.runtime[i] = node.result.runtime
                if hasattr(node.result, 'provenance'):
                    self._result.provenance.insert(i, node.result.provenance)
            returncode.insert(i, err)
            if self.outputs:
                for key, _ in list(self.outputs.items()):
                    rm_extra = (self.config['execution']
                                ['remove_unnecessary_outputs'])
                    if str2bool(rm_extra) and self.needed_outputs:
                        if key not in self.needed_outputs:
                            continue
                    values = getattr(self._result.outputs, key)
                    if not isdefined(values):
                        values = []
                    if node.result.outputs:
                        values.insert(i, node.result.outputs.get()[key])
                    else:
                        values.insert(i, None)
                    defined_vals = [isdefined(val) for val in values]
                    if any(defined_vals) and self._result.outputs:
                        setattr(self._result.outputs, key, values)
    
        if self.nested:
            for key, _ in list(self.outputs.items()):
                values = getattr(self._result.outputs, key)
                if isdefined(values):
                    values = unflatten(values, filename_to_list(getattr(self.inputs, self.iterfield[0])))
                setattr(self._result.outputs, key, values)
    
        if returncode and any([code is not None for code in returncode]):
            msg = []
            for i, code in enumerate(returncode):
                if code is not None:
                    msg += ['Subnode %d failed' % i]
                    msg += ['Error:', str(code)]
            raise Exception('Subnodes of node: %s failed:\n%s' %
>                           (self.name, '\n'.join(msg)))
E           Exception: Subnodes of node: myfunc failed:
E           Subnode 0 failed
E           Error:
E           dummy_func() got an unexpected keyword argument 'WRONG'
E           Subnode 1 failed
E           Error:
E           dummy_func() got an unexpected keyword argument 'WRONG'
E           Subnode 2 failed
E           Error:
E           dummy_func() got an unexpected keyword argument 'WRONG'

/code/nipype/nipype/pipeline/engine/nodes.py:1232: Exception

During handling of the above exception, another exception occurred:

tmpdir = local('/tmp/pytest-of-mathias/pytest-2/test_mapnode_crash30')

    @pytest.mark.skipif(sys.version_info < (3,0),
                       reason="the famous segfault #1788")
    def test_mapnode_crash3(tmpdir):
        """Test mapnode crash when mapnode is embedded in a workflow"""
        node = pe.MapNode(niu.Function(input_names=['WRONG'],
                                       output_names=['newstring'],
                                       function=dummy_func),
                          iterfield=['WRONG'],
                          name='myfunc')
        node.inputs.WRONG = ['string{}'.format(i) for i in range(3)]
        wf = pe.Workflow('testmapnodecrash')
        wf.add_nodes([node])
        wf.base_dir = tmpdir.strpath
        with pytest.raises(RuntimeError):
>           wf.run(plugin='Linear')

/code/nipype/nipype/pipeline/engine/tests/test_utils.py:399: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/code/nipype/nipype/pipeline/engine/workflows.py:591: in run
/code/nipype/nipype/pipeline/plugins/linear.py:52: in run
    crashfile = report_crash(node)
/code/nipype/nipype/pipeline/plugins/tools.py:58: in report_crash
    savepkl(crashfile, dict(node=node, traceback=traceback))
/code/nipype/nipype/utils/filemanip.py:623: in savepkl
    pkl_file = gzip.open(filename, 'wb')
/home/mathias/miniconda2/envs/dev3/lib/python3.6/gzip.py:53: in open
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AttributeError("'GzipFile' object has no attribute 'fileobj'") raised in repr()] GzipFile object at 0x7f7fe991fe10>
filename = '/code/nipype/crash-20171031-113736-mathias-myfunc-3723030d-f1e8-4a4a-8d5c-e72fefa81e12.pklz', mode = 'wb', compresslevel = 9
fileobj = None, mtime = None

    def __init__(self, filename=None, mode=None,
                 compresslevel=9, fileobj=None, mtime=None):
        """Constructor for the GzipFile class.
    
            At least one of fileobj and filename must be given a
            non-trivial value.
    
            The new class instance is based on fileobj, which can be a regular
            file, an io.BytesIO object, or any other object which simulates a file.
            It defaults to None, in which case filename is opened to provide
            a file object.
    
            When fileobj is not None, the filename argument is only used to be
            included in the gzip file header, which may include the original
            filename of the uncompressed file.  It defaults to the filename of
            fileobj, if discernible; otherwise, it defaults to the empty string,
            and in this case the original filename is not included in the header.
    
            The mode argument can be any of 'r', 'rb', 'a', 'ab', 'w', 'wb', 'x', or
            'xb' depending on whether the file will be read or written.  The default
            is the mode of fileobj if discernible; otherwise, the default is 'rb'.
            A mode of 'r' is equivalent to one of 'rb', and similarly for 'w' and
            'wb', 'a' and 'ab', and 'x' and 'xb'.
    
            The compresslevel argument is an integer from 0 to 9 controlling the
            level of compression; 1 is fastest and produces the least compression,
            and 9 is slowest and produces the most compression. 0 is no compression
            at all. The default is 9.
    
            The mtime argument is an optional numeric timestamp to be written
            to the last modification time field in the stream when compressing.
            If omitted or None, the current time is used.
    
            """
    
        if mode and ('t' in mode or 'U' in mode):
            raise ValueError("Invalid mode: {!r}".format(mode))
        if mode and 'b' not in mode:
            mode += 'b'
        if fileobj is None:
>           fileobj = self.myfileobj = builtins.open(filename, mode or 'rb')
E           PermissionError: [Errno 13] Permission denied: '/code/nipype/crash-20171031-113736-mathias-myfunc-3723030d-f1e8-4a4a-8d5c-e72fefa81e12.pklz'

/home/mathias/miniconda2/envs/dev3/lib/python3.6/gzip.py:163: PermissionError
---------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------
171031-11:37:35,900 workflow INFO:
	 Workflow testmapnodecrash settings: ['check', 'execution', 'logging']
171031-11:37:35,901 workflow INFO:
	 Running serially.
171031-11:37:35,902 workflow INFO:
	 Executing node testmapnodecrash.myfunc in dir: /tmp/pytest-of-mathias/pytest-2/test_mapnode_crash30/testmapnodecrash/myfunc
171031-11:37:35,904 workflow INFO:
	 Executing node _myfunc0 in dir: /tmp/pytest-of-mathias/pytest-2/test_mapnode_crash30/testmapnodecrash/myfunc/mapflow/_myfunc0
171031-11:37:35,905 workflow INFO:
	 Running node "_myfunc0" ("nipype.interfaces.utility.wrappers.Function").
171031-11:37:36,183 workflow INFO:
	 Executing node _myfunc1 in dir: /tmp/pytest-of-mathias/pytest-2/test_mapnode_crash30/testmapnodecrash/myfunc/mapflow/_myfunc1
171031-11:37:36,185 workflow INFO:
	 Running node "_myfunc1" ("nipype.interfaces.utility.wrappers.Function").
171031-11:37:36,458 workflow INFO:
	 Executing node _myfunc2 in dir: /tmp/pytest-of-mathias/pytest-2/test_mapnode_crash30/testmapnodecrash/myfunc/mapflow/_myfunc2
171031-11:37:36,459 workflow INFO:
	 Running node "_myfunc2" ("nipype.interfaces.utility.wrappers.Function").
171031-11:37:36,735 workflow ERROR:
	 Node myfunc failed to run on host mathias-N501VW.
171031-11:37:36,735 workflow ERROR:
	 Saving crash info to /code/nipype/crash-20171031-113736-mathias-myfunc-3723030d-f1e8-4a4a-8d5c-e72fefa81e12.pklz
  File "/code/nipype/nipype/pipeline/plugins/linear.py", line 43, in run
  File "/code/nipype/nipype/pipeline/engine/nodes.py", line 407, in run
    self._run_interface()
  File "/code/nipype/nipype/pipeline/engine/nodes.py", line 1326, in _run_interface
    updatehash=updatehash))
  File "/code/nipype/nipype/pipeline/engine/nodes.py", line 1232, in _collate_results
    (self.name, '\n'.join(msg)))
Exception: Subnodes of node: myfunc failed:
Subnode 0 failed
Error:
dummy_func() got an unexpected keyword argument 'WRONG'
Subnode 1 failed
Error:
dummy_func() got an unexpected keyword argument 'WRONG'
Subnode 2 failed
Error:
dummy_func() got an unexpected keyword argument 'WRONG'

=============================================================== pytest-warning summary ================================================================
WI9 None could not create cache path /code/nipype/.cache/v/cache/lastfailed
```

djarecka:
@mgxd : those errors are from tests that you not listed previously, it might mean that the fix works, but more tests require changes

djarecka:
@mgxd but thanks for checking! I'll test it on my own, makes sense to just change the permission.

djarecka:
I have still problem with one test in read-only systems, it's a well-known `nipype/pipeline/engine/tests/test_utils.py::test_mapnode_crash3`. I believe this is because nipype tries to write crashfile in the nipype directory. Is this really ok, that the default directory for crashfile is not cwl?

satra:
@djarecka - the location of where crash files are dumped can be changed via config.

djarecka:
@satra I noticed that I can change ` wf.config["execution"]["crashdump_dir"]`, can change the directory in this one test. Was just wondering if this is expected behavior

satra:
> @satra I noticed that I can change wf.config["execution"]["crashdump_dir"], can change the directory in this one test. Was just wondering if this is expected behavior

yes

djarecka:
@satra  ok, i change this one test. everything should work now in read-only systems

djarecka:
i mean, except the one test that it's failing regardless of permissions... will debug it  again today

nipype/interfaces/base.py
```diff
@@ -123,10 +123,10 @@ class Bunch(object):
     --------
     >>> from nipype.interfaces.base import Bunch
     >>> inputs = Bunch(infile='subj.nii', fwhm=6.0, register_to_mean=True)
-    >>> inputs # doctest: +ALLOW_UNICODE
+    >>> inputs
```
mgxd: why remove unicode?

```
Expected:
    Bunch(fwhm=6.0, infile='subj.nii', register_to_mean=True)
Got:
    Bunch(fwhm=6.0, infile=u'subj.nii', register_to_mean=True)
```

nipype/interfaces/base.py
```diff
@@ -1505,18 +1518,18 @@ class must be instantiated with a command argument
     >>> from nipype.interfaces.base import CommandLine
     >>> cli = CommandLine(command='ls', environ={'DISPLAY': ':1'})
     >>> cli.inputs.args = '-al'
-    >>> cli.cmdline # doctest: +ALLOW_UNICODE
+    >>> cli.cmdline
     'ls -al'
 
     # Use get_traitsfree() to check all inputs set
-    >>> pprint.pprint(cli.inputs.get_traitsfree())  # doctest: +NORMALIZE_WHITESPACE +ALLOW_UNICODE
```
mgxd: we should add these tags back in?
```
Expected:
    {'args': '-al',
     'environ': {'DISPLAY': ':1'},
     'ignore_exception': False}
Got:
    {'args': '-al', 'environ': {'DISPLAY': ':1'}, 'ignore_exception': False}
```

nipype/interfaces/base.py
djarecka: I set this in `pytest.ini`: https://github.com/nipy/nipype/blob/master/pytest.ini#L4

is it failing?

nipype/interfaces/base.py
djarecka: the same here

nipype/interfaces/base.py
djarecka: You'd have to show me where it's failing. It works on my py2 and all tests for PR were also passing

nipype/interfaces/base.py
oesteban: Seems like `+ALLOW_UNICODE` did work right?

djarecka: @oesteban @mgxd - so the way how Mathias is testing than I always do:
```
conda create -n testnipype python=2 pip
source activate testnipype
conda install -y nipype
pip install -U https://github.com/nipy/nipype/archive/master.zip
python -c "import nipype; nipype.test()"
```
Indeed, if you test like this, pytest is testing nipype installed in the environment, i.e. `....../envs/testnipype/lib/python2.7/site-packages/nipype/interfaces/base.py` and there is no `pytest.ini` to collect.

djarecka: I'm just wondering if the `pytest.ini` shouldn't be in`nipype/nipype`. any thoughts?

djarecka:
@oesteban , @mgxd : talked to @satra and will just modify `nipype.test()`

PR #2274, opened 2017-11-10
Memoize version check

### LINKED ISSUE #2273: ANTS, AFNI always check version before execution
### Summary
Interfaces checking versions in a custom `Info` or `Base` class need to run the version request command line every time a derived interface is run.

### Expected behavior
Version is only checked once, the first time one of these interfaces is run. It is cached after that.

### Script/Workflow details
Any workflow. This is becoming a problem in fmriprep though.

effigies:
Fixes #2273.

- Create `PackageInfo` template that reads version information from a file or command output, passes to a subclass-defined parser, and stores the result for future calls
- Example subclasses for AFNI, ANTs and FreeSurfer

### DIFF
--- a/nipype/interfaces/afni/base.py
+++ b/nipype/interfaces/afni/base.py
@@ -14,45 +14,26 @@ from ... import logging, LooseVersion
 from ...utils.filemanip import split_filename, fname_presuffix
 
 from ..base import (
-    CommandLine, traits, CommandLineInputSpec, isdefined, File, TraitedSpec)
+    CommandLine, traits, CommandLineInputSpec, isdefined, File, TraitedSpec,
+    PackageInfo)
 from ...external.due import BibTeX
 
 # Use nipype's logging system
 IFLOGGER = logging.getLogger('interface')
 
 
-class Info(object):
+class Info(PackageInfo):
     """Handle afni output type and version information.
     """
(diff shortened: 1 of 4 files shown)


satra:
@effigies - should we use memoize here, or just call the version check again? in the unlikely, but possible scenario that someone changes a software on the system.

effigies:
The specific purpose is to memoize so that when calling an interface many times that runs a version check under the hood, we're not spawning an additional process each time.

Under what circumstances would you want to actively support changing versions during a run, and is it worth accounting for that situation by default? Most of the effects are going to be for input spec versioning, which will happen at workflow construction time. If you swap in an incompatible version after that, it should break when it's time to run. But significantly changing the environment is one of those cases where I feel like breaking is both expected and probably a good alternative to silently pressing on.

The one case I can see is if you're working in a Python console, and upgrade software, and for whatever reason you really don't want to start a new console. With this change, you could do `interfaces.ants.base.Info._version = None`, but we could also add a method or function to clear the cache.

satra:
@effigies - i'll merge this in a second. the situation really deals with environment where people have long running jobs, but underlying software can get updated without people knowing. while this is not common, as i mentioned earlier, i think with versioning, just like file hashing we cannot rely on the external environment to remain constant.

Issue #1242, opened 2015-10-09
Incorrect use of logging module causing poor performance


jamescasbon:
This code is in engine... 

        logger.debug('setting nodelevel(%s) input %s = %s' % (str(self),
                                                              parameter,
                                                              str(val)))

The `str` calls here are not needed.   The logger should call str *if needed*.  By doing this, you pay the cost of string representing inputs even when debug logging is not activated.  For very large objects, this can be significant.

oesteban:
Indeed, we should specify parameters as logger arguments instead of using formatting (http://docs.pylint.org/features.html, W1201)

PR #2276, opened 2017-11-12
[REF] Fix use of logging module in utils, algorithms and interfaces.

tsalo:
Partially addresses #1242.

- Switch from string formatting to using string format arguments as logging function parameters.

I went through `nipype.utils`, `nipype.algorithms`, and `nipype.interfaces`, fixing the logging calls where I found them.


satra:
@tsalo - there is one conflict due to a merge? could you please resolve and update this PR?

PR #2278, opened 2017-11-12
ENH: Enable users to add a run lock to a node

effigies:
This PR proposes a `run_lock` that can be added to nodes at instantiation. The lock is an object that can be called with a filename (the working directory for the node) and return a context manager.

The targeted problem is when the same node needs to be run in a shared working directory in multiple processes (possibly across multiple nodes). If one process has exclusive access to the working directory (e.g. by running first), then all other processes will reuse that directory, while simultaneous runs will result in a race condition.

In this formulation, it is the responsibility of the workflow programmer to provide (possibly passed to the end user to specify) a suitable locking mechanism for their needs. In all cases, it should have the property that no blocking occurs if the lock is called with different paths.

Related: poldracklab/fmriprep#821, [neurostars #901](https://neurostars.org/t/fmriprep-errors-and-error-recovery/901)

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -56,6 +56,15 @@ from .base import EngineBase
 logger = logging.getLogger('workflow')
 
 
+class _null_lock(object):
+    def __init__(self, path):
+        pass
+    def __enter__(self):
+        pass
+    def __exit__(self, *args, **kwargs):
+        pass
+
+
 class Node(EngineBase):
     """Wraps interface objects for use in pipeline
 
@@ -80,7 +89,7 @@ class Node(EngineBase):
     def __init__(self, interface, name, iterables=None, itersource=None,
                  synchronize=False, overwrite=None, needed_outputs=None,
                  run_without_submitting=False, n_procs=None, mem_gb=0.20,
-                 **kwargs):
+                 run_lock=_null_lock, **kwargs):
         """
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -56,6 +56,15 @@
 logger = logging.getLogger('workflow')
 
 
+class _null_lock(object):
```
oesteban: Wouldn't it be nice to have this stub in `nipype.utils.filemanip`?

effigies: Except it doesn't do anything. It's certainly not manipulating files. I can put it somewhere else, but as a hidden variable used once, it's not obvious to me that there's a better spot than the file it's used in.

effigies:
> Wouldn't it be clearer if interfaces locked their working directory themselves?.

Without a specific proposal, it's hard to say, but I suspect not. Re-running interfaces is a feature of `Node` objects, so by the time you get to interface-specific code, you may have already made the decision of whether the node needs to be re-run.

Further, the type of locking you do is highly context-dependent. Within-process and between-(cluster-)node are more or less suited to different locks.

> I think it'd be great that, if you run an interface in an IPython terminal, and try to use that as base_dir for another interface you got an error saying it is locked. DYT?

Wouldn't this just make resuming workflows impossible? I'm not sure what model you have in mind, though. Feel free to elaborate.

effigies:
@satra @chrisfilo I wrote this to be inert by default, but as this is pretty close to core nipype functionality, I don't want to go forward on this proposal without both of your input.

nipype/pipeline/engine/nodes.py
```diff
@@ -80,7 +89,7 @@ class Node(EngineBase):
     def __init__(self, interface, name, iterables=None, itersource=None,
                  synchronize=False, overwrite=None, needed_outputs=None,
                  run_without_submitting=False, n_procs=None, mem_gb=0.20,
-                 **kwargs):
+                 run_lock=_null_lock, **kwargs):
```
chrisfilo: Could you add some documentation how to use this parameter? Possibly with examples?

effigies:

PR #2279, opened 2017-11-13
ENH: Enable recon-all -FLAIR / -FLAIRpial options

effigies:
The invocation of the pial surface correction differs based on whether FreeSurfer thinks it's using a T2 or a FLAIR image. We should thus support both pathways.

- Adds FLAIR_file input to ReconAll
- Adds use_FLAIR (mutually exclusive with use_T2) input to ReconAll


nipype/interfaces/freesurfer/preprocess.py
```diff
@@ -628,10 +628,17 @@ class ReconAllInputSpec(CommandLineInputSpec):
                        argstr="-hemi %s")
     T1_files = InputMultiPath(File(exists=True), argstr='-i %s...',
                               desc='name of T1 file to process')
-    T2_file = File(exists=True, argstr="-T2 %s", min_ver='5.3.0',
+    T2_file = File(exists=True, argstr="-T2 %s",
+                   min_ver='5.3.0',
                    desc='Convert T2 image to orig directory')
-    use_T2 = traits.Bool(argstr="-T2pial", min_ver='5.3.0',
-                         desc='Use converted T2 to refine the cortical surface')
+    FLAIR_file = File(exists=True, argstr="-FLAIR %s",
```
satra: can we use a single input i.e. T2, perhaps with:
1.  a `requires` set for `use*` 
2. a description update saying T2 file can be T2 or FLAIR
?

freesurfer doesn't really distinguish between the inputs in the recon-all process, only the use options change some thresholds to fixing surfaces.

effigies: When adding the images, `-T2 <file>` vs `-FLAIR <file>` dictates whether it's converted to `orig/T2raw.mgz` or `orig/FLAIRraw.mgz` Then `-T2pial` and `-FLAIRpial` draw from those files as well as set different options to `mris_make_surfaces`. (Also note that `-T2` and `-FLAIR` are not mutually exclusive, but `-T2pial` and `-FLAIRpial` are.)

Also, recall that someone may use this in multiple stages. The `-T2 <file>` and `-T2pial` options may not be passed in the same interface, so dependencies between these options aren't necessarily going to be straightforward.

Given these considerations, do you see a way of simplifying this?

satra: @effigies - i was going to say we could set the appropriate argument in `_format_arg`. but since this interface offers the possibility of passing any of the `Do*` arguments, perhaps this is cleaner at this point. so i'm ok with the way this PR is currently implemented.

PR #2284, opened 2017-11-16
[ENH] Set maxtasksperchild in MultiProc Pool

oesteban:
To avoid the workers to grow too big. Offer one new plugin_args to modify this option.

Now testing if this alleviates in any ways the memory issues we are experiencing in poldracklab/fmriprep#833.

### DIFF
--- a/doc/users/plugins.rst
+++ b/doc/users/plugins.rst
@@ -82,6 +82,10 @@ Optional arguments::
   exceed the total amount of resources available (memory and threads), when
   ``False`` (default), only a warning will be issued.
 
+  maxtasksperchild : refresh the workers after this specific number of nodes
+  run (default: 10).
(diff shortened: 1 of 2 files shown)


doc/users/plugins.rst
```diff
@@ -82,6 +82,10 @@ Optional arguments::
   exceed the total amount of resources available (memory and threads), when
   ``False`` (default), only a warning will be issued.
 
+  maxtasksperchild : refresh the workers after this specific number of nodes
+  run (default: 10).
```
effigies: Actually, this could be clearer. Perhaps: "number of nodes to run on each worker before refreshing (default: 10)."

oesteban:
This shouldn't be harmful. Merging...

PR #2285, opened 2017-11-16
[REF,FIX] Revision of the resource monitor

oesteban:
- [x] Allow to reuse a resource monitor file (e.g. when we reuse a partially cached work directory), with accompanying new nipype option.
- [x] Add a new `[monitoring]` section to the Nipype config file, simplified option names and updated option deprecation logic to allow for new sections. 
- [x] Allow for changing the default summary file with all the profiling traces.


oesteban:
Tests passing https://circleci.com/gh/oesteban/nipype/714

mgxd:
Are we trying to get this in for `0.14`? If so, ping @satra for review

oesteban:
Yes, this would be nice to have in 0.14

oesteban:
@satra, this would be the last thing before releasing 0.14.0 - isn't it?

PR #2289, opened 2017-11-17
[ENH] Revising use of subprocess.Popen

oesteban:
This PR:

- [x] Revises the use of `Popen` in `CommandLineInterface` ensuring they are terminated and garbage-collected.
- [x] Tidy up a bit the code:
  - [x] Removes unused code for getting the version of the command (`interfaces.base`)
  - [x] Make sure the ordering of imports is okay (`interfaces.base`)
  - [x] Comment out historical hacks that potentially can be removed (`interfaces.base`)
  - [x] Minor modifications in `engine.workflows` and `plugins.multiproc`.
- [x] Simplifies `get_dependencies` and executes it only iff the option `execution. get_linked_libs` is `True` (which is the default value).


nipype/interfaces/base.py
```diff
@@ -1619,21 +1629,6 @@ def raise_exception(self, runtime):
...
-            out_environ = self._get_environ()
-            env.update(out_environ)
-            proc = sp.Popen(' '.join((cmdname, flag)),
-                            shell=True,
-                            env=env,
-                            stdout=sp.PIPE,
-                            stderr=sp.PIPE,
-                            )
-            o, e = proc.communicate()
-            return o
```
effigies: This is an API change. Even if not used internally, someone may depend on this method. Does nipype have a process for deprecating?

nipype/interfaces/base.py
```diff
@@ -1452,12 +1448,18 @@ def _process(drain=0):
...
     else:
-        proc.communicate()  # Discard stdout and stderr
+        stdout, stderr = proc.communicate()
+        if output == 'allatonce':  # Discard stdout and stderr otherwise
+            result['stdout'] = read_stream(stdout, logger=iflogger)
+            result['stderr'] = read_stream(stderr, logger=iflogger)
+
+    runtime.returncode = proc.returncode
+    proc.terminate()  # Ensure we are done
+    gc.collect()  # Force GC for a cleanup
```
effigies: Goal here is just to clean up `proc` internals, not `stdout`/`stderr`, etc?

nipype/interfaces/base.py
```diff
@@ -1467,21 +1469,26 @@ def get_dependencies(name, environ):
...
+
+    if cmd is None:
         return 'Platform %s not supported' % sys.platform
-    o, e = proc.communicate()
+
+    try:
+        proc = sp.Popen(
+            cmd(name), stdout=sp.PIPE, stderr=sp.PIPE, shell=True,
+            env=environ, close_fds=True)
+        o, e = proc.communicate()
```
effigies: I'd replace `e` with `_`, since it's unused.

nipype/interfaces/base.py
oesteban: Maybe we should start using https://pypi.python.org/pypi/deprecation ?

nipype/interfaces/base.py
```diff
@@ -1452,12 +1448,28 @@ def _process(drain=0):
             result['merged'] = result['stdout']
             result['stdout'] = []
     else:
-        proc.communicate()  # Discard stdout and stderr
+        stdout, stderr = proc.communicate()
```
effigies: Maybe make these `stdoutstr`, `stdinstr`, to be consistent with `if output.startswith('file')`, so that the same `del` statement will work in all cases.

Alternately, it might be simplest to just run `gc.collect()` in `CommandLine._run_interface`, after this function is called and everything is allowed to go out of scope.

nipype/interfaces/base.py
```diff
@@ -1452,12 +1448,30 @@ def _process(drain=0):
             result['merged'] = result['stdout']
             result['stdout'] = []
     else:
-        proc.communicate()  # Discard stdout and stderr
+        stdoutstr, stderrstr = proc.communicate()
+        if output == 'allatonce':  # Discard stdout and stderr otherwise
+            result['stdout'] = read_stream(stdoutstr, logger=iflogger)
+            result['stderr'] = read_stream(stderrstr, logger=iflogger)
+        del stdoutstr
+        del stderrstr
```
effigies: Sorry, meant for these to replace the `del stdout` and `del stderr` below, since `stdoutstr` and `stderrstr` can be created in the other branch.

I guess, though, that those only exist under certain sub-branches. Not sure how you want to go about this.

nipype/interfaces/base.py
```diff
@@ -634,16 +635,16 @@ def __deepcopy__(self, memo):
...
+        #     if key in self.__dict__.keys():
+        #         _ = getattr(self, key)
         # clone once
         dup = self.clone_traits(memo=memo)
-        for key in self.copyable_trait_names():
-            try:
-                _ = getattr(dup, key)
-            except:
-                pass
+        # for key in self.copyable_trait_names():
```
satra: dynamic traits won't work without these. i'm not sure we have a test for it but this is the relevant issue and PR.

https://github.com/enthought/traits/pull/373

nipype/interfaces/base.py
```diff
@@ -1620,6 +1642,8 @@ def _get_environ(self):
         return getattr(self.inputs, 'environ', {})
 
     def version_from_command(self, flag='-v'):
+        iflogger.warning('version_from_command member of CommandLine was '
+                         'Deprecated in nipype-1.0.0 and deleted in 2.0.0')
```
satra: we should perhaps keep this in the 1.0 branch so deleted in 1.1

PR #2301, opened 2017-11-22
[FIX] MultiProc mishandling crashes

### LINKED ISSUE #2300: [BUG] MultiProc crashes are not correctly handled
### Summary
Since the latest refactor, MultiProc calls `self._report_crash` with a `traceback` argument and such function is not defined:

```
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/workflows.py", line 591, in run
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/plugins/base.py", line 182, in run
self._send_procs_to_workers(updatehash=updatehash, graph=graph)
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/plugins/multiproc.py", line 250, in _send_procs_to_workers
self._report_crash(self.procs[jobid], traceback=traceback)
TypeError: _report_crash() got an unexpected keyword argument 'traceback'
```

I'll propose a fix today.

oesteban:
Fixes #2300


nipype/pipeline/plugins/base.py
```diff
@@ -197,13 +197,18 @@ def _get_result(self, taskid):
     def _submit_job(self, node, updatehash=False):
         raise NotImplementedError
 
-    def _report_crash(self, node, result=None):
-        tb = None
+    def _report_crash(self, node, result=None, traceback=None):
```
satra: is the change in signature because multiproc is not adding traceback to result? i.e., could this remain the same and `'traceback'` added to `result`?

satra:
in general this is fine. i was just wondering why the signature of the function is changing.

oesteban:
You were right, I found that `run_without_submitting` wasn't clearing up the job from the queue and, since `_clean_queue` calls `_report_crash` then the `_report_crash` calls were indeed unnecessary.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -275,10 +277,13 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
                 logger.debug('Running node %s on master thread',
                              self.procs[jobid])
                 try:
-                    self.procs[jobid].run()
```
oesteban: I also caught this missing `updatehash`

oesteban:
This is working out with fmriprep 👍

PR #2303, opened 2017-11-26
[MAINT] Reorganize nipype.interfaces.base

oesteban:
This PR intends to 1) split `nipype/nipype/interfaces/base.py` in a more digestible python submodule; 2) remove miscellaneous/utility functions and place them in modules where they are more visible or they fit better. After all, nipype code should be more readable after these changes;
3) completely isolate the `traits_extension` module, so users are now expected to use `base.traits` for all traits in nipype. That would also help the transition towards traits or alternatives.

This PR does not intend to change any functionality. The idea is making it easier to transition code of Nipype Interfaces from 1.0 to 2.0. Also, it will be very appreciated in a "detachable" nipype `CommandLine` I'm working on (the `run` member will still work exactly the same, but you can just `start` an interface and let it run in the background while you keep using your ipython shell, for instance).

#### New module structure

```
nipype.interfaces.base/
                  |- __init__          - Makes classes available the way they used to be available
                  |- core              - The interfaces proper, cleaned up of other code
                  |- specs             - input/output specs to be used by interfaces
                  |- support           - some miscellaneous utilities needed by interfaces
                  \- traits_extension  - same old module, with few additions
```

**This should not be merged before releasing 0.14.0.**

List of changes:

- [x] Moved `nipype.interfaces.base._exists_in_path` -> `nipype.utils.filemanip.which`. Additionally, this function has been modified to rely on `shutil.which` on python >= 3.3. This function does not return `bool, str` anymore, to make it more similar to `shutil.which`. If a command is not found in PATH, it returns `None` (vs. `False, None` in the former version). Uses of the old `_exists_in_path` have been revised for this change. 
- [x] Moved `nipype.interfaces.base._canonicalize_env` -> `nipype.utils.filemanip.canonicalize_env`.
- [x] Moved `nipype.interfaces.base.get_depencencies` -> `nipype.utils.filemanip.get_dependencies`.
- [x] Moved some nipype traits (`OutputMultiPath`, `InputMultiPath`, etc) and some monkeypatches to tratis into the new `traits_extension` submodule.
- [x] Moved interface specs (`BaseTraitedSpec`, `TraitedSpec`, `DynamicTraitedSpec`, etc) into a `nipype.interfaces.base.specs` module. 
- [x] Moved `nipype.interfaces.base.load_template` to `nipype.interfaces.fsl.model` which was the only module that relied on it. A stub for backwards compatibility with a deprecation warning has been left in place. I also moved the `script_templates` folder to a (IMHO) better location under the top folder. Also, the templates are now accessed using the `package_resources` module.
- [x] Created a base submodule called `support` with utilities to the base interfaces (the `Bunch`, `InterfaceResult`, etc.).
- [x] Revised `nifty-` interfaces to use the `version_from_command` property of nipype interfaces (instead of re-implementing the same thing).
- [x] Minor editions of small issues found in files that I had to touch along the way (mostly unicode prefixes in tests, unused imports, etc).
- [x] Added _some_ documentation to the new files.


oesteban:
It'd be great to consider this one ASAP before it gets stale

@satra - I asked Dorota and Mathias for reviews, but your big-picture opinion would be greatly appreciated :)

djarecka:
@oesteban - completely missed the request... Will check in the morning and see if I have any comments.

oesteban:
@djarecka sure, no rush!

nipype/interfaces/base/traits_extension.py
```diff
@@ -334,3 +352,117 @@ def has_metadata(trait, metadata, value=None, recursive=True):
                 count += has_metadata(handler, metadata, recursive)
 
     return count > 0
+
+
+class MultiPath(traits.List):
```
djarecka: after @satra suggestions i've started calling it MultiObject

nipype/interfaces/base/traits_extension.py
```diff
@@ -334,3 +352,117 @@ def has_metadata(trait, metadata, value=None, recursive=True):
...
+        elif len(value) == 1:
+            return value[0]
+        else:
+            return value
+
+    def set(self, object, name, value):
+        self.set_value(object, name, value)
+
+
+class InputMultiPath(MultiPath):
```
djarecka: this class don't have any new method so i was wondering if we can decrease number of classes, e.g. leave `InputMultiPath` and `OutputMultiPath`

oesteban:
Thanks a lot @djarecka!

In general, my answer is that the intent of this PR is to reorganize modules avoiding functionality modifications.

I'll open one or two issues to be addressed after this PR is merged, based on your comments here. Most of those things you mention are worth revising and fixing afterwards. For the rest of your comments not covered by those follow-up issues, I'll make the corresponding changes here.

nipype/interfaces/base/core.py
```diff
@@ -2,663 +2,65 @@
...
-class TraitedSpec(BaseTraitedSpec):
-    """ Create a subclass with strict traits.
-
-    This is used in 90% of the cases.
-    """
-    _ = traits.Disallow
-
-
 class Interface(object):
     """This is an abstract definition for Interface objects.
```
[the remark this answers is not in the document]
oesteban: Back in a day, I proposed something similar: https://github.com/oesteban/nipype/blob/fe4578b4461ee8aee63c86eaf47de5ac4985fae6/nipype/interfaces/base/interfaces.py#L62-L101

In this case, using the abstract interfaces that you build using `traits.api.Interface` and those `@provides` decorators.

oesteban:
Actually, I think it would be better if you opened the issues. After all you are the expert on traits and loose ends of nipype interfaces!

djarecka:
@oesteban - sure, we can open a new pr. I like the reorganization and the separation of the `base.py` file. I added my comments, since you were removing some of the functions, but I can copy/paste my comments to a new issue. 

The only suggestions that I had for the reorganization (and at the end forgot to write it ;-) was about `PackageInfo` (https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/core.py#L1215) - could be also moved to `support.py` IMO.

oesteban:
Did I remove something without providing an alternative for it? Do you have an example?

djarecka:
@oesteban - no, it's not what i meant, i don't think so you modify anything important. I just saw removing/changes in `_exists_in_path`, `no_nifty_package`, etc. I agree, it's better to include my suggestions in a new pr.

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -65,8 +69,10 @@ def test_callback_exception(tmpdir):
     assert so.statuses[1][1] == 'exception'
 
 def test_callback_multiproc_normal(tmpdir):
+    tmpdir.chdir()
+
     so = Status()
-    wf = pe.Workflow(name='test', base_dir=tmpdir.strpath)
+    wf = pe.Workflow(name='test')
```
satra: perhaps `base_dir` should be set to local wd. in this case the working directory will be in some other tmp directory

setup.py
```diff
@@ -106,7 +106,7 @@ def main():
         pjoin('workflows', 'data', '*'),
         pjoin('pipeline', 'engine', 'report_template.html'),
         pjoin('external', 'd3.js'),
-        pjoin('interfaces', 'script_templates', '*'),
+        pjoin('script_templates', '*'),
```
satra: since the script_templates are all fsl model related, perhaps this can be moved into `interfaces/fsl/model/script_templates`

satra:
@oesteban - overall looks great!

nipype/interfaces/base/support.py
```diff
@@ -0,0 +1,303 @@
...
+    """Custom error for interfaces"""
+
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return '{}'.format(self.value)
+
+
+class Bunch(object):
```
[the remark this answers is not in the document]
oesteban: Yep, this would be great to add in @djarecka's issue.

oesteban:
Thanks! @mgxd may you do the honors?

nipype/interfaces/base/support.py
```diff
@@ -0,0 +1,303 @@
...
+    interface : class type
+        A copy of the `Interface` class that was run to generate this result.
+    inputs :  a traits free representation of the inputs
+    outputs : Bunch
+        An `Interface` specific Bunch that contains all possible files
+        that are generated by the interface.  The `outputs` are used
+        as the `inputs` to another node when interfaces are used in
+        the pipeline.
+    runtime : Bunch
+
```
mgxd: extra line?

nipype/interfaces/niftyreg/tests/test_reg.py
```diff
@@ -4,13 +4,17 @@
...
 
-from nipype.interfaces.niftyreg import (no_nifty_package, get_custom_path,
-                                        RegAladin, RegF3D)
-from nipype.testing import example_data
+from ....utils.filemanip import which
+from ....testing import example_data
+from .. import (get_custom_path, RegAladin, RegF3D)
+
+
+def no_nifty_tool(cmd=None):
```
mgxd: maybe define in `tests/utils.py` and import in each test?

nipype/utils/filemanip.py
```diff
@@ -663,3 +671,90 @@ def dist_is_editable(dist):
...
+    Return the path to an executable which would be run if the given
+    cmd was called. If no cmd would be called, return ``None``.
+
+    Code for Python < 3.3 is based on a code snippet from
+    http://orip.org/2009/08/python-checking-if-executable-exists-in.html
+
+    """
+
+    if pathext is None:
+        pathext = os.environ.get("PATHEXT", "").split(os.pathsep)
```
mgxd: os.getenv cleaner?

nipype/interfaces/base/support.py
oesteban: removed 👍

satra:
@oesteban - shouldn't many of the base tests move into the base directory as well?

oesteban:
Working on it

oesteban:
this PR will require a huge `make specs` before merging, just let me know when you are ready to make that happen.

satra:
@oesteban - changes look good to me. thanks for the effort.

satra:
when updating the autotests make sure to add __init__.py to the new test folders.

PR #2312, opened 2017-11-29
rel: 0.14.0

mgxd:
TODO after merge
* [ ] git tag
* [ ] deploy to Pypi
* [ ] update [conda build](https://github.com/conda-forge/nipype-feedstock/pull/8) hash
* [ ] change `nipype/info.py` and 0.14.1-dev? and add a new header "Upcoming Release" to CHANGES


doc/conf.py
```diff
@@ -82,7 +82,7 @@
 # The short X.Y version.
 version = nipype.__version__
 # The full version, including alpha/beta/rc tags.
-release = "0.14.0-rc1"
+release = "0.14.0"
```
oesteban: do we want this to be different from `version`?

oesteban:
I'd go for `0.14.1-dev` for the next release, and try to release it in two weeks.

doc/conf.py
satra: once version is updated, this can stay at `0.14.0`. so just for release this is good, and then post release, we don't have to touch this.

PR #2315, opened 2017-11-30
ENH: Explicit garbage collection in MultiProc

effigies:
With large numbers of processors, moderate (~1GB) increases in memory consumption can produce large spikes in memory committed by the kernel when a batch of processes is forked. For example, if 16 1-core nodes are executed in one round of the MultiProc scheduler, then 16x the virtual memory footprint of the main process is committed by the kernel - so a 500MB increase in main process size would correspond to an 8GB increase in committed memory. For systems which disallow over-commitment of memory, these brief spikes can produce crashes even if the maximum resident memory never approaches the system limits.

This PR adds garbage collection at two points to reduce the memory footprint before forking and after running nodes in the main process, which has the most potential to accumulate objects that need to be freed.

- Garbage collect before submitting jobs
- Garbage collect after `run_without_submitting` jobs
- PEP8/pyflakes cleanup

Happy to argue/discuss alternatives. We found that unsetting `run_without_submitting` on some nodes that were a little more resource-intensive than we originally thought reduced spikes in main process memory consumption. (Would you say this is an accurate characterization?)

### DIFF
--- a/nipype/pipeline/plugins/multiproc.py
+++ b/nipype/pipeline/plugins/multiproc.py
@@ -122,13 +122,16 @@ class MultiProcPlugin(DistributedPluginBase):
         non_daemon = self.plugin_args.get('non_daemon', True)
         maxtasks = self.plugin_args.get('maxtasksperchild', 10)
         self.processors = self.plugin_args.get('n_procs', cpu_count())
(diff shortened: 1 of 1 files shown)


oesteban:
@effigies would you like to place this in the 0.14.1 milestone?

PR #2323, opened 2017-12-01
[ENH] Logging - indicate whether the node was cached

oesteban:
This PR introduces a minimal modification to the logging trace that happens when a node has been run, in order to provide more detailed information and whether the node was cached or actually run.


oesteban:
@satra, so far I'm finding this pretty useful

Issue #2320, opened 2017-12-01
suggesting changes to `interfaces.base`


djarecka:
### Summary
I copy/paste my comments and questions from #2303, so might be included in a separate pr. Comments are welcome:

- [ ] is anyone using `MpiCommandLine`? couldn't find any examples in nipypehttps://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/core.py#L1145

- [ ] `_check_version_requirements` does not check the max_version, this block has wrong indentation: https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/core.py#L403

- [ ] I was a bit confused about `BaseFile` and `File`, etc., do we really need both? Also I believe that we're assuming that `BaseUnicode` uses the fast validator (https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/traits_extension.py#L98), but from traits documentation I understood that `Unicode` (not `BaseUnicode`) uses the C-level fast validator (http://docs.enthought.com/traits/traits_api_reference/trait_types.html?highlight=baseunicode#traits.trait_types.BaseUnicode. Am I missing something?   

- [ ]  couldn't find a place where we're using `_requires_warn` (https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/specs.py#L114)

- [ ]  the same about `_hash_infile` (https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/specs.py#L158)

- [x] changing names to "list" https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/specs.py#L263

- [ ] is this method (https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/support.py#L171) still working? 
is it for this package: http://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.pretty.html#extending ?
they suggest `_repr_pretty_`

- [ ] should be called `MultiObject` instead of `MultiPath`? https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/traits_extension.py#L357

- [ ] should we keep only 2 classes `InputMultiPath` and `OutputMultiPath` (the first one doesn't have any new methods)?

**updates**

- [ ] refining and simplifying `Bunch` class (using `hastraits`) 
- [ ] [Check on this `synchronize`](https://github.com/nipy/nipype/pull/2325#discussion_r159650088)
[...]

PR #2325, opened 2017-12-01
[MAINT] Cleaning / simplify Node

oesteban:
This PR uses some code I had under the hood before 0.14.0 that simplifies `Node`. These are the changes:

- [x] Move `nipype.pipeline.engine.utils.make_output_dir` to `nipype.utils.filemanip.makedirs` and change the function signature to be consistent with the os.makedirs of Python >=3.3. The behavior has been barely modified (if `exists_ok=False` falls back to the traditional os.makedirs; runs former nipype code otherwise).
- [x] Cache output directory with first call to `output_dir()`.
- [x] Simplify `hash_exists` - should implement the same in a hopefully more reliable way (less branches)
- [x] Simplify `Node.run` - again, reducing code branches.
- [x] Simplify `Node._run_interface`, moving the chdirs to `Node.run`. Seems to fix poldracklab/fmriprep#868, however I haven't checked deep enough.
- [x] Pep8 fixes, tidy up imports, etc.


nipype/pipeline/engine/nodes.py
```diff
@@ -189,12 +188,11 @@ def interface(self):
...
     @property
     def result(self):
-        if self._result:
-            return self._result
-        else:
-            cwd = self.output_dir()
-            result, _, _ = self._load_resultfile(cwd)
-            return result
+        # Cache first
+        if not self._result:
```
satra: we don't want to do this i think. this will increase memory consumption. we want results to be loaded on the fly.

nipype/pipeline/engine/nodes.py
```diff
@@ -189,12 +188,11 @@ def interface(self):
 
     @property
     def result(self):
-        if self._result:
-            return self._result
```
satra: i think this was a legacy from before. we should check when `self._result` is actually not None.

nipype/pipeline/engine/nodes.py
```diff
@@ -277,23 +274,29 @@ def hash_exists(self, updatehash=False):
...
+        if op.exists(outdir):
+            # Find previous hashfiles
+            hashfiles = glob(op.join(outdir, '_0x*.json'))
+            if len(hashfiles) > 1:  # Remove hashfiles if more than one found
+                logger.info('Removing hashfiles (%s) and forcing node to rerun',
+                            ', '.join(['"%s"' % op.basename(h) for h in hashfiles]))
+                for hf in hashfiles:
+                    os.remove(hf)
+
+            if updatehash and len(hashfiles) == 1:
```
satra: i don't think the second clause is necessary or the os.remove statement.

the intent of `updatehash==True` was to always update the hash and return true. not just when there was only one hashfile. it was intended to be used cautiously.

nipype/utils/filemanip.py
```diff
@@ -663,3 +664,54 @@ def dist_is_editable(dist):
...
+        try:
+            os.makedirs(path)
+        except OSError:
+            fmlogger.debug("Problem creating directory %s", path)
+            if not os.path.exists(path):
+                raise OSError('Could not create directory %s' % path)
+    return path
+
+
+def emptydirs(path):
```
satra: docstring

nipype/pipeline/engine/nodes.py
oesteban: Okay, I'll see how to improve the process

oesteban: Done, no `self._result` anymore.

oesteban:
@djarecka, @mgxd another maintenance PR for you to review, sorry about that. I don't have more of these for now... if it is of any help. The idea of these two refactorings is making things ready for the code rodeo and to scrub some rust off.

Here I'm changing some of the functionality of `Node`, so this PR requires a deeper scrutiny from you.

oesteban:
https://github.com/poldracklab/fmriprep/issues/868#issuecomment-349877524

nipype/pipeline/engine/nodes.py
```diff
@@ -277,23 +271,28 @@ def hash_exists(self, updatehash=False):
...
-            for file in glob(op.join(outdir, '_0x*.json')):
-                os.remove(file)
-            self._save_hashfile(hashfile, hashed_inputs)
-        return op.exists(hashfile), hashvalue, hashfile, hashed_inputs
+        logger.debug('Node hash value: %s', hashvalue)
+
+        if op.exists(outdir):
+            # Find previous hashfiles
+            hashfiles = glob(op.join(outdir, '_0x*.json'))
+            if len(hashfiles) > 1:  # Remove hashfiles if more than one found
```
satra: `or (hashfiles and hashfiles[0] != hashfile)`

this would ensure that two hashfiles are never left in the directory.

also we should check for any unfinished hashfiles in the directory and raise an exception. i.e. tell the user that you can update a hashfile in an unfinished hashfile situation.

oesteban: Regarding the second idea: are you sure we want to error? We don't want to just dismiss any viable hashfile and clear up the folder (if an unfinished hashfile was found)?

I'm pushing a commit for this, with error. Let me know if we just want to clear up the folder and return no hashfiles.

oesteban: BTW @satra, what do you think about passing the responsibilities of hashing/caching on to the interface?

The `run()` of the interface would have a `force_run` argument (so it can be done dynamically) and a `updatehash` to also support that option.

It'd be great to have the interfaces cache themselves.

satra:
@oesteban - passing hashing to interfaces is a bigger discussion of merging node and interface into a common baseclass. let's leave that out of this and talk about separately. @djarecka has finally gotten back to the engine revision. let's review the base class question after that's completed.

the dichotomy for caching has often been around whether we should be caching or users do it. i think making memcache be the default way interfaces run would be a good thing, but that's a drastic change, so we leave that for 2.0.

oesteban:
Oh, sure thing. I wasn't proposing that for this PR :)

Have you thought about rising an error or just go on when an unfinished hashfile is found in the output directory (when it shouldn't)?

satra:
@oesteban - i haven't reviewed it yet - will do a bit later. i think the only caveats are related to when an interface can continue. i just have to remind myself of the full logic and make sure this initial check doesn't hamper that. we may have to specifically check if the only json file in that section is the unfinished one. 

i'm thinking the error should be raised if there is an unfinished hash file or no hashfile and the user asks to update hash in that directory. 

the multiple hashfile should really no longer exist and was a function of an old MapNode bug that was resolved several moons ago.

nipype/pipeline/engine/nodes.py
```diff
@@ -277,23 +271,37 @@ def hash_exists(self, updatehash=False):
...
-            for file in glob(op.join(outdir, '_0x*.json')):
-                os.remove(file)
-            self._save_hashfile(hashfile, hashed_inputs)
-        return op.exists(hashfile), hashvalue, hashfile, hashed_inputs
+        logger.debug('Node hash value: %s', hashvalue)
+
+        if op.exists(outdir):
+            # Find unfinished hashfiles and error if any
+            unfinished = glob(op.join(outdir, '_0x*_unfinished.json'))
+            if unfinished:
```
satra: `unfinished and updatehash`

nipype/pipeline/engine/nodes.py
```diff
@@ -304,124 +312,118 @@ def run(self, updatehash=False):
...
+                    exp_hash = exp_hash_file_base[len('_0x'):]
+                    logger.debug("Previous node hash = %s", exp_hash)
+                    try:
+                        prev_inputs = load_json(exp_hash_paths[0])
+                    except:
+                        pass
+                    else:
+                        logging.logdebug_dict_differences(prev_inputs,
+                                                          hashed_inputs)
+            if not force_run and str2bool(self.config['execution']['stop_on_first_rerun']):
```
satra: perhaps stop_on_first_rerun should change the logging from debug to info, so that people don't always have to switch to debug mode to see the differences.

nipype/pipeline/engine/nodes.py
```diff
@@ -304,124 +312,118 @@ def run(self, updatehash=False):
...
+        except:
+            logger.warning('[Node] Exception "%s" (%s)', self.fullname, outdir)
+            # Tear-up after error
+            os.remove(hashfile_unfinished)
+            os.chdir(cwd)
+            raise
+
+        # Tear-up after success
+        shutil.move(hashfile_unfinished, hashfile)
+        self.write_report(report_type='postexec', cwd=outdir)
```
satra: not in this PR, but in a future PR, we can change write_report to take a user defined function.

nipype/pipeline/engine/nodes.py
```diff
@@ -1039,10 +1038,12 @@ def __init__(self, interface, iterfield, name, serial=False, nested=False, **kwa
         name : alphanumeric string
             node specific name
         serial : boolean
-            flag to enforce executing the jobs of the mapnode in a serial manner rather than parallel
+            flag to enforce executing the jobs of the mapnode in a serial
+            manner rather than parallel
         nested : boolea
```
satra: boolean

nipype/pipeline/engine/nodes.py
```diff
@@ -1039,10 +1038,12 @@ def __init__(self, interface, iterfield, name, serial=False, nested=False, **kwa
...
-            support for nested lists, if set the input list will be flattened before running, and the
-            nested list structure of the outputs will be resored
+            support for nested lists, if set the input list will be flattened
+            before running, and the nested list structure of the outputs will
+            be resored
```
satra: restored

satra: `support for nested lists. If set, ... before running and the nested ... restored`

satra:
@oesteban - just a few minor comments - otherwise looks good.

oesteban:
Thanks, I'll be working on this along next week :)

nipype/utils/filemanip.py
```diff
@@ -673,6 +674,65 @@ def dist_is_editable(dist):
     return False
 
 
+def makedirs(path, exist_ok=False):
```
djarecka: I wonder weather using the same name as the `os` can be confusing.

nipype/pipeline/engine/nodes.py
```diff
@@ -1334,5 +1339,5 @@ def _run_interface(self, execute=True, updatehash=False):
             for path in dirs2remove:
                 shutil.rmtree(path)
         else:
-            self._result = self._load_results(cwd)
+            result = self._load_results(cwd)
```
djarecka: you don't return `result`, is this ok?

nipype/pipeline/engine/nodes.py
oesteban: In fmriprep we have now more logging levels. We could have a VERBOSE level (more info than INFO, for instance these messages), and use DEBUG for deep debugging info.

WDTY?

nipype/pipeline/engine/nodes.py
```diff
@@ -258,15 +260,7 @@ def set_input(self, parameter, val):
...
-        val = None
-        if self._result:
-            val = getattr(self._result.outputs, parameter)
-        else:
-            cwd = self.output_dir()
-            result, _, _ = self._load_resultfile(cwd)
-            if result and result.outputs:
-                val = getattr(result.outputs, parameter)
-        return val
+        return getattr(self.result.outputs, parameter, None)
```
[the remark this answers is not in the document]
oesteban: `self.result` is a property that always loads the results files, so as not to keep it in memory.

nipype/pipeline/engine/nodes.py
oesteban: thanks, this has been revised :)

nipype/pipeline/engine/utils.py
```diff
@@ -274,18 +273,16 @@ def _write_detailed_dot(graph, dotfilename):
                 else:
                     outport = cd[0][0]
                 inport = cd[1]
-                ipstrip = 'in' + replacefunk(inport)
-                opstrip = 'out' + replacefunk(outport)
+                ipstrip = 'in%s' % _replacefunk(inport)
```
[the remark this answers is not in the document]
oesteban: I don't think this formatting is deprecated, is it?

nipype/utils/filemanip.py
oesteban: The idea is to be able to remove this safely when we only support Python 3

oesteban:
this is ready for a deep review

nipype/pipeline/engine/nodes.py
oesteban: I modified a little how this works. With fmriprep it goes great (but we have the extra VERBOSE level). This way one can avoid the extremely verbose DEBUG, and still get the differences that make an interface be run.

satra:
@oesteban - the changes look good to me. thanks for the extensive refactoring and clean up.

oesteban:
Yes, I though some housekeeping was necessary before the code rodeo so you guys can do the hard work :D.

There are still some little things here and there that can be improved on this PR, but I think that will come naturally with use. I've been using these changes in FMRIPREP for two releases now and tested rather extensively, and so far so good. Since I merged it in our experimental mirror of nipype, I haven't seen poldracklab/fmriprep#868 replicate.

I'll not rush merging to allow @djarecka and @mgxd have a look if the can/want.

oesteban:
In the last commit I had to add a patch since we were seeing a regression of poldracklab/fmriprep#868. Maybe worth giving this a thought, @satra: seems like (AFAIK only in MultiProc) some other process is removing the CWD where that worker was. I've added debug traces where shutil.rmtree is called (if there weren't already) to help understand this problem.

satra:
@oesteban - is there a repeatable example? or is this happening randomly?

oesteban:
It's hard to reproduce, no repeatable at all. However, I was running fmriprep on 235 subjects (each an individual process exclusively run in a node), and ~80 of those had the problem.

I'm now testing on those with this patch to see if it fixes the issue or creates new ones. I think 80 is enough to replicate the problem at least once (or better said: enough to make nipype catch the exception and test if the patch has side effects).

I don't have any clue yet of where we are deleting those folders.

nipype/pipeline/engine/utils.py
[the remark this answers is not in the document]
djarecka: It's not deprecated right now, but `format` is "intended as a replacement for the existing '%' string", not sure what is a typical timeframe... 
@satra at some point asked me to use `format` whenever I change the code.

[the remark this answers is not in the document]
oesteban: I can totally change this to format, if we see it important. The idea was not to use concatenation which is less efficient and error prone (with unicode and bytes)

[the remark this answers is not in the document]
djarecka: no, it was mostly my question

nipype/utils/misc.py
```diff
@@ -229,3 +239,57 @@ def normalize_mc_params(params, source):
...
+    # Add topical message
+    if diff:
+        diff.insert(0, "Dictionaries had differing keys:")
+
+    diffkeys = len(diff)
+
+    # Values in common keys would differ quite often,
+    # so we need to join the messages together
+    for k in new_keys.intersection(old_keys):
+        same = False
```
djarecka: not needed

djarecka:
@oesteban - thanks for this PR! I've tried to review the changes again today and just had some small questions/comments. 
After this is merged I'll check which points from #2320 are still relevant.

nipype/pipeline/engine/utils.py
```diff
@@ -904,7 +1142,7 @@ def _standardize_iterables(node):
     fields = set(node.inputs.copyable_trait_names())
     # Flag indicating whether the iterables are in the alternate
     # synchronize form and are not converted to a standard format.
-    synchronize = False
+    # synchronize = False  # OE: commented out since it is not used
```
djarecka: is these comments are valid for `node.synchronize` instead of `synchronize`. would be good to clean it so it's clear

nipype/pipeline/engine/nodes.py
```diff
@@ -1156,8 +1037,9 @@ def _make_nodes(self, cwd=None):
                         base_dir=op.join(cwd, 'mapflow'),
                         name=nodename)
             node.plugin_args = self.plugin_args
-            node._interface.inputs.trait_set(
+            node.interface.inputs.trait_set(
```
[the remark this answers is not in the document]
oesteban: in general, it is preferable to access public members rather than private members (marked generally with underscore) from outside the object.

nipype/pipeline/engine/nodes.py
```diff
@@ -304,143 +354,109 @@ def run(self, updatehash=False):
         updatehash: boolean
             Update the hash stored in the output directory
```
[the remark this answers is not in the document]
oesteban: If @satra agrees, we can remove it. I think it is an ancient feature.

The idea being that, when you run with `updatehash=True`, even if the hash of some input changed, the interface will not be run again and reuse the old output as output.

For instance, you forgot to mark with `nohash=True` some of your inputs (say the number of threads), but you don't want to rerun interfaces that are already cached. Then you would change the number of threads and run with `updatehash=True`. When nipype finds some of your interfaces with this problem cached, they are not run again and the hash of this input gets updated.

I'm not sure the example is a use case but I understand the feature that way.

nipype/pipeline/engine/nodes.py
```diff
@@ -304,143 +354,109 @@ def run(self, updatehash=False):
...
-        """
-        if len(param) > 32:
-            return sha1(param.encode()).hexdigest()
+        # by rerunning we mean only nodes that did finish to run previously
+        if hash_exists and not isinstance(self, MapNode):
+            logger.debug('[Node] Rerunning "%s"', self.fullname)
+            if not force_run and str2bool(self.config['execution']['stop_on_first_rerun']):
+                raise Exception('Cannot rerun when "stop_on_first_rerun" is set to True')
+
+        # Hashfile while running, remove if exists already
```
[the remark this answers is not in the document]
oesteban: I'll check.

nipype/pipeline/engine/nodes.py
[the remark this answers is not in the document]
satra: it's true that this is rarely used currently, but when people were using interactive workflows and moving around the working directory this became a very useful thing. examples are:

1. you run a workflow and move the working directory from one filesystem to another. all the hashes will change
2. you go into the working directory to debug something and messed things up. you can quickly rerun an update hash that updates your cache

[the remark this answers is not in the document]
djarecka: if we decide to leave it, can we extend the docstring?

nipype/pipeline/engine/nodes.py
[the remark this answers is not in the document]
djarecka: ok, yes, it is for `node`, not for `self`...

nipype/pipeline/engine/nodes.py
```diff
@@ -246,54 +253,97 @@ def output_dir(self):
...
-                params_str = [self._parameterization_dir(p) for p in params_str]
+                params_str = [_parameterization_dir(p) for p in params_str]
             outputdir = op.join(outputdir, *params_str)
-        return op.abspath(op.join(outputdir, self.name))
+
+        self._output_dir = op.abspath(op.join(outputdir, self.name))
+        return self._output_dir
 
     def set_input(self, parameter, val):
         """ Set interface input value"""
```
mgxd: excess space

nipype/pipeline/engine/nodes.py
```diff
@@ -246,54 +253,97 @@ def output_dir(self):
...
-        else:
-            cwd = self.output_dir()
-            result, _, _ = self._load_resultfile(cwd)
-            if result and result.outputs:
-                val = getattr(result.outputs, parameter)
-        return val
+        return getattr(self.result.outputs, parameter, None)
 
     def help(self):
         """ Print interface help"""
```
mgxd: Space

nipype/pipeline/engine/nodes.py
```diff
@@ -610,177 +539,131 @@ def _load_results(self, cwd):
...
+            self.config,
+            dirs2keep=dirs2keep
+        )
+        _save_resultfile(result, outdir, self.name)
+
+        return result
 
-    def _copyfiles_to_wd(self, outdir, execute, linksonly=False):
+    def _copyfiles_to_wd(self, execute=True, linksonly=False):
         """ copy files over and change the inputs"""
```
mgxd: space

nipype/pipeline/engine/nodes.py
```diff
@@ -1156,8 +1037,9 @@ def _make_nodes(self, cwd=None):
                         base_dir=op.join(cwd, 'mapflow'),
                         name=nodename)
             node.plugin_args = self.plugin_args
-            node._interface.inputs.trait_set(
+            node.interface.inputs.trait_set(
                 **deepcopy(self._interface.inputs.get()))
+            node.interface.resource_monitor = self._interface.resource_monitor
```
mgxd: `self.interface.resource_monior`

nipype/pipeline/engine/utils.py
```diff
@@ -4,39 +4,50 @@
...
 import sys
-from future import standard_library
-standard_library.install_aliases()
+import pickle
 from collections import defaultdict
-
+import re
 from copy import deepcopy
 from glob import glob
+from distutils.version import LooseVersion
```
mgxd: `from ... import LooseVersion`

nipype/pipeline/engine/utils.py
```diff
@@ -46,39 +57,268 @@
...
+                err = [result.runtime.traceback]
+
+            err += format_exception(*sys.exc_info())
+            err = '\n'.join(err)
+        finally:
+            yield i, result, err
+
+
+def write_report(node, report_type=None, is_mapnode=False):
+    """
```
mgxd: Can just make this one line if not expanding

nipype/pipeline/engine/workflows.py
```diff
@@ -930,13 +911,13 @@ def _get_dot(self, prefix=None, hierarchy=None, colored=False,
...
-                    '#00A300', '#B4FFB4', '#E6FFE6', # G
-                    '#0000FF', '#B4B4FF'] # loop B
+        colorset = ['#FFFFC8',                        # Y
+                    '#0000FF', '#B4B4FF', '#E6E6FF',  # B
+                    '#FF0000', '#FFB4B4', '#FFE6E6',  # R
+                    '#00A300', '#B4FFB4', '#E6FFE6',  # G
+                    '#0000FF', '#B4B4FF']             # loop B
         if level > len(colorset) - 2:
-            level = 3 # Loop back to blue
+            level = 3  # Loop back to blue
```
mgxd: extra space?

nipype/pipeline/engine/nodes.py
```diff
@@ -610,177 +539,131 @@ def _load_results(self, cwd):
...
+        logger.debug('copying files to wd [execute=%s, linksonly=%s]',
+                     execute, linksonly)
+
+        outdir = self.output_dir()
+        if execute and linksonly:
+            olddir = outdir
+            outdir = op.join(outdir, '_tempinput')
+            makedirs(outdir, exist_ok=True)
+
+        for info in self.interface._get_filecopy_info():
```
mgxd: `self._interface` and `self.interface` should be equal - but as an added layer of security wouldn't it be better to keep the usage consistent?

nipype/pipeline/engine/workflows.py
oesteban: Nope, apparently there is some PEP somewhere advising to have 2 spaces before the inline comments

oesteban:
I'm seeing this from time to time:

```
File "/usr/local/miniconda/lib/python3.6/multiprocessing/forkserver.py", line 178, in main
_serve_one(s, listener, alive_r, handler)
File "/usr/local/miniconda/lib/python3.6/multiprocessing/forkserver.py", line 212, in _serve_one
code = spawn._main(child_r)
File "/usr/local/miniconda/lib/python3.6/multiprocessing/spawn.py", line 115, in _main
self = reduction.pickle.load(from_parent)
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/__init__.py", line 13, in <module>
from .utils.config import NipypeConfig
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/utils/__init__.py", line 4, in <module>
from .config import NUMPY_MMAP
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/utils/config.py", line 82, in <module>
""" % (homedir, os.getcwd())
FileNotFoundError: [Errno 2] No such file or directory
```

Will let you know when I find why this is happening.

nipype/pipeline/engine/nodes.py
oesteban: I would opt for `self._interface`. I've made sure no `self.interface` remain.

nipype/utils/misc.py
oesteban: you sure? how would you modify this?

nipype/pipeline/engine/utils.py
oesteban: I didn't want to remove it since the comment above seemed relevant. But I agree that once we clarify the intent of this synchronize, we should clean up these comments.

djarecka: that was my guess that the these comments are relevant also for `node.synchronize` but wasn't sure if entirely.

nipype/utils/misc.py
djarecka: it's too late in Europe to be sure about anything, but regardless what happens in this loop, `same` should be defined/changed either in `try` part or `except` part, so looked to me that it doesn't have to be defined before `try/except`

nipype/pipeline/engine/utils.py
oesteban: Added to #2320

nipype/utils/misc.py
oesteban: Added to #2320. Even though this PR is changing a lot already, I'd leave this catch for a new one.

oesteban:
Alright, this is looking good. The `FileNotFoundError` seems to be under control with the new code, I'll keep an eye on these kinds of errors. Unless you want me to go one extra round of checking, this is ready to hit merge.

mgxd:
@oesteban - while using this PR

```
--- Logging error ---
  File "/opt/conda/envs/neuro/lib/python3.6/logging/__init__.py", line 992, in emit
    msg = self.format(record)
  File "/opt/conda/envs/neuro/lib/python3.6/logging/__init__.py", line 838, in format
    return fmt.format(record)
  File "/opt/conda/envs/neuro/lib/python3.6/logging/__init__.py", line 575, in format
    record.message = record.getMessage()
  File "/opt/conda/envs/neuro/lib/python3.6/logging/__init__.py", line 338, in getMessage
    msg = msg % self.args
TypeError: not enough arguments for format string
Call stack:
  File "/opt/conda/envs/neuro/bin/mindboggle123", line 315, in <module>
    mbFlow.run(plugin=args.plugin)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/workflows.py", line 574, in run
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/plugins/linear.py", line 43, in run
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 443, in run
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 520, in _run_interface
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 594, in _run_command
    self._interface.__class__.__name__)
Message: '[Node] Running "%s" ("%s.%s"), a CommandLine Interface with command:\nantsCorticalThickness.sh -a /datastore/sub-BANDA001/anat/sub-BANDA001_T1w.nii.gz -m /output/work/Mindboggle123/antsCorticalThickness/T_template0_BrainCerebellumProbabilityMask.nii.gz -e /opt/data/OASIS-30_Atropos_template/T_template0.nii.gz -d 3 -f /opt/data/OASIS-30_Atropos_template/T_template0_BrainCerebellumExtractionMask.nii.gz -s nii.gz -o /output/ants_subjects/sub-BANDA001/ants -p nipype_priors/BrainSegmentationPrior%02d.nii.gz -t /opt/data/OASIS-30_Atropos_template/T_template0_BrainCerebellum.nii.gz -j 1'
Arguments: ('antsCorticalThickness', 'nipype.interfaces.ants.segmentation', 'antsCorticalThickness')
```

mgxd:
perhaps we should move towards `.format`, since any command containing `%` will raise this error

nipype/utils/misc.py
effigies: @djarecka I agree that you can remove this one line. `same` is defined before being accessed in all branches here.

(This comment in response to #2387.)

nipype/pipeline/engine/utils.py
djarecka: @effigies my concerns were more about the comments explaining `synchronize`. Does it still apply to `node.synchronize` after all changes? I don't fully understand the comments and never used it, so if you can confirm (or suggest changes) that would be great.

effigies: Oh, sorry. It does read to me that there was intent to switch synchronize behavior with this flag, and that the two lines above only apply to that case. I think I would remove them, because the comment below seems to accurately describe the operation.

PR #2331, opened 2017-12-04
ENH: Profile VMS as well as RSS

effigies:
Given that some systems are limited on virtual memory allocation, and not actual memory consumption, being able to profile both memory metrics would be useful

### DIFF
--- a/nipype/utils/profiler.py
+++ b/nipype/utils/profiler.py
@@ -100,9 +100,12 @@ class ResourceMonitor(threading.Thread):
 
     def run(self):
         """Core monitoring function, called by start()"""
+        start_time = time()
+        wait_til = start_time
         while not self._event.is_set():
             self._sample()
-            self._event.wait(self._freq)
+            wait_til += self._freq
+            self._event.wait(max(0, wait_til - time()))
 
 
 # Log node stats function
--- a/nipype/utils/profiler.py
+++ b/nipype/utils/profiler.py
@@ -94,7 +94,7 @@ class ResourceMonitor(threading.Thread):
             except psutil.NoSuchProcess:
                 pass
 
-        print('%f,%f,%f,%f' % (time(), rss / _MB, cpu, vms / _MB),
+        print('%f,%f,%f,%f' % (time(), cpu, rss / _MB, vms / _MB),
               file=self._logfile)
         self._logfile.flush()
 
--- a/nipype/utils/profiler.py
+++ b/nipype/utils/profiler.py
@@ -67,11 +67,14 @@ class ResourceMonitor(threading.Thread):
 
     def _sample(self, cpu_interval=None):
         cpu = 0.0
-        mem = 0.0
+        rss = 0.0
+        vms = 0.0
         try:
             with self._process.oneshot():
                 cpu += self._process.cpu_percent(interval=cpu_interval)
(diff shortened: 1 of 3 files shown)


nipype/interfaces/base.py
```diff
@@ -1134,8 +1134,9 @@ def run(self, **inputs):
 
                     runtime.prof_dict = {
                         'time': vals[:, 0].tolist(),
-                        'mem_gb': (vals[:, 1] / 1024).tolist(),
+                        'rss_gb': (vals[:, 1] / 1024).tolist(),
                         'cpus': vals[:, 2].tolist(),
+                        'vms_gb': (vals[:, 3] / 1024).tolist(),
```
satra: just a nitpick here - should we make the `gb` be `gib` or `GiB` to maintain consistency with the term Gibibytes.

effigies: I guess I'd either prefer `GiB` (since `gib` does not say "gibibyte" to me) or just to switch to decimal factors and keep variable names as `gb`. I'm not sure whether changing the variable name or the log base would be least disruptive, so I'll leave that to your call @oesteban @satra.

satra: i think both are disruptive. since more people will use the name, i think changing the name ensures that people using it accounts for it. changing the log base implies that people using it may interpret the data incorrectly. it's possible this is already happening.

my vote would be for changing the name.

oesteban: I don't think there are a critical mass of people to maintain a name that poorly represents the concept. Let's change it to `GiB` before more people start using the monitor, following @effigies preference :+1: .

effigies:
I'm leaving alone non-profiler uses of `_gb`, as these go pretty deep and will definitely break people's code when annotating their nodes. That refactor will need a deprecation schedule and may be better to do post-1.0.

nipype/utils/profiler.py
```diff
@@ -85,11 +88,13 @@ def _sample(self, cpu_interval=None):
...
                     cpu += child.cpu_percent()
-                    mem += child.memory_info().rss
+                    mem_info = child.memory_info()
+                    rss += mem_info.rss
+                    vms += mem_info.vms
             except psutil.NoSuchProcess:
                 pass
 
-        print('%f,%f,%f' % (time(), (mem / _MB), cpu),
+        print('%f,%f,%f,%f' % (time(), rss / _MB, cpu, vms / _MB),
```
oesteban: a little nitpick here: maybe best putting cpu right after the timestamp? - sorry I didn't suggest this in the first round.

effigies: Sure. I was trying to leave anything that depends on the old columns usable, but I agree that I'd rather have `time,cpu,rss,vms`.

oesteban: let's make the change now, better than after people start using this for real.

effigies:
@oesteban What do you think about moving `run()` to constant-interval instead of constant-wait (as in #2317) here as well?

https://github.com/effigies/nipype/blob/a15e08600798e56bcd616cd81fc8f918e8b1b0c5/nipype/utils/profiler.py#L101-L105

Might as well do it in this PR if we're going to do it.

oesteban:
Sounds very good :+1:

PR #2334, opened 2017-12-10
FIX: some fixes to afni.Deconvolve

mvdoc:
1. If 3dDeconvolve is run with `--x1D_stop`, the output file is not created. The interface currently fails because it assumes that the output file exists. This PR checks whether x1D stop is set, and changes the `exists` flag of `output_spec.out_file`.
2. `force_TR` should be float not int and should precede `-input`
3. fix formatting of `ortvec`


mvdoc:
My attempt to dynamically change the `exists` flag of `out_file` didn't really work, so I set `exists` to False. But if you have any suggestion on how I can modify dynamically a trait in output spec, I feel that would be better.

mgxd:
@mvdoc how about leaving `exists=False` within the outputspec, but removing `outfile` from outputs in [`Deconvolve._list_outputs`](https://github.com/mvdoc/nipype/blob/e95c0a0ca8ef84f4fea1eb77ce731861b9527e9d/nipype/interfaces/afni/model.py#L293) if `x1d_stop` is set to True?

mvdoc:
Done @mgxd. Let me know. Pinging @effigies as well.

nipype/interfaces/afni/tests/test_deconvolve.py
```diff
@@ -0,0 +1,10 @@
+"""Test afni deconvolve"""
```
effigies: Please rename this file to "test_extra_Deconvolve".

This isn't yet a part of documentation, but we realized that lowercase tests like this block building the auto test on OSX. Using "extra" marks it as something to have in addition to the auto-test. And just for consistency's sake, let's use the capitals.

effigies:
Also this needs a `make specs` update.

mvdoc:
@effigies: done

PR #2336, opened 2017-12-10
FIX,REF: BIDSDataGrabber

### LINKED ISSUE #2333: BIDSDataGrabber doesn't populate outputspec when outfields is defined
### Summary
When defining custom fields BIDSDataGrabber doesn't add the corresponding traits to output_spec, so it's not possible to link nodes in a workflow.

### Actual behavior
```python
In [1]: from nipype.interfaces.bids_utils import BIDSDataGrabber

In [2]: bg = BIDSDataGrabber(infields=['infield1', 'infield2'], outfields=['outfield1', 'outfield2'])

In [3]: bg._outputs()
Out[3]: 
```
### Expected behavior

```python
In [1]: from nipype.interfaces.bids_utils import BIDSDataGrabber

In [2]: bg = BIDSDataGrabber(infields=['infield1', 'infield2'], outfields=['outfield1', 'outfield2'])

In [3]: bg._outputs()
Out[3]: 

outfield1 = <undefined>
outfield2 = <undefined>

```

I started working on a fix for this, and it seems to me that perhaps subclassing IOBase instead of BaseInterface might be a solution. But I was wondering what you think it would be better.

mvdoc:
Fixes #2333 .

- Moved `BIDSDataGrabber` to `interfaces.io` for consistency, and also for using `IOBase` as a base class (otherwise failed for circular import)
- Fixed a bug by which if `infields` was set, it was populated anyway by Pybids defaults
- `BIDSDataGrabber` is now subclass of `IOBase` instead of `BaseInterface`
- Added some tests to check that outputs are populated correctly.


nipype/interfaces/io.py
```diff
@@ -2630,3 +2637,116 @@ def _list_outputs(self):
...
+            Indicates the input fields to be dynamically created
+        """
+        super(BIDSDataGrabber, self).__init__(**kwargs)
+
+        if not isdefined(self.inputs.output_query):
+            self.inputs.output_query = {"func": {"modality": "func"},
+                                        "anat": {"modality": "anat"}}
+
+        # If infields is empty, use all BIDS entities
+        if infields is None and have_pybids:
```
mgxd: good catch

nipype/interfaces/bids_utils.py
```diff
@@ -1,147 +0,0 @@
-# -*- coding: utf-8 -*-
```
mgxd: I am okay with this change - but removing this is an api change from `0.14` @satra WDYT

effigies:
@mvdoc This would be good to get fixed before 1.0 (hoping to release tomorrow).

mvdoc:
!!! 1.0 !!!

OK, will fix asap

mvdoc:
OK as for @mgxd's comment I added the interface to `__init__.py`. Let me know @effigies

PR #2345, opened 2017-12-17
Afni verbosity

salma1601:
add AFNI verbosity options when available


nipype/interfaces/afni/preprocess.py
```diff
@@ -438,6 +438,12 @@ class AllineateInputSpec(AFNICommandInputSpec):
         traits.Enum(*_dirs),
         argstr='-nwarp_fixdep%s',
         desc='To fix non-linear warp dependency along directions.')
+    verb = traits.Bool(
```
effigies: Feel free to say no, but would this make more sense as "verbose" instead of "verb"? I think that'll make interfaces a little friendlier to new users, while experienced AFNI users would know "verb" means "verbose", so it shouldn't cause confusion.

nipype/interfaces/afni/preprocess.py
```diff
@@ -438,6 +438,12 @@ class AllineateInputSpec(AFNICommandInputSpec):
...
+        argstr='-verb',
+        desc='Print out verbose progress reports.')
+    quiet = traits.Bool(
+        argstr='-quiet',
+        desc='Don\'t print out verbose progress reports.')
```
effigies: Preference: Use double-quotes instead of escaping apostrophes within single quotes.

nipype/interfaces/afni/preprocess.py
salma1601: yes I agree, I was hesitating on this one. Is that OK if I change all the AFNI `verb` (even introduced before my changes) to `verbose` ?

effigies: Hmm. No, that would require deprecating, because somebody depends on this behavior. If it's consistent with existing interfaces, `verb` seems fine.

salma1601: Ok

nipype/interfaces/afni/utils.py
```diff
@@ -1377,6 +1380,9 @@ class MaskToolInputSpec(AFNICommandInputSpec):
              'or using the labels in {R,L,A,P,I,S}.',
         argstr='-fill_dirs %s',
         requires=['fill_holes'])
+    verbose = traits.Int(
```
effigies: Note that you have `verbose` in `Copy`. If it's okay to mix-and-match (i.e., there are some interfaces that already have `verbose`), I'd prefer that you use `verbose` for all new additions. If this would be the first interface with `verbose`, I'd say switch it to `verb` for consistency.

nipype/interfaces/afni/preprocess.py
salma1601: Ok

effigies:
Looks good. Happy to merge when tests pass.

nipype/interfaces/afni/utils.py
salma1601: Sorry I am not sure to follow. I added `verbose` to `Allineate`, `Copy`, `MaskTool` and`Tcat`. These interfaces had no verbosity parameter before. I didn't touch the other ones which had already a `verb` or a `verbose` parameter as you told me.

effigies: Sorry, I think I wrote this just as you made those changes. I wasn't sure if you were going to keep it at `verb`, but the changes you made look good. You can safely ignore that last message. (Also this one.)

PR #1392, opened 2016-03-03
WIP: antsRegistrationSyNQuick.sh wrapper

fliem:
Hi guys,

I created a wrapper for antsRegistrationSyNQuick.sh. It's my first interface, so it would be great if you had suggestions for improvements (I marked a couple of issues in the code with # todo).

Additionally, I'd like to create a wrapper for antsRegistrationSyN.sh.
antsRegistrationSyNQuick.sh and antsRegistrationSyN.sh  do not have completely identical input arguments:
-t (transform type) has different valid values in SyN and SyNQuick.
-r is setting a different variable
SyNQuick: -r: histogram bins for mutual information in SyN stage (default = 32);
Syn: -r:  radius for cross correlation metric used during SyN stage (default = 4)
What would be the most elegant way to have interfaces for both? Two separat interfaces?

PR #2347, opened 2017-12-22
antsRegistrationSyNQuick (continuation of #1392)

djarecka:
continuing #1392 

not sure where to find a good description of RegistrationSyNQuick, [here](https://github.com/ANTsX/ANTs/blob/master/Scripts/antsRegistrationSyNQuick.sh) I can find only a few usage examples.


mgxd:
@djarecka anymore additions or is this ready?

nipype/interfaces/ants/tests/test_auto_Registration.py
```diff
@@ -1,157 +0,0 @@
-# AUTO-GENERATED by tools/checkspecs.py - DO NOT EDIT
```
mgxd: I think we still wanna keep this

djarecka: I didn't remove it manually. I mentioned in #2349 that since we have a non-auto test, this test has been removed by `make specs`. What happen when you run `make specs`?

nipype/interfaces/ants/registration.py
```diff
@@ -1350,3 +1351,81 @@ def aggregate_outputs(self, runtime=None, needed_outputs=None):
...
+                            usedefault=True, desc='image dimension (2 or 3)')
+    fixed_image = InputMultiPath(File(exists=True), mandatory=True, argstr='-f %s',
+                                 desc='Fixed image or source image or reference image')
+    moving_image = InputMultiPath(File(exists=True), mandatory=True, argstr='-m %s',
+                                  desc='Moving image or target image')
+    output_prefix = Str("transform", usedefault=True, argstr='-o %s',
+                        desc="A prefix that is prepended to all output files")
+
+    # todo ANTSCommandInputSpec already has this, but I can't figure out how to set it without defining it again
+    num_threads = traits.Int(default_value=1, desc='Number of threads (default = 1)', argstr='-n %d')
```
satra: you should check if the script checks for the environment variables that is set by the base class instead of this argument. if so the underlying commands in the script will use it. 

but since this script has an explicit flag for threads, you should over-ride the base class.

nipype/interfaces/ants/registration.py
```diff
@@ -1350,3 +1351,81 @@ def aggregate_outputs(self, runtime=None, needed_outputs=None):
...
+    warped_image = File(exists=True, desc="Warped image")
+    inverse_warped_image = File(exists=True, desc="Inverse warped image")
+    out_matrix = File(exists=True, desc='Affine matrix')
+    forward_warp_field = File(exists=True, desc='Forward warp field')
+    inverse_warp_field = File(exists=True, desc='Inverse warp field')
+
+
+class RegistrationSynQuick(ANTSCommand):
+    """
+    Examples
```
satra: you can just use some of the ants registration examples here in much simplified form.

nipype/interfaces/ants/registration.py
djarecka: @satra - don't understand how this should be changed. Should `num_threads` be in `RegistrationSynQuickInputSpec` (as it is now), or over-ride somewhere else?

In [this script](https://github.com/ANTsX/ANTs/blob/master/Scripts/antsRegistrationSyNQuick.sh) the default value is `-n 1`, so that's why probably the author of changes used `default_value=1.

satra: it should be in:

> `RegistrationSynQuickInputSpec` (as it is now)

since the script explicitly overrides this: https://github.com/ANTsX/ANTs/blob/master/Scripts/antsRegistrationSyNQuick.sh#L388

djarecka:
something went wrong, will open a new pr

PR #2349, opened 2017-12-29
small changes in  FSSourceInputSpec (fixes #2116)

### LINKED ISSUE #2116: FreeSurferSource doesn't check subjects_dir
### Summary
`FreeSurferSource` doesn't check `subjects_dir` is an existing path.

```
from nipype.interfaces.io import FreeSurferSource
fs = FreeSurferSource()
fs.inputs.subjects_dir = 'path/to/no/existing/directory'
fs.inputs.subject_id = 'sub-01'
res = fs.run()
```

### Actual behavior
Doesn't give an error.

### Expected behavior
Should check if directory exists (and possibly if there are files for `subject_id` ?).

djarecka:
adding exists=True for subjects_dir in FSSourceInputSpec
fixes #2116


nipype/interfaces/tests/test_io.py
```diff
@@ -407,6 +407,13 @@ def test_freesurfersource():
     assert fss.inputs.subjects_dir == Undefined
 
 
+def test_freesurfersource_incorrectdir():
+    fss = nio.FreeSurferSource()
+    with pytest.raises(Exception) as err:
```
oesteban: Doesn't it work using `with pytest.raises(TraitError)` ?

djarecka: you're right, after importing `TraitError` it works...

oesteban:
this is a bit weird: why the only changing specs are unrelated to freesurfer? I'd expect FS' interfaces to change... WDYT?

djarecka:
@oesteban - I assumed that it was not run in previous PR for some time, but could be wrong... in my other PR (#2350) I also had changes in those files after running `make specs`, and this was partailly the reason why I didn't run it here

oesteban:
Maybe @satra will correct me, but the idea for the auto tests is to be updated with the PR that modifies interfaces, so that we clearly see how the interface changed. Is that correct?

If so, then we would need to keep both PRs separate and update specs accordingly.

djarecka:
@oesteban - not sure if I understand your point. I didn't mix my PRs, I only wrote that I had exactly the same problem in my other PR, i.e., `make specs` changed autotests that were not related to my changes. In both cases two autotests were deleted: `nipype/interfaces/ants/tests/test_auto_Registration.py` and `nipype/interfaces/niftyseg/tests/test_auto_PatchMatch.py`.

My guess was that simply `make specs` were not run with the PRs that changed those 2 interfaces, so I just included the changes. I've just checked that `make specs` removes exactly the same files even when run in my master (with clean working tree). Can you please check it on your master, I might be doing something wrong...

djarecka:
@oesteban - the mystery solved, `make specs` works differently on my OSX (is not case sensitive). @effigies is addressing this issue in #2374

effigies:
Is this ready to merge? @oesteban @djarecka

djarecka:
@effigies yes

PR #2350, opened 2017-12-29
trait changes in RealignInputSpec (fixes #2120)

### LINKED ISSUE #2120: different error message from SPM interaces
### Summary
#1949 introduced `ImageFileSPM` that gives a nice error message, when input file is `nii.gz`, but `Realign` gives still the old error message " the path  '/data/ds000114/sub-03/func/sub-03_task-fingerfootlips_bold.nii.gz' does not exist."

Compare these two codes:

```
from nipype.interfaces.spm import Smooth
smooth = Smooth(in_files='/existing/file.nii.gz')
smooth.run()
```

and

```
from nipype.interfaces.spm import Realign
realign = Realign(in_files='/existing/file.nii.gz')
realign.run()
```

### Actual behavior
The first code gives a nice error:
`TraitError: /existing/file.nii.gz is not included in allowed types: .img, .nii, .hdr`
the second gives:
`TraitError: Each element of the 'in_files' trait of a RealignInputSpec instance must be a list of items which are an existing file name or an existing file name, but a value of '/existing/file.nii.gz' <class 'str'> was specified.`

### Expected behavior

Both examples should give the same error.

djarecka:
changing in_files in RealignInputSpec so it's similar to SmoothInputSpec
fixes #2120


djarecka:
@satra, @oesteban, @mgxd : CircleCi stops always with "Exception: Cannot rerun when 'stop_on_first_rerun' is set to True" when runs examples. Any suggestions what should I change?

djarecka:
I'll close it and open the new PR without changes in auto tests (#2375)

PR #2355, opened 2018-01-05
STY: deleting tailing spaces and adding missing newlines at end of files

miykael:
I looked at all textfiles, using ``find . -type f -exec grep -Iq . {} \; -and -print`` and made sure that there were...
 - no tailing spaces per line
 - each file ended with a newline
 - no file started with an empty new line


miykael:
I decided to update the hash assertion check in https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/tests/test_support.py#L60 where `nipype/testing/data/realign_json.json` has a newline at the end of the file.

PR #2364, opened 2018-01-10
fix: logging error if % in node cmd

mgxd:
Fixes [#2325 (comment)](https://github.com/nipy/nipype/pull/2325#issuecomment-356485950).

- move some logging to `.format` substitution

Eventually, we should be thorough and replace most `%` substitutions

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -577,25 +577,25 @@ class Node(EngineBase):
             self._originputs = deepcopy(self._interface.inputs)
             self._copyfiles_to_wd(execute=execute)
 
-        message = '[Node] Running "%s" ("%s.%s")'
+        message = '[Node] Running "{}" ("{}.{}")'
         if issubclass(self._interface.__class__, CommandLine):
             try:
                 cmd = self._interface.cmdline
             except Exception as msg:
-                result.runtime.stderr = '%s\n\n%s' % (
+                result.runtime.stderr = '{}\n\n{}'.format(
                     getattr(result.runtime, 'stderr', ''), msg)
                 _save_resultfile(result, outdir, self.name)
                 raise
             cmdfile = op.join(outdir, 'command.txt')
             with open(cmdfile, 'wt') as fd:
                 print(cmd + "\n", file=fd)
-            message += ', a CommandLine Interface with command:\n%s' % cmd
-        logger.info(message, self.name, self._interface.__module__,
-                    self._interface.__class__.__name__)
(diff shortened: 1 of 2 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -577,25 +577,25 @@ def _run_command(self, execute, copyfiles=True):
...
                     getattr(result.runtime, 'stderr', ''), msg)
                 _save_resultfile(result, outdir, self.name)
                 raise
             cmdfile = op.join(outdir, 'command.txt')
             with open(cmdfile, 'wt') as fd:
                 print(cmd + "\n", file=fd)
-            message += ', a CommandLine Interface with command:\n%s' % cmd
-        logger.info(message, self.name, self._interface.__module__,
-                    self._interface.__class__.__name__)
+            message += ', a CommandLine Interface with command:\n{}'.format(cmd)
```
oesteban: why not `cmd.replace('%', '%%')`?

oesteban: Actually, this problem now will happen the same if the command has braces...

mgxd: good call - I would prefer to not alter the `cmd` if possible, what about substitution first?

oesteban: `cmd.replace('%', '%%')` is not an in-place operation, why not just do it here?

nipype/pipeline/engine/nodes.py
```diff
@@ -577,7 +577,9 @@ def _run_command(self, execute, copyfiles=True):
             self._originputs = deepcopy(self._interface.inputs)
             self._copyfiles_to_wd(execute=execute)
 
-        message = '[Node] Running "{}" ("{}.{}")'
+        message = '[Node] Running "{}" ("{}.{}")'.format(
```
oesteban: None of these variables should contain percent characters. Why not letting the log module do the work?

mgxd: Better to do the substitution first (before cmd is added), since a rouge `%` or `{}` will lead to a logging error

nipype/pipeline/engine/nodes.py
mgxd: sure, we can replace to cover against substitution (`%%`) / escape characters (`\", \\`, etc), but isn't `.format` cleaner?

oesteban: It is just to use python's logging standards, but I'm not particularly attached to one solution or another 👍

PR #2368, opened 2018-01-10
[FIX] MultiProc starting workers at dubious wd

oesteban:
When ``maxtasksperchild`` is set at a very low value (e.g. 1), workers are created very often, sometimes at working directories deleted after the interface cleanup. That would trigger an ``OSError`` when calling ``os.getcwd()`` during ``nipype.config`` import.

This PR sets an initializer for the workers that just changes to the appropriate working directory before the worker is spun up. Fixes poldracklab/fmriprep#868

The reason why one would be interested in ``maxtasksperchild=1`` is memory consumption. Using this option in addition to ``multiprocessing`` in ``'forkserver'`` mode ensures memory is freed after each interface has been run.


nipype/pipeline/plugins/multiproc.py
```diff
@@ -140,19 +151,28 @@ def __init__(self, plugin_args=None):
...
+                     self.processors, self.memory_gb, self._cwd)
 
         NipypePool = NonDaemonPool if non_daemon else Pool
         try:
-            self.pool = NipypePool(processes=self.processors,
-                                   maxtasksperchild=maxtasks)
+            self.pool = NipypePool(
+                processes=self.processors,
+                maxtasksperchild=maxtasks,
+                initializer=_init_worker,
```
effigies: Is there a reason this shouldn't just be `os.chdir`?

nipype/pipeline/plugins/multiproc.py
```diff
@@ -128,6 +135,10 @@ def __init__(self, plugin_args=None):
         self._task_obj = {}
         self._taskid = 0
 
+        # Cache current working directory and make sure we
+        # change to it when workers are set up
+        self._cwd = os.getcwd()
```
effigies: Is this the workflow base directory or the CWD of the shell? Could this cause things to dump into the local directory?

nipype/pipeline/plugins/multiproc.py
oesteban: not really

nipype/pipeline/plugins/multiproc.py
oesteban: It shouldn't because interfaces handle their WD. I think this is fixing an edge case for fmriprep where we are spinning up and killing workers all the time.

oesteban:
@satra I don't see potential side effects to this PR, but it'd be great if you took a peek through your window to the future since other humans we don't have it.

oesteban:
Hold off merging, I still need to refine a minor nit.

nipype/pipeline/plugins/multiproc.py
oesteban: Ok it turns out that some `SimpleInterface`s are writing to the workflow base directory :(

effigies: Well that's... suboptimal.

oesteban: It's actually bothering, because it means that without this patch, those interfaces are being run in some other unexpected path.

oesteban: Apparently it is a problem only with `SimpleInterface`

effigies:
@oesteban Still need to hold off?

oesteban:
No, once I identified that the problem is the `SimpleInterface` we will need to fix that. I have taken care of it in FMRIPREP proper meanwhile. So this is okay to go

PR #2376, opened 2018-01-17
[MAINT] Cleanup EngineBase

oesteban:
A quick cleanup of the `EngineBase` class:

* [x] Added new tests for this base class
* [x] Refactored `test_utils.py`, and moved some tests to `test_nodes.py` and `test_workflows.py`
* [x] Cleaned up tests:
  * [x] Avoid using lambdas when testing exceptions (not necessary anymore with pytest)
  * [x] Parametrize `test_workflows.py::test_outputs_removal_wf` adding `MultiProc` and removing for loops


nipype/pipeline/engine/base.py
```diff
@@ -18,17 +18,15 @@
...
 import re
 import numpy as np
-from ... import logging
+from future import standard_library
+
 from ...interfaces.base import DynamicTraitedSpec
 from ...utils.filemanip import loadpkl, savepkl
 
-logger = logging.getLogger('workflow')
+standard_library.install_aliases()
```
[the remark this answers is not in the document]
oesteban: The aliases would modify the behavior of imports, so I guess that is the reason. Anyways, I moved it because flake8 suggested so. I can roll back at any moment

nipype/pipeline/engine/base.py
```diff
@@ -47,35 +45,37 @@ def __init__(self, name=None, base_dir=None):
...
 
-    @property
-    def outputs(self):
-        raise NotImplementedError
+    @name.setter
+    def name(self, name):
+        if name and bool(re.match(r'^[\w_]+$', name)):
+            self._name = name
+        else:
+            raise ValueError('[Workflow|Node] name "%s" is not valid.' % name)
```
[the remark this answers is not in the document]
oesteban: I think we were supporting also dashes `-` so I'll add them.

oesteban:
Testing with FMRIPREP this seems to work correctly.

nipype/pipeline/engine/base.py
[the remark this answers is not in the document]
effigies: @mgxd @satra Are you familiar enough with `future.standard_library.install_aliases` to know whether there's a semantic change being made by reordering here?

You can also add a `# noqa` comment to pacify flake8, if you do return it.

nipype/pipeline/engine/base.py
```diff
@@ -18,18 +18,14 @@
                         absolute_import)
 from builtins import object
 
-from future import standard_library
-standard_library.install_aliases()
```
effigies: I meant not to bother removing or reordering, and just tolerate the PEP8 violation. Unless it's causing problems, let's not change this here.

nipype/pipeline/engine/base.py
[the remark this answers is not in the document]
djarecka: some comments on the order are here: https://pypi.python.org/pypi/future

All examples [here](http://python-future.org/imports.html) have it at the very beginning

nipype/pipeline/engine/base.py
oesteban: I've checked locally with Python 2 and this aliases are not necessary. Let's get rid of them whenever possible.

nipype/pipeline/engine/tests/test_nodes.py
```diff
@@ -0,0 +1,292 @@
...
+workflow.run()
+
+workflow.run()
+'''
+
+# Node
+
+
+def test_node_init():
+    with pytest.raises(Exception):
```
djarecka: we can change change `Exception` to `TypeError`

oesteban:
Good to go!

nipype/pipeline/engine/base.py
djarecka: @oesteban @effigies I'm not sure if I understand, does `standard_library.install_aliases()`  at the end of the import block (e.g. in node.py or workflow.py) changes anything?

effigies: I don't believe it does.

PR #2378, opened 2018-01-18
FIX: Collect jobids in a numpy 1.9 compatible manner

### LINKED ISSUE #2372: Multiproc plugin with numpy 1.9.0
### Summary
I am trying to use the Multiproc plugin but it does not send the tasks to the workers. I traced the reason for this to a numpy expression in _send_procs_to_workers():
`jobids = np.nonzero(~self.proc_done & (self.depidx.sum(0) == 0))[1]`
in https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/multiproc.py (line 216)

### Actual behavior

jobids is of type numpy.matrixlib.defmatrix.matrix with shape: (1,1) 

### Expected behavior

jobids should be of type numpy.ndarray with shape (1,)

I can achieve this behavior with:
`jobids = np.asarray(np.nonzero(~self.proc_done & (self.depidx.sum(0) == 0))[1]).flatten()`

Then everything is fine and works for all my pipelines.

I am using numpy version 1.9.0 which is still matching nipype's requirements, isn't it?

### Platform details:

{'nibabel_version': '2.1.0', 'sys_executable': '/usr/bin/python', 'networkx_version': '1.11', 'numpy_version': '1.9.0', 'sys_platform': 'linux2', 'sys_version': '2.7.9 (default, Jun 29 2016, 13:08:31) \n[GCC 4.9.2]', 'commit_source': 'repository', 'commit_hash': '8384523', 'pkg_path': '/home/klipsy1.psych.tu-dresden.de/foerster/nipype/nipype', 'nipype_version': '0.14.1-dev+g8384523', 'traits_version': '4.6.0', 'scipy_version': '0.14.0'}
0.14.1-dev+g8384523

effigies:
Fixes #2372.

- Return to a more explicit technique for collecting available job IDs, which does not depend on newer numpy

cc @divetea @oesteban @mgxd

### DIFF
--- a/nipype/pipeline/plugins/multiproc.py
+++ b/nipype/pipeline/plugins/multiproc.py
@@ -223,7 +223,9 @@ class MultiProcPlugin(DistributedPluginBase):
 
         # Check to see if a job is available (jobs with all dependencies run)
         # See https://github.com/nipy/nipype/pull/2200#discussion_r141605722
-        jobids = np.nonzero(~self.proc_done & (self.depidx.sum(0) == 0))[1]
+        # See also https://github.com/nipy/nipype/issues/2372
+        jobids = np.flatnonzero(~self.proc_done &
+                                (self.depidx.sum(axis=0) == 0).__array__())
 
         # Check available resources by summing all threads and memory used
         free_memory_gb, free_processors = self._check_resources(


effigies:
Requesting review. @oesteban would prefer your eyes, if you have a couple minutes.

oesteban:
LGTM, merging.

PR #2379, opened 2018-01-18
ENH: Add _cmd_prefix class variable to CommandLine

### LINKED ISSUE #1415: Prefixed names for interface functions
At least one of the many functions nipype has interfaces for - FSL's `cluster` - collides with another function from another package that shares its name. To make matters worse, this other package (graphviz) is a direct dependency of FSL (for the full description of the conflict see [this issue](https://github.com/gentoo-science/sci/issues/606)).

One of the most intuitive ways to fix this is by adding a prefix to the function which is called by fewer other packages. In this case that would be FSL's `cluster`. I have actually noticed that many sysadmins consistently add a prefix to all functions from FSL or AFNI.

So, should I as a package maintainer decide to prefix FSL's `cluster` as `fsl_cluster` on Gentoo - how could I make sure nipype is aware that it should call that when one invokes the `Cluster()` interface?

effigies:
This PR adds a `_cmd_prefix` which is a string that is prepended (with no space) to the command. This will permit users to add installation-specific prefixes, whether they be installation directories that are not in the path or idiosyncratic names that have been chosen by a package manager.

The idea is that users could do something like:

```Python
import nipype
nipype.interfaces.ants.AntsCommand._cmd_prefix = '/opt/ants/'
```

or

```Python
import nipype
nipype.interfaces.fsl.FSLCommand._cmd_prefix = 'fsl_'
```

Fixes #1415.

cc @satra @TheChymera

### DIFF
--- a/nipype/interfaces/afni/base.py
+++ b/nipype/interfaces/afni/base.py
@@ -294,13 +294,10 @@ class AFNIPythonCommandInputSpec(CommandLineInputSpec):
 class AFNIPythonCommand(AFNICommand):
     @property
     def cmd(self):
-        if spawn.find_executable(super(AFNIPythonCommand,
-                                       self).cmd) is not None:
-            return spawn.find_executable(super(AFNIPythonCommand, self).cmd)
-        else:
-            return super(AFNIPythonCommand, self).cmd
+        orig_cmd = super(AFNIPythonCommand, self).cmd
+        found = spawn.find_executable(orig_cmd)
+        return found if found is not None else orig_cmd
 
     @property
-    def cmdline(self):
-        return "{} {}".format(self.inputs.py27_path,
-                              super(AFNIPythonCommand, self).cmdline)
+    def _cmd_prefix(self):
+        return "{} ".format(self.inputs.py27_path)
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -988,13 +988,25 @@ class CommandLine(BaseInterface):
 
         # which $cmd
         executable_name = self.cmd.split()[0]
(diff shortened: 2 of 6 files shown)


effigies:
@satra If you want to have a look at this to make sure it's what you were thinking, that'd be great.

nipype/interfaces/afni/base.py
```diff
@@ -294,13 +294,10 @@ class AFNIPythonCommandInputSpec(CommandLineInputSpec):
 class AFNIPythonCommand(AFNICommand):
     @property
     def cmd(self):
-        if spawn.find_executable(super(AFNIPythonCommand,
-                                       self).cmd) is not None:
-            return spawn.find_executable(super(AFNIPythonCommand, self).cmd)
-        else:
-            return super(AFNIPythonCommand, self).cmd
+        orig_cmd = super(AFNIPythonCommand, self).cmd
+        found = spawn.find_executable(orig_cmd)
```
satra: i think `find_executable` may need to be updated unless we mandate that the prefix has to be a path prefix returning a fully formed path. 

(at least in terms of our present docs).

satra:
in general this looks good.

effigies:
@satra Testing found what I think you were thinking of, which was a `which` function. Can you have a quick look at this latest commit?

effigies:
@satra I opted to go with [`shlex.split(prefix + cmd)`](https://docs.python.org/3/library/shlex.html#shlex.split), which should be more reliable than any parsing strategy we choose.

effigies:
This is ready, as far as I'm concerned.

effigies:
@satra This is the only thing left to review for 1.0.

PR #2268, opened 2017-10-31
No longer concatenating extensions

TheChymera:
`graph.png` and `graph.dot` side by side is a lot better than `graph.dot.png` and `graph.dot` - for both tab completion, and software which has trouble intelligently parsing extensions (e.g. LaTeX's `\includegraphics`).

PR #2381, opened 2018-01-19
No longer concatenating extensions

TheChymera:
graph.png and graph.dot side by side is a lot better than graph.dot.png and graph.dot - for both tab completion, and software which has trouble intelligently parsing extensions (e.g. LaTeX's \includegraphics).

reiteration of #2268

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -1314,18 +1314,19 @@ def export_graph(graph_in,
 def format_dot(dotfilename, format='png'):
     """Dump a directed graph (Linux only; install via `brew` on OSX)"""
     if format != 'dot':
-        cmd = 'dot -T%s -O \'%s\'' % (format, dotfilename)
+        dot_base =  os.path.splitext(dotfilename)[0]
+        formatted_dot = '{}.{}'.format(dot_base, format)
+        cmd = 'dot -T{} -o"{}" "{}"'.format(format, formatted_dot, dotfilename)
         try:
             CommandLine(cmd, resource_monitor=False).run()
         except IOError as ioe:
             if "could not be found" in str(ioe):
-                raise IOError(
-                    "Cannot draw directed graph; executable 'dot' is unavailable"
-                )
+                raise IOError("Cannot draw directed graph; executable 'dot' is unavailable")
             else:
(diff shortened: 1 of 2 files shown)


nipype/pipeline/engine/utils.py
```diff
@@ -1314,18 +1314,17 @@ def export_graph(graph_in,
 def format_dot(dotfilename, format='png'):
     """Dump a directed graph (Linux only; install via `brew` on OSX)"""
     if format != 'dot':
-        cmd = 'dot -T%s -O \'%s\'' % (format, dotfilename)
+        dot_base =  os.path.splitext(dotfilename)[0]
+        formatted_dot = '{}.{}'.format(dot_base, format)
+        cmd = 'dot -T{} -o"{}" "{}"'.format(format, formatted_dot, dotfilename)
```
mgxd: I think this needs to be uppercase O

effigies: Returning `formatted_dot` breaks if `format == 'dot'`. One option is to revert the return line, and update these three lines to:

```Python
dot_base =  os.path.splitext(dotfilename)[0]
orig_dot = dotfilename
dotfilename = '{}.{}'.format(dot_base, format)
cmd = 'dot -T{} -o"{}" "{}"'.format(format, dotfilename, orig_dot)
```

TheChymera: nope, `-ofile` is what we should be using.

TheChymera: thanks for the advice, but I would recommend the current approach. Having `dotfilename` mean both the original and the formatted filename is confusing, but I would rather not construct it implicitly inline with some other operation for the sake of clarity.

If the filename for some reason can't be split, I'd much rather like the traceback to serve me up this line directly:

```
dot_base =  os.path.splitext(dotfilename)[0]
```

effigies: The issue is that you're returning a value from an unassigned name in the case where `format == 'dot'`. If you don't want to reassign a variable, you can return the `formatted_dot` from within the `if` block, or `return dotfilename if format == 'dot' else formatted_dot`, or whatever, but the tests will need to pass before we can merge this.

Issue #2380, opened 2018-01-19
TestSignalExtraction.test_signal_extr_traits_valid failing


djarecka:
### Summary

I just noticed that TestSignalExtraction.test_signal_extr_traits_valid is failing on my OSX with the message:
```
E           traits.trait_errors.TraitError: The trait 'out_file' of a SignalExtractionOutputSpec instance is an existing file name, but the path  '/private/var/folders/5s/hk2j7ylx54b2clm9bx1mpvj00000gq/T/tmp_2bektow/SignalExtraction/signals.tsv' does not exist.
```
I can see that the file `signals.tsv` is being created but is being deleted [here](https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/nodes.py#L630). 

@oesteban I'll try to figure out why this happens on my OSX and not on CircleCI (or other Linux we tried today), but if you have any suggestions, please let me know.

oesteban:
Looks a lot like poldracklab/fmriprep#936

oesteban:
@djarecka  could you try on you OSX the PR referenced above? Thanks!

djarecka:
it could be related to the temporary directory paths IMO.

i'm debugging it right now, and I can see that `needed file` [here](https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/utils.py#L1416) contains `signals.tsv` as `/private/var/folders/5s/hk2j7ylx54b2clm9bx1mpvj00000gq/T/tmpuuucmm6n/SignalExtraction/signals.tsv`
but file from `walks_files(cwd)` has `/var/folders/5s/hk2j7ylx54b2clm9bx1mpvj00000gq/T/tmpuuucmm6n/SignalExtraction/signals.tsv`.

djarecka:
just run with your `oesteban-maint/node-hash` and the test is still failing

djarecka:
@oesteban not sure if this helps, but only `signals.tsv` in `needed_files` has `/private` at the beginning, i.e.:
```
(Pdb) needed_files
[...]

PR #2384, opened 2018-01-19
FIX: Change to interface workdir within Interface.run() instead Node

oesteban:
Addresses #2380 and https://github.com/nipy/nipype/pull/2368#discussion_r160794434

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -33,7 +33,7 @@ from dateutil.parser import parse as parseutc
 
 from ... import config, logging, LooseVersion
 from ...utils.provenance import write_provenance
-from ...utils.misc import trim, str2bool
+from ...utils.misc import trim, str2bool, rgetcwd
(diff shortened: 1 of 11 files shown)


oesteban:
I've added many reviewers just to make sure we are not missing out any details.

nipype/interfaces/base/core.py
```diff
@@ -471,7 +480,8 @@ def run(self, **inputs):
             env['DISPLAY'] = config.get_display()
 
         runtime = Bunch(
-            cwd=os.getcwd(),
+            cwd=cwd,
+            prevcwd=syscwd,
```
djarecka: I couldn't find a place where this is being used

oesteban: Oh, forgot to add it to the report.rst

oesteban:
Removed the CHANGES entry:

```
* FIX: Change to interface workdir within Interface.run() instead Node (https://github.com/nipy/nipype/pull/2384)
```

PR #2202, opened 2017-10-02
ENH: generate Dockerfiles with neurodocker + migrate to CircleCI 2.0

kaczmarj:
Fixes #2144.

Changes proposed in this pull request:

Addresses (or will address) these items from the original issue:
> 1. generate dockerfiles using neurodocker
> 2. check if dockerfile hash has changed or force update is set (need to determine this from PR)
> 3. if force update, regenerate dockerfile
> 4. build docker image if needed or reload minimized image
> 5. run all tests via neurodocker reprozip (in parallel)
> 6. combine all reprozip packs into a single pack
> 7. generate minimized docker image using reprounzip and cache this image

Other changes:
- Migrate to CircleCI 2.0.

This is a work-in-progress. It would be great to have some feedback on what I have so far, namely dockerfile generation and the migration to circleci 2.0.

Parts of the code refer to my personal dockerhub repo (`kaczmarj/nipype`) but those will be changed when this PR is closer to complete. Also, the `master` neurodocker docker image is used now, but I will change this to `v0.3.2` once I release that version.

I am in the process of reworking neurodocker-reprozip's container-minimization functionality (ViDA-NYU/reprozip#274). I hope to finish that soon for this PR. I have already added a command to merge reprozip pack files (kaczmarj/neurodocker#69).

Just to note, I propose using `neurodebian:stretch-non-free` as the base image, instead of `ubuntu:xenial-20161213`.

Todo
-----
- [x] ~~generate dockerfiles using neurodocker~~
- [ ] check if dockerfile hash has changed or force update is set (need to determine this from PR)
    - checking is implemented, but forcing update is not.
- [ ] if force update, regenerate dockerfile
- [x] ~~build docker image if needed or reload minimized image~~
- [ ] run all tests via neurodocker reprozip (in parallel)
- [ ] combine all reprozip packs into a single pack
- [ ] generate minimized docker image using reprounzip and cache this image

PR #2386, opened 2018-01-20
TST/CI: Split Circle workflow

effigies:
Updating with todo list from comment thread:

* [x] Properly place codecov outputs, move codecov to its own entry in jobs (https://github.com/nipy/nipype/pull/2386#issuecomment-363870964)
* [x] Save coverage artifacts (https://github.com/nipy/nipype/pull/2386#issuecomment-363871713)
* [x] Save built docs as artifacts (https://github.com/nipy/nipype/pull/2386#issuecomment-363871713)
* [x] Test summaries (https://github.com/nipy/nipype/pull/2386#issuecomment-363872361)

----

#### Original text

This PR tracks an attempt to split out the testing process into a workflow of single-container jobs using the Circle 2.0 workflows. The goal is to reduce the amount of time containers are allocated but idle, and allow multiple PRs to do some simultaneous testing.

Follow up to #2202.

Comments and suggestions welcome. cc @kaczmarj


effigies:
This is working. There's a definite slowdown due to saving and loading images, so perhaps builds should be moved back into the test jobs. Still, it looks like we can achieve concurrency across PRs.

effigies:
From the [latest build](https://circleci.com/workflow-run/4f08c3bc-a38b-4500-bf67-de204ec74cbc):

17:58 to build images
15:14 to tarball images and persist to workspace

12:38 -12:47 to attach to workspace and load images

So for a cost of about 5 minutes per test job, we can skip a 30 minute bottleneck. Seems like a good trade-off.

.circleci/config.yml
```diff
@@ -120,31 +153,159 @@ jobs:
...
+            bash /home/circleci/nipype/.circleci/test_py3_fmri_spm_dartel_linear_l2.sh
+      - store_artifacts:
+          path: /home/circleci/work/tests
+
+  test_fmri_spm_nested_multiproc:
+    docker:
+      - image: docker:17.10.0-ce-git
+    working_directory: /home/circleci/nipype
+    steps:
+      - setup_remote_docker
```
oesteban: the remote docker might be useful when building, but it is not necessary for running. I may be wrong, but I think that remote docker does not allow mounting, and you need to do `docker cp`. The different jobs may be just using the local docker install. IMHO this line can be removed on the test jobs.

.circleci/config.yml
```diff
@@ -120,31 +153,159 @@ jobs:
...
+          command: |
+            docker load < /tmp/docker/nipype-base-latest-py36-py27.tar.gz
+      - checkout:
+          path: /home/circleci/nipype
+      - run:
+          name: Get codecov
+          command: |
+            pip install --no-cache-dir codecov
+      - run:
+          name: Download test data
```
oesteban: Maybe we can even split further and take this out as a different job?

.circleci/config.yml
```diff
@@ -120,31 +153,159 @@ jobs:
...
-          no_output_timeout: 4h
+          no_output_timeout: 2h
           environment:
             WORKDIR: /home/circleci/work
           command: |
             mkdir -p "$WORKDIR"
             chmod -R 777 "$WORKDIR"
-            bash /home/circleci/nipype/.circleci/tests.sh
+            bash /home/circleci/nipype/.circleci/pytests.sh
       - store_artifacts:
```
oesteban: I'd use `store_test_results` here, so we can have the test summary.

oesteban:
@kaczmarj could you have a quick look? Thanks!

effigies:
There are some issues with codecov not pointing to the right files. I think I want to move the codecov to its own entry in Circle, which would let us simplify the script structure, too.

I just haven't found much time to figure out where the coverage files are/should be.

satra:
also we should ensure that some of the artifacts are being saved, such as the built docs and the coverage files.

oesteban:
oh, one more: use set up the test summaries (https://circleci.com/docs/2.0/configuration-reference/#store_test_results) so that circle can show us number of tests and time in the `test_build` job

effigies:
Well, this branch is open to pushes from team members...

Anyway, I can try to find some time later this week, but I think it's going to take a couple periods of sustained attention, so no definite time frame.

kaczmarj:
@effigies @oesteban - I will look soon. This is an exciting pr!

kaczmarj:
Sorry for the delay. This looks great to me. One thing I would suggest is to reduce duplication in in `.circleci/config.yml` with yaml pointers. Common procedures can be defined once and referenced multiple times.

For example:

```yaml
_modify_nipype_version: &modify_nipype_version
  name: Modify Nipype version if necessary
  command: |
    if [ "$CIRCLE_TAG" != "" ]; then
      sed -i -E "s/(__version__ = )'[A-Za-z0-9.-]+'/\1'$CIRCLE_TAG'/" nipype/info.py
    fi
# ...
- run: *modify_nipype_version
# ...
- run: *modify_nipype_version
```

I can try to add this over the weekend. Otherwise this looks great!

effigies:
That is super handy to know. If you just want to ping this thread when
you're going to work on something, that'll keep me from doing the same
thing if I happen to have time to work on it at the same time. (I'll do the
same.)

kaczmarj:
@effigies - I am working on this now

effigies:
This is ready for review. Docs and coverage are in the artifacts. Codecov is pushing pull request numbers again, which should start showing up again once this hits master. And test summaries show up for the `test_pytest` entry.

satra:
thanks @effigies - this looks good.

PR #2387, opened 2018-01-21
changes in Interfaces base (closes #2320)

### LINKED ISSUE #2320: suggesting changes to `interfaces.base`

djarecka:
changes in interface base listed in #2320:

- [x] `_check_version_requirements` does not check the max_version, this block has wrong indentation

- [x] I was a bit confused about `BaseFile` and `File`, etc., do we really need both? Also I believe that we're assuming that `BaseUnicode` uses the fast validator (https://github.com/oesteban/nipype/blob/ref/interface-base/nipype/interfaces/base/traits_extension.py#L98), but from traits documentation I understood that `Unicode` (not `BaseUnicode`) uses the C-level fast validator (http://docs.enthought.com/traits/traits_api_reference/trait_types.html?highlight=baseunicode#traits.trait_types.BaseUnicode. Am I missing something?   

- [x]  couldn't find a place where we're using `_requires_warn` 

- [x]  the same about `_hash_infile` 

- [x] is `__pretty__` still working? 
is it for this package: http://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.pretty.html#extending ?
they suggest `_repr_pretty_`

- [x] [Check on this `synchronize`](https://github.com/nipy/nipype/pull/2325#discussion_r159650088)

- [x] [Check the logic of this](https://github.com/nipy/nipype/pull/2325#discussion_r159647953)


effigies:
Is this intended for 1.0, 1.0 maintenance, or 2.0?

djarecka:
@effigies this are for 1.0, but don't know if we have to wait to include everything before release. Was planning to do it last week, but didn't happen..

Notice, that these are independent points that were suggested by me or @satra during #2303 revision. If you or @satra find time to review it we can merge before finishing all points (I'm trying to keep them in separate commits) and I can open a new PR.
that would eliminate a bug in `_check_version_requirements`.

djarecka:
do you know anyone using `MpiCommandLine`? should I remove it?

effigies:
Okay, I can have a look.

As for MpiCommandLine, I would generally prefer not to remove functionality that someone may be using without a deprecation cycle. i.e. We should have had a few releases where we raised warnings that it was going to go away, which would give people time to complain or find another approach before we actually pull the rug out from under them. So unless there's a really compelling reason to remove it, I would leave it. If people think it's not worth supporting, we can deprecate it so it can be removed later.

djarecka:
@effigies ok, I removed `MpiCommandLine` from the list. 

Note that my last commits renamed ` MultiPath` (as suggested by @satra some time ago), and removed `BaseFile`/`BaseDirectory` (because they were not used and didn't differ from `File`/`Directory`). Hope that's ok, but if you think that would cause problems, I can revert

nipype/interfaces/base/traits_extension.py
```diff
@@ -54,7 +54,7 @@ class Str(Unicode):
 traits.DictStrStr = DictStrStr
 
 
-class BaseFile(BaseUnicode):
+class File(BaseUnicode):
```
effigies: I'm unclear on the original justification for this separation, but dropping `BaseFile` seems fine to me.

nipype/interfaces/base/specs.py
```diff
@@ -165,29 +151,6 @@ def _deprecated_warn(self, obj, name, old, new):
                             '%s' % trait_spec.new_name: new
                         })
 
-    def _hash_infile(self, adict, key):
```
effigies: Unused private method. Good to remove, as far as I'm concerned. Again, @satra, feel free to override if this looks like a mistake.

nipype/interfaces/base/support.py
```diff
@@ -168,10 +168,8 @@ def _get_bunch_hash(self):
         sorted_dict = to_str(sorted(dict_nofilename.items()))
         return dict_withhash, md5(sorted_dict.encode()).hexdigest()
 
-    def __pretty__(self, p, cycle):
-        """Support for the pretty module
-
-        pretty is included in ipython.externals for ipython > 0.10"""
+    def _repr_pretty_(self, p, cycle):
```
effigies: I'm okay with this, but we should verify that we can pretty print Bunches in IPython with this change. I don't know if the API changed beyond what method it looks for in your objects.

if it doesn't work quickly, I'd say remove the function altogether and create an issue to fix and re-add, rather than waste time on this today.

effigies:
@djarecka Review complete. I'll look at the max_ver issue locally. I don't see how your change can be fixing it, but if the test fails before and works after, I'll allow it.

djarecka:
@effigies ok, I'm just working actually on `Bunch`, but I can definitely set this as a new PR for future versions

djarecka:
@effigies re max_ver:
notice how [names](https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/core.py#L399) are defined. They should be redefined for max version check.

The best way to see  it doesn't work  is to see what is the `Exception` that is really being raised [here](https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/tests/test_core.py#L239). After [fixing it](https://github.com/nipy/nipype/pull/2387/files) the test will fail

nipype/interfaces/base/support.py
effigies: It looks like what it does is prevents an infinite loop during formating. Could you test with a bunch that contains itself?

djarecka: yes, you're right, for a bunch that contains itself it makes a difference! But only with the new name `_repr_pretty_`. 
And you don't actually have to import pretty in ipython.

djarecka:
@satra @effigies if I understood all your comments correctly this should be ready for final review and merging if all tests pass

PR #2389, opened 2018-01-22
rel: some housekeeping for 1.0

mgxd:
- updated zenodo/mailmap
- add missing autotests from `make specs`


effigies:
Looks good. We'll need a changelog update, but whether that's here or directly on master at release time is fine with me.

nipype/info.py
```diff
@@ -148,6 +148,7 @@ def get_nipype_gitversion():
     'pydotplus',
     'pydot>=%s' % PYDOT_MIN_VERSION,
     'packaging',
+    'yapf'
```
satra: `yapf` is not a runtime dependency. it has been added to a new label called specs below.

mgxd: ah didn't see this earlier - sorry for the clutter

effigies:
We may be good to go. Do we need to do any documentation updates before release? We'll need to establish processes for contributing to 1.x-maint vs 2.0-dev (or whatever we call the branches).

satra:
@effigies - till we get the engine in place (which should be highest priority for @djarecka in the next few weeks), we can go with current contribution process and do updates on the 1.x branch (leave master as is). as soon as the engine is ready and tested, we can create the 2.0 branch and start moving things over to the new engine.

satra:
@effigies - the only other thing i would do is to check and update the tutorial. i know @djarecka now has automated testing of the tutorial. it should be checked against master perhaps before 1.0 is checked off.

djarecka:
i can do it

satra:
use current master (10.12pm eastern or later)

djarecka:
unfortunately my tests are failing on graphs, it's likely not related to nipype itself, but I'll debug it tomorrow:
https://circleci.com/gh/djarecka/nipype_tutorial/34

djarecka:
@satra @effigies have any changes to `write_graph` where intended? Did we want to have a shorter name of files? After pep8 changes it's harder to use `blame` option in github...

`write_graph` is now inconsistent, if I call with `graph2use='orig'` I have `workflow.dot.png`, but if I use `graph2use='colored'` then I have file with a shorter name `workflow.png`.

effigies:
#2381 was meant to remove the `.dot`. I'll have a look.

effigies:
Good to merge and move tutorial discussion to #2397?

PR #2409, opened 2018-01-26
ENH: Re-enable spm.Realign to take lists of lists of files

### LINKED ISSUE #2406: With NiPype Memory with version 0.12.1 and >=0.13.0
Hi Nipype team,

### Summary
Trying to understand what is causing this difference between versions 0.12.1 and >=0.13.0. Is this with Memory or Realign. I am suspecting that this is the problem. But, I am sure I am missing some important details on this ?
 We are experiencing this in package pypreprocess which is reliant on Nipype. See full traceback in https://github.com/neurospin/pypreprocess/issues/288
### Actual behavior
```python
Nipype version:0.13.0
  File "report_nipype.py", line 11, in <module>
    print(realign)
  File "/home/kamalakar/miniconda2/envs/pypreprocess0_13/lib/python2.7/site-packages/nipype/caching/memory.py", line 98, in __repr__
    self.base_dir)
ValueError: Single '}' encountered in format string
```
### Expected behavior
```python
Nipype version:0.12.1
PipeFunc(nipype.interfaces.spm.preprocess.Realign, base_dir=/tmp/nipype_mem)
```
```python
import nipype
print("Nipype version:{0}".format(nipype.__version__))

import nipype.interfaces.spm as spm
from nipype.caching import Memory

mem = Memory(base_dir='/tmp/')
realign = mem.cache(spm.Realign)
print(realign)
```
Thanks a lot for your time.

effigies:
Fixes #2406. Essentially reverts #2375, but using `InputMultiPath` instead of `Either`.

- Re-enables spm.Realign to take lists of lists, using nested `InputMultiPath`s.

### DIFF
--- a/nipype/interfaces/base/traits_extension.py
+++ b/nipype/interfaces/base/traits_extension.py
@@ -322,15 +322,11 @@ class MultiPath(traits.List):
 
         newvalue = value
 
+        inner_trait = self.inner_traits()[0]
         if not isinstance(value, list) \
-            or (self.inner_traits() and
-                isinstance(self.inner_traits()[0].trait_type,
-                           traits.List) and not
-                isinstance(self.inner_traits()[0].trait_type,
-                           InputMultiPath) and
-                isinstance(value, list) and
-                value and not
-                isinstance(value[0], list)):
+            or (isinstance(inner_trait.trait_type, traits.List) and
+                not isinstance(inner_trait.trait_type, InputMultiPath) and
+                not isinstance(value[0], list)):
             newvalue = [value]
         value = super(MultiPath, self).validate(object, name, newvalue)
 
--- a/nipype/interfaces/base/traits_extension.py
+++ b/nipype/interfaces/base/traits_extension.py
@@ -214,25 +214,18 @@ class ImageFile(File):
         """
         self.types = types
         self.allow_compressed = allow_compressed
-        self._exts = None
         super(ImageFile, self).__init__(value, filter, auto_set, entries,
                                         exists, **metadata)
 
     def info(self):
-        existing='n existing' if self.exists else ''
+        existing = 'n existing' if self.exists else ''
         comma = ',' if self.exists and not self.allow_compressed else ''
-        uncompressed=' uncompressed' if not self.allow_compressed else ''
-        with_ext = ' (valid extensions: [{}])'.format(', '.join(self.exts)) \
-            if self.types else ''
+        uncompressed = ' uncompressed' if not self.allow_compressed else ''
+        with_ext = ' (valid extensions: [{}])'.format(
+            ', '.join(self.grab_exts())) if self.types else ''
         return 'a{existing}{comma}{uncompressed} file{with_ext}'.format(
             existing=existing, comma=comma, uncompressed=uncompressed,
             with_ext=with_ext)
-
-    @property
-    def exts(self):
-        if self.types and self._exts is None:
-            self._exts = self.grab_exts()
-        return self._exts
 
     def grab_exts(self):
         # TODO: file type validation
@@ -260,10 +253,11 @@ class ImageFile(File):
         """
         validated_value = super(ImageFile, self).validate(object, name, value)
         if validated_value and self.types:
-            if not any(validated_value.endswith(x) for x in self.exts):
+            _exts = self.grab_exts()
+            if not any(validated_value.endswith(x) for x in _exts):
                 raise TraitError(
                     args="{} is not included in allowed types: {}".format(
-                        validated_value, ', '.join(self.exts)))
+                        validated_value, ', '.join(_exts)))
         return validated_value
 
 
--- a/nipype/interfaces/base/traits_extension.py
+++ b/nipype/interfaces/base/traits_extension.py
@@ -214,8 +214,25 @@ class ImageFile(File):
         """
         self.types = types
         self.allow_compressed = allow_compressed
+        self._exts = None
         super(ImageFile, self).__init__(value, filter, auto_set, entries,
                                         exists, **metadata)
+
+    def info(self):
+        existing='n existing' if self.exists else ''
+        comma = ',' if self.exists and not self.allow_compressed else ''
+        uncompressed=' uncompressed' if not self.allow_compressed else ''
+        with_ext = ' (valid extensions: [{}])'.format(', '.join(self.exts)) \
+            if self.types else ''
+        return 'a{existing}{comma}{uncompressed} file{with_ext}'.format(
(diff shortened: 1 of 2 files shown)


effigies:
Before working on these tests, is there a semantic difference between `[['in_fileA.nii', 'in_fileB.nii']]` and `[['in_fileA.nii'], ['in_fileB.nii']]`, and which would be the correct interpretation of `['in_fileA.nii', 'in_fileB.nii']`?

effigies:
@mgxd @djarecka I'm undoing your consensus here, so I'd appreciate your input.

@satra Looks like you initially implemented support for a series of 3D files for Realign, so you might understand better than us the best way to go about this.

djarecka:
@effigies - i will take a look later today, but i'm still not sure how my changes could be responsible for issue with ver 0.13... but will read more about the issue.

satra:
this was the original formulation:

https://github.com/nipy/nipype/blob/8946bcab9d0e2f24e5364e42d4a7766e00237cb8/nipype/interfaces/spm/preprocess.py#L122

so as a first pass we could switch back to this. 

in terms of tests we would really have to check this against the matlab file that's created, so that would be a little involved.

satra:
there are a few things here:

1. the error message that comes from `InputMultiPath` is correct, but doesn't reflect the `ImageFileSPM` error, so that should be a fix in `InputMultiPath`, not the SPM interface.
2. SPM accepts a list of list of 3d nifti files where each inner list corresponds to a run. while most people use 4D files these days, the support from SPM should not be revoked.
3. nesting `InputMultiPath` would be fine, if that works properly. i'm not sure that works fine. someone should check.

effigies:
1. Sounds good.
2. Agreed.
3. I guess the failing test builds suggest that it's at least not working exactly as before, so it would likely break workflows, even if the behavior is still supported.

nipype/interfaces/base/traits_extension.py
```diff
@@ -322,15 +338,11 @@ def validate(self, object, name, value):
...
-                isinstance(self.inner_traits()[0].trait_type,
-                           traits.List) and not
-                isinstance(self.inner_traits()[0].trait_type,
-                           InputMultiPath) and
-                isinstance(value, list) and
-                value and not
-                isinstance(value[0], list)):
+            or (isinstance(inner_trait.trait_type, traits.List) and
+                not isinstance(inner_trait.trait_type, InputMultiPath) and
+                not isinstance(value[0], list)):
```
effigies: Just to clarify:

* `self.inner_traits()` is always `True` for a `traits.List` (default value: `(traits.Any(),)`)
* `isinstance(value, list)` is always `True` at this point in the short-circuit
* `bool(value)` is always True, or we would have returned `Undefined` already

These changes aren't really related to the rest of the PR, but I thought I would be modifying code in here at first, so factoring the `inner_trait` local variable led to this.

effigies:
New error message:

```Python
In [1]: from nipype.interfaces import spm

In [2]: realign = spm.Realign()

In [3]: realign.inputs.in_files = ['test.nii.gz', 'test.img']
TraitError                                Traceback (most recent call last)
<ipython-input-3-b45a253e48c2> in <module>()
----> 1 realign.inputs.in_files = [['test.nii.gz', 'test.img']]

/data/cjmarkie/Projects/nipype/nipype/interfaces/base/traits_extension.py in validate(self, object, name, value)
    345             newvalue = [value]
    346 
--> 347         value = super(MultiPath, self).validate(object, name, newvalue)
    348         # try:
    349         # except TraitError as excp:

~/.anaconda3/lib/python3.6/site-packages/traits/trait_types.py in validate(self, object, name, value)
   2334                 return value
   2335 
-> 2336             return TraitListObject( self, object, name, value )
   2337 
   2338         self.error( object, name, value )

~/.anaconda3/lib/python3.6/site-packages/traits/trait_handlers.py in __init__(self, trait, object, name, value)
   2311             except TraitError as excp:
   2312                 excp.set_prefix( 'Each element of the' )
-> 2313                 raise excp
   2314 
   2315         self.len_error( len( value ) )

~/.anaconda3/lib/python3.6/site-packages/traits/trait_handlers.py in __init__(self, trait, object, name, value)
   2303                 validate = trait.item_trait.handler.validate
   2304                 if validate is not None:
-> 2305                     value = [ validate( object, name, val ) for val in value ]
   2306 
   2307                 list.__setitem__(self, slice(0, 0), value )

~/.anaconda3/lib/python3.6/site-packages/traits/trait_handlers.py in <listcomp>(.0)
   2303                 validate = trait.item_trait.handler.validate
   2304                 if validate is not None:
-> 2305                     value = [ validate( object, name, val ) for val in value ]
   2306 
   2307                 list.__setitem__(self, slice(0, 0), value )

~/.anaconda3/lib/python3.6/site-packages/traits/trait_handlers.py in validate(self, object, name, value)
   1981             except TraitError:
   1982                pass
-> 1983         return self.slow_validate( object, name, value )
   1984 
   1985     def slow_validate ( self, object, name, value ):

~/.anaconda3/lib/python3.6/site-packages/traits/trait_handlers.py in slow_validate(self, object, name, value)
   1989             except TraitError:
   1990                pass
-> 1991         self.error( object, name, value )
   1992 
   1993     def full_info ( self, object, name, value ):

~/.anaconda3/lib/python3.6/site-packages/traits/trait_handlers.py in error(self, object, name, value)
    170         """
    171         raise TraitError( object, name, self.full_info( object, name, value ),
--> 172                           value )
    173 
    174     def full_info ( self, object, name, value ):

TraitError: Each element of the 'in_files' trait of a RealignInputSpec instance must be a list of
items which are an existing, uncompressed file (valid extensions: [.img, .nii, .hdr]) or an existing,
uncompressed file (valid extensions: [.img, .nii, .hdr]), but a value of 'test.nii.gz' <class 'str'> was
specified.
```

effigies:
Last change I think makes the grammar a little smoother.

```
TraitError: Each element of the 'in_files' trait of a RealignInputSpec instance must be an
existing, uncompressed file (valid extensions: [.hdr, .img, .nii]) or a list of items which
are an existing, uncompressed file (valid extensions: [.hdr, .img, .nii]), but a value of
'test.nii.gz' <class 'str'> was specified.
```

nipype/interfaces/base/traits_extension.py
djarecka: @effigies what if `value = []`?

djarecka: ok, sorry it would return undefined in l.331

djarecka:
lgtm

PR #2410, opened 2018-01-26
[ENH] Delay crashing if exception is raised in local hash check

oesteban:
This PR forces running a node if an Exception is raised when checking its hash. In that situation, a warning with the traceback is logged.


oesteban:
@satra, can you anticipate any issues derived from this PR?

PR #2418, opened 2018-01-30
FIX: argstr for mask in BlurToFWHM

mvdoc:
Fix argstr for mask input in BlurToFWHM

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -930,7 +930,7 @@ class BlurToFWHMInputSpec(AFNICommandInputSpec):
     mask = File(
         desc='Mask dataset, if desired. Voxels NOT in mask will be set to zero '
         'in output.',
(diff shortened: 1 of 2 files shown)


effigies:
This appears to be changing the API, since `-blurmaster` is a valid argument. It's possible someone depends on this misnaming.

Maybe a compromise is to deprecate `mask`, add `mask_file` to correspond to `-mask`, and a `blurmaster_file` to replace the old `mask`.

mvdoc:
I'm not sure I follow, because `-blurmaster` is another argument already present in the inputspec. Right now if somebody specifies both a blurmaster and a mask, it will fail, because it will be converted as `-blurmaster blurmaster.nii.gz -blurmaster mask.nii.gz`. Besides, `3dBlurToFWHM` would use the smoothness of the mask as target smoothness, and if it's a binary mask that wouldn't work.

PR #2425, opened 2018-02-03
fix: updated custom search to current api [skip ci]

satra:
the custom search was broken since google changed its api. this updates it to a new one.


satra:
this is not as great as i would like to see, but i think this at least works. currently search is broken. so i'm merging this and will refine after the current one is online for master.

PR #2426, opened 2018-02-03
ENH+FIX: Add 3dTproject AFNI interface, Fix OneDToolPy, Add -noFDR flag to 3dDeconvolve

mvdoc:
- Add `interfaces.afni.preprocess.TProject` wrapping `3dTproject`
- Fix `out_file` for `OneDToolPy` if `censor_motion` is set
- Add `-noFDR` flag to `3dDeconvolve`
- Change `jobs` to `num_threads` in `3dDeconvolve` for consistency with other interfaces


mvdoc:
@effigies @mgxd I'm trying to fix 3dDeconvolve in that if I specify a number of jobs, MultiProc still believe it needs only one proc, see example below (`-jobs` was set to 24)

```
[MultiProc] Running 1 tasks, and 0 jobs ready. Free memory (GB): 1360.18/1360.38, Free processors: 23/24.
```

I tried to set `num_threads` but it didn't work. How should I fix this?

Thanks

effigies:
You can set `n_procs=24` as a `Node()` argument. But perhaps it would be better to change the `jobs` input to `num_threads` to be consistent with other interfaces, since that is automatically synced with `node.n_procs`.

nipype/interfaces/afni/preprocess.py
```diff
@@ -2406,6 +2406,170 @@ class TNorm(AFNICommand):
...
+                ++ For catenated datasets, each run gets a separate set
+                   set of pp+1 Legendre polynomial regressors.
+                ++ Use of -polort -1 is not advised (if data mean != 0),
+                   even if -ort contains constant terms, as all means are
+                   removed.""",
+        argstr="-polort %d")
+    bandpass = traits.Tuple(
+        traits.Float, traits.Float,
+        desc="""Remove all frequencies EXCEPT those in the range""",
+        argstr='-bandpass %f %f')
```
effigies: Does it make sense to use `%g`, which will generally give you nicer formatting? e.g. `2` vs `2.0000` or `1.395` vs `1.3949999999`.

There are some other instances of `%f` in here. Your call whether to switch to `%g`.

mvdoc:
Thanks for the comments @effigies . Added some changes

nipype/interfaces/afni/model.py
```diff
@@ -153,7 +153,7 @@ class DeconvolveInputSpec(AFNICommandInputSpec):
         'instead of the bucket dataset, if possible.',
         argstr='-cbucket %s')
     out_file = File(desc='output statistics file', argstr='-bucket %s')
-    jobs = traits.Int(
+    num_threads = traits.Int(
         desc='run the program with provided number of sub-processes',
         argstr='-jobs %d')
```
effigies: Might want to set this to `nohash=True`, unless there's reason to think the number of threads affects the output.

satra:
@effigies - do you think the error we see is likely because of pybids changes, since we pull pybids from github master. i think i saw a reorganization of the dataset email somewhere.

'/home/travis/pybids/bids/grabbids/tests/data/ds005'

effigies:
Yes, that changed in the latest PRs. I'll push a fix in the morning if I
don't find time tonight.

satra:
@mvdoc - could you please merge with master and push? that should fix the travis error

mvdoc:
@satra done

Issue #2348, opened 2017-12-22
Remove dependency on distutils


oesteban:
distutils is used here and there and it would be really easy to remove this dependency.

PR #2441, opened 2018-02-14
[FIX/REF] Switch to new MRtrix3 `labelconvert` CL tool

jguillon:
Changes old MRtrix3 `labelconfig` to `labelconvert`.

http://mrtrix.readthedocs.io/en/latest/reference/commands/labelconvert.html

- Refactor class name to `LabelConvert`
- Remove now useless CL arguments


satra:
@jguillon - you don't need to include the actual freesurfer color LUT, since the testing is only to ensure a filename exists. you can keep the file, but empty out the contents. that will reduce the size of the repo.

satra:
@jguillon - i just emptied it out directly. will wait for at least the travis tests to pass and merge.

effigies:
@satra You may have already been planning this, but it woul be good to merge as a squashed commit, so the contents aren't sitting in the history. (Only a few kilobytes, but still. Nipype is already a heavy clone.)

satra:
@effigies - yes i was planning on that for this one.

effigies:
Quick question before merging: Is the old `labelconfig` tool no longer possible to use? Would it make sense to provide this as a new interface, rather than replacing the old one?

jguillon:
If, in general, `nipype` stays compatible with old versions of the wrapped softwares, yes it would make sense. But it could be complicated to respect this policy all the time... What about SPM8 ? And Freesurfer 5.X ?

satra:
nipype is fully compatible with SPM5/8 and FreeSurfer 5.X. we take care of this through versioning of interfaces/packages, and introducing fields that are dependent on `min_ver` and `max_ver` as noted here: http://nipype.readthedocs.io/en/latest/devel/interface_specs.html#common

jguillon:
Amazing! Then, let's keep the "old" `labelconfig` interface.  But do you have an equivalent of those `min_ver`, `max_ver` attributes but for an entire interface and not just a trait?

satra:
@jguillon - there are a few ways of doing whole interfaces:

1. overwrite `_run_interface` and/or add a version check in there
example: https://github.com/nipy/nipype/blob/db267032616ba05b5fd465491da6134e46041276/nipype/interfaces/fsl/model.py#L1208

2. create a new interface:
example: https://github.com/nipy/nipype/blob/db267032616ba05b5fd465491da6134e46041276/nipype/interfaces/spm/preprocess.py#L660

effigies:
@jguillon Have you had a chance to look at this further? Let us know if you have any more questions.

jguillon:
Hi again! Sorry I didn't have much time these last days.

I think I'll go with the simple solution of keeping the old interface of `LabelConvert` along with the new one. If that's fine with you?

As a side question, don't you ever check (and force) for the version of SPM when instantiating a SPM12-specific interface like this one:
https://github.com/nipy/nipype/blob/db267032616ba05b5fd465491da6134e46041276/nipype/interfaces/spm/preprocess.py#L785 
or did I miss something? Didn't test at all, just a curiosity.

effigies:
Just restoring the old one and keeping your new one alongside is definitely easiest. That works for me. (If we want to add in deprecating code, we can push to your branch before merging.)

nipype/interfaces/mrtrix3/connectivity.py
```diff
@@ -228,3 +228,91 @@ def _list_outputs(self):
...
+        position=-1,
+        usedefault=True,
+        desc='output file after processing')
+    spine = File(
+        argstr='-spine %s',
+        desc='provide a manually-defined '
+        'segmentation of the base of the spine where the streamlines'
+        ' terminate, so that this can become a node in the connection'
+        ' matrix.')
+    nthreads = traits.Int(
```
effigies: This should be `num_threads`, to help nipype automatically do resource_management.

nipype/interfaces/mrtrix3/connectivity.py
```diff
@@ -228,3 +228,91 @@ def _list_outputs(self):
...
+
+    Example
+    -------
+
+    >>> import nipype.interfaces.mrtrix3 as mrt
+    >>> labels = mrt.LabelConvert()
+    >>> labels.inputs.in_file = 'aparc+aseg.nii'
+    >>> labels.inputs.in_config = 'mrtrix3_labelconfig.txt'
+    >>> labels.inputs.in_lut = 'FreeSurferColorLUT.txt'
+    >>> labels.cmdline                               # doctest: +ELLIPSIS
```
effigies: No ellipsis in the `cmdline`, so I think we can drop the doctest directive here.

nipype/interfaces/mrtrix3/connectivity.py
```diff
@@ -228,3 +228,91 @@ def _list_outputs(self):
...
+    input_spec = LabelConvertInputSpec
+    output_spec = LabelConvertOutputSpec
+
+    def _parse_inputs(self, skip=None):
+        if skip is None:
+            skip = []
+
+        if not isdefined(self.inputs.in_config):
+            from distutils.spawn import find_executable
+            path = find_executable(self._cmd)
```
effigies: We're looking at dropping the `distutils` dependency (#2348). Can you use `nipype.utils.filemanip.which` instead?

PR #2450, opened 2018-02-18
Afni nwarpadjust

salma1601:
add interface for AFNI 3dNwarpAdjust


nipype/interfaces/afni/utils.py
```diff
@@ -1516,6 +1516,92 @@ def _list_outputs(self):
...
+    out_file = File(
+        desc='Output mean dataset, only needed if in_files are also given. '
+             'The output dataset will be on the common grid shared by the '
+             'source datasets.',
+        argstr='-prefix %s',
+        mandatory=False,
+        name_source='in_files',
+        name_template='%s_NwarpAdjust',
+        keep_extension=True,
+        xand=['in_files'])
```
satra: you can remove this since `in_files` is mandatory. also there is no `xand` in the current spec. so `make specs` would have returned a violation.

nipype/interfaces/afni/utils.py
```diff
@@ -1516,6 +1516,92 @@ def _list_outputs(self):
...
+        argstr='-source %s',
+        desc='List of input 3D datasets to be warped by the adjusted warp '
+             'datasets.  There must be exactly as many of these datasets as '
+             'there are input warps.')
+    out_file = File(
+        desc='Output mean dataset, only needed if in_files are also given. '
+             'The output dataset will be on the common grid shared by the '
+             'source datasets.',
+        argstr='-prefix %s',
+        mandatory=False,
```
satra: here and in the previous input field `mandatory=False` is not necessary.

nipype/interfaces/afni/utils.py
```diff
@@ -1516,6 +1516,92 @@ def _list_outputs(self):
...
+    >>> adjust.cmdline
+    '3dNwarpAdjust -nwarp func2anat_InverseWarp.nii.gz func2anat_InverseWarp.nii.gz func2anat_InverseWarp.nii.gz func2anat_InverseWarp.nii.gz func2anat_InverseWarp.nii.gz'
+    >>> res = adjust.run()  # doctest: +SKIP
+
+    """
+    _cmd = '3dNwarpAdjust'
+    input_spec = NwarpAdjustInputSpec
+    output_spec = AFNICommandOutputSpec
+
+    def _parse_inputs(self, skip=None):
```
satra: i think this function can be skipped as well.

nipype/interfaces/afni/utils.py
salma1601: actually I am trying to make `in_files` not mandatory, because what is needed here is only the warps

effigies: I believe what you want is `requires=['in_files']`. See https://nipype.readthedocs.io/en/latest/devel/interface_specs.html.

effigies:
Hi @salma1601 do you have some time to address @satra's comments? If you have any questions, please let us know.

salma1601:
@effigies when I remove the `_parse_inputs` function, the command line has -prefix <undefined> so I think it should be maintained no ?

effigies:
Please merge master to fix the tests.

I'm not very familiar with `_parse_inputs`, so I can't say without trying with and without your `_parse_inputs` method whether it's required. Have you tried removing it and verifying that your interface no longer works?

nipype/interfaces/afni/utils.py
```diff
@@ -1516,6 +1516,90 @@ def _list_outputs(self):
...
+        if not self.inputs.in_files:
+            if skip is None:
+                skip = []
+            skip += ['out_file']
+        return super(NwarpAdjust, self)._parse_inputs(skip=skip)
+
+    def _gen_filename(self, name):
```
mgxd: this probably isn't necessary either since you're calling `name_source` in the inputspec

nipype/interfaces/afni/utils.py
```diff
@@ -1516,6 +1516,90 @@ def _list_outputs(self):
...
+        if name == 'out_file':
+            return self._list_outputs()[name]
+        return None
+
+    def _list_outputs(self):
+        outputs = self.output_spec().get()
+
```
mgxd: perhaps this can be cleaned up to just

```python
...
if self.inputs.in_files:
    outputs['out_file'] = os.path.abspath(self.inputs.out_file)
return outputs
```

nipype/interfaces/afni/utils.py
```diff
@@ -1516,6 +1516,90 @@ def _list_outputs(self):
         return outputs
 
 
+class NwarpAdjustInputSpec(CommandLineInputSpec):
```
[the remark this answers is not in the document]
salma1601: done

nipype/interfaces/afni/utils.py
salma1601: done

nipype/interfaces/afni/utils.py
salma1601: it doesn't work when I don't give exactly how to build `outputs['out_file']` from `in_files`: the command line is correct but it fails at returning the outputs and I have an error

salma1601:
>I'm not very familiar with _parse_inputs, so I can't say without trying with and without your _parse_inputs method whether it's required. Have you tried removing it and verifying that your interface no longer works?

Yes I tried, the command line is wrong without _parse_inputs
`3dNwarpAdjust -prefix <undefined>_NwarpAdjust -nwarp /tmp/out_warp.nii.gz /tmp/out_warp.nii.gz /tmp/out_warp.nii.gz /tmp/out_warp.nii.gz /tmp/out_warp.nii.gz`

oesteban:
@satra I think @salma1601 just discovered a bug here: if a trait with `name_source` points to an undefined input, then it is replaced anyways. We should address that first.

satra:
@oesteban - yes we should track this, but since the `out_file` has `requires`, it should fail without `in_files` being present. the mandatory inputs check is passing, which it shouldn't or out_file shouldn't be generated.

oesteban:
maybe the ordering: filling the `name_template` is triggered before the requires?

oesteban:
Actually, the `requires` should be independent (and not used in this case: I should be able to overwrite the `out_file` field myself without providing `in_files`)

satra:
we should figure out the logic here. 

if requires is present then setting or using out_file to generate out_file needs in_files. 

otherwise the logic here is stating,

if out_file use out_file
if undefined(out_file) and in_files then use name_source
else (currently not defined)

perhaps in this scenario, there should be a default `out_file` (using `usedefault=True`) and no `namesource`.

effigies:
@satra @oesteban Is this something likely to be resolved this week? And should this hold up the current PR, or should we merge and create an issue for fixing this bug that's being worked around?

effigies:
Just a bump on this. My vote is to fix the `name_source` issues in a separate PR (probably next release). Any objections? @satra @oesteban

effigies:
Two thumbs up, so in it goes. I'll create an issue for 1.0.3.

PR #2454, opened 2018-02-20
CI/DOCKER: Hard code miniconda 4.3.31

effigies:
Travis builds on master are failing, apparently because of today's miniconda release, which bundled asn1crypto 0.24. Possibly rolling back to 0.23 will fix the issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,6 +183,28 @@ jobs:
       - run: *_run_codecov_coverage
       - store_artifacts: *store_artifacts_kwds
       - store_test_results: *store_artifacts_kwds
+      - run:
+          name: Build docs (py36)
+          no_output_timeout: 30m
+          environment: *test_environment
+          command: bash -ux /home/circleci/nipype/.circleci/test_py3_docs.sh
+      - store_artifacts:
+          path: /home/circleci/work/docs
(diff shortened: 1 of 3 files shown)


satra:
@effigies - perhaps it would be good to use wheels for all the installers. i think they all exist now. also i'm not sure if the icu issue is an issue still. may be worthwhile to synchronize this with the docker builds.

i think some of this comes from the time it takes to update conda forge packages.

effigies:
I'm not really sure what would be involved in such a change. If it's non-trivial, perhaps we should discuss Thursday. The 0.23 install had the same issue, so I've tried reverting to the most recent release of Miniconda. Assuming that passes, is that an acceptable stop-gap until then?

effigies:
(It looks to be passing.)

effigies:
Actually, it looks like the new Miniconda is breaking the Circle builds, too. They're retrying building the Docker images. I'm going to hard code the old one there, too.

This might be worth escalating, but I'm not sure where the best place to make noise would be.

effigies:
Okay, going to merge this to get tests passing again. I'll open a PR that reverts the hard-coding, so we can see if miniconda gets an update that fixes this, or move on to Satra's wheel installer approach.

PR #2455, opened 2018-02-21
CI: Update Travis builds, Docker to use latest miniconda

effigies:
Tracks fixes to miniconda. Tests should be periodically rerun, to see if Continuum is fixing things on their end.

Follow-up to #2454.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_install:
   fi;
 
 # handle python operations separately to reduce timeouts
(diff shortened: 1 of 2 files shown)


effigies:
Travis is fixed since 4.4.11 came out, but Neurodocker doesn't currently support updating conda, except through a specific command. It seems to make more sense to update Neurodocker, so I've opened kaczmarj/neurodocker#144.

Issue #2456, opened 2018-02-22
In CompCor, extraction of unit information from image header fails


eort:
### Summary
Hi, 
When using fmriprep (1.0.6), I bumped into an error in the confound workflow. For both aCompCor as well as tCompCor the script failed when trying to extract TR information from the header of the input image. After having a look at the part of the code that threw the error, I think the problem is that nipype uses nibabel's `get_xyzt_units()`to directly extract the information from the image, rather than the header, so that the script crashes with an `AttributeError`. 

### Actual behavior
The script excepts with an `ValueError`, caused by an `AttributeError` when running the block:
              
```
         try:
                    TR = imgseries.header.get_zooms()[3]
                    if imgseries.get_xyzt_units()[1] == 'msec':
                        TR /= 1000
                except (AttributeError, IndexError):
                    TR = 0

                if TR == 0:
                    raise ValueError(
                        '{} cannot detect repetition time from image - '
                          'Set the repetition_time input'.format(self._header))
```

### Expected behavior
The line `imgseries.get_xyzt_units()[1] == 'msec'` should evaluate to true or false

Please put URL to code or code here (if not too long).
https://github.com/nipy/nipype/blob/master/nipype/algorithms/confounds.py#L535

### Platform details:
I ran that code from within fmriprep v1.0.6

I'll send a pull request with a suggested fix.

effigies:
Good catch, thanks. Opened #2457.

effigies:
Oh, if you're going to submit a PR, feel free. (I didn't read that far before making my edit.)

eort:
And I didn't read yours in time, now there are 2 PRs...

effigies:
No worries. I closed mine in favor of yours.

PR #2458, opened 2018-02-22
Fix: Extract unit information from image header in CompCor

eort:
Fixes [#2456](https://github.com/nipy/nipype/issues/2456) .

- extract unit information from image header, rather than from image directly


effigies:
LGTM

Issue #2436, opened 2018-02-12
Matlab command cannot be found


jykim24:
### Summary
Hello, I'm new to nipype and I am following preprocessing tutorial in the NypypeBeginnersGuide, and I cannot seem to run the workflow or write a graph.
The workflow unzips the file, but crashes at slice timing.
The crash file created states that matlab command cannot be found, but I have matlab 2016b and SPM12 installed. 
Also, the error tells that 'No command 'dot' found' when I try to write a graph. 

When testing nipype with nipype.test(), a fauilure occurs in test_Commandline().
I've deleted and reinstalled nipype twice, but this problem occurred both times.

How can I resolve these issues?

### Actual behavior
When trying to write a graph,
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-9-026c77eb0f4a> in <module>()
      1 # 7. run workflow
----> 2 pp.write_graph(graph2use='flat')

~/anaconda3/lib/python3.6/site-packages/nipype/pipeline/engine/workflows.py in write_graph(self, dotfilename, graph2use, format, simple_form)
    448                 dotfilename=dotfilename,
    449                 format=format,
--> 450                 simple_form=simple_form)
    451 
    452         logger.info(

~/anaconda3/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py in export_graph(graph_in, base_dir, show, use_execgraph, show_connectinfo, dotfilename, format, simple_form)
   1285 
   1286     # Convert .dot if format != 'dot'
-> 1287     outfname, res = _run_dot(out_dot, format_ext=format)
   1288     if res is not None and res.runtime.returncode:
   1289         logger.warning('dot2png: %s', res.runtime.stderr)

~/anaconda3/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py in _run_dot(dotfilename, format_ext)
   1328     cmd = 'dot -T{} -o"{}" "{}"'.format(format_ext, formatted_dot, dotfilename)
   1329     res = CommandLine(cmd, terminal_output='allatonce',
-> 1330                       resource_monitor=False).run()
   1331     return formatted_dot, res
   1332 

~/anaconda3/lib/python3.6/site-packages/nipype/interfaces/base/core.py in run(self, cwd, **inputs)
    513 
    514         try:
--> 515             runtime = self._run_interface(runtime)
    516             outputs = self.aggregate_outputs(runtime)
    517         except Exception as e:

[...]

PR #2469, opened 2018-02-28
TEST: Parallelize pytest

effigies:
Was curious how much parallelizing could squeeze down pytest runtime, acknowledging that it's already not super long (~10 minutes out of a 40 minute build and test cycle).

This led down a doctest-shaped rabbit hole, because some workers were running the module-level doctests that changed directories and others were running the function-level doctests, and the context wasn't being shared so there were a couple stochastic crashes each build. So I figured out how to set a fixture that would `chdir` into the `testing/data` directory on doctests and only doctests. This seems cleaner than the old way of doing it anyway.

- Add `pytest-xdist` to testing dependencies
- Add `-n auto` to pytest config
- Add `in_testing` fixture to `chdir` for doctests
- Remove old header `chdir` boilerplate

### DIFF
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -25,19 +25,19 @@ logging = Logging(config)
 
 
 class NipypeTester(object):
-    def __call__(self, doctests=True):
+    def __call__(self, doctests=True, parallel=True):
         try:
             import pytest
         except:
             raise RuntimeError(
                 'py.test not installed, run: pip install pytest')
-        params = {'args': []}
-        if doctests:
-            params['args'].append('--doctest-modules')
-        nipype_path = os.path.dirname(__file__)
-        params['args'].extend(
-            ['-x', '--ignore={}/external'.format(nipype_path), nipype_path])
-        pytest.main(**params)
+        args = []
+        if not doctests:
+            args.extend(['-p', 'no:doctest'])
+        if not parallel:
+            args.append('-n0')
+        args.append(os.path.dirname(__file__))
+        pytest.main(args=args)
 
 
 test = NipypeTester()
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -3,13 +3,6 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 '''
 Algorithms to compute confounds in :abbr:`fMRI (functional MRI)`
-
-    Change directory to provide relative paths for doctests
-    >>> import os
-    >>> filepath = os.path.dirname(os.path.realpath(__file__))
-    >>> datadir = os.path.realpath(os.path.join(filepath, '../testing/data'))
-    >>> os.chdir(datadir)
-
 '''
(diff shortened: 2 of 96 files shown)


oesteban:
On my experience those two cores will perform as just one....

effigies:
Well, looks like we're dropping from 6:30 to about 4:15, so about a 33% speedup, which isn't too bad for 2 cores.

djarecka:
I think `nipype.test()` won't work.

also, one of the reason why I didn't remove imports etc. from files to auto-fixture was that i thought it will be harder for users to follow these examples. Not sure, how important it is..

effigies:
> I think `nipype.test()` won't work.

Yeah, I get one failure, but I'm not sure if that's related. I've never tried testing that way. I'll check on master.

> also, one of the reason why I didn't remove imports etc. from files to auto-fixture was that i thought it will be harder for users to follow these examples. Not sure, how important it is..

Doesn't seem super useful to me, since those docstrings are never rendered in the published docs. But I'm okay being overruled here.

satra:
the docstrings are meant to guide individuals - they are not meant to execute a real computation. so i'm +1 for fixture as long as it doesn't get in the way of the docstrings (which at least in this PR it doesn't look like).

djarecka:
@effigies - i also don't usually test like that, but @satra always wanted to include this option, so users can easily test `nipype`. I don't remember right now (would have to read again and check my previous PR), but it is related how pytest searching for root directory etc. Never fully understood all details..

regarding docstrings, I just wanted explain my point, I thought that they should be design in a way that users can search the file, copy-paste and try to execute

effigies:
Okay, the failure introduced by this in `nipype.test()` is in the following test:

https://github.com/nipy/nipype/blob/6ca791d9c5ec0efb9f56cc9e44758d6e53ffb800/nipype/pipeline/plugins/tests/test_callback.py#L49-L67

The failing line is:

djarecka:
@effigies sorry, I think this time it's not related to directory structure, you didn't change anything with it, but doesn't like the flag `-n auto`. do you need it?

djarecka:
@satra ok

djarecka:
(almost) all tests are passing on my OSX when I removed `-n auto` from `pytest.ini`

* almost because I've recreated the error in `test_core.py` mentioned in  #2436

effigies:
Oh, I see. We were overwriting some of what was in the `pytest.ini`. For consistency, I dropped most of that to run the same tests inside as outside, but added arguments for disabling both doctests and parallelism.

The 3.4 and 3.5 issue that's showing up on Travis is that dictionaries don't have guaranteed order pre-3.6. See pytest-dev/pytest#1075. Someone there suggests adding `pytest-env`, but I'd rather not add yet another dependency if I don't have to.

effigies:
Went ahead and used `pytest-env` because I couldn't come up with anything better without wasting more effort than avoiding it was worth. This is ready for review.

PR #2471, opened 2018-02-28
ENH: Add interpolation order parameter to NiftyReg's RegTools

### LINKED ISSUE #2470: NiftyReg's RegTools is missing interpolation order argument

fepegar:
Fixes #2470

Note: I ran `make check-before-commit` and got the following:
```
(nipype)  nipype   add-interp-to-niftyreg-regtools  make check-before-commit                                                                                           ◂◂◂◂◂◃◃◃◃◃
env PYTHONPATH=".:" python tools/checkspecs.py
  File "tools/checkspecs.py", line 15, in <module>
    from yapf.yapflib.yapf_api import FormatCode
ModuleNotFoundError: No module named 'yapf'
make: *** [specs] Error 1
```
I fixed it with `pip install yapf`.

### DIFF
--- a/nipype/interfaces/niftyreg/regutils.py
+++ b/nipype/interfaces/niftyreg/regutils.py
@@ -122,7 +122,7 @@ warpfield.nii -res im2_res.nii.gz'
     # Need this overload to properly constraint the interpolation type input
     def _format_arg(self, name, spec, value):
         if name == 'inter_val':
-            inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 5}
+            inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 4}
             return spec.argstr % inter_val[value]
         else:
             return super(RegResample, self)._format_arg(name, spec, value)
@@ -338,7 +338,7 @@ class RegTools(NiftyRegCommand):
     # Need this overload to properly constraint the interpolation type input
     def _format_arg(self, name, spec, value):
         if name == 'inter_val':
-            inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 5}
+            inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 4}
             return spec.argstr % inter_val[value]
         else:
             return super(RegTools, self)._format_arg(name, spec, value)
--- a/nipype/interfaces/niftyreg/regutils.py
+++ b/nipype/interfaces/niftyreg/regutils.py
@@ -295,6 +295,15 @@ class RegToolsInputSpec(NiftyRegCommandInputSpec):
         desc=desc,
         argstr='-smoG %f %f %f')
 
+    # Interpolation type
+    inter_val = traits.Enum(
+        'NN',
+        'LIN',
+        'CUB',
+        'SINC',
+        desc='Interpolation order to use to warp the floating image',
+        argstr='-interp %d')
+
 
 class RegToolsOutputSpec(TraitedSpec):
     """ Output Spec for RegTools. """
--- a/nipype/interfaces/niftyreg/regutils.py
+++ b/nipype/interfaces/niftyreg/regutils.py
@@ -335,6 +335,14 @@ class RegTools(NiftyRegCommand):
     output_spec = RegToolsOutputSpec
     _suffix = '_tools'
 
+    # Need this overload to properly constraint the interpolation type input
+    def _format_arg(self, name, spec, value):
+        if name == 'inter_val':
+            inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 5}
+            return spec.argstr % inter_val[value]
+        else:
+            return super(RegTools, self)._format_arg(name, spec, value)
+
 
 class RegAverageInputSpec(NiftyRegCommandInputSpec):
     """ Input Spec for RegAverage. """
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -558,7 +558,7 @@
       "name": "Flandin, Guillaume"
     },
     {
-      "affiliation": "Stereotaxy Core, Brain & Spine Institute",
+      "affiliation": "University College London",
       "name": "P\u00e9rez-Garc\u00eda, Fernando",
       "orcid": "0000-0001-9090-3024"
     },
--- a/nipype/interfaces/niftyreg/tests/test_auto_RegTools.py
+++ b/nipype/interfaces/niftyreg/tests/test_auto_RegTools.py
@@ -24,6 +24,7 @@ def test_RegTools_inputs():
             argstr='-in %s',
             mandatory=True,
         ),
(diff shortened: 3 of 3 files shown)


nipype/interfaces/niftyreg/regutils.py
```diff
@@ -295,6 +295,15 @@ class RegToolsInputSpec(NiftyRegCommandInputSpec):
...
         argstr='-smoG %f %f %f')
 
+    # Interpolation type
+    inter_val = traits.Enum(
+        'NN',
+        'LIN',
+        'CUB',
+        'SINC',
+        desc='Interpolation order to use to warp the floating image',
+        argstr='-interp %d')
```
mgxd: you'll want to change this to `%s`

fepegar: Thanks for the feedback. I've checked the source code and this parameter (of the command line tool) expects `0`, `1`, `3` or `4`, as in `RegResample`:
https://github.com/nipy/nipype/blob/045b28ef9056fac1107bc4f0707859d043f3bfd1/nipype/interfaces/niftyreg/regutils.py#L61-L68

So I basically copied that section from `RegResample`. Should anything be modified for both interfaces? I don't know much about traits, but it feels like the current code can only produce `0`, `1`, `2` or `3`.

Also, I'm a bit new to PRs. To change something, should I just commit in my local fork and push to the same branch normally?

effigies: You'll need to copy this part as well:

https://github.com/nipy/nipype/blob/045b28ef9056fac1107bc4f0707859d043f3bfd1/nipype/interfaces/niftyreg/regutils.py#L123-L128

fepegar: I see. Although shouldn't line 124 be
`inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 4}`
instead of
`inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 5}`
?

mgxd: I'm not sure why 2 and 4 are skipped, perhaps @mmodat has some input.

```python
# If change is needed, it should look something like this
inter_val = {'NN': 0, 'LIN': 1, 'CUB': 2, 'SINC': 3}
```

effigies: Looks like it. https://cmiclab.cs.ucl.ac.uk/mmodat/niftyreg/blob/master/reg-apps/reg_resample.cpp#L65

Their documentation could certainly be clearer on the point.

nipype/interfaces/niftyreg/regutils.py
```diff
@@ -326,6 +335,14 @@ class RegTools(NiftyRegCommand):
     output_spec = RegToolsOutputSpec
     _suffix = '_tools'
 
+    # Need this overload to properly constraint the interpolation type input
+    def _format_arg(self, name, spec, value):
+        if name == 'inter_val':
+            inter_val = {'NN': 0, 'LIN': 1, 'CUB': 3, 'SINC': 5}
```
effigies: Can you go ahead and change `5` to `4` here and in `RegResample`?

fepegar:
I've added what @effigies [commented](https://github.com/nipy/nipype/pull/2471#discussion_r171356843). Waiting for @mmodat to see if Enum values need to be modified.

nipype/interfaces/niftyreg/regutils.py
fepegar: Sure

effigies:
Can you `make specs` and add the auto test?

fepegar:
@effigies I don't know what those are. Could you please point me to an example or reference?

effigies:
In the source directory, run: `make specs`. This should update the following files:

* [nipype/interfaces/niftyreg/tests/test_auto_RegTools.py](https://github.com/nipy/nipype/blob/master/nipype/interfaces/niftyreg/tests/test_auto_RegTools.py)
* [nipype/interfaces/niftyreg/tests/test_auto_RegResample.py](https://github.com/nipy/nipype/blob/master/nipype/interfaces/niftyreg/tests/test_auto_RegResample.py)

Add and commit those.

fepegar:
Done, but only [test_auto_RegTools.py](https://github.com/nipy/nipype/blob/master/nipype/interfaces/niftyreg/tests/test_auto_RegTools.py) was updated.

Btw `make specs` worked only on Python 2.

effigies:
Yup, you're right. We didn't change the spec on the other one.

`make specs` should work on both (#1813). What issue did you have with Python 3?

fepegar:
```
(nipype)  nipype   add-interp-to-niftyreg-regtools  make specs                                                                                                         ◂◂◂◂◃◃◃◃◃◃
env PYTHONPATH=".:" python tools/checkspecs.py
  File "tools/checkspecs.py", line 14, in <module>
    from nipype.interfaces.base import BaseInterface
  File "/Users/fernando/git/nipype/nipype/__init__.py", line 12, in <module>
    from .utils.config import NipypeConfig
  File "/Users/fernando/git/nipype/nipype/utils/__init__.py", line 4, in <module>
    from .config import NUMPY_MMAP
  File "/Users/fernando/git/nipype/nipype/utils/config.py", line 25, in <module>
    from future import standard_library
ModuleNotFoundError: No module named 'future'
make: *** [specs] Error 1
```

effigies:
Ah. That can be resolved with `pip3 install future`.

fepegar:
In that case, shouldn't `[yapf](https://github.com/nipy/nipype/pull/2471#issue-172036419)` and `[future](https://github.com/nipy/nipype/pull/2471#issuecomment-369424859)` be included the requirements?

satra:
@fepegar - this https://github.com/nipy/nipype/blob/master/nipype/info.py#L134 clarifies how bits and pieces are installed. the basic install is intended for running, not developing nipype.

`pip install nipype` should always install `future` since that is a requirement (if you do python setup.py install, then it depends on your version of setuptools, how the dependencies are installed).

`pip install nipype[tests]` should install some additional bits for tests
`pip install nipype[specs]` should install all the pieces required for `make specs`.
`pip install nipype[all]` will install everything

we hope to clarify user/developer documentation on testing in the next update (end of march).

fepegar:
I'd expect anything needed by `make specs` to be installed when I run [`python setup.py develop`](https://github.com/nipy/nipype/blob/master/CONTRIBUTING.md#making-a-change).

effigies:
We should probably update that recommendation. Everything I've read for a few years has recommended using `pip` over running `setup.py` directly.

satra:
@fepegar - unfortunately that's not how `python setup.py develop` works. there is no standard way to install the dev modules.

@effigies  - we can create a `dev` option for extras-requires and then update docs to do: `pip install -e .[dev]` . perhaps `dev` should include `tests`, `specs`, and `doc` building, but no external dependencies - nipy/dipy.

PR #2495, opened 2018-03-11
FIX: mark .niml.dset as special extension in utils.filemanip

mvdoc:
All AFNI commands work also on files with extension `.niml.dset` (for surfaces).

- consider `*.niml.dset` as special extension to correctly split filenames


oesteban:
LGTM

PR #2496, opened 2018-03-13
FIX: Some install_requires are actually only setup_requires

ashgillman:
`pytest` and `mock` were previously listed under `install_requires`. This meant that if creating a packages that depends on `nipype`, you need to have these dependencies installed, but really they are only used during building, testing and installing (I believe, a quick grep looks like it).

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -143,11 +143,14 @@ REQUIRES = [
     'prov==%s' % PROV_VERSION,
     'click>=%s' % CLICK_MIN_VERSION,
     'funcsigs',
-    'pytest>=%s' % PYTEST_MIN_VERSION,
-    'mock',
     'pydotplus',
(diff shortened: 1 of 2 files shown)


oesteban:
they are used only for testing, are we using them in building and installing? could you point me to that?

satra:
@ashgillman - the reason we have those pieces is because anyone currently can do:

```python
import nipype
nipype.test()
```

i feel that having this available is good in case one wants to test a particular install.

ashgillman:
Ah I see. I noticed this because I maintain the Nipype inclusion in the Nix package manager, https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/nipype/default.nix

I listed pytest and mock as build-only requirements (this is usually the case), but I can change that :)

Issue #2216, opened 2017-10-06
AFNI Allineate not generating prefix argument


ostanley:
### Summary
When in a pipeline allineate does not generate filenames. This is using a version after the allineate fix.

### Actual behaviour
The allineate call now fails with a runtime error.

### Expected behaviour
Allineate to generate an output file name

### How to replicate the behaviour
Run a script with an input 1D file and an input file to motion correct

### Script/Workflow details
    moco = pe.MapNode(interface=afni.Allineate(), name='moco', iterfield=['in_file', 'in_matrix'])
    moco.inputs.outputtype = 'NIFTI_GZ'
    preprocphase.connect([(inputspec, img2float, [('input_phase', 'in_file')]),
                          (img2float, prepphase, [('out_file', 'phase')]),
                          (inputspec, moco, [('motion_par', 'in_matrix')]),
                          (prepphase, moco, [('detrended_phase', 'in_file')]),
                          (inputspec, applyalign, [('run_txfm','in_matrix_file')]),
                          (moco, applyalign, [('out_file', 'in_file')]),
                          (applyalign, maskfunc, [('out_file', 'in_file')]),
                          (inputspec, maskfunc, [('mask_file', 'in_file2')]),
                          (maskfunc, outputspec,[('out_file', 'proc_phase')])
                          ])

Please put URL to code or code here (if not too long).

### Platform details:

{'nibabel_version': '2.1.0', 'sys_executable': '/cfmm/home/ostanle2/.virtualenvs/nipype2/bin/python', 'networkx_version': '1.9.1', 'numpy_version': '1.12.0', 'sys_platform': 'linux2', 'sys_version': '2.7.10 (default, Oct 28 2015, 11:22:11) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-16)]', 'commit_source': u'installation', 'commit_hash': u'4d08138', 'pkg_path': '/cfmm/home/ostanle2/.virtualenvs/nipype2/lib/python2.7/site-packages/nipype-1.0.0.dev0+g4d08138-py2.7.egg/nipype', 'nipype_version': u'1.0.0-dev', 'traits_version': '4.6.0', 'scipy_version': '0.18.1'}
1.0.0-dev
171005-22:46:15,712 interface INFO:
	 Closing display (if virtual)

### Execution environment

- My python environment outside container

effigies:
Can you share the text of the error?

ostanley:
[...]

PR #2502, opened 2018-03-20
[FIX] fix afni.allineate interface

### LINKED ISSUE #2499: issues with afni.allineate w/ solution (and possible issue with genfile parameter)
###  Summary

Hi all. Just started learning nipype, and it's been pretty great so far. I've only run into issues with the `afni.allineate` interface. I was able to fix it on my own, but I thought I should post an issue here for awareness (plus I'm not sure how to do a pull request...)

There are 2 problems I've encountered. The first is that `afni.allineate` consistently fails the hash check when placed in a workflow and always reruns. I think this is because the `hash_files` parameter needs to be set to `False`.

```
out_file = File(
    desc='output file from 3dAllineate',
    argstr='-prefix %s',
    genfile=True,
    hash_files=False, # <-- when this is added it starts working again
    xor=['allcostx']) 
```
Referencing the docs, I think this is an appropriate change:
> `hash_files`
To be used with inputs that are defining output filenames. When this flag is set to false any Nipype will not try to hash any files described by this input. This is useful to avoid rerunning when the specified output file already exists and has changed.

The second issue is that the prefix (`out_file`) argument does not get generated when used in a workflow. This issue is also referenced here: #2216, but with a temporary workaround.

My understanding is that when `genfile` is set to `True`, the `_gen_filename()` method for the interface gets called when the user does not specify a value. Looking at the code, the `_gen_filename()` does seems to implemented, but not correctly:

```
def _gen_filename(self, name): # this gets called when `out_file` is not defined
    if name == 'out_file':
        return self._list_outputs()[name] # <-- we are returning the value of `out_file`, which is undefined...
    return None
```

So I changed it to this:

```
def _gen_filename(self, name): # this gets called when `out_file` is not defined
    if name == 'out_file':
        return self._gen_fname(self.inputs.in_file,op.dirname(self.inputs.in_file),suffix='_allineate')
    return None
```

Which should define the `out_file` to be in the same location and have the same name as the `in_file`, but with the '_allineate' suffix.

When I run this, the problem seems to be fixed, as the prefix is now defined:

```
[Node] Running "3dallineate_orig" ("nipype.interfaces.afni.preprocess.Allineate"), a CommandLine Interface with command:
3dAllineate -source /home/vana/Projects/p3/tmp/P3/_subject_id_sub-CTS200/3dallineate_orig/orig_out.nii.gz -prefix /home/vana/Projects/p3/tmp/P3/_subject_id_sub-CTS200/3dallineate_orig/orig_out_allineate.nii.gz -1Dmatrix_save FSorig.XFM.FS2MPR.aff12.1D -overwrite -base /home/vana/Projects/p3/dataset/sub-CTS200/anat/sub-CTS200_T1w.nii.gz
```
But downstream nodes don't seem to notice that the `out_file` is actually defined now... (Note: I have  a node whose in_file is connected to the out_file of the allineate node)

```
 [Node] Error on "P3.3drefit1" (/home/vana/Projects/p3/tmp/P3/_subject_id_sub-CTS200/3drefit1)
  File "./preproc.py", line 293, in <module>
    wf.run()
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/pipeline/engine/workflows.py", line 602, in run
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/pipeline/plugins/linear.py", line 44, in run
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 487, in run
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 571, in _run_interface
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 638, in _run_command
    cmd = self._interface.cmdline
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/interfaces/base/core.py", line 935, in cmdline
    self._check_mandatory_inputs()
  File "/home/vana/.local/lib/python3.6/site-packages/nipype/interfaces/base/core.py", line 389, in _check_mandatory_inputs
    raise ValueError(msg)
ValueError: Refit requires a value for input 'in_file'. For a list of required inputs, see Refit.help()
```

After looking at some of the fsl interfaces (and some trial-and-error), I figured out that the downstream nodes seem to be getting the information for their inputs from the `_list_outputs` method of the upstream node, and ignore `_gen_filename`. So when I do this:

```
def _gen_outfilename(self):
    out_file = self.inputs.out_file
    if not isdefined(out_file) and isdefined(self.inputs.in_file) and not isdefined(self.inputs.allcostx):
        out_file = op.abspath(self._gen_fname(self.inputs.in_file,op.dirname(self.inputs.in_file),suffix='_allineate'))
    return out_file

def _list_outputs(self):

    outputs['out_file'] = self._gen_outfilename()

    # other stuff below this ...

    if name == 'out_file':
        return self._gen_outfilename()
    return None
```

Everything works! But then the question becomes, what's the point of the `genfile` parameter if I have to define a method that checks if the input is defined anyway?

I'm not sure how to do a pull request to this repo, so I've attached my git patch below. (This is on nipype 1.0.1)
[allineate.patch.tar.gz](https://github.com/nipy/nipype/files/1826031/allineate.patch.tar.gz)

vanandrew:
Fixes #2499, #2216 

-See linked issue #2499

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -218,7 +218,8 @@ class AllineateInputSpec(AFNICommandInputSpec):
     out_file = File(
         desc='output file from 3dAllineate',
         argstr='-prefix %s',
-        genfile=True,
+        name_template='%s_allineate',
+        name_source='in_file',
         hash_files=False,
         xor=['allcostx'])
     out_param_file = File(
@@ -425,11 +426,11 @@ class AllineateInputSpec(AFNICommandInputSpec):
     _dirs = ['X', 'Y', 'Z', 'I', 'J', 'K']
     nwarp_fixmot = traits.List(
         traits.Enum(*_dirs),
-        argstr='-nwarp_fixmot%s',
+        argstr='-nwarp_fixmot%s...',
         desc='To fix motion along directions.')
     nwarp_fixdep = traits.List(
         traits.Enum(*_dirs),
-        argstr='-nwarp_fixdep%s',
+        argstr='-nwarp_fixdep%s...',
         desc='To fix non-linear warp dependency along directions.')
     verbose = traits.Bool(
         argstr='-verb', desc='Print out verbose progress reports.')
@@ -466,7 +467,6 @@ class Allineate(AFNICommand):
     '3dAllineate -source functional.nii -prefix functional_allineate.nii -1Dmatrix_apply cmatrix.mat'
     >>> res = allineate.run()  # doctest: +SKIP
 
-    >>> from nipype.interfaces import afni
     >>> allineate = afni.Allineate()
     >>> allineate.inputs.in_file = 'functional.nii'
     >>> allineate.inputs.reference = 'structural.nii'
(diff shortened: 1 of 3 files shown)


nipype/interfaces/afni/preprocess.py
```diff
@@ -465,31 +467,29 @@ class Allineate(AFNICommand):
...
     >>> allineate.cmdline
     '3dAllineate -source functional.nii -base structural.nii -allcostx |& tee out.allcostX.txt'
     >>> res = allineate.run()  # doctest: +SKIP
+
+    >>> allineate = afni.Allineate()
+    >>> allineate.inputs.in_file = 'functional.nii'
+    >>> allineate.inputs.reference = 'structural.nii'
+    >>> allineate.inputs.nwarp_fixmot = ['X', 'Y']
+    >>> allineate.cmdline
+    '3dAllineate -source functional.nii -nwarp_fixmotX -nwarp_fixmotY -prefix functional_allineate -base structural.nii'
```
effigies: A question: Is `-prefix functional_allineate` the correct default, or should it be `functional_allineate.nii`? 
If it should have the extension, we should add `keep_extension=True` to the `out_file` trait spec.

@tsalo @salma1601 You might also be good people to chime in here.

nipype/interfaces/base/core.py
```diff
@@ -1072,6 +1072,12 @@ def _filename_from_source(self, name, chain=None):
         if not isdefined(retval) or "%s" in retval:
             if not trait_spec.name_source:
                 return retval
+
+            # Do not generate filename when excluded by other inputs
+            if trait_spec.xor and any(isdefined(getattr(self.inputs, field))
+                                      for field in trait_spec.xor):
+                return retval
```
effigies: @satra @oesteban Does this constitute a full fix of #2506?

effigies:
Ah, sorry. Forgot to `make specs`. Pushing it directly.

nipype/interfaces/base/core.py
satra: @effigies - this is a good addition, but i'm not sure this addresses #2506, which does not have any `xor`.  isn't it the case in #2506 that `out_file` should be populated if `in_files` is defined and set to some default otherwise?

i think this PR is fine, just not sure if it addresses #2506.

effigies: Ah, right. I had it kind of backwards, but a similar fix to this should resolve #2506. I'll merge and propose a quick PR for that one.

PR #2512, opened 2018-03-26
DOC: dcm2niix - details for the output filename pattern, minor typo fix

yarikoptic:
found uncommitted on the drive, paid price of rebasing and fixing conflicts ;)


nipype/interfaces/dcm2nii.py
```diff
@@ -275,7 +275,24 @@ class Dcm2niixInputSpec(CommandLineInputSpec):
         xor=['source_names'])
     out_filename = traits.Str(
         argstr="-f %s",
-        desc="Output filename")
+        desc="Output filename template ("
+             "%a=antenna (coil) number, " \
```
effigies: You don't need these backslashes, since you're inside parentheses.

effigies:
Since it's a small documentation fix, and it looks like we'll need to wait for #2513 to be fixed to release 1.0.2, I'm marking this 1.0.2.

PR #2463, opened 2018-02-23
[WIP] fix: dtitk interface specs

satra:
@kesshijordan - this involves a bunch of changes. but i don't have a good understanding of the interfaces. i.e. where the inputs can actually be better specified:

- replace traits.Str with more specific Int/Float/List/etc
- there were many places where exists=True and mandatory=False were there. these seemed quite contradictory. in nipype exists only goes with File/Directory traits and mandatory is for required inputs. it would be helpful to comment on each spec or simply send a PR to my PR
- we will probably not merge this till after the monday release

Fixes #2462

also at this point i haven't regenerated the autospecs to simplify reviewing.

Issue #2506, opened 2018-03-23
Fix name_source to handle missing source


effigies:
As discussed in #2450, the `name_source` logic is not working as expected.

Consider the following input spec:

```Python
class NwarpAdjustInputSpec(AFNICommandInputSpec):
    warps = InputMultiPath(
        File(exists=True),
        minlen=5,
        mandatory=True,
        argstr='-nwarp %s',
        desc='List of input 3D warp datasets')
    in_files = InputMultiPath(
        File(exists=True),
        minlen=5,
        argstr='-source %s',
        desc='List of input 3D datasets to be warped by the adjusted warp '
             'datasets.  There must be exactly as many of these datasets as '
             'there are input warps.')
    out_file = File(
        desc='Output mean dataset, only needed if in_files are also given. '
             'The output dataset will be on the common grid shared by the '
             'source datasets.',
        argstr='-prefix %s',
        name_source='in_files',
        name_template='%s_NwarpAdjust',
        keep_extension=True,
        requires=['in_files'])
```

The goal is to automatically derive an `out_file` value iff `in_files` is set. There was some discussion over whether the traits are correct as is (e.g. @oesteban suggested that `requires` is not needed for `out_file`), but it did seem to be agreed that the goal should be achievable purely using traits without overriding interface methods.

effigies:
Figure I should bump this before we get too close to release. Does anybody have time to figure this out?

@oesteban @satra Are there open issues to discuss, or is it a question of implementation?

PR #2514, opened 2018-03-27
FIX: DTITK Interface

### LINKED ISSUE #2462: some specs need fixing
```
nipype.interfaces.afni.preprocess:TProject:Inputs:mask:exist
nipype.interfaces.dtitk.registration:AffineTask:Inputs:fixed_file:exists
nipype.interfaces.dtitk.registration:AffineTask:Inputs:moving_file:exists
nipype.interfaces.dtitk.registration:AffineTask:Inputs:similarity_metric:exists
nipype.interfaces.dtitk.registration:AffineTask:Inputs:useInTrans:mandatory=False
nipype.interfaces.dtitk.registration:ComposeXfmTask:Inputs:in_aff:exists
nipype.interfaces.dtitk.registration:ComposeXfmTask:Inputs:in_aff:mandatory=False
nipype.interfaces.dtitk.registration:ComposeXfmTask:Inputs:in_df:exists
nipype.interfaces.dtitk.registration:ComposeXfmTask:Inputs:in_df:mandatory=False
nipype.interfaces.dtitk.registration:ComposeXfmTask:Inputs:out_file:exists
nipype.interfaces.dtitk.registration:ComposeXfmTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:fixed_file:exists
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:fixed_file:mandatory=False
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:ftol:exists
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:legacy:exists
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:mask:exists
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:mask:mandatory=False
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:moving_file:exists
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:moving_file:mandatory=False
nipype.interfaces.dtitk.registration:DiffeoTask:Inputs:n_iters:exists
nipype.interfaces.dtitk.registration:RigidTask:Inputs:fixed_file:exists
nipype.interfaces.dtitk.registration:RigidTask:Inputs:moving_file:exists
nipype.interfaces.dtitk.registration:RigidTask:Inputs:similarity_metric:exists
nipype.interfaces.dtitk.registration:RigidTask:Inputs:useInTrans:mandatory=False
nipype.interfaces.dtitk.registration:affScalarVolTask:Inputs:in_volume:exists
nipype.interfaces.dtitk.registration:affScalarVolTask:Inputs:in_volume:mandatory=False
nipype.interfaces.dtitk.registration:affScalarVolTask:Inputs:in_xfm:exists
nipype.interfaces.dtitk.registration:affScalarVolTask:Inputs:in_xfm:mandatory=False
nipype.interfaces.dtitk.registration:affScalarVolTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:in_target:exists
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:in_target:mandatory=False
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:in_tensor:exists
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:in_tensor:mandatory=False
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:in_xfm:exists
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:in_xfm:mandatory=False
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:out_file:exists
nipype.interfaces.dtitk.registration:affSymTensor3DVolTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_flip:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_flip:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_interp:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_interp:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_target:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_target:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_type:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_type:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_volume:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_volume:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_vsize:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_vsize:mandatory=False
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_xfm:exists
nipype.interfaces.dtitk.registration:diffeoScalarVolTask:Inputs:in_xfm:mandatory=False
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:in_target:exists
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:in_target:mandatory=False
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:in_tensor:exists
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:in_tensor:mandatory=False
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:in_xfm:exists
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:in_xfm:mandatory=False
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:out_file:exists
nipype.interfaces.dtitk.registration:diffeoSymTensor3DVolTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.utils:BinThreshTask:Inputs:in_file:exists
nipype.interfaces.dtitk.utils:BinThreshTask:Inputs:in_file:mandatory=False
nipype.interfaces.dtitk.utils:BinThreshTask:Inputs:in_numbers:exists
nipype.interfaces.dtitk.utils:BinThreshTask:Inputs:in_numbers:mandatory=False
nipype.interfaces.dtitk.utils:BinThreshTask:Inputs:out_file:exists
nipype.interfaces.dtitk.utils:BinThreshTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:in_file:exists
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:in_target:exists
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:in_target:mandatory=False
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:in_voxsz:exists
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:in_voxsz:mandatory=False
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:origin:exists
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:origin:mandatory=False
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:out_file:exists
nipype.interfaces.dtitk.utils:SVAdjustVoxSpTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:in_arraysz:exists
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:in_arraysz:mandatory=False
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:in_file:exists
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:in_voxsz:exists
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:in_voxsz:mandatory=False
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:out_file:exists
nipype.interfaces.dtitk.utils:SVResampleTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.utils:TVAdjustOriginTask:Inputs:origin:exists
nipype.interfaces.dtitk.utils:TVAdjustOriginTask:Inputs:origin:mandatory=False
nipype.interfaces.dtitk.utils:TVAdjustOriginTask:Outputs:out_file:exists
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Inputs:origin:exists
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Inputs:origin:mandatory=False
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Inputs:target:exists
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Inputs:target:mandatory=False
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Inputs:vsize:exists
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Inputs:vsize:mandatory=False
nipype.interfaces.dtitk.utils:TVAdjustVoxSpTask:Outputs:out_file:exists
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:in_arraysz:exists
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:in_arraysz:mandatory=False
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:in_file:exists
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:in_voxsz:exists
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:in_voxsz:mandatory=False
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:out_file:exists
nipype.interfaces.dtitk.utils:TVResampleTask:Inputs:out_file:mandatory=False
nipype.interfaces.dtitk.utils:TVtoolTask:Inputs:in_file:exists
nipype.interfaces.dtitk.utils:TVtoolTask:Inputs:in_file:mandatory=False
nipype.interfaces.dtitk.utils:TVtoolTask:Inputs:in_flag:exists
nipype.interfaces.dtitk.utils:TVtoolTask:Inputs:in_flag:mandatory=False
```

kesshijordan:
Fixes #2462 

- This is addressing the code review from @satra and @effigies  #2463 

Closes #2463.


kesshijordan:
THEY PASSED!!! Thank you so much for your help, @satra and @effigies !!!

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
-                >>> import nipype.interfaces.dtitk as dtitk
-                >>> node = dtitk.ComposeXfmTask()
-                >>> node.inputs.in_df = 'ants_Warp.nii.gz'
-                >>> node.inputs.in_aff= 'ants_Affine.txt'
-                >>> node.run() # doctest: +SKIP
-                """
+    >>> import nipype.interfaces.dtitk as dtitk
+    >>> node = dtitk.ComposeXfmTask()
+    >>> node.inputs.in_df = 'im1.nii'
+    >>> node.inputs.in_aff= 'im2.nii'
```
effigies: It would probably be more informative to use a file of the right type (e.g. `*.aff`). Note that you can use any filename in [`nipype/testing/data/`](https://github.com/nipy/nipype/tree/master/nipype/testing/data) (and you can add empty files to that directory to make these usefully realistic).

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
-    in_xfm = traits.Str(desc='transform to apply', exists=True,
-                        mandatory=False, position=1, argstr="-trans %s")
-    in_target = traits.Str(desc='', exists=True, mandatory=False, position=2,
-                           argstr="-target %s")
-    out_file = traits.Str(desc='', exists=True, mandatory=False, position=3,
-                          argstr="-out %s", name_source="in_tensor",
-                          name_template="%s_affxfmd.nii.gz")
+    in_file = File(desc='moving tensor volume', exists=True,
+                   argstr="-in %s", mandatory=True)
+    out_file = traits.Str(desc='output filename', exists=True,
```
effigies: This should be a `File` without `exists=True`.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+                          argstr="-out %s", name_source="in_file",
+                          name_template="%s_affxfmd", keep_extension=True)
+    transform = File(exists=True, argstr="-trans %s",
+                     xor=['target', 'translation', 'euler', 'deformation'], desc='transform to apply: specify an input transformation  file; parameters input will be ignored',)
```
effigies: Newline before `desc`.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+    interpolation = traits.Enum('LEI', 'EI', usedefault=True,
+                                argstr="-interp %s",
+                                desc='Log Euclidean/Euclidean Interpolation')
+    reorient = traits.Enum('PPD', 'NO', 'FS', argstr='-reorient %s',
+                           usedefault=True)
```
effigies: A `desc` would help make this clearer.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+    target = File(exists=True, argstr="-target %s", xor=['transform'],
+                  desc='output volume specification read from the target volume if specified')
+    translation = traits.Tuple((0, 0, 0), desc='translation (x,y,z) in mm',
```
effigies: I'm not sure if `traits.Tuple` will derive the correct types from `(0, 0, 0)`. I'd use `traits.Float()` (3 times), and skip the `(0, 0, 0)`. Since you're not using the default value, you don't need to set it.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+                               argstr='-translation %g %g %g',
+                               xor=['transform'])
+    euler = traits.Tuple((0, 0, 0), desc='(theta, phi, psi) in degrees',
```
effigies: `traits.Tuple((traits.Float(), traits.Float(), traits.Float()), ...)`

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+                         xor=['transform'], argstr='-euler %g %g %g')
+    deformation = traits.Tuple((1, 1, 1, 0, 0, 0), desc='(xx,yy,zz,xy,yz,xz)',
```
effigies: Similarly. Also, for one this long, it might be clearer to use multiplication on a 1-tuple:

`traits.Tuple((traits.Float(),) * 6, ...)`

However you like to do it is fine, though.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
-    in_xfm = traits.Str(desc='transform to apply', exists=True,
-                        mandatory=False,
-                        position=1, argstr="-trans %s")
-    in_target = traits.Str(desc='', position=2, argstr="-target %s")
-    out_file = traits.Str(desc='', mandatory=False, position=3,
-                          argstr="-out %s", name_source="in_volume",
-                          name_template="%s_affxfmd.nii.gz")
+    in_file = File(desc='moving scalar volume', exists=True,
+                   argstr="-in %s", mandatory=True)
+    out_file = traits.Str(desc='output filename', exists=True,
```
effigies: `File`, no `exists=True`.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+                          argstr="-out %s", name_source="in_file",
+                          name_template="%s_affxfmd", keep_extension=True)
+    transform = File(exists=True, argstr="-trans %s",
+                     xor=['target', 'translation', 'euler', 'deformation'], desc='transform to apply: specify an input transformation  file; parameters input will be ignored',)
+    interpolation = traits.Enum(0, 1, usedefault=True,
+                                argstr="-interp %s",
+                                desc='0=trilinear (def); 1=nearest neighbor')
```
effigies: Before this goes into a release, you could consider making this:

```Python
interpolation = traits.Enum('trilinear', 'nearest', usedefault=True,
                            argstr='-interp %s',
                            desc='use trilinear or nearest neighbor interpolation')
```

And then, below:

```Python
class affScalarVolTask:
    ...
        if name == 'interpolation':
            value = {'trilinear': 0, 'nearest': 1}[value]
        super(affScalarVolTask, self)._format_arg(name, spec, value)
```

This is one way we can make interfaces a little more user friendly. Your call whether it's worth it. Your `desc` makes it clear, as is.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+    target = File(exists=True, argstr="-target %s", xor=['transform'],
+                  desc='output volume specification read from the target volume if specified')
+    translation = traits.Tuple((0, 0, 0), desc='translation (x,y,z) in mm',
+                               argstr='-translation %g %g %g',
+                               xor=['transform'])
+    euler = traits.Tuple((0, 0, 0), desc='(theta, phi, psi) in degrees',
+                         xor=['transform'], argstr='-euler %g %g %g')
+    deformation = traits.Tuple((1, 1, 1, 0, 0, 0), desc='(xx,yy,zz,xy,yz,xz)',
+                               xor=['transform'],
+                               argstr='-deformation %g %g %g %g %g %g')
```
effigies: See earlier comments RE `traits.Float()`.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
+"""
+
+from ..base import TraitedSpec, CommandLineInputSpec, traits, isdefined, File
 from ...utils.filemanip import fname_presuffix
 import os
 from .base import CommandLineDtitk
 
+__docformat__ = 'restructuredtext'
 
 class RigidInputSpec(CommandLineInputSpec):
```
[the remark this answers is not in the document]
kesshijordan: You were correct in your comment "This is the actual interface that performs a task"; I took away "task" from all interfaces so it's now `Rigid` and `RigidInputSpec` to follow the convention.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,419 @@
...
         outputs = self.output_spec().get()
-        outputs['out_file_xfm'] = self.inputs.moving_file.replace(
-            '.nii.gz', '_diffeo.df.nii.gz')
-        outputs['out_file'] = self.inputs.moving_file.replace(
-            '.nii.gz', '_diffeo.nii.gz')
+        splitlist = os.path.basename(self.inputs.moving_file).split('.')
+        basename = splitlist[0]
+        termination = '.'+'.'.join(splitlist[1:])
+        outputs['out_file_xfm'] = basename+'_diffeo.df'+termination
+        outputs['out_file'] = basename+'_diffeo'+termination
```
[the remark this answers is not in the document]
kesshijordan: I replaced this with `fname_presuffix` in all of the relevant interfaces.

nipype/interfaces/dtitk/registration.py
kesshijordan: I removed `exists=True` for all `out_file` fields in the inputs specs, and changed the `traits.Str` to `traits.File`.

nipype/interfaces/dtitk/registration.py
kesshijordan: I'm guessing 'NO' is not reorienting, but not positive. 

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4213863/

nipype/interfaces/dtitk/registration.py
effigies: Just a note that we use `File` (imported from `nipype.interfaces.base`) not `traits.File`. It has slightly improved error messages compared to `traits.File`.

kesshijordan:
https://github.com/kesshijordan/shared_notebooks/blob/master/Kesh_Test_DTITK_Interface_Registration_Utils.ipynb

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
+                   if specified')
+    translation = traits.Tuple((traits.Float(), traits.Float(),
+                               traits.Float()),
+                               desc='translation (x,y,z) in mm',
+                               argstr='-translation %g %g %g',
+                               xor=['transform'])
+    euler = traits.Tuple((traits.Float(), traits.Float(), traits.Float()),
+                         desc='(theta, phi, psi) in degrees',
+                         xor=['transform'], argstr='-euler %g %g %g')
+    deformation = traits.Tuple(traits.Tuple((traits.Float(),) * 6, ...),
```
effigies: Oh, sorry. I meant:

```Python
deformation = traits.Tuple((traits.Float(),) * 6, 
                           desc='(xx,yy,zz,xy,yz,xz)', xor=['transform'],
                           argstr='-deformation %g %g %g %g %g %g')
```

The `...` was just to avoid copying and pasting all the other stuff. :-)

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
+    interpolation = traits.Enum('LEI', 'EI', usedefault=True,
+                                argstr="-interp %s",
+                                desc='Log Euclidean/Euclidean Interpolation')
+    reorient = traits.Enum('PPD', 'NO', 'FS', argstr='-reorient %s',
+                           usedefault=True, desc='Reorientation strategy: \
+                           preservation of principal direction, no \
+                           reorientation, or finite strain')
+    target = File(exists=True, argstr="-target %s", xor=['transform'],
+                  desc='output volume specification read from the target volume\
+                   if specified')
```
effigies: For strings that go on beyond line lengths, the preferred way to do this would be:

```Python
    desc='output volume specification read from the target volume'
    ' if specified'
```

As long as you're within parentheses, newlines don't matter, and two strings sitting next to each other are automatically combined into a single string.

nipype/interfaces/dtitk/registration.py
```diff
@@ -165,7 +165,8 @@ class ComposeXfmInputSpec(CommandLineInputSpec):
                   argstr="-aff %s", mandatory=True)
     out_file = File(desc='output path',
                     argstr="-out %s",  name_source="in_df",
-                    name_template="%s_aff.df", keep_extension=True)
+                    name_template="%s_affdf.nii")
+    # keep_extension is keeping the .df but not .nii; need to figure out
```
effigies: This looks like it might be a bug in `name_template` parsing. (So possibly worth considering along with any fixes for #2506.)

nipype/interfaces/dtitk/registration.py
kesshijordan: Thanks. This doesn't work for the doctests, though (see, below)... is there a different syntax I should use?

Expected:

   'deformationScalarVolume -in im1.nii -interp 0 -out im1_diffeoxfmd.nii'
    ' -trans im_warp.df.nii'
Got:
    'deformationScalarVolume -in im1.nii -interp 0 -out im1_diffeoxfmd.nii -trans im_warp.df.nii'

effigies: Ah, yeah, the doctests aren't as nice about these things. But they should be forgiving on whitespace. You may be able to throw in a newline without closing and reopening the quotes, but I'm not positive of that. If you go beyond the line limit in the doctests, I wouldn't quibble over it.

effigies:
Just checking in... is this in my court to review, or do you have things you're planning to do first?

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
+    mask_file = File(desc="mask", exists=True,  position=2, argstr="%s")
+    legacy = traits.Enum(1, desc="legacy parameter; always set to 1",
+                         usedefault=True, mandatory=True,
+                         position=3, argstr="%d")
+    n_iters = traits.Int(6, desc="number of iterations",
+                         exists=True, mandatory=True,
+                         position=4, argstr="%d", usedefault=True)
+    ftol = traits.Float(0.002, desc="iteration for the optimization to stop",
+                        exists=True, mandatory=True, position=5, argstr="%g",
+                        usedefault=True)
```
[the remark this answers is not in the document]
kesshijordan: Just looked up the documentation; this is only for files, not for parameters.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
-                          position=2, argstr="-out %s",  name_source="in_df",
-                          name_template="%s_comboaff.nii.gz")
+    in_df = File(desc='diffeomorphic warp file', exists=True,
+                 argstr="-df %s", copyfile=False, mandatory=True)
+    in_aff = File(desc='affine transform file', exists=True,
+                  argstr="-aff %s", mandatory=True)
+    out_file = File(desc='output path',
+                    argstr="-out %s",  name_source="in_df",
+                    name_template="%s_affdf.nii")
+    # keep_extension is keeping the .df but not .nii; need to figure out
```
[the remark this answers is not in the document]
kesshijordan: Is this something someone is working on and I should keep an eye on the PR to incorporate when it's finished, or am I adding the fix to my DTITK PR?

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
+                                desc='Log Euclidean/Euclidean Interpolation')
+    reorient = traits.Enum('PPD', 'NO', 'FS', argstr='-reorient %s',
+                           usedefault=True, desc='Reorientation strategy: '
+                           'preservation of principal direction, no '
+                           'reorientation, or finite strain')
+    target = File(exists=True, argstr="-target %s", xor=['transform'],
+                  desc='output volume specification read from the target '
+                  'volume if specified')
+    translation = traits.Tuple((traits.Float(), traits.Float(),
+                               traits.Float()),
```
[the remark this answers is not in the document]
kesshijordan: I have flake8 checking on my editor (Atom)... when I align the first character on the new line to either the second `(` or the `t` in `traits.Float()`, it doesn't complain (though it does outside of those two locations). Also, if the tuple doesn't continue onto the next line, flake8 has an error when aligning to the inside of the tuple, so I suppose it's just when the tuple continues to the new line. I've switched the two cases in which the tuple continues on a second line to align with the inside of the tuple, but left the others as they are. Is that correct?

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
-    samplingZ = traits.Float(mandatory=True, position=5, argstr="%s",
-                             desc="dist between samp points (mm)",
-                             default_value=4)
-    ftol = traits.Float(mandatory=True, position=6, argstr="%s",
-                        desc="cost function tolerance", default_value=0.01)
-    useInTrans = traits.Float(mandatory=False, position=7, argstr="%s",
-                              desc="to initialize with existing xfm set as 1",
-                              default_value=1)
+                                    desc="similarity metric", usedefault=True)
+    samplingXYZ = traits.Tuple((4, 4, 4), mandatory=True, position=3,
```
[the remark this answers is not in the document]
kesshijordan: For class names, though, you want `CamelCase` (with the first character capitalized), right? (link, below)

http://nipype.readthedocs.io/en/latest/devel/cmd_interface_devel.html

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
-                """
+    Example
+    -------
+
+    >>> from nipype.interfaces import dtitk
+    >>> node = dtitk.ComposeXfm()
+    >>> node.inputs.in_df = 'im_warp.df.nii'
+    >>> node.inputs.in_aff= 'im_affine.aff'
+    >>> node.cmdline #doctest:
+    'dfRightComposeAffine -aff im_affine.aff -df im_warp.df.nii -out\
```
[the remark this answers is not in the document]
kesshijordan: Yup, you're right. Just tested it and it works without the backslash.

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,431 @@
...
 class ComposeXfmInputSpec(CommandLineInputSpec):
-    in_df = traits.Str(desc='diffeomorphic file.df.nii.gz', exists=True,
-                       mandatory=False, position=1, argstr="-df %s")
-    in_aff = traits.Str(desc='affine file.aff', exists=True, mandatory=False,
-                        position=0, argstr="-aff %s")
-    out_file = traits.Str(desc='output_path', exists=True, mandatory=False,
-                          position=2, argstr="-out %s",  name_source="in_df",
-                          name_template="%s_comboaff.nii.gz")
+    in_df = File(desc='diffeomorphic warp file', exists=True,
+                 argstr="-df %s", copyfile=False, mandatory=True)
```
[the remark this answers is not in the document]
kesshijordan: I'm not sure about this... @satra changed it in #2463. I can look into it.

nipype/interfaces/dtitk/registration.py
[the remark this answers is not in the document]
effigies: Correct.

effigies:
Great, thanks for making those changes. Since this is so big, I'll have another read through in a couple days, but I suspect this is just about done. Thanks a lot for your patience.

nipype/interfaces/dtitk/registration.py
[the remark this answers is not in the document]
kesshijordan: I switched all of the class names from camelCase to CamelCase, since those were mixed, too.

kesshijordan:
Thanks for all the time you have put into this review!

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,430 @@
...
+
+class DiffeoSymTensor3DVolInputSpec(CommandLineInputSpec):
+    in_file = File(desc='moving tensor volume', exists=True,
+                   argstr="-in %s", mandatory=True)
+    out_file = File(desc='output filename',
+                    argstr="-out %s", name_source="in_file",
+                    name_template="%s_diffeoxfmd", keep_extension=True)
+    transform = File(exists=True, argstr="-trans %s",
+                     mandatory=True, desc='transform to apply')
+    df = traits.Str('FD', argstr="-df %s", usedefault=True)
```
effigies: Description?

And if `-df` were omitted, it would act as if `-df FD` were passed? Are there no other options?

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,430 @@
...
+    interpolation = traits.Enum('LEI', 'EI', usedefault=True,
+                                argstr="-interp %s",
+                                desc='Log Euclidean/Euclidean Interpolation')
+    reorient = traits.Enum('PPD', 'FS', argstr='-reorient %s',
+                           usedefault=True, desc='Reorientation strategy: '
+                           'preservation of principal direction or finite '
+                           'strain')
```
effigies: Just to be clear, no reorientation is not an option for this command? And if you omitted the argument, it would either cause an error or default to `PPD`?

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,430 @@
...
+        return super(DiffeoSymTensor3DVol, self)._format_arg(name, spec, value)
+
+
+class DiffeoScalarVolInputSpec(CommandLineInputSpec):
+    in_file = File(desc='moving scalar volume', exists=True,
+                   argstr="-in %s", mandatory=True)
+    out_file = File(desc='output filename',
+                    argstr="-out %s", name_source="in_file",
+                    name_template="%s_diffeoxfmd", keep_extension=True)
+    transform = transform = File(exists=True, argstr="-trans %s",
```
effigies: `transform = transform =`

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,430 @@
...
+                                 mandatory=True, desc='transform to apply')
+    target = File(exists=True, argstr="-target %s", xor=['voxel_size'],
+                  desc='output volume specification read from the target '
+                  'volume if specified')
+    voxel_size = traits.Tuple((traits.Float(), traits.Float(), traits.Float()),
+                              desc='xyz voxel size (superseded by target)',
+                              argstr="-vsize %g %g %g", xor=['target'])
+    flip = traits.Tuple((traits.Int(), traits.Int(), traits.Int()),
+                        argstr="-flip %d %d %d")
```
effigies: Description? Also, my guess is this is limited to 0 and 1... Maybe it should be `traits.Tuple((traits.Bool, traits.Bool, traits.Bool), ...)`? The `argstr` could stay the same.

I would add a `flip` input in the docstring to make sure that the `argstr` works properly.

nipype/interfaces/dtitk/utils.py
```diff
@@ -1,367 +1,302 @@
...
-                          argstr="-out %s")
-    origin = traits.Str(desc='xyz voxel size', exists=True, mandatory=False,
-                        position=4, argstr='-origin %s')
-    target = traits.Str(desc='target volume', exists=True, mandatory=False,
-                        position=2, argstr="-target %s")
-    vsize = traits.Str(desc='resampled voxel size', exists=True,
-                       mandatory=False, position=3, argstr="-vsize %s")
+    in_file = File(desc="tensor volume to modify", exists=True,
+                   mandatory=True, argstr="-in %s")
+    out_file = File(genfile=True, desc='output path',
```
effigies: I think `genfile` is an older mechanism than `name_template`, and no longer recommended. So we'd just drop `genfile=True` and keep this trait otherwise as is. @satra is this correct?

nipype/interfaces/dtitk/utils.py
```diff
@@ -1,367 +1,302 @@
...
-    in_file = traits.Str(desc='', exists=True, mandatory=False, position=0,
-                         argstr="%s")
-    out_file = traits.Str(desc='', exists=True, mandatory=False, position=1,
-                          argstr="%s")
-    in_numbers = traits.Str(desc='LB UB inside_value outside_value',
-                            exists=True, mandatory=False, position=2,
-                            argstr="%s")
+    in_file = File(desc='Image to threshold/binarize', exists=True,
+                   position=0, argstr="%s", mandatory=True)
+    out_file = File(desc='',  position=1, argstr="%s",
```
effigies: Empty description.

nipype/interfaces/dtitk/utils.py
```diff
@@ -1,367 +1,302 @@
...
+                    keep_extension=True, name_source='in_file',
+                    name_template='%s_thrbin')
+    lower_bound = traits.Float(0.01, position=2, argstr="%g", mandatory=True)
+    upper_bound = traits.Float(100, position=3, argstr="%g", mandatory=True)
+    inside_value = traits.Float(1, position=4, argstr="%g", usedefault=True,
+                                mandatory=True)
+    outside_value = traits.Float(0, position=5, argstr="%g", usedefault=True,
+                                 mandatory=True)
```
effigies: Descriptions for these four numbers would be good.

nipype/interfaces/dtitk/utils.py
```diff
@@ -1,367 +1,302 @@
...
+        >>> node = dtitk.BinThresh()
+        >>> node.inputs.in_file = 'im1.nii'
+        >>> node.inputs.lower_bound = 0
+        >>> node.inputs.upper_bound = 100
+        >>> node.inputs.inside_value = 1
+        >>> node.inputs.outside_value = 0
+        >>> node.cmdline
+        'BinaryThresholdImageFilter im1.nii im1_thrbin.nii 0 100 1 0'
         >>> node.run() # doctest: +SKIP
         """
```
effigies: De-indent this block (up to `Example`).

nipype/interfaces/dtitk/registration.py
kesshijordan: Here is the documentation. I think that is the case... "df" usually means "deformation field" in this package, but I'm not sure what FD stands for (I didn't find it in the documentation).

nipype/interfaces/dtitk/registration.py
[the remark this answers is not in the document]
kesshijordan: Thanks; I'll work on this and put the note in #2506.

nipype/interfaces/dtitk/registration.py
kesshijordan: I think that is the case (see, below, from documentation). It looks like the available options are specified to the right and "FS" or "PPD" are the only two options.

-reorient type: string(s) required: N	default: PPD FS or PPD

nipype/interfaces/dtitk/registration.py
kesshijordan: I'm guessing this is a binary x y z to flip each axis, but I haven't found that documented.

` -flip          type: integer(s)   required: N	default: 0 0 0 `

kesshijordan:
@effigies I haven't found descriptions for a lot of the flags in the documentation. I may see if I can get ahold of the author and see if he can look this over.

nipype/interfaces/dtitk/utils.py
kesshijordan: We still want to use `genfile=True` for `TVtool`, right? (this is one that we are building the filename from multiple inputs)

effigies: Yes, that one is fine to leave, since we're not modifying it to use `name_template`.

nipype/interfaces/dtitk/registration.py
effigies: Cool. This looks good then.

nipype/interfaces/dtitk/registration.py
effigies: Gotcha. Okay, this makes sense. It sets the default, but because it's `Str` not `Enum`, it lets users who know more than we do set an alternative.

effigies:
Here's a mixin you can put in `dtitk/base.py`

```Python
class DTITKRenameMixin(object):
    def __init__(self, *args, **kwargs):
        classes = [cls.__name__ for cls in self.__class__.mro()]
        dep_name = classes[0]
        rename_idx = classes.index('DTITKRenameMixin')
        new_name = classes[rename_idx + 1]
        warnings.warn('The {} interface has been renamed to {}\n'
                      'Please see the documentation for DTI-TK '
                      'interfaces, as some inputs have been '
                      'added or renamed for clarity.'
                      ''.format(dep_name, new_name),
                      DeprecationWarning)
        super(DTITKRenameMixin, self).__init__(*args, **kwargs)
```

And then for classes that have been renamed, we can just have a big sequence of dummy classes like so:

`dtitk/registration.py`:
```Python
class RigidTask(DTITKRenameMixin, Rigid):
    pass
...
class diffeoScalarVolTask(DTITKRenameMixin, DiffeoScalarVol):
    pass
```

`dtitk/utils.py`:
```Python
class TVAdjustVoxSpTask(DTITKRenameMixin, TVAdjustVoxSp):
    pass
...
class BinThreshTask(DTITKRenameMixin, BinThresh):
    pass
```

@satra @mgxd @djarecka I suspect you might have preferences for how deprecations are handled. Feel free to suggest any modifications (e.g. do we want to set a version that these will be removed by?).

nipype/interfaces/dtitk/registration.py
```diff
@@ -1,394 +1,446 @@
...
-    out_file = traits.File(desc='cheese', exists=True)
-
-
-class affSymTensor3DVolTask(CommandLineDtitk):
+    def _gen_filename(self, name):
+        if name != 'out_file':
+            return
+        return fname_presuffix(os.path.basename(
+                               self.inputs.in_df).split('.')[0],
+                               suffix='_affdf.df.nii')
```
effigies: So you originally had (a10da5d00df9006ef5bfaf0c80eee8bd5156c273): `name_template="%s_aff.df", keep_extension=True`

I think we can achieve that pattern as follows:

```Python
return fname_presuffix(os.path.basename(self.inputs.in_df),
                       suffix='_aff.df', keep_ext=True)
```

kesshijordan: In that case, fname_presuffix still keeps the .df, so we end up with an extra extension (`filename.df_aff.df.nii`). I wanted to make it match the other warps ".df.nii" extension with something like "_affdf" prepended so the user knows this warp includes both the affine and diffeomorphic transforms together (`filename_affdf.df.nii` or `filename_affdf.df.nii.gz`). Is the extra extension okay or would it be better to match the other warp files that come from just the diffeo transform?

effigies: Oh, I see. I misunderstood the situation (and I'm now no longer sure that fixing #2506 will let us do what you want). There are two perspectives here, IMO.

1) File names are semantically meaningful, and as long as we're going to impose a convention within the DTI-TK interfaces, we should respect it.

2) Nipype is often used to bring tools together that have zero respect for each others' conventions, so attempting to manipulate names semantically will fail in some cases. Naming should largely be handled at the `DataSink` level.

I think we can sort of split the difference here:

```Python
path, base, ext = split_filename(self.inputs.in_df)
suffix = '_affdf'
if base.endswith('.df'):
    suffix += '.df'
    base = base[:-3]
return fname_presuffix(base, suffix=suffix, keep_ext=True)
```

This way if you get a file named `x.df.nii`, you'll get `x_affdf.df.nii`, but if you just get `x.nii`, you'll get `x_affdf.nii`.

Did I get your convention right? And does this seem like a reasonable way to achieve that?

kesshijordan: Yup. Thanks! I think it's a good idea to match the diffeomorphic naming convention.

nipype/interfaces/dtitk/utils.py
```diff
@@ -1,367 +1,333 @@
...
-                                                    self.inputs.in_file.
-                                                    split(".")[1:]))
-        outputs["out_file"] = os.path.abspath(outputs["out_file"])
-        return outputs
 
-    def _gen_filename(self, name):
-        if name == "out_file":
-            return self._list_outputs()["out_file"]
-        return None
+class BinThreshTASK(DTITKRenameMixin, BinThresh):
```
effigies: This `TASK` is all caps.

kesshijordan:
@effigies can I test the automatic naming in a workflow before this gets merged?

effigies:
Sounds good. Just say when you're ready.

kesshijordan:
Thanks, @effigies . Something isn't right and I'm trying to figure out what it is... in the report.rst for nodes with multiple outputs that are named automatically, the transform is not getting logged as an output. If I look at the .json, "needed_outputs" is only "out_file"... do I need to do something with "aggregate_outputs" ?

effigies:
You can update the [configuration](https://nipype.readthedocs.io/en/latest/users/config_file.html) to set `remove_unnecessary_outputs` to `false`, and see if that resolves the issue.

If `remove_unnecessary_outputs` is `true`, then any file that isn't fed into some other node or otherwise marked as needed will be removed. I don't remember the details of how to mark as needed, but you shouldn't need to mess with it. I set it to `false` while developing, to make sure that I can see all the outputs without re-running the node, but then make sure it's still `true` by default for final pipelines.

kesshijordan:
Thanks for the suggestion, @effigies. That was helpful in debugging. I think I figured out what's been causing problems and how to fix it. It seems like the outputs aren't moving between node directories correctly in registration nodes. Let me know if you think this is a good fix.

DTI-TK saves the outputs in the same directory as the inputs for the functions that don't take an output path (e.g. Rigid registration), so I addressed this by setting copyfile=False for both fixed and moving images (which means that the execution inputs are paths to the input files in the node working directory). Then, the outputs (image and rigid transform) were put in the node working directory instead of the input node directory (which is what we want). We still have an issue, however, in that there is no input for the initialization transform (to provide the Affine node with the Rigid node's output transform), so it needs to be copied to the Affine node working directory. 

I addressed this by making the boolean flag instructing the use of a precomputed transform to initialize the registration (use_in_transform) into a file that takes in the .aff transform file. It adds the string, which triggers using the transform just like the "1" flag would, and leaves it off it it isn't set (thus not triggering "use transform". Then I can set copyfile=True to ensure that the transform makes it into the new directory even though it isn't specified as an input to the command line.

Cases:
nothing input to Affine.inputs.initialize_xfm:
`dti_rigid_reg ...working_dir/dtitk_reg/rigid_node/Ptensor_resampled.nii.gz ...working_dir/dtitk_reg/rigid_node/Ttensor_resampled.nii.gz EDS 4 4 4 0.01`

Rigid.outputs.out_file_xfm connected to Affine.inputs.initialize_xfm
`dti_affine_reg ...working_dir/dtitk_reg/affine_node/Ptensor_resampled.nii.gz ...working_dir/dtitk_reg/affine_node/Ttensor_resampled.nii.gz EDS 4 4 4 0.01 ...working_dir/dtitk_reg/affine_node/Ttensor_resampled.aff`

One thing to be addressed with this approach is that we need to enforce the DTITK format or it's a silent failure (if it can't open the .aff transform file, it just skips it and still performs the registration with a different initialization strategy)
I was thinking: can we search the stderr for keywords to raise an error if the .aff file is not found?
`stderr: filename.aff doesn't exist or can't be opened` prints while it runs

kesshijordan:
Yes; that is my understanding of the issue. Your suggestion works for running the interface alone, but in the pipeline it doesn't put the symlink to .aff in the working directory. It works if you put copyfile=False outside the File(), but that breaks if you set it to 1. Do you know why copyfile=False would behave differently inside of `traits.Either`?

copyfile=False as argument to `File` inside of `traits.Either` sets the .aff in the rigid node and it can't find the .aff in the working directory.
`dti_affine_reg ...working_dir/dtitk_reg/affine_node/Ptensor_resampled.nii.gz ...working_dir/dtitk_reg/affine_node/Ttensor_resampled.nii.gz EDS 4 4 4 0.01 ...working_dir/dtitk_reg/rigid_node/Ttensor_resampled.aff`

copyfile=False as argument to `traits.Either` works, but then breaks if you set it to 1 because it can't copy the non-file
`dti_affine_reg ...working_dir/dtitk_reg/affine_node/Ptensor_resampled.nii.gz ...working_dir/dtitk_reg/affine_node/Ttensor_resampled.nii.gz EDS 4 4 4 0.01 ...working_dir/dtitk_reg/affine_node/Ttensor_resampled.aff`

effigies:
Oh, so if you specify the location, it uses that location as both input and output? I get it now. Out of curiosity, is it still a symlink after running? If so, I suspect it's overwriting the file back in the rigid node, in which case you'll want `copyfile=True`.

And I guess let's drop the `Either` and just use `File`. The `copyfile` in `Either` issue is probably a bug, but fixing it is definitely out-of-scope for this PR.

kesshijordan:
It still saves the .aff to the node working directory (because the moving file gets symlinked and that is the path used in the execution input, which is the basis for the output `<moving>.aff`). You can't specify the location of the output on the command line (there's no field)... the only thing it does is guess what the .aff would be from the moving filename and use it if the flag is set to anything and the file exists. Since I'm setting `copyfile=False` on the fixed and moving images, I think it's symlinking those files to the node working directory. (the original inputs are the previous node directories, but the execution inputs are the node working directory). That's why I need to symlink the `initialize_xfm` path to the working directory; because it will be looking there. If it doesn't find it (as is the case with `Either`), it still generates the output by the same name `<moving>.aff`, but does not use the previous affine to initialize. I'm cheating by specifying an input the command line function can't take, but if I set copyfile=False it symlinks and works the way we want it to. 

I don't see the symlinks after running (I think they're deleted after the node processes), but I can infer that they're there during processing from the 'Execution Inputs' section of the `report.rst file` (and the output filename `<moving>.aff` would not be in the directory if that were not the case).

If we drop `Either` and just use `File`, then I can leave `exists` as False so we can still use the interface by itself with the flag `'1'`. It won't work with `initialize_xfm=True` or `initialize_xfm=1` like we had it before, though... it must be a string because we're tricking it into thinking the flag is a filename and providing anything at all in that field to the command line function will trigger the desired behavior (using the pre-calculated transform in the working directory). Does that sound good? It's a little gross but I'm not sure how else to do it without changing `copyfile` in `Either`.

kesshijordan:
I think the intended behavior is to overwrite the previous .aff file behind the scenes to cut down on processing, but I don't think that happens backward through the symlink in our case because the "execution" paths are all to the working directory so the `<moving>.nii` file has a path in the current directory `affine/<moving>.nii` (I assume this is a symlink that gets deleted). That means the transform save path is `affine/<moving>.aff` so I think it's writing in the new directory and keeping the transform, since it's passed to the next step. The way it's written, though, the flag is set to `copyfile=True` for the transform, but the fixed and moving files have `copyfile=False`. I think it copies the rigid transform over, symlinks the fixed and moving files so that the saved transform will be in the right place (`affine/<moving>.aff`), runs the function initializing with the copied .aff in the working directory, and then overwrites the copied .aff and deletes the symlinks. Does that sound right? The affines are different in rigid vs affine when the images are different sizes so I think it's working as it should.

I agree that setting `exists=True` makes more sense to the user... it ensures that the user knows exactly what file they're initializing with. It can be confusing when the function is finding its own file in the working directory and I think that would cause problems if the user doesn't realize what is happening. Worst case scenario, they have to figure out the filename, which they should have anyway since they're using it to initialize (most software packages ask for the xfm path so this is more in line with normal use)... or, as you mentioned, they could add `affine.inputs.args` if they really want to use a 1.

`    initialize_xfm = File(copyfile=True, desc="Initialize w/DTITK-FORMAT"
                          "affine", position=5, argstr="%s", exists=True)`

I made that change, and added simple workflows for performing affine and diffeomorphic registration to make sure a user has a functioning pipeline example that adheres to dtitk's requirements (e.g. array size dimension zero must be a power of 2). Does that look good?

effigies:
Nice. https://148-59684390-gh.circle-artifacts.com/0/home/circleci/work/docs/interfaces/generated/workflows.dmri/dtitk.tensor_registration.html

This looks great. If you're all set, I'm happy to merge this.

PR #2516, opened 2018-03-28
CI: Add retry script for Docker commands

### LINKED ISSUE #2489: loops in CircleCI
### Summary
I'm not sure if [for loops in config.yml](https://github.com/nipy/nipype/blob/master/.circleci/config.yml#L51) change anything, I think it does not retry if the command fails. I mostly noticed this when I tried to use the same logic in the Nipype Tutorial repo, where `datalad` command fails quite often. 

A simple example showing that CircleCI tries only once is [here](https://circleci.com/gh/djarecka/nipype_tutorial/64?utm_campaign=build-failed&utm_medium=email&utm_source=notification)

effigies:
Fixes #2489.

- Add `retry_cmd.sh` script
- Replace loops in `.circleci/config.yml`

### DIFF
--- a/tools/retry_cmd.sh
+++ b/tools/retry_cmd.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+#
+# retry_cmd.sh [-n NLOOPS] [-s SLEEP] CMD
+#
+# Retry command until success or pre-specified number of failures
+#
+# 2018 Chris Markiewicz
+# Released into public domain
+
+NLOOPS=3
+TOSLEEP=5
+
+while true; do
+    case "$1" in
+        -n ) NLOOPS="$2"; shift 2 ;;
+        -s ) TOSLEEP="$2"; shift 2 ;;
+        -- ) shift; break ;;
+        * ) break ;;
+    esac
+done
+
+RET=0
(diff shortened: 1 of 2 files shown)


effigies:
There are some limits to this script WRT multi-command behavior.

```Python
$ retry_cmd.sh "echo TEST && exit 2"

File: tools/retry_cmd.sh [hunk 1]
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+#
+# retry_cmd.sh [-n NLOOPS] [-s SLEEP] CMD
+#
+# Retry command until success or pre-specified number of failures
+#
+# 2018 Chris Markiewicz
+# Released into public domain
+
```
oesteban: Doesn't this need to set `set +o pipefail` here to work? After some digging, I think the original problem reported in #2489 is just that the steps are run with `set -eo pipefile` (circle has started to run it with 2.0). Using both -eo will make the script to exit even with that `|| true` pipe.

effigies: No, because those apply to each instance of the shell. It doesn't infect subprocesses.

PR #2517, opened 2018-03-28
ENH: Add MultiObject, ensure/simplify_list; alias old names for 1.x compatibility

effigies:
For the sake of smoothing the transition to 2.0, I propose that we introduce `Input/OutputMultiObject` in 1.x, but leave `MultiPath` traits in place. Downstream projects with dependency >=1.0.3 will be able to write interfaces that require one less change in the future.


oesteban:
Maybe some deprecation warning would be nice here? So to induce people moving into these new `MultiObject`?

effigies:
Well, we're not deprecating in 1.x, but `FutureWarning` might be appropriate.

On the other hand, I want to be careful about warnings that are targeted to developers, but will have a primary effect of annoying or scaring end users.

satra:
talk of other bad semantics, my other favorite one is `filename_to_list`. i think similar to scikit-learn it should be called `ensure_list` and perhaps change `list_to_filename` to `simplify_list`

let's keep warnings out, but update our docs where we reference these objects to the new ones.

effigies:
`Input/OutputMultiPath` and `filename_to_list/list_to_filename` don't appear in the docs, so there's not much left to do.

Incidentally, I couldn't find `ensure_list` in scikit-learn source. Do you have a reference? It would be nice to use consistent naming with other projects.

satra:
@effigies - scikit-learn has ensure_2d as a keyword in some of its functions.

but perhaps the closest is `np.asarray` and that made me think `aslist` might be a good alternative to `ensure_list` and shorter?

effigies:
Hmm. `aslist` feels more ambiguous to me than `ensure_list`. Pybids uses `listify`. We could do `listify`/`unlistify`, but I'm also fine with `ensure`/`simplify`.

nipype/utils/filemanip.py
```diff
@@ -573,13 +573,17 @@ def list_to_filename(filelist):
         return filelist[0]
 
 
+filename_to_list = ensure_list
+list_to_filename = simplify_list
```
satra: perhaps we generate warnings for these as well as multiobject to have people start using the new objects.

```python
def filename_to_list(filelist):
    logger.warn('Please use `ensure_list` instead')
    return ensure_list(filelist)
```

satra:
i would be ok with `listify`/`unlistify` if we wanted to conform with pybids - but fine with the current names as well.

nipype/utils/filemanip.py
effigies: From https://github.com/nipy/nipype/pull/2517#issuecomment-378368408:

> let's keep warnings out, but update our docs where we reference these objects to the new ones.

I think before we start adding warnings for things we don't plan to drop support for until 2.0, we should try to come up with a coherent plan for warnings and deprecations. While I want to make new names available for people trying to future proof, I don't want to annoy users unduly.

I can open an issue on the subject, where we can try to map out a plan. Seem reasonable?

satra: roger.

PR #2520, opened 2018-03-30
FIX: Validate Function interface input names when set

### LINKED ISSUE #1114: Function interface and input_names
I'm not sure if this is worth an issue, but it's something that can make debugging a complicated pipeline a little difficult: if you have forget to add an input for a function node to its `input_names` at initialization, a function node will still allow you to set it using `node.inputs.in1 = 'foo'` and this `in1` will show up in the node's report in the working directory, even though the node itself, when it runs, will be unable to see this input (and so it will either use a default value, if it's provided, or fail because there aren't enough inputs).

For example:

    def func(in1,in2=0):
        return in1+in2

    f_node = pe.Node(interface=util.Function(input_names=['in1'],output_names=['sum'],function=func),name='node')
    f_node.inputs.in1 = 1
    f_node.inputs.in2 = 1

Then, if you try to run this node, it will run without any errors and return a `1`, instead of the expected `2`. Furthermore, if you look in the report  in the node's directory, `in2 = 1` will be listed as an input; loading the node using `node=loadpkl(_node.pklz)` and checking `node.inputs` will show the same thing. It's only if you run the node and debug it will you notice that `in2=0`.

Doesn't it make sense to prevent users from being able to add values to `node.inputs` that aren't listed in `input_names`? Why does this behavior happen?

effigies:
This is serious overkill, in the sense that I think the better solution is simply to ignore `input_names` altogether, but it was fun to figure out how to do this.

Fixes #1114.

- Adds a new `CheckHidden` trait type that delegates deciding whether a trait can be set based on the name to the interface class.
- Defines "hidden" traits of `Function` to be those that are in the function protocol but not `input_names`
- Sets any post-initialization traits as `CheckHidden` traits

Closes #2525.


nipype/interfaces/utility/tests/test_wrappers.py
```diff
@@ -50,6 +52,78 @@ def increment_array(in_array):
...
+def test_function_string_reset():
+    fstr = "def f(a, b):\n\treturn a + b"
+    gstr = "def g(a, b, c):\n\treturn a + b - c"
+    hstr = "def h(b, c):\n\treturn b - c"
+
+    fun = utility.Function(function=fstr)
+    assert fun._input_names == ['a', 'b']
+    fun.inputs.a = 5
+    fun.inputs.b = 3
+    fun.inputs.c = 2  # Unused trait
```
satra: @effigies - i'm curious as to your thought process as to why this should not raise an `Exception`

effigies: Because a whitelist would be a pain to maintain. IPython adds attributes to objects (or checks for them, which causes a promiscuous HasTraits to add them), and any number of environments might do something like that. Node checks for or sets num_threads, too.

It seemed to me that making sure we don't lie about the inputs to a function is more important than policing extra attributes that wouldn't be expected to change the interface's behavior.

Willing to be overruled, here, but I think we'll need a whitelist to do this right, otherwise.

satra: would this work?

https://docs.python.org/2.7/library/inspect.html#inspect.getargspec (in both py2 and py3)

if we get list of function variables using the function above, then:

1. for the scenario where input_names are specified, we restrict to names
2. for the scenario where all inputs are allowed, we restrict to the function variables, with any **kwargs allowing for any name.

```python
In [2]: from inspect import getargspec

In [11]: def func(a, b, c=1, **kwargs):
    ...:     pass

In [12]: getargspec(func)
Out[12]: ArgSpec(args=['a', 'b', 'c'], varargs=None, keywords='kwargs', defaults=(1,))

In [13]: def func(a, b, c=1):
    ...:     pass

In [14]: getargspec(func)
Out[14]: ArgSpec(args=['a', 'b', 'c'], varargs=None, keywords=None, defaults=(1,))
```

inside the traits object setter we could use:
```python
if name not in argspec.args and argspec.keywords is None:
    raise ValueError('{ } is not a valid input for the function.'.format(name))
```

for adding traits:
```python
if argspec.keywords is None and set(input_names) - set(argspec.args):
    raise ValueError('input_names containers undefined variable: {}'.format(set(input_names) - set(argspec.args)))

# add the input_names as traits
```

if we wanted the full possibilities, in py3 one can specifically use:

https://docs.python.org/3/library/inspect.html#inspect.getfullargspec

```
In [20]: from inspect import getfullargspec

In [22]: def func(a, b, *, c=1, **kwargs):
    ...:     pass

In [23]: getfullargspec(func)
Out[23]: FullArgSpec(args=['a', 'b'], varargs=None, varkw='kwargs', defaults=None, kwonlyargs=['c'], kwonlydefaults={'c': 1}, annotations={})
```

effigies: I can have a look at this. At the least, it looks like a plausible replacement for the custom `_get_varnames` I'm using here.

I'm not sure how this resolves the issue of environments like IPython attempting to plop things like `_ipython_canary_method_should_not_exist_` into every object it sees. Did you have something in mind?

satra: i think that may be because of how `_input_names` is currently used when running the interface.  i think a lot of this can be taken care of with traits. i'll try to push a rough version on my bus ride in 30 mins.

effigies: Oh, I do like the possibility of using annotations to make function interfaces typed. Perhaps for Nipype 2, `SimpleInterface` and `Function` could be largely merged:

```Python
def func(a: int, b: float, *, c=1: int) -> {'out1': int, 'out2': float):
    ...
```

`Function(func)` would then have:

```Python
class input_spec:
    a = traits.Int()
    b = traits.Float()
    c = traits.Int(1, usedefault=True)

class output_spec:
    out1 = traits.Int()
    out2 = traits.Float()
```

It would probably make sense to try to make `function_str` immutable once set, for simplicity.

satra:
sorry this took a little longer: see #2525 (i'm pretty sure it won't work across py27 and py36, but just wanted to get the concepts across, and perhaps we should create those tests.

effigies:
Okay. So if I can summarize your specific proposals from #2525, it looks like the following:

* Only permit unknown traits to be set if `**kwargs` (or equivalent) is in function parameters.
* Disallow `input_names` to exclude parameters that do not have default values.
* Use `inspect.getfullargspec` when available to enable checking for keyword-only arguments

These seem reasonable modifications to my current approach. I do still prefer the delegating trait to throwing errors in callback functions. I'm assuming that you saw the mess of error messaging that occurred when you did that?

Are there other salient points that you think are worth incorporating? I'll also make some comments on your PR with my thoughts, given my experience playing with this PR.

satra:
@effigies - correct.

two possible additions to your list:

- Make keyword args with default values part of the immutable input spec (as currently implemented in my PR)
- Allow users to override any keyword parameter, even if not in input names. 

i don't have a strong feeling about either of these. 

yes, throwing errors in callback functions is good. feel free to close my PR, it was less of a WIP than a concept demo.

effigies:
> throwing errors in callback functions is good

Just to be clear, my experience is that it's actually bad. Do you prefer using the callbacks to a delegating traits? (Or is that delegation what you mean by callback, here?)

satra:
> (Or is that delegation what you mean by callback, here?)

yes sorry - that's what i intended to mean.

effigies:
@satra A couple questions:

1) Is the ability to change `function_str` an actually desired feature?
2) Do we want to support `*args`? My inclination is not to make a fuss about its presence, but since we pass all arguments by name, it will never be populated. (You can't populate it with `args=(...)`, so it would have to be done positionally.)

With regard to:

>  - Make keyword args with default values part of the immutable input spec (as currently implemented in my PR)
> - Allow users to override any keyword parameter, even if not in input names.

These seem mutually exclusive. And I interpret the latter as saying "Ignore `input_names`; always respect values set in `inputs`." If that's the case, I'm fully in favor of the latter. I think it would be immensely simplifying, and is along the lines of what I was trying to suggest in https://github.com/nipy/nipype/issues/1114#issuecomment-347880666.

effigies:
I've pushed an updated version. The changes:

* `function_str` can be set, either via `Function(function=f)` or `Function().inputs.function_str = getsource(f)`, but once set cannot be changed.
* If `function` and `input_names` are set in `__init__`, then arguments with default values that are not in `input_names` are immutable and raise `TraitError` on attempts to change.
* Inputs that don't match arguments can only be set if the function takes a `**kwargs` parameter.
* All inputs are passed to the function.

In response to your last post, the following could still be done:

* If function takes `*args`, we could add an `args` parameter. This would require re-parsing the function at runtime, and sorting all of the inputs to ensure correct placement. I'm inclined not to bother until someone has a need for this functionality.
* Make immutable/hidden inputs modifiable with `raise_on_override` or similar. This feels like overkill. People can just drop `input_names` if they want access to all inputs.

nipype/interfaces/utility/tests/test_wrappers.py
```diff
@@ -50,6 +52,74 @@ def increment_array(in_array):
...
+    fun2 = utility.Function()
+    fun2.inputs.function_str = fstr
+
+    for fun in (fun1, fun2):
+        assert fun._input_names == ['a', 'b']
+        fun.inputs.a = 5
+        fun.inputs.b = 3
+        assert fun.run().outputs.out == 8
+
+        fun.inputs.function_str = gstr
```
satra: does this not raise an error?

nipype/interfaces/utility/wrappers.py
```diff
@@ -25,9 +25,91 @@
 iflogger = logging.getLogger('interface')
 
 
-class FunctionInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
+try:
+    from inspect import getfullargspec
+except ImportError:
+    def getfullargspec(func):
```
satra: nice backport :)

nipype/interfaces/utility/tests/test_wrappers.py
effigies: Logs a warning.

satra: i would just raise error just like the other immutable traits. if we have decided that this is not possible, we should throw an exception.

effigies: Well, I implemented those just by making `traits.Enum(value)`. For this I went ahead and made a full-blown `SetOnce` trait. You can pass it an inner trait, so `function_str = SetOnce(traits.Str)`.

Added tests to make sure you can pass the same value multiple times without causing an error.

satra:
this looks good. the one additional thing may be to add a few examples showing the options in the doctests.

it seems like certain things are failing in the tests (or we are catching reuse of functions :) ).

effigies:
Yeah, working on the tests. I'll push and annotate changes in other tests.

nipype/pipeline/engine/tests/test_utils.py
```diff
@@ -157,70 +157,3 @@ def test_provenance(tmpdir):
...
-
-
-@pytest.mark.skipif(
-    sys.version_info < (3, 0), reason="the famous segfault #1788")
-def test_mapnode_crash(tmpdir):
-    """Test mapnode crash when stop_on_first_crash is True"""
-    cwd = os.getcwd()
-    node = pe.MapNode(
-        niu.Function(
-            input_names=['WRONG'],
```
effigies: @satra These tests depend on bad input names. I don't know if they can be rewritten to test the segfaulting code path, or disallowing `input_name` mismatches prevents it entirely.

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -315,7 +315,6 @@ def func2(a):
         name='n2')
     w1 = pe.Workflow(name='test')
     modify = lambda x: x + 1
-    n1.inputs.a = 1
```
effigies: Not sure if there was a reason for these?

effigies:
Hmm. Something is causing tests to time out, and it's not the new trait like I guessed. I'll try to track down specifically which are failing.

nipype/interfaces/utility/wrappers.py
```diff
@@ -25,8 +25,91 @@
...
+        from copy import deepcopy
+        values = deepcopy(self.trait_get(), memo)
+        dup = self.clone_traits(memo=memo)
+        orig_traits = self.traits()
+        dup_traits = dup.traits()
+        for name, value in values.items():
+            if name not in dup_traits:
+                dup.add_trait(name, deepcopy(orig_traits[name], memo))
+        dup.trait_set(**values)
+        return dup
```
effigies: Okay, it's the input spec. There's something about moving to a (less) dynamic input spec that breaks the following tests:

* `nipype/pipeline/plugins/tests/test_multiproc_nondaemon.py` (all)
* `nipype/pipeline/engine/tests/test_engine.py::test_serial_input
* `nipype/pipeline/engine/tests/test_workflows.py::test_outputs_removal_wf[MultiProc-*-*]`

It's not throwing errors, and I'm having trouble tracking down the specific code that is deadlocking.

effigies:
I don't think I'm going to resolve these issues this week. Punting to next month's release.

nipype/pipeline/engine/tests/test_engine.py
satra: nope.

nipype/pipeline/engine/nodes.py
```diff
@@ -619,7 +618,6 @@ def _run_command(self, execute, copyfiles=True):
             inputs=self._interface.inputs.get_traitsfree())
 
         if copyfiles:
-            self._originputs = deepcopy(self._interface.inputs)
```
satra: is this no longer being used in report.rst ?

effigies:
```Shell
$ grep -rI _originputs nipype
nipype/pipeline/engine/nodes.py:        self._originputs = None
nipype/pipeline/engine/nodes.py:            self._originputs = deepcopy(self._interface.inputs)
```

satra: ah yes - this was changed in oscar's node refactor.

effigies:
LOL.

effigies:

PR #2521, opened 2018-03-30
FIX: Build cmdline from working directory

### LINKED ISSUE #2518: expert.opts are being written to the root directory
### Summary
`expert.opts` file is being written to the root directory which is not writable under singularity. 
See https://neurostars.org/t/singularity-fmriprep-export-opts-error/1491 for more details.

effigies:
This fix supports an earlier ridiculous hack (#1857) that was broken by properly managing working directories (#2368).

Fixes #2518.

- Construct `cmdline` within the working directory, for the probably one case in which the `cmdline` depends on the working directory.
- Adds and tests an `indirectory` context manager, because I hate the `os.getcwd(); os.chdir(); CONTENT; os.chdir()` pattern.

There may be a more elegant way to get input traits into a file, but it's not clear to me what that is at this point, so I went for the cheap solution.

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -27,7 +27,7 @@ from ...utils.misc import flatten, unflatten, str2bool, dict_diff
 from ...utils.filemanip import (md5, FileNotFoundError, filename_to_list,
                                 list_to_filename, copyfiles, fnames_presuffix,
                                 loadpkl, split_filename, load_json, makedirs,
-                                emptydirs, savepkl, to_str)
+                                emptydirs, savepkl, to_str, indirectory)
 
 from ...interfaces.base import (traits, InputMultiPath, CommandLine, Undefined,
                                 DynamicTraitedSpec, Bunch, InterfaceResult,
@@ -627,7 +627,8 @@ class Node(EngineBase):
             self._interface.__class__.__name__)
         if issubclass(self._interface.__class__, CommandLine):
             try:
-                cmd = self._interface.cmdline
+                with indirectory(outdir):
+                    cmd = self._interface.cmdline
             except Exception as msg:
                 result.runtime.stderr = '{}\n\n{}'.format(
                     getattr(result.runtime, 'stderr', ''), msg)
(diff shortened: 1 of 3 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -627,7 +627,8 @@ def _run_command(self, execute, copyfiles=True):
             self._interface.__class__.__name__)
         if issubclass(self._interface.__class__, CommandLine):
             try:
-                cmd = self._interface.cmdline
+                with indirectory(outdir):
```
oesteban: does outdir always exist at this point?

effigies: We create the `command.txt` file in line 637, just below here, and copy files to working directory just above.

A quick scan didn't show me where the directory is created, but I'm reasonable confident it exists.

oesteban: The directory is created here, inside `run()`:

https://github.com/effigies/nipype/blob/555a905f623be290a16c5fede1ea3a1558ac1f57/nipype/pipeline/engine/nodes.py#L469

Therefore, it is not granted that it will exist at this point. We could add a `makedirs` call to the context manager.

oesteban: but you can always call cmdline manually before run.

inside a workflow this does not apply, but many people use the interfaces and call cmdline in their ipython interactive sessions.

effigies:
This is for when a node is run, not an interface. Dropping files in the
local cwd is standard for interfaces.

satra:
@effigies - that is indeed an important conceptual distinction.

interfaces do whatever the underlying command would do without trying to change the behavior of the command.

nodes have additional policies (fixed working directory), which is why interfaces have some additional metadata for nodes (`copyfile` in particular).

effigies:
Not quite sure if we're all on the same page here. Just to clarify: It is my position that this doesn't need changing at this point. `Node` will create the directory prior to the `indirectory()` context, so `makedirs` is not required, and running the interface will not hit the `indirectory()` context manager code, nor should it, as interfaces do not change directories.

@satra @oesteban If you disagree or are expecting changes, please let me know.

oesteban:
Gotcha, no disagreement here

PR #2522, opened 2018-03-30
ENH: Add an activation count map interface

oesteban:
A simple calculation of ACMs, adapted from @jokedurnez's code: https://github.com/poldracklab/CNP_task_analysis/blob/61c27f5992db9d8800884f8ffceb73e6957db8af/CNP_2nd_level_ACM.py


nipype/algorithms/stats.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Adapted from: https://github.com/poldracklab/CNP_task_analysis/\
+    blob/61c27f5992db9d8800884f8ffceb73e6957db8af/CNP_2nd_level_ACM.py
+    """
+    input_spec = ACMInputSpec
+    output_spec = ACMOutputSpec
+
+    def _run_interface(self, runtime):
+        allmaps = nb.concat_images(
+            [nb.load(f) for f in self.inputs.in_files])
```
effigies: I assume you want to `.get_data()` somewhere here, before making numeric comparisons.

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,68 @@
...
+    blob/61c27f5992db9d8800884f8ffceb73e6957db8af/CNP_2nd_level_ACM.py
+    """
+    input_spec = ACMInputSpec
+    output_spec = ACMOutputSpec
+
+    def _run_interface(self, runtime):
+        allmaps = nb.concat_images(
+            [nb.load(f) for f in self.inputs.in_files]).get_data()
+        acm_pos = (allmaps > self.inputs.threshold).astype(
+            float).mean(3)
```
effigies: What about:

```Python
acm_pos = np.mean(allmaps > self.inputs.threshold, axis=3, dtype=np.float32)
```

I think it reads more cleanly, and it'll make any line breaks cleaner looking, too. (I also prefer to use explicit `axis` and `dtype` keywords, just to avoid having to look up the order when I read. And I prefer to use actual dtypes, just to avoid depending on numpy magic that they can change at any time.)

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Adapted from: https://github.com/poldracklab/CNP_task_analysis/\
+    blob/61c27f5992db9d8800884f8ffceb73e6957db8af/CNP_2nd_level_ACM.py
+    """
+    input_spec = ACMInputSpec
+    output_spec = ACMOutputSpec
+
+    def _run_interface(self, runtime):
+        allmaps = nb.concat_images(
+            [nb.load(f) for f in self.inputs.in_files]).get_data()
```
effigies: Just realizing you can drop the load step here. concat_images will take file names.

effigies:
LGTM. Have you tested it with some data, just to be sure?

effigies:
I guess, really, we could write a test that randomly generates some small (e.g. 10x10x10) images and runs the interface. Shouldn't add significant run time. Your call if it seems worth the effort.

oesteban:
I'll add the test :+1:

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,66 @@
...
+                          axis=3, dtype=np.float32)
+        acm_diff = acm_pos - acm_neg
+
+        nii = nb.load(self.inputs.in_files[0])
+        self._results['out_file'] = os.path.join(
+            runtime.cwd, 'acm_diff.nii.gz')
+        self._results['acm_pos'] = os.path.join(
+            runtime.cwd, 'acm_pos.nii.gz')
+        self._results['acm_neg'] = os.path.join(
+            runtime.cwd, 'acm_neg.nii.gz')
```
effigies: Maybe let's make this:

```Python
template_fname = self.inputs.in_files[0]
ext = split_filename(template_fname)[2]
fname_fmt = os.path.join(runtime.cwd, 'acm_{}' + ext).format

self._results['out_file'] = fname_fmt('diff')
self._results['acm_pos'] = fname_fmt('pos')
self._results['acm_neg'] = fname_fmt('neg')

img = nb.load(template_fname)
```

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,66 @@
...
+
+        nb.Nifti1Image(
+            acm_diff, nii.affine, nii.header).to_filename(self._results['out_file'])
+        nb.Nifti1Image(
+            acm_pos, nii.affine, nii.header).to_filename(self._results['acm_pos'])
+        nb.Nifti1Image(
+            acm_neg, nii.affine, nii.header).to_filename(self._results['acm_neg'])
```
effigies: And then we would make these:

```Python
img.__class__(acm_diff, img.affine, img.header).to_filename(self._results['out_file'])
img.__class__(acm_pos, img.affine, img.header).to_filename(self._results['acm_pos'])
img.__class__(acm_neg, img.affine, img.header).to_filename(self._results['acm_neg'])
```

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,66 @@
...
+                                  'corresponds to a two-sided test of p<.10')
+
+
+class ACMOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='output activation count map')
+    acm_pos = File(exists=True, desc='positive activation count map')
+    acm_neg = File(exists=True, desc='negative activation count map')
+
+
+class ACM(SimpleInterface):
```
effigies: I think I'd prefer a more descriptive name like `ActivationCount`.

effigies:
See also: oesteban/nipype#7

oesteban:
Some second opinion about this addition to nipype? @djarecka, @mgxd, @satra or @chrisfilo ?

Maybe a quick double check on the correctness from @jokedurnez ?

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,68 @@
...
+from ..utils.filemanip import split_filename
+
+
+class ActivationCountInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiPath(File(exists=True), mandatory=True,
+                              desc='input file, generally a list of z-stat maps')
+    threshold = traits.Float(1.65, usedefault=True,
+                             desc='binarization threshold. A z-value of 1.65 '
+                                  'corresponds to a two-sided test of p<.10')
+
```
djarecka: I'm wondering if it's better to leave `threshold ` without default value, so users have to really choose one....

nipype/algorithms/tests/test_stats.py
```diff
@@ -0,0 +1,22 @@
...
+    for fname in in_files:
+        nb.Nifti1Image(np.random.normal(size=(5, 5, 5)),
+                       np.eye(4)).to_filename(fname)
+
+    acm = ActivationCount(in_files=in_files)
+    res = acm.run()
+    diff = nb.load(res.outputs.out_file)
+    pos = nb.load(res.outputs.acm_pos)
+    neg = nb.load(res.outputs.acm_neg)
+    assert np.allclose(diff.get_data(), pos.get_data() - neg.get_data())
```
djarecka: should we also check some statistics `pos.get_data` itself?

effigies: The main things I can think of that should be true of performing this on some random noise would be `np.all(pos.get_data() > 0)` and `np.all(neg.get_data() > 0)`.

nipype/algorithms/stats.py
effigies: I don't have an opinion here.

nipype/algorithms/tests/test_stats.py
djarecka: here is an example: https://github.com/oesteban/nipype/pull/9/commits/6e53f016f819bd6330fa63f7d9002cd6c565432b

effigies: That looks valid, but I'm not fully clear on what assurances this gets us for the interface. I guess, roughly, that Gaussian noise is being filtered as we'd expect?

I'm okay if this test goes in, but my overall feeling is that a useful test asserts properties that should be true given any real or simulated inputs, and this seems tied to the specific distribution of simulated random variates.

djarecka: you usually use specific cases for tests, and often some simple cases, so you know the answer.

The interface should work for any real or simulated inputs, but I'm checking here for a specific distribution (exactly the same as @oesteban used in the original test). Obviously I'm not able to predict answers for completely random set of numbers, but for the normal distribution I can.

I just thought that this adds extra checks to the original @oesteban test, that was only checking `pos-neg == diff`, but I won't insist to include it.

djarecka: we can also just generate a small array and just calculate the expected output of the interface.

nipype/algorithms/stats.py
oesteban: Makes sense, I'll update accordingly.

effigies:
@oesteban I think you said you're going to drop the threshold. Anything else before merging?

oesteban:
Provided that the tests pass, this should be good to go.

nipype/algorithms/stats.py
```diff
@@ -0,0 +1,68 @@
...
+)
+from ..utils.filemanip import split_filename
+
+
+class ActivationCountInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiPath(File(exists=True), mandatory=True,
+                              desc='input file, generally a list of z-stat maps')
+    threshold = traits.Float(
+        mandatory=True, desc='binarization threshold. E.g. a threshold of 1.65 '
+                             'corresponds to a two-sided test of p<.10')
```
effigies: "two-sided Z-test"?

PR #2533, opened 2018-04-10
changing arguments names to default values for traits (closes #2532)

### LINKED ISSUE #2532: threshold in `fsl.FILM_GLS`
### Summary

I'm wondering why the `threshold` trait for `FILM_GLS` for `5.07` (and 5.05) takes `-1000` as [default value](https://github.com/nipy/nipype/blob/master/nipype/interfaces/fsl/model.py#L648). Is this just a typo? 

motivated by [this qst](https://neurostars.org/t/nipype-filmgls-threshold-vs-fsl-film-gls-command/1528)

djarecka:


satra:
also `usedefault=False` should be caught by check specs.

nipype/interfaces/afni/utils.py
```diff
@@ -1696,13 +1697,14 @@ class NwarpCatInputSpec(AFNICommandInputSpec):
     inv_warp = traits.Bool(
         desc='invert the final warp before output', argstr='-iwarp')
     interp = traits.Enum(
+        'wsinc5',
         'linear',
         'quintic',
         'wsinc5',
```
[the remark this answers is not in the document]
djarecka: what do you mean? this element needs to be removed (so "linear" will be default) or `usedefault=True` should be removed?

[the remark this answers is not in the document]
djarecka: I have two more traits as this one as `interp` or `ainterp`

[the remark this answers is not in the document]
satra: `'wsinc5'` is listed twice - only once is necessary - for enum's the first element automatically becomes default.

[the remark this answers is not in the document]
djarecka: oh, you are right, I forgot to remove the second one here. I've changed the order because `Enum` takes the first element as a default, and in the previous version it was `default='wsinc5'`.

Since the previous way of setting default was not working doctests were ignoring these elements, so I'd have to change a few doctests,  unless you believe that these traits shouldn't have default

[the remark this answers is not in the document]
djarecka: I can create a list of all doctests I'd have to change to pass now the tests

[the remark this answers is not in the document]
satra: sounds good - just don't do `make specs` yet. that should happen after everything has been reviewed.

djarecka:
@satra - what do you mean that `usedefault=False` should be caught by check `specs`? there are several places in nipype that uses it

satra:
@djarecka - just like `mandatory=False` is not useful, `usedefault=False` is also not useful. hence if a default should not be used, those keywords should be left out. `make specs` reports all the inconsistencies of any keyword. if it is not catching `usedefault=False`, we should add it.

djarecka:
@satra - ok, so you mean warnings. Indeed `make specs` returns warnings for `mandatory=False` (we have around 40 of them), but we don't have warnings for `usedefault=False`

djarecka:
Ok, so in the last commit I changes tests in order to pass them, so would ask for extra checking:

-  in `ants/resampling.py::ApplyTransforms` I added `float` to `cmdline`
- in `afni/utils.py::[NwarpApply, NwarpCat]` I dded `interp` and `ainterp`
- in `niftyseg/em.py::EM` I added `bc_order`, `bc_thresh`,  `max_iter` and  `min_iter` 
- had to change hash in `base/tests/test_core.py` due to changest in `ants.Registration`

djarecka:
As I mentioned at the beginning there are traits that have specified default values and do not have `usedefault=True`. I'm guessing that authors who are specifying default values would like them to be used so it's a bug, but I'd need a few more hours to find all places. In most cases the name of the argument (`default`, `default_value`, etc.)  was not used, so it's hard to search...

satra:
@djarecka - are you going to add warning for `usedefault=False`

satra:
other than that this looks good.

djarecka:
@satra - oh, yes, I can add warning. I thought that I can just remove them too.

I was thinking whether to search other traits that don't have `usedefault=True` and define some default values, just not sure how long it can take. can do only for most popular interfaces, and can always open a new pr

djarecka:
@satra many default values in traits are set just using the first position, like [here](https://github.com/nipy/nipype/blob/master/nipype/interfaces/spm/preprocess.py#L83).

give me one day, will try to check a few directories and see how it goes.

satra:
@djarecka  - we do remove them whenever we ask people to run `make specs` and we review, but we should improve that process. we should really start adding bots to github.

djarecka:
we have many warnings for `exists`, e.g. `nipype.interfaces.dtitk.registration:AffineTask:Inputs:fixed_file:exists`. Why do we need them? What is wrong with this trait?

satra:
this is being cleaned up in #2514

djarecka:
@satra - so you don't want me to clean `dtitk`? I've already cleaned `mandatory=False`. can always revert it...

Also, I'm still not sure when `exists` should be reported in the `checkspecs.py`.

satra:
@djarecka - don't clean dtitk.

djarecka:
ok, reverted my changes to dtitk

I added a simple check for default values, but it's not perfect. I wasn't sure what is a good way to check if an author of an interface defines any default, so I'm checking `if trait.default`, so I'm checking if the default value seen by `traits` is not the default "default value", i.e. it's not 0, False, [], etc.
Unfortunately, sometimes in nipype default value is set to 0 or False, and my check won't be able to catch this.

Because the way how traits treats default for various trait types, I also had to make several exceptions: 
- for `Enum` traits always treat the first value as default, so have no way of checking if the author wanted to have default value or not (excluded from the check)
- for `Tuple` traits takes tuple of default values of inner traits as default tuple, but doesn't use it even if `usedefault=True` (excluded from the check
- for Range traits assumes that `default == low` if no `value` is provided (if `low == default` I don't return any warning) 

@satra you can check the list of warnings, if my rules and the list make sense, I can correct the interfaces

satra:
looks reasonable - could you please merge with master, run make specs, and update the PR?

djarecka:
ok, I merged and updated the code to remove most warnings. We had many places without `usedefault`, hopefully this is really what it should be...

I also exclude traits that have `xor` or `requires`, IMO it doesn't make sense to provide any default value for this kind of traits, but I just kept them unchanged.

One trait that is still in warnings that I wasn't sure about is this output name: https://github.com/nipy/nipype/blob/master/nipype/interfaces/ants/segmentation.py#L1263
with `%d` inside string. Should this have `usedefault=True`?

djarecka:
so unfortunately I have to again change some tests, I changed hash in one of them, but I'm not sure what to do with:
- `workflows/rsfmri/fsl/tests/test_resting.py`
- `algorithms/tests/test_CompCor.py`
they're comparing to array of numbers that i don't know where does it come from. Is it any "ground truth" or can I change it?

nipype/interfaces/fsl/aroma.py
```diff
@@ -28,7 +28,8 @@ class ICA_AROMAInputSpec(CommandLineInputSpec):
         xor=['feat_dir'],
         desc='volume to be denoised')
     out_dir = Directory(
-        'out', genfile=True, argstr='-o %s', desc='output directory')
+        'out', usedefault=True, genfile=True,
```
effigies: Judging by `_gen_filename`, this wasn't intended to be a default value. @jdkent can you clarify your intent here?

nipype/interfaces/fsl/model.py
```diff
@@ -56,7 +56,7 @@ class Level1DesignInputSpec(BaseInterfaceInputSpec):
         desc=("which regressors to make orthogonal e.g., "
               "{1: {0:0,1:0,2:0}, 2: {0:1,1:1,2:0}} to make the second "
               "regressor in a 2-regressor model orthogonal to the first."),
-        default={})
+        value={}, usedefault=True)
```
effigies: `{}` is already the default.

nipype/interfaces/mrtrix3/tracking.py
```diff
@@ -262,7 +262,8 @@ class Tractography(MRTrix3Base):
     >>> tk.inputs.roi_mask = 'mask.nii.gz'
     >>> tk.inputs.seed_sphere = (80, 100, 70, 10)
     >>> tk.cmdline                               # doctest: +ELLIPSIS
-    'tckgen -algorithm iFOD2 -mask mask.nii.gz -seed_sphere \
+    'tckgen -algorithm iFOD2 -samples 4 -output_seeds out_seeds.nii.gz \
+-mask mask.nii.gz -seed_sphere \
```
effigies: Just a note you can drop the backslashes from this output string.

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -183,6 +183,7 @@ class TensorMetricsInputSpec(CommandLineInputSpec):
         argstr='-value %s', desc='output selected eigenvalue(s) file')
     component = traits.List(
         [1, 2, 3],
+        usedefault=True,
```
effigies: From the [docs](https://mrtrix.readthedocs.io/en/latest/reference/commands/tensor2metric.html):

> **-num sequence** specify the desired eigenvalue/eigenvector(s). Note that several eigenvalues can be specified as a number sequence. For example, ‘1,3’ specifies the principal (1) and minor (3) eigenvalues/eigenvectors (default = 1).

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -266,7 +267,7 @@ class ComputeTDIInputSpec(CommandLineInputSpec):
         desc='specify output image data type')
     use_dec = traits.Bool(argstr='-dec', desc='perform mapping in DEC space')
     dixel = File(
-        'dixels.txt',
+        'dixels.txt', usedefault=True,
```
effigies: It's not clear to me from the [docs](https://mrtrix.readthedocs.io/en/latest/reference/commands/tckmap.html) that this is set by default.

nipype/interfaces/niftyseg/em.py
```diff
@@ -54,16 +54,21 @@ class EMInputSpec(CommandLineInputSpec):
...
+        default_value=100,
+        usedefault=True,
         desc='Maximum number of iterations')
 
     min_iter = traits.Int(
-        argstr='-min_iter %s', default=0, desc='Minimun number of iterations')
+        argstr='-min_iter %s',
+        default_value=0,
+        usedefault=True,
+        desc='Minimun number of iterations')
```
effigies: Minimum

nipype/algorithms/confounds.py
```diff
@@ -391,12 +391,13 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
         'extraction')
     use_regress_poly = traits.Bool(
         True,
+        usedefault=True,
```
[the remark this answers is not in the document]
djarecka: ok, will remove it. Actually `DeprecationWarning` is not displayed after my changes, only when you're changing later, so it's better to remove the default

nipype/interfaces/afni/utils.py
```diff
@@ -1633,11 +1635,10 @@ class NwarpApplyInputSpec(CommandLineInputSpec):
         'tricubic',
         'quintic',
         'triquintic',
-        'wsinc5',
         desc='specify a different interpolation method than might '
         'be used for the warp',
         argstr='-ainterp %s',
-        default='wsinc5')
+        usedefault=True)
```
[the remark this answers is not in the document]
djarecka: ok, will remove

nipype/interfaces/afni/utils.py
```diff
@@ -1841,7 +1842,6 @@ class OneDToolPyInputSpec(AFNIPythonCommandInputSpec):
     show_cormat_warnings = traits.File(
         desc='Write cormat warnings to a file',
         argstr="-show_cormat_warnings |& tee %s",
-        default="out.cormat_warn.txt",
         usedefault=False,
```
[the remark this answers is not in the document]
djarecka: yes, should add warnings

nipype/interfaces/fsl/model.py
djarecka: that's true, sometimes I just kept it if author wanted to be explicit, but can remove it

nipype/interfaces/mrtrix3/tracking.py
djarecka: it doesn't work without

nipype/interfaces/mrtrix3/utils.py
djarecka: ok, can keep just the first one

nipype/interfaces/mrtrix3/utils.py
djarecka: we can remove `usedefault=True`, but I'd also remove the filename, imo it might me misleading

nipype/interfaces/fsl/aroma.py
jdkent: haha, my intent when I made this was to get it to work, I kluged the out_dir definition from [melodic](https://github.com/nipy/nipype/blob/5c3c4137da078a5682e46960c5623b2b4690e692/nipype/interfaces/fsl/model.py#L1606) and [your comment](https://github.com/nipy/nipype/pull/1993/commits/fafdf220669490e57dbf6d2768681b239c74a3be#r114595613) from when I first tried making the wrapper. what a blast from the past!

djarecka: so should `out_dir` have a default value? or should i remove `'out', usedefault=True,`?

effigies: Okay, so in practice, it's been setting `out_dir=os.getcwd()`, as opposed to `'out'`. Which is the preferable default?

* `out_dir = Directory('out', usedefault=True, mandatory=True, argstr='-o %s', desc='output directory')`
* `out_dir = Directory(genfile=True, argstr='-o %s', desc='output directory')`

I have a slight preference to the former, just to avoid cluttering the local directory (though maybe 'aroma_output' would be a better default). In which case we should also drop `_gen_filename()` method.

effigies: But it is a default value. It's just specified with a method rather than a trait...

Sorry if I'm being dense, I'm just not following the logic. Anyway, either solution is fine. And @jdkent's is less work.

djarecka: yes, `_gen_filename` will generate an `out_dir`, but I agree with @effigies if we do have a default value it might be better to explicitly do it in the trait definition. Unless you really prefer `os.getcwd` as default.

jdkent: After reading @effigies comment, I agree with him and you. So I reverse my vote, I have a preference for this:
```
out_dir = Directory('out', usedefault=True, mandatory=True, argstr='-o %s', desc='output directory')
```

which means we can delete `_gen_filename`

djarecka: ok, so if we all agree, I'll remove  `_gen_filename` and will use `out` as default

effigies: Yup, sounds good.

djarecka: @effigies - just an extra question, do we use `mandatory=True` if we already have a default value and `usedefault=True`?

effigies: I think it's a good idea. The reason I suggest it is because somebody could set it as `Undefined` manually or (god knows why) pipe in a value that fails to be defined from a previous node.

```Python
In [1]: from nipype.interfaces.base import *

In [2]: class A(BaseInterface):
   ...:     class input_spec(TraitedSpec):
   ...:         a = traits.Str('test', usedefault=True)
   ...:     def _run_interface(self, runtime):
   ...:         print(self.inputs.a)
   ...:         return runtime
   ...:     

In [3]: a = A()

In [4]: a.inputs
Out[4]: 

a = test

In [5]: a.inputs.a = Undefined

In [6]: _ = a.run()
<undefined>
```

So unless it really is optional, I would recommend flagging it `mandatory`.

djarecka: sure, I can add it. I was just asking because in many places that we use `usedefault=True`  we don't have `mandatory=True` (and I believe some of them are mandatory). 

But that was just general question, adding  `mandatory=True` won't hurt for sure, so will do it.

effigies: Ah. Yeah, I think we should probably add `mandatory=True` to those cases, when we see them. Alas, nipype interface metadata will remain inaccurate for a while longer.

effigies:
It explicitly says optional. I would drop it and add it to `desc` as an example. Like:

```Python
out_intensity_fusion_name_format = traits.Str(
    argstr="",
    desc='Optional intensity fusion image file name format. '
         '(e.g. "antsJointFusionIntensity_%d.nii.gz")')

atsuch:
Hello @djarecka @satra, @effigies,

I think there might be an issue here. My lab mate noticed a considerable change in the output of FSL topup, and found the culprit to be the `usedefault=True` introduced @ 7871798.

Previously, command for the popup with minimum required input was;

  'topup --config=b02b0.cnf --datain=topup_encoding.txt \
--imain=b0_b0rev.nii --out=b0_b0rev_base --iout=b0_b0rev_corrected.nii.gz \
--fout=b0_b0rev_field.nii.gz --jacout=jac --logout=b0_b0rev_topup.log \
--rbmout=xfm --dfout=warpfield'

But because of the `usedefault=True`, the same minimum input gives;

'topup --config=b02b0.cnf --datain=topup_encoding.txt \
--fwhm=8.000000 --imain=b0_b0rev.nii --miter=5 \
--out=b0_b0rev_base --iout=b0_b0rev_corrected.nii.gz \
--fout=b0_b0rev_field.nii.gz --jacout=jac --logout=b0_b0rev_topup.log \
--rbmout=xfm --dfout=warpfield --miter=1 --splineorder=3 --subsamp=1 \
--warpres=10.000000'

All the additional arguments (e.g. --fwhm, --miter, --warpres..) are optional, yet set by default in the new version of nipype (>1.0.3). While FSL provides default values for these args, some of these are default values IF these options are chosen (e.g. 8mm smoothing IF --fwhm is set). Some of these values may be default used regardless of whether the optional arg was included or not, but otherwise it's introducing a whole set of arguments the user did not intend to use with the command.

We found this to be the case for topup at least, but I think the options should be reviewed to make sure unintended options are not added to the command...

djarecka:
@atsuch - thank you for reporting this issue. I added `usedefault=True` in places where I found that the default value was set by the author of the interface and at the end the value is not used (I assumed that this is misleading and it might be even unintentional).

I understand that you main concern is that at least some of these additional arguments should not be used as default for `topup` command. Could you please tell me which argument should be used as default (including the values) in your opinion? 

cc: @oesteban - since you are the author of this interface, please let me know if you have any comments.

atsuch:
Hi @djarecka , @oesteban,

I'm not expert in this command either. We just noticed the difference in the output when using the newer nipype in our pipeline. I read the FSL topup user guide quickly here (https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/topup/TopupUsersGuide), and found it a bit ambiguous with regard to the default values for --fwhm, --miter, and --warpres. Even though calling the `topup` in terminal tells you that default for these values are set to 8, 5, 10, respectively, the user guide does not mention these default values, and I'm not sure why you'd want to smooth it at 8mm for example, if you're not sub-sampling your image. 

The fact that we get different results with or without setting these values suggest to me that these are not true default values, but we need to test it a bit more. We are running the command with and without some of these options to check the log and output. 

By the way, I just noticed that --miter argument repeat itself (5 and then 1) in the generated topup command. Do you have any idea why and which value is actually being used??

djarecka:
@atsuch - thank you for testing this more! Indeed the log is interesting, and it doesn't look like the default values are always used. I'd be happy to know what FSL team thinks about it.

From `nipype` perspective it probably makes sense just to remove the default values. I can always provide the FSL default value from the FSL help (assuming that they are correct...?), but do not set them on the nipype level. 

And answering your question about `miter` used twice. It is a bug, [`reg_lambda`](https://github.com/nipy/nipype/blob/78717980a6b8ff6f232717870d1324d5b3d5ca05/nipype/interfaces/fsl/epi.py#L223) uses the same name of the argument for the `topup` command. Before I started using the default values probably not many people used it, so no one noticed. Reading the `topup` help, I believe tha argument name should be `--lambda`

djarecka:
@atsuch - I've created a PR #2637 with changes for `topup`, let me know if you have any comments

weningerleon:
Hi all,

I also noticed this new behaviour when updating to a newer version of nipype.
However, I found this new behaviour quite problematic: If the user creates a config file, the custom settings in the config file are overwritten by the nipype default values. As far as I see, there is no way to deactivate the default values of nipype, so config files cannot be used any more if any nipype default values are given.

In comparison, the fsl-default values are overwritten by the config file, and the config file can be overwritten by the values given in the command.

So I would suggest deactivating the nipype default values again.

PR #2538, opened 2018-04-11
ENH: Add LibraryBaseInterface

effigies:
Similar to CLI interfaces, interfaces built on Python libraries (such as dipy, nilearn, nipy, etc.) have some common features and requirements, and it may be useful to unify this into a base class.

Here I suggest a minimal `LibraryBaseInterface` class which:

* Checks that the library is importable
* Optionally checks for additional imports required by the interface (intended for requirements that go beyond the library's dependencies)
* Defines `version` property by the library `__version__`, which should make it trivial to use `min_ver` and `max_ver` in input specs

As a quick demonstration, I've simplified `DipyBaseInterface`.

If this is something worth pursuing, I'll have a look at other library-based interface collections and see what useful things can be standardized and abstracted up into this class.

@satra @chrisfilo @oesteban Thoughts?


oesteban:
Looks like a great idea to me

satra:
@effigies +1  - i think anytime we can abstract usefully and reduce boilerplate is a welcome change. i'm finishing up a new issue post on just this - it came up day before when i was with some folks in DC.

effigies:
A few thoughts from playing with this:

* There were a few different idioms for performing these checks, mostly involving test imports at various points in the file or the interfaces themselves. This seems to me to be a good argument for a more unified approach.
* The file-scoped `have_libXYZ` variables have been centralized and are now imported into their original files, in case anybody's depending on them. I would suggest we eliminate these outside of test files, in the future.
* From playing, the first `importlib.import_module` might take a little time, but every time after that is instantaneous, so we shouldn't be shy about keeping our namespaces clean for the sake of efficiency. And this will help us delay these test imports until actually needed, which should give us a speed boost on `import nipype`, as a lot of things are being loaded somewhere in the hierarchy now.
* Style guide suggestion: No file-level imports that are not direct nipype dependencies. Imports inside `_run_interface()` are totally fine, efficiency-wise, and help make it clearer what the scope of each interface is.

Anyway, I think I've done what I'm likely to with this PR. If it seems valuable, cool. If there are other library-based interface sets that I've missed (I mostly discovered these by grepping for `package_check`) let me know. If it seems like it's not doing enough work to justify it, feel free to close.

nipype/interfaces/base/core.py
```diff
@@ -1267,6 +1267,35 @@ def _format_arg(self, name, spec, value):
...
+        if check_import:
+            import importlib
+            failed_imports = []
+            for pkg in (self._pkg,) + tuple(self.imports):
+                try:
+                    importlib.import_module(pkg)
+                except ImportError:
+                    failed_imports.append(pkg)
+            if failed_imports:
+                iflogger.warn('Unable to import %s; %s interface may fail to '
```
satra: is this going to generate a lot of warnings when we create docs? or `make specs`? i.e. things that simply create an instance of an interface:

`InterfaceX()`

satra:
@effigies - this looks good to me.


for future work (in 2.0) we could consider simplifying many of these interfaces through the py3 annotation framework

nipype/interfaces/base/core.py
effigies: Yeah, potentially. I guess we could set a class-level flag to only display once per class? Or do you have another preferred approach?

satra: one option would be to remove this if. let's merge it and see how annoying it gets.

Issue #2548, opened 2018-04-22
MultiProc goes into infinite loop (resource management related)


chrisfilo:
### Actual behavior

In particular occasions MultiProc claims a job is running, but actually goes into an infinite loop and never finishes.

### Expected behavior.

Tasks should either run and finish or an exception should be raised if there aren't enough resources to run the task.

### How to replicate the behavior

Run MRIQC 0.10.4 on https://openneuro.org/datasets/ds001338/versions/00001 with docker being allocated 7GB of RAM and 2 CPUs

Full command to use
```
mriqc //d/data/ParanoiaStory_20180421_154028/ParanoiaStory/ //d/scratch/mriqc_out participant --participant_label tb3858 --n_procs 2 --verbose
```

chrisfilo:
The job MRIQC is running forever in this particular occasion is `compute_tsnr`.

oesteban:
My bet is that, due to a MemoryError, one process had to get killed. In the latest refactor of MultiProc, we tried to make sure that child processes killed were reported. However, ([and this may be a design problem of multiprocessing](https://stackoverflow.com/a/45456445)) if the worker is killed, a new one is spun up silently (https://github.com/python/cpython/blob/master/Lib/multiprocessing/pool.py#L406). I guess the node kept running as a zombie for a little while since the pool was running in non-daemon mode.

I'll start writing a unit test for this:
  1. Spin up a workflow, and run MultiProc with `n_proc=2`.
  2. Start two endless nodes, which store their PID in a common place.
  3. Query the PIDs of the workers
  4. Kill one worker
  5. Check whether Nipype ends with error or keeps going silently.

WDYT @satra?

oesteban:
A proof of concept that successfully replicates this problem:

```Python
from nipype.pipeline import engine as pe
from nipype.interfaces import utility as niu

wf = pe.Workflow('testworkflow')
def _print():
    import os
    from time import sleep
    while True:
        print(os.getpid())
        sleep(10)

wf.add_nodes([
    pe.Node(niu.Function(function=_print), name='node1'),
    pe.Node(niu.Function(function=_print), name='node2')
])
wf.run('MultiProc', plugin_args={'n_procs': 2})
```
Then I killed one (or both) PIDs printed during runtime. Killing the processes kept Nipype indefinitely polling without error.

oesteban:
Okay, since the PID of the worker and the node are the same, I wanted to check whether this happens because the death of the worker, the child or both:

```Python
[...]

PR #2549, opened 2018-04-24
FIX: Do not generate filename when required fields are missing

### LINKED ISSUE #2506: Fix name_source to handle missing source

effigies:
Fixes #2506.

- Do not generate filename from `name_source` when required fields are missing

This doesn't address the situation where `name_source='in_files'`, but `requires` does not include `'in_files'`. Is it always desired that an undefined `name_source` implies an undefined derivative trait, or should we require `requires` to ensure this behavior?

Will add tests.

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -1123,6 +1123,9 @@ class CommandLine(BaseInterface):
                 base = self._filename_from_source(ns, chain)
                 if isdefined(base):
                     _, _, source_ext = split_filename(base)
+                else:
+                    # Do not generate filename when required fields are missing
+                    return retval
 
             chain = None
             retval = name_template % base
--- a/nipype/interfaces/base/tests/test_specs.py
+++ b/nipype/interfaces/base/tests/test_specs.py
@@ -274,6 +274,55 @@ def test_cycle_namesource2(setup_file):
     assert '%s_generated_mootpl' % nme in res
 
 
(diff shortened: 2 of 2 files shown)


effigies:
This is ready for review, though an unrelated test timed out. I wonder if these intermittent timeouts are related to #2548.

PR #2550, opened 2018-04-25
fix: undeprecate dcm2niix source_names option

satra:
Fixes dcm2niix issue in heudiconv: https://github.com/nipy/heudiconv/pull/189

- restore source_names to allow sending a collection of dicom files rather than a directory to dcm2niix

rationale:

with node, these files get copied to the current working directory and dcm2niix will only need to convert the local files

as an interface, this allows users to point at a set of files. it's possible in this case that all files in the directory get converted, not just the files pointed to. i have tried to include documentation to guide in this scenario.


nipype/interfaces/dcm2nii.py
```diff
@@ -399,7 +414,7 @@ def _format_arg(self, opt, spec, val):
                 spec.argstr += ' n'
                 val = True
         if opt == 'source_names':
-            return spec.argstr % val[0]
+            return spec.argstr % (os.path.dirname(val[0]) or '.')
```
mgxd: since the dicoms are copied to the node, couldn't this just be `'.'`

satra: not when used as an interface.

nipype/interfaces/dcm2nii.py
```diff
@@ -263,15 +263,18 @@ class Dcm2niixInputSpec(CommandLineInputSpec):
         argstr="%s",
         position=-1,
         copyfile=False,
-        deprecated='1.0.2',
-        new_name='source_dir',
         mandatory=True,
+        desc=('A set of filenames to be converted. Note that the current '
```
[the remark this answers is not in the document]
satra: roger

PR #2552, opened 2018-04-25
[wip]Fix/deprecated 2551 (closes #2551 #2508)

### LINKED ISSUE #2508: LSFPlugin sets ignore_exception=True on line 57 of lsf.py plugin causing workflow to crash due to deprecation
The _is_pending function in the LSFPlugin when running the command sets the ignore_exception=True on line 57.  See below:

```
def _is_pending(self, taskid):
        """LSF lists a status of 'PEND' when a job has been submitted but is
        waiting to be picked up, and 'RUN' when it is actively being processed.
        But _is_pending should return True until a job has finished and is
        ready to be checked for completeness. So return True if status is
        either 'PEND' or 'RUN'"""
        cmd = CommandLine(
            'bjobs', resource_monitor=False, terminal_output='allatonce')
        cmd.inputs.args = '%d' % taskid
        # check lsf task
        oldlevel = iflogger.level
        iflogger.setLevel(logging.getLevelName('CRITICAL'))
        result = cmd.run(ignore_exception=True)
        iflogger.setLevel(oldlevel)
        # logger.debug(result.runtime.stdout)
        if 'DONE' in result.runtime.stdout or 'EXIT' in result.runtime.stdout:
            return False
        else:
            return True
```
This results in workflows not executing with newer versions of nipype (I'm using version 1.0.1) exiting with the following error:

> TraitError: Input ignore_exception in interface CommandLine is deprecated. Will be removed or raise an error as of release 1.0.0 

I removed ```ignore_exception=True``` from my copy of nipype and was able to execute the workflow with no problems.

### LINKED ISSUE #2551: deprecated traits in BaseInterfaceInputSpec and CommandLineInputSpec
### Summary

we have two traits that are marked ad `deprecated='1.0.0'` in `interfaces/base/specs.py`:
- [ignore_exception in BaseInterface](https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/specs.py#L322)
- [terminal_output in CommandLine](https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/specs.py#L374)

At the same time we're using these traits in some plugins, e.g. [here](https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/condor.py#L57) and it leads to [errors](https://neurostars.org/t/traiterror-for-condor-plugin/1567/5)

I'm not sure why we wanted to deprecate these traits, and if we still want to do it now, or we want to set a different nipype version.

Related to #2508

djarecka:
removing `terminal_output` and `ignore_exception`when creating `CommandLine` instances. These traits are deprecated in `CommandLineInputSpec`

Closes #2551 
Closes #2508

### DIFF
--- a/nipype/interfaces/afni/base.py
+++ b/nipype/interfaces/afni/base.py
@@ -87,7 +87,8 @@ class Info(PackageInfo):
         Could be made more fancy to allow for more relocatability'''
         clout = CommandLine(
             'which afni',
-            resource_monitor=False).run()
+            resource_monitor=False,
+            terminal_output='allatonce').run()
         if clout.runtime.returncode is not 0:
             return None
 
--- a/nipype/interfaces/afni/base.py
+++ b/nipype/interfaces/afni/base.py
@@ -87,8 +87,7 @@ class Info(PackageInfo):
         Could be made more fancy to allow for more relocatability'''
         clout = CommandLine(
             'which afni',
-            resource_monitor=False,
-            terminal_output='allatonce').run()
+            resource_monitor=False).run()
         if clout.runtime.returncode is not 0:
             return None
 
--- a/nipype/interfaces/afni/base.py
+++ b/nipype/interfaces/afni/base.py
@@ -87,6 +87,7 @@ class Info(PackageInfo):
         Could be made more fancy to allow for more relocatability'''
         clout = CommandLine(
             'which afni',
+            ignore_exception=True,
             resource_monitor=False,
             terminal_output='allatonce').run()
         if clout.runtime.returncode is not 0:
--- a/nipype/interfaces/afni/base.py
+++ b/nipype/interfaces/afni/base.py
@@ -87,7 +87,6 @@ class Info(PackageInfo):
         Could be made more fancy to allow for more relocatability'''
         clout = CommandLine(
             'which afni',
-            ignore_exception=True,
             resource_monitor=False,
             terminal_output='allatonce').run()
         if clout.runtime.returncode is not 0:
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -1314,7 +1314,8 @@ class PackageInfo(object):
                 try:
                     clout = CommandLine(
                         command=klass.version_cmd,
-                        resource_monitor=False).run()
+                        resource_monitor=False,
+                        terminal_output='allatonce').run()
                 except IOError:
                     return None
 
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -445,7 +445,7 @@ class BaseInterface(Interface):
             r['path'] = self.__module__
             due.cite(**r)
 
-    def run(self, cwd=None, **inputs):
+    def run(self, cwd=None, ignore_exception=None, **inputs):
         """Execute this interface.
 
         This interface will not raise an exception if runtime.returncode is
@@ -463,6 +463,10 @@ class BaseInterface(Interface):
         that was executed, provenance information and, if successful, results
         """
         from ...utils.profiler import ResourceMonitor
+
+        # if ignore_exception is not provided, taking self.ignore_exception
+        if ignore_exception is None:
+            ignore_exception = self.ignore_exception
 
         # Tear-up: get current and prev directories
         syscwd = rgetcwd(error=False)  # Recover when wd does not exist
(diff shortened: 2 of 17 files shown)


effigies:
Ah, `terminal_output` is still a valid argument to a `CommandLine` interface and `ignore_exception` is still valid for `BaseInterface`, and they do valid things. It's just that they are no longer going to be inputs.

See:

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/core.py#L173-L174

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/core.py#L906

So what's happening at `.run(ignore_exception=True)` is that it's hitting:

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/core.py#L475

I think the easiest thing would be to add `ignore_exception` to `BaseInterface.run`:

```Python
def run(self, cwd=None, ignore_exception=None, **inputs):
    if ignore_exception is None:
        ignore_exception = self.ignore_exception
```

And then just change this line to use the local variable `ignore_exception`:

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/core.py#L534-L535

We should also remove:

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/core.py#L922-L925

and

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/core.py#L953-L955

effigies:
So I would actually undo all of the changes in the current diff, and replace

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/specs.py#L321-L328

with `BaseInterfaceSpec = TraitedSpec`. And just remove this block:

https://github.com/nipy/nipype/blob/a1ab5181370b69490df151838187edf9b0382c44/nipype/interfaces/base/specs.py#L367-L381

This will require a huge `make specs`, and probably some doctest updates.

djarecka:
I'm completely confused. Why do we deprecate these traits, what does it actually mean? we just want to use it differently? only to some methods?

effigies:
I'm pretty sure this was @oesteban deprecating, but the reason is that they make more sense as attributes than as inputs to be set in the input spec. For instance, nobody passes `ignore_exception` from one node to another. So the idea is just to remove the traits and anything that tries to set them while keeping the old behavior.

djarecka:
Ok, so I understand that should remove from input, but be sure that it can be passed to all methods that are using them and to `__init__` (this is already done).

I will revert my changes and try again

djarecka:
I added bot `ignore_exception` and `terminal_output` to `run` method. If you don't want `terminal_output` to be chosen when calling `run`, let me know.

djarecka:
I added `ignore_exception` to `run` method. I was wondering if we also want to have `terminal_output` in `run` (and pass to `run_interface` etc.) but probably it's enough that it can be set in `__init__`. Also, shouldn't we have a different default value different, not `None`? so we can remove this [if](https://github.com/djarecka/nipype/blob/22e9dcaee191814fb68100b68b1fc40fa53dbcfb/nipype/interfaces/base/core.py#L923) and we always have `self.terminal_output`.

effigies:
> I'm not sure about removing ignore_exception and termianl_output from specs.py. I think it might be good to keep it this error for some time.

Sure. It at least gives an informative error. We may want to set a milestone to remove it, just so we don't forget, but that's fine with me. @satra do you agree?

> Also, shouldn't we have a different default value different [for `terminal_output`], not `None`?

Doing it the way we currently are allows a subclass to set a different default like so:

```Python
class NewCommand(CommandLine):
    _terminal_output = 'file_stdout'
```

If we set a default in `__init__`, the subclass would have to override `__init__` to change the default.

So I would keep it as is.

satra:
@effigies - fine with me.

effigies:
Any preferred timeline? 1.1? 1.2? Given that we've only seen two issues, both from internal bugs, not external use of these inputs, I think we can fully remove them after a reasonably short interval.

satra:
let's do 1.1

satra:
we can schedule 1.1 for end of june, right after hbm

PR #2558, opened 2018-04-27
FIX for [#2557](https://github.com/nipy/nipype/issues/2557#issuecomment-384834681)

### LINKED ISSUE #2557: Crashdump directory is not properly set in workflow config file
Crash files are written to the current directory despite setting the crashdump directory of a workflow with `workflow.config = {"execution": {"crashdump_dir": 'crashdump'}}`.

I suspect the problem is line 594 in [pipeline/engine/workflows.py](https://github.com/nipy/nipype/blob/3c43f0c5041eaa127cbfb2e6a5ee765357aa262e/nipype/pipeline/engine/workflows.py/) when the workflow config file is being merged with the node config file. The default behaviour of `merge_dict` is that in case of a key conflict the second dictionary has precedence, so that the user-defined setting from the workflow level is being overwritten with the default settings from the node level. 

As far as I can see the ideal behaviour would be that node settings override workflow settings, but additionally, explicitly defined settings override default-settings.

eort:
Fixes #2557 

- Don't override empty node config file with the default config information

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -413,7 +413,6 @@ class Node(EngineBase):
 
         if self.config is None:
             self.config = {}
-        self.config = merge_dict(deepcopy(config._sections), self.config)
 
         outdir = self.output_dir()
         force_run = self.overwrite or (self.overwrite is None and


eort:
closed in place of #2559

PR #2559, opened 2018-04-27
fix: propagate explicit workflow config to nodes in a workflow and allow nodes to overwrite

satra:
Fixes # 2557.

- Simply set the node config to None on initialization
- This picks up the default when run is called.

@eort - please see if this branch solves your problem


eort:
@satra  Yes, that solves it.

Issue #2553, opened 2018-04-25
In-node reporting


effigies:
In [niworkflows](https://github.com/poldracklab/niworkflows/), we've implemented a reporting mechanism that allows interfaces to generate images or HTML snippets to describe what was done and allow a user to assess the quality.

The approach can be seen in the [`ReportCapableInterface`](https://github.com/poldracklab/niworkflows/blob/master/niworkflows/interfaces/report_base.py), which adds `_post_run_hook` and `_generate_report` methods which have access to both the inputs and the outputs of the interface.

We currently use these as mixins to augment standard nipype interfaces, but they could as easily be used in built-in interfaces as well.

Is this something worth trying to bring upstream - both the reporting mechanism itself and the specific report generation code for various operations (e.g. registration and segmentation)?

PR #2560, opened 2018-04-27
ENH: ReportCapableInterface mix-in/base interface

effigies:
Related (closes?): #2553

- Adds pre- and post-run hooks to BaseInterface, which is intended to allow mix-ins to be more flexible about where exactly they fall within the method resolution stack.
- Creates `ReportCapableInterface` as a mix-in that is also written to work as a base class as well.

This factors out what seemed to be the essential parts of the niworkflows [`ReportCapableInterface`](https://github.com/poldracklab/niworkflows/blob/9d4163e30dae3caf7560313ef353d4177a38004f/niworkflows/interfaces/report_base.py#L23-L108).

Differences:

* Compression is removed as an option - without the actual figure generation and compression machinery, this would be under-supported. If these are worth adding, we can do so, but in point of fact we're considering undoing some of our compression steps as they make our reports Chrome-only.
* `_post_run_hook` is now the driver of this functionality, and classes using this mix-in should tail-call `return super(..., self)._post_run_hook(runtime)`.
* `_generate_error_report` was removed. I don't think this can actually be hit in niworkflows, as `CommandLine` interfaces that had bad return values raised exceptions.

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -517,7 +517,9 @@ class BaseInterface(Interface):
         outputs = None
 
         try:
+            runtime = self._pre_run_hook(runtime)
             runtime = self._run_interface(runtime)
+            runtime = self._post_run_hook(runtime)
             outputs = self.aggregate_outputs(runtime)
         except Exception as e:
             import traceback
@@ -652,6 +654,28 @@ class BaseInterface(Interface):
         iflogger.debug('saving inputs {}', inputs)
         with open(json_file, 'w' if PY3 else 'wb') as fhandle:
             json.dump(inputs, fhandle, indent=4, ensure_ascii=False)
+
+    def _pre_run_hook(self, runtime):
+        """
+        Perform any pre-_run_interface() processing
+
+        Subclasses may override this function to modify ``runtime`` object or
+        interface state
+
+        MUST return runtime object
+        """
+        return runtime
+
+    def _post_run_hook(self, runtime):
+        """
+        Perform any post-_run_interface() processing
+
+        Subclasses may override this function to modify ``runtime`` object or
+        interface state
+
+        MUST return runtime object
+        """
+        return runtime
 
 
 class SimpleInterface(BaseInterface):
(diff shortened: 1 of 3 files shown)


nipype/interfaces/base/core.py
```diff
@@ -517,7 +517,9 @@ def run(self, cwd=None, ignore_exception=None, **inputs):
         outputs = None
 
         try:
+            runtime = self._pre_run_hook(runtime)
             runtime = self._run_interface(runtime)
+            runtime = self._post_run_hook(runtime)
```
effigies: @satra Are you comfortable with this change to `BaseInterface`?

satra: for now yes - we may have to add some prov stuff when i refactor that piece.

nipype/interfaces/mixins/reporting.py
```diff
@@ -0,0 +1,59 @@
...
+
+from ... import logging
+from ..base import (
+    File, traits, BaseInterface, BaseInterfaceInputSpec, TraitedSpec)
+
+iflogger = logging.getLogger('interface')
+
+
+class ReportCapableInputSpec(BaseInterfaceInputSpec):
+    generate_report = traits.Bool(False, usedefault=True,
```
oesteban: As for `terminal_output` and `ignore_errors`, this should not be an input. Should be a state variable of the instance.

nipype/interfaces/mixins/reporting.py
```diff
@@ -0,0 +1,59 @@
...
+    _out_report = None
+
+    def _post_run_hook(self, runtime):
+        runtime = super(ReportCapableInterface, self)._post_run_hook(runtime)
+
+        # leave early if there's nothing to do
+        if not self.inputs.generate_report:
+            return runtime
+
+        self._out_report = os.path.abspath(self.inputs.out_report)
```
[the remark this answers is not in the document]
effigies: I see three options:

1) `chdir(runtime.cwd)` between each step - ugly, but forgiving
2) `RuntimeError` if `os.getcwd() != runtime.cwd` after one of these steps is run
3) Reject the idea of cwd assurances altogether and resolve this path relative to `runtime.cwd`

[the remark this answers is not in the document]
effigies: Actually, I'm going to do 3 for now. @satra, let me know if you think we should do any work to either mitigate or error out if `_*_run_hook` and `_run_interface` change the CWD on us.

PR #2561, opened 2018-04-29
FIX: DTITK nonlinear workflow origin reslicing

kesshijordan:
I was testing the workflow on some more difficult registrations (e.g. child-to-adult from different datasets) and found that the reslicing can cause problems when images have origins far from 0,0,0. This is just a few lines addition to the workflow to move all images to origin zero prior to any reslicing.

Fixes #2514 .

- Move all images to origin zero prior to reslicing for the nonlinear tensor registration DTITK workflow


kesshijordan:
@effigies and @satra is it possible to get this in to the 1.0.3 or is it too late?

effigies:
I would prefer to hold off until 1.0.4, at this point.

effigies:
As a non-DTI-TK user, this looks fine to me.

My main question would be whether you want to restore the origin at the end of the pipeline, for viewing. (I'm assuming that (0, 0, 0) is in voxel space, not RAS.)

kesshijordan:
For viewing purposes, I think the important thing is that it matches the fixed image. Now that you mention it, I should probably reslice the moving image/move origin to match the original fixed image (the fixed image has to be resliced to a dimension of power 2 to use the diffeomorphic registration). That way the registered volume would match the fixed image input, instead of having two resliced images (fixed & moving) as output that do not match the input fixed image.

Before reslicing/move origin
NIFTI Intent Code: NIFTI_INTENT_SYMMATRIX
NIFTI Orientation Code: RPI
Volume Info of /Users/kesshijordan/ref_data/test_tensors/child2_SC_EC_tensor.nii.gz
size: 110x110x69, voxel size: 2x2x2, origin: [-55.5361, 48.0527, 28.5598]

After reslicing/move origin
NIFTI Intent Code: NIFTI_INTENT_SYMMATRIX
NIFTI Orientation Code: LPI
Volume Info of resliced/fixed/child1_SC_EC_tensor_avs_resampled.nii.gz
size: 128x128x64, voxel size: 2.2x2.2x2.2, origin: [0, 0, 0]

effigies:
> I should probably reslice the moving image/move origin to match the original fixed image 

Cool, that makes sense to me.

effigies:
Hi @kesshijordan just a bump. We're going to try to freeze by Friday, if you want to try to get this in by then.

kesshijordan:
Hi, @effigies. I was trying to add a workflow for applying the diffeomorphic transform (since the origin has to be moved and image resliced, applying the saved transform is not a matter of simply using the saved .df.nii file), but I'm having trouble figuring out why it isn't working. I have the other points we discussed fixed here and took out the apply_transform workflow; I'll put it back in when I manage to debug it. For now I just have a note that you need to reslice/move origin on any new images if you want to apply the workflow transform to another volume.

PR #2563, opened 2018-04-30
FIX: Print UID in crashfile if login name is unavailable

effigies:
Related: poldracklab/fmriprep#1060

- Use a stand-in login name of `UID<UID>` if `getpass.getuser()` fails.

We had a user use the `-u` flag for Docker, which allowed them to run fMRIPrep and get outputs owned by their local user. If that UID doesn't exist for a user within the container, `getpass.getuser()` fails, and crashfiles cannot be written properly.


satra:
lgtm

effigies:
DARTEL was timing out again, so I killed it. Can rebase this to re-run if needed.

PR #2566, opened 2018-05-03
FIX: Workaround to ICA-AROMA change of directory

oesteban:
As reported here https://github.com/maartenmennes/ICA-AROMA/issues/25#issuecomment-386169296, ICA-AROMA changes of directory to allow for relative imports.

That is confusing Nipype's interpretation of the input. This workaround fixes poldracklab/fmriprep#1090.


effigies:
Sorry for the delay.

PR #2572, opened 2018-05-11
ENH: Reorient interface

effigies:
This is a pure nibabel/numpy interface to reorient an image to any `RAS` ... `IPL`. It also produces an affine transform that describes the reorientation.

Based off of [`fmriprep.interfaces.images.Reorient`](https://github.com/poldracklab/fmriprep/blob/6f4c69bc2dc60c1cbe102b039e3484a5fd49ed59/fmriprep/interfaces/images.py#L294-L337), but does not assume RAS target orientation or nibabel >= 2.2.

I created a `interfaces.image` file, since this didn't seem to fall into existing interface categories. Is there a better place to put this?


oesteban:
+1000, and I think the new module is fine.

effigies:
Ah, no. This is the case where there's a flip but not a reordering. The dimension info doesn't include direction, just axis.

effigies:
@oesteban Good to merge?

PR #2585, opened 2018-05-18
FSL slice

salma1601:
added interface for `fslslice`


nipype/interfaces/fsl/utils.py
```diff
@@ -262,6 +262,70 @@ def _format_arg(self, name, trait_spec, value):
         return super(Smooth, self)._format_arg(name, trait_spec, value)
 
 
+class SliceInputSpec(FSLCommandInputSpec):
+    in_file = File(exists=True, argstr="%s", position=0, mandatory=True,
+                   desc="input filename")
```
effigies: Since this places outputs in the same directory as `in_file` (unless `out_base_name` is defined), you should add `copyfile=False` to this trait.

nipype/interfaces/fsl/utils.py
```diff
@@ -262,6 +262,70 @@ def _format_arg(self, name, trait_spec, value):
...
+            out_dir = os.getcwd()
+        else:
+            out_dir = os.path.dirname(self.inputs.in_file)
+            in_base, _ = os.path.splitext(self.inputs.in_file)
+            if '.gz' in self.inputs.in_file:
+                in_base, _ = os.path.splitext(in_base)
+
+            outbase = in_base + '_slice*'
+        outputs['out_files'] = sorted(glob(os.path.join(out_dir,
+                                                        outbase + ext)))
```
effigies: This doesn't account for if `out_base_name` is an absolute path. You can simplify this a bit, too, using `fname_presuffix`:

```Python
suffix = '_slice_*' + ext
if isdefined(self.inputs.out_base_name):
    fname_template = os.abspath(self.inputs.out_base_name + suffix)
else:
    fname_template = fname_presuffix(self.inputs.in_file, suffix=suffix, use_ext=False)

outputs['out_file'] = sorted(glob(fname_template))
```

PR #2586, opened 2018-05-19
enh: add interface for LabelGeometryMeasures

satra:
- new ants interface for LabelGeometryMeasures


nipype/interfaces/ants/utils.py
```diff
@@ -357,3 +357,60 @@ class ComposeMultiTransform(ANTSCommand):
...
+        position=0,
+        desc='image dimension (2 or 3)')
+    label_image = File(
+        argstr='%s',
+        position=1,
+        mandatory=True,
+        desc='label image to use for extracting geometry measures')
+    intensity_image = File(
+        value='[]',
+        exists=True,
```
effigies: `exists=True` seems weird, given a default value of `'[]'`. I guess it applies only to non-default values?

satra: `exists` only applies to non-default values (could be considered a bug). i'm taking advantage of that here, without writing a `_format_arg` function, but something to keep in mind.

satra:
related gist: https://gist.github.com/satra/ca140fe2cb41659b1b9666909df19cf4

PR #2587, opened 2018-05-20
CI: Move PyPI deployment to Circle 

### LINKED ISSUE #2313: Automate deployment from CircleCI
Seems like Circle have recently added [the tag filtering that was missing in Circle 2.0](https://circleci.com/docs/2.0/workflows/#git-tag-job-execution).

Using that, we could automate deployment to Docker Hub and to Pypi as we currently do with fmriprep and mriqc. Additionally, we would also generate the new conda hash.

Finally this would really simplify the process, since it would be automatically triggered by cutting a new release through github.

effigies:
Closes #2313.

- Moves PyPI deployment to CircleCI

The Travis builds are a little weird in that, on tagging, the first to successfully complete will upload to PyPI, while the rest will fail. This seems unnecessary, and we might as well use the dependencies allowed by Circle 2.0 workflows.

This will need @satra's PyPI credentials stored properly in Circle before it makes sense to merge.

Testing everything but the `twine` upload.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,7 +339,7 @@ jobs:
           command: |
             pip install twine future
             python setup.py sdist bdist_wheel
-            twine upload dist/*
+            # twine upload dist/*
 
 workflows:
   version: 2
@@ -368,11 +368,11 @@ workflows:
             - test_py3_fmri_fsl_spm
             - test_py3_fmri_spm_dartel_multiproc
       - deploy_pypi:
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /.*/
+          # filters:
+          #   branches:
+          #     ignore: /.*/
+          #   tags:
+          #     only: /.*/
           requires:
             - test_pytest
             - test_fmri_spm_nested_fsl_feeds
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,7 +299,7 @@ jobs:
       - run: *_run_codecov_smoke
       - store_artifacts: *store_artifacts_kwds
 
-  deploy:
+  deploy_dockerhub:
     docker:
       - image: docker:17.10.0-ce-git
     steps:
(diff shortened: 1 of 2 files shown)


satra:
@effigies - just let me know whenever my credentials are required.

alternatively, @chrisfilo  can you add @effigies  and @mgxd to pypi as maintainers. i think you created that one.

effigies:
You can set the `TWINE_USERNAME` and `TWINE_PASSWORD` environment variables in the [Circle project](https://circleci.com/gh/nipy/nipype/edit#env-vars) whenever.

satra:
@effigies - done

@effigies and @mgxd - if you send me your pypi usernames, i can add you as maintainers of the nipype project.

effigies:
@satra cjmarkie

mgxd:
@satra mgxd

.circleci/config.yml
```diff
@@ -357,3 +369,14 @@ workflows:
             - test_fmri_spm_nested_fsl_feeds
             - test_py3_fmri_fsl_spm
             - test_py3_fmri_spm_dartel_multiproc
+      - deploy_pypi:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /.*/
```
satra: for tags - should we only do master?

effigies: What's the goal you're trying to achieve? I assume not to push a package every time master is updated.

satra: i think i may be misunderstanding how circle bit works - i thought the tags section should look only for tags applied to the master branch, since tags could potentially be applied on any branch. i'll read up on my train ride in a bit.

effigies: So by default Circle will build on all branches and no tags. In order to restrict it from some branches, you can match with `ignore` and `only`, so here I'm saying don't run this on any branches. And then to run on all tags (regardless of branch) we use `only: /.*/`.

Which is to say, this job will be run iff we have a tag.

PR #2593, opened 2018-05-23
DOCs: fix/improve descriptions for commands and interfaces

yarikoptic:
- minor fix/impovement for N4BiasFieldCorrection input parameters description
- tried to unify FSLCommand interfaces (to check what nipypers think about unifying them all up in a similar fashion ;))
  - I think "uses " provides no additional information
  - make them consistent and descriptive
  - short description is not a complete sentence (more of continuation of "Thisthing is ...", so removed trailing periods


nipype/interfaces/ants/segmentation.py
```diff
@@ -266,10 +266,15 @@ class N4BiasFieldCorrectionInputSpec(ANTSCommandInputSpec):
     input_image = File(
         argstr='--input-image %s',
         mandatory=True,
-        desc=('image to apply transformation to (generally a '
-              'coregistered functional)'))
-    mask_image = File(argstr='--mask-image %s')
-    weight_image = File(argstr='--weight-image %s')
+        desc=('input for bias correction. Negative values or values close to '
+            'zero should be processed prior to correction'))
```
satra: `processed` - how? if there is some additional info we can add that would be great.

nipype/interfaces/fsl/preprocess.py
```diff
@@ -113,7 +113,7 @@ class BETOutputSpec(TraitedSpec):
 
 
 class BET(FSLCommand):
-    """Use FSL BET command for skull stripping.
+    """FSL BET command for skull stripping
```
satra: we could use the nipype term `interface` or `wrapper`

nipype/interfaces/ants/segmentation.py
yarikoptic: I don't even know (yet).  This was copy pasted from the --help output of the tool and IMHO is better than nothing ;-)

nipype/interfaces/fsl/preprocess.py
yarikoptic: Just choose ;-)  I saw such phrasing used in other ones, and decided to go with it.  The point that it is an interface could somewhat be deduced from the module path, but I would also kinda like that being explicit, but then making it explicit that it is interface to the command (instead of to e.g. Python function/module) also has value.
In my case, I will use it for description of a flywheel gear, and there it would not matter either it is "interface" or "wrapper" (somewhat assumed), so I would need to strip it away or replace to something like above

satra: let's use `wrapper`

effigies:
@yarikoptic Do you have a couple minutes for this one? Are there still open discussion topics?

mgxd:
merging this for 1.1, we can always add more documentation ;)

yarikoptic:
sorry guys I've failed to finish it on my own - fell off the raidar  ;-) thanks for merging!

PR #2594, opened 2018-05-24
ENH: initial connectome workbench support

mgxd:
- initial support for connectome workbench interfaces
- interface MetricResample (`wb_command -metric-resample`)
- added some empty files to show clearer use-case within doctest


nipype/interfaces/workbench/metric.py
```diff
@@ -0,0 +1,147 @@
...
+    fs_LR.L.midthickness_va_avg.32k_fs_LR.shape.gii'
+    """
+    input_spec = MetricResampleInputSpec
+    output_spec = MetricResampleOutputSpec
+    _cmd = 'wb_command -metric-resample'
+
+    def _format_arg(self, opt, spec, val):
+        if opt == "metric_out":
+            # ensure generated filename is assigned to trait
+            self.inputs.trait_set(metric_out=val)
```
mgxd: bug? without this `trait_set` call, `metric_out` remained as `Undefined` in `_list_outputs`

mgxd: bug? `metric_out` was displayed correctly in `cmdline`, but without this `trait_set` call, remained as `Undefined` in `_list_outputs

effigies:
You need to add an ` nipype/interfaces/workbench/tests/__init__.py` to pacify the tests.

nipype/interfaces/workbench/metric.py
```diff
@@ -0,0 +1,147 @@
...
+    fs_LR.L.midthickness_va_avg.32k_fs_LR.shape.gii'
+    """
+    input_spec = MetricResampleInputSpec
+    output_spec = MetricResampleOutputSpec
+    _cmd = 'wb_command -metric-resample'
+
+    def _format_arg(self, opt, spec, val):
+        if opt == "out_file":
+            # ensure generated filename is assigned to trait
+            self.inputs.trait_set(out_file=val)
```
effigies: Sorry, I meant to say that I think this (and the `_list_outputs` entry) shouldn't be necessary if you make the input spec match the output spec.

nipype/interfaces/workbench/metric.py
```diff
@@ -0,0 +1,143 @@
...
+
+class MetricResampleOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc="the output metric")
+    roi_file = File(desc="ROI of vertices that got data from valid source vertices")
+
+
+class MetricResample(WBCommand):
+    """
+    Resample a metric file to a different mesh
+
```
oesteban: I think it would help documentation to copy a bit more from their docs:

```
      Resample a metric file to a different mesh

      Resamples a metric file, given two spherical surfaces that are in
      register.  If ``ADAP_BARY_AREA`` is used, exactly one of -area-surfs or
      ``-area-metrics`` must be specified.

      The ``ADAP_BARY_AREA`` method is recommended for ordinary metric data,
      because it should use all data while downsampling, unlike ``BARYCENTRIC``.
      The recommended areas option for most data is individual midthicknesses
      for individual data, and averaged vertex area metrics from individual
      midthicknesses for group average data.

      The ``-current-roi`` option only masks the input, the output may be slightly
      dilated in comparison, consider using ``-metric-mask`` on the output when
      using ``-current-roi``.

      The ``-largest option`` results in nearest vertex behavior when used with
      ``BARYCENTRIC``.  When resampling a binary metric, consider thresholding at
      0.5 after resampling rather than using ``-largest``.
```

effigies: @oesteban Want to just push this change? It's after 5pm here.

oesteban: Sure

PR #2595, opened 2018-05-25
 FIX: Update logging levels in enable_debug_mode 

### LINKED ISSUE #1169: config.enable_debug_mode() does not work as advertised
config.enable_debug_mode() is not equivalent to:

```
mkdir ~/.nipype
echo "[logging]" > ~/.nipype/nipype.cfg
echo "workflow_level = DEBUG" ~/.nipype/nipype.cfg
echo "interface_level = DEBUG" >> ~/.nipype/nipype.cfg
echo "filemanip_level = DEBUG" >> ~/.nipype/nipype.cfg
```

effigies:
Fixes #1169

- Update logging levels in enable_debug_mode 
- Add a test

### DIFF
--- a/doc/users/config_file.rst
+++ b/doc/users/config_file.rst
@@ -237,16 +237,23 @@ Debug configuration
 
 To enable debug mode, one can insert the following lines::
 
-  from nipype import config, logging
+  from nipype import config
   config.enable_debug_mode()
-  logging.update_logging(config)
 
 In this mode the following variables are set::
 
   config.set('execution', 'stop_on_first_crash', 'true')
   config.set('execution', 'remove_unnecessary_outputs', 'false')
+  config.set('execution', 'keep_inputs', 'true')
   config.set('logging', 'workflow_level', 'DEBUG')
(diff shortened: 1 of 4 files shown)


satra:
i don't think this was really a bug. if we treat it as such (and i am fine with the change) we should update this as well:

http://nipype.readthedocs.io/en/latest/users/config_file.html#debug-configuration

effigies:
It does seem weird to `enable_debug_mode` and only change the config without changing the log levels the config refers to.

That said, if `utils_level` wasn't intended to be part of it, I'm okay with removing that. That may be too noisy. Will update the docs.

satra:
also this docs: http://nipype.readthedocs.io/en/latest/users/debug.html

so we did have proper instructions for debug mode.

i would use this opportunity to clean up and make debug mode as useful as possible. so go ahead with your changes.

effigies:
Any further comments? Hoping to merge these last four PRs in batch when @mgxd's tests pass.

satra:
lgtm

PR #2596, opened 2018-05-25
FIX: Set default result in DistributedPluginBase._clean_queue

achetverikov:
Fixing this: https://github.com/nipy/nipype/issues/2570

- calls to _clean_queue in plugins/base.py now always have the information about exceptions when it's available

### DIFF
--- a/nipype/pipeline/plugins/base.py
+++ b/nipype/pipeline/plugins/base.py
@@ -154,9 +154,7 @@ class DistributedPluginBase(PluginBase):
                     result = self._get_result(taskid)
                 except Exception:
                     notrun.append(
-                        self._clean_queue(
-                            jobid,
-                            graph))
+                        self._clean_queue(jobid, graph))
                 else:
                     if result:
                         if result['traceback']:
@@ -218,9 +216,9 @@ class DistributedPluginBase(PluginBase):
 
         if self._status_callback:
             self._status_callback(self.procs[jobid], 'exception')
-		if result is None:
-			result={'result': None,
-					'traceback': '\n'.join(format_exception(*sys.exc_info()))}
+        if result is None:
+            result = {'result': None,
+                      'traceback': '\n'.join(format_exception(*sys.exc_info()))}
 
         if str2bool(self._config['execution']['stop_on_first_crash']):
             raise RuntimeError("".join(result['traceback']))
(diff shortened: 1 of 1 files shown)


effigies:
Please remove the unrelated auto test. That is being added elsewhere, and will currently cause this PR to fail.

effigies:
Also, I think a cleaner solution would be to set this default in `_clean_queue`, as it should not crash when its default argument is set.

https://github.com/nipy/nipype/blob/9eaa2a32c8cb3569633a79d6f7968270453f9aed/nipype/pipeline/plugins/base.py#L220-L227

I would suggest adding at line 225:

```Python
if result is None:
    result={'result': None,
            'traceback': '\n'.join(format_exception(*sys.exc_info()))}
```

effigies:
@achetverikov Are you going to submit an updated PR? I'd like to get all fixes in today for the next release, and this seems like a quick one.

achetverikov:
check the updated commit

nipype/pipeline/plugins/base.py
```diff
@@ -156,11 +156,7 @@ def run(self, graph, config, updatehash=False):
                     notrun.append(
                         self._clean_queue(
                             jobid,
-                            graph,
-                            result={
-                                'result': None,
-                                'traceback': '\n'.join(format_exception(*sys.exc_info()))
-                            }))
+                            graph))
```
effigies: Maybe let's pull all this back onto one line:

```Python
self._clean_queue(jobid, graph)
```

nipype/pipeline/plugins/base.py
```diff
@@ -222,6 +218,9 @@ def _clean_queue(self, jobid, graph, result=None):
 
         if self._status_callback:
             self._status_callback(self.procs[jobid], 'exception')
+		if result is None:
+			result={'result': None,
+					'traceback': '\n'.join(format_exception(*sys.exc_info()))}
```
effigies: Looks like you're using tabs, not spaces. Do you think you could update this to use 4-space indentation? Also, there should be spaces around the `=`, and `'traceback'` should start on the same column as `'result'`.

(Sorry if this is a pain in your editor.)

achetverikov:
done, I think

achetverikov:
you're doing great work guys, I'm glad to be of help

PR #2598, opened 2018-05-26
RF: Futures-based MultiProc

### LINKED ISSUE #2548: MultiProc goes into infinite loop (resource management related)

effigies:
This is a first pass at basing MultiProc off of the `concurrent.futures` module.

I don't expect this to pass the 2.7 tests, but it will be interesting to see if we can get Python 3 working with about half an hour of effort.

Fixes #2548


effigies:
Huh, apparently futures gets bundled with Python 2.7 in conda.

Should probably still add it to the dependencies, but this looks to be passing, so far.

satra:
looks good to me. we should check it in a basic system python 2.7 environment and see if that requires any backports or a specific version of py2.7.

effigies:
Is there a reason we're using Miniconda in Travis? It supports vanilla Python installations, and I think all dependencies can be installed with `pip`. Nibabel has [URLs](https://github.com/nipy/nibabel/blob/master/.travis.yml#L20-L21) for wheels for anything that would otherwise need compilation.

satra:
perhaps we should switch since we can pip install all of nipype's requirements at this point.

effigies:
Opened a PR to test independently of these changes. Will merge into this branch if everything goes well.

nipype/pipeline/plugins/multiproc.py
```diff
@@ -74,25 +75,6 @@ def run_node(node, updatehash, taskid):
...
-    def _get_daemon(self):
-        return False
-
-    def _set_daemon(self, value):
-        pass
-
-    daemon = property(_get_daemon, _set_daemon)
-
-
-class NonDaemonPool(pool.Pool):
```
oesteban: I think we still want nondaemon processes, we will need to investigate how to do that with concurrent.futures

nipype/pipeline/plugins/multiproc.py
```diff
@@ -149,30 +129,19 @@ def __init__(self, plugin_args=None):
...
-                     'mem_gb=%0.2f, cwd=%s)', 'non' * int(non_daemon),
+        logger.debug('[MultiProc] Starting (n_procs=%d, '
+                     'mem_gb=%0.2f, cwd=%s)',
                      self.processors, self.memory_gb, self._cwd)
 
-        NipypePool = NonDaemonPool if non_daemon else Pool
-        try:
-            self.pool = NipypePool(
-                processes=self.processors,
-                maxtasksperchild=maxtasks,
```
oesteban: Apparently, there is no way to define a maxtasksperchild with concurrents.futures (https://stackoverflow.com/a/42229574). We should maybe consider loky.

Max tasks per child is important for FMRIPREP, as memory gets cleared up every time a worker is shut down and spun up.

nipype/pipeline/plugins/multiproc.py
effigies: It will require using undocumented API, if it is possible.

What purpose does it serve, by the way?

nipype/pipeline/plugins/multiproc.py
effigies: So if `concurrent.futures` doesn't have the features we need, perhaps it makes more sense to try to push them to fix the bug in `multiprocessing.Pool`?

effigies:
@oesteban Just a bump on this. If we want to have a futures-based plugin, we should decide what we're doing here.

effigies:
@oesteban, I've made a copy of the old `MultiProc` as `LegacyMultiProc`.

The point of this change is to seamlessly fix the crashing problem for people for whom it can be fixed, so using a new plugin name will not really get us where we want (or allow us to test widely). So instead I'm going to risk breaking for people, and we can tell them they can restore the old behavior by using `LegacyMultiProc`, and in the meantime ask them to help us debug the new `MultiProc`.

Does this seem reasonable to everybody?

cc @chrisfilo since #2548 was yours.

oesteban:
This is a great addition :). Have you checked whether it has any influence on the overall throughput? (maybe on fmriprep proper, is it faster?)

effigies:
I don't think it will significantly improve throughput. I don't believe job submission has generally been a major bottleneck, except with extremely short jobs, and then it's been the 2s sleep.

But we can compare the fmriprep test runtimes in that pinned branch with the latest master test runtimes.

effigies:
Maybe I spoke too soon? ds054 and ds210 both have a multi-minute speed-up on the workflow-running step (comparing https://circleci.com/workflow-run/035e3177-0d20-4714-ae9f-b19f1f0a1664 with https://circleci.com/workflow-run/cdcfd49e-e523-4021-a358-900955e679ab). Unclear if that will be a consistent improvement, though.

oesteban:
> That's encouraging.

Indeed. I had the suspicion that this PR would improve performance. My understanding is that the control thread is much more efficient with the futures.concurrent implementation. My only fear is memory, as workers are not automatically reset. And yet we should go ahead since this fixes #2548 in theory.

effigies:
Sounds good. Taking the plunge.

PR #2600, opened 2018-05-27
CI: Stop using Miniconda on Travis

effigies:
Suggested in #2598. Will allow us to test more vanilla installations of Python.

Should speed up Travis builds, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,24 @@ python:
 - 3.6
 
 env:
-- INSTALL_DEB_DEPENDECIES=true NIPYPE_EXTRAS="doc,tests,fmri,profiler" CI_SKIP_TEST=1
-- INSTALL_DEB_DEPENDECIES=false NIPYPE_EXTRAS="doc,tests,fmri,profiler" CI_SKIP_TEST=1
-- INSTALL_DEB_DEPENDECIES=true NIPYPE_EXTRAS="doc,tests,fmri,profiler,duecredit,ssh" CI_SKIP_TEST=1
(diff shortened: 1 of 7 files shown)


satra:
perhaps we can do something through the pytest config, just like we are doing doctests? we could also update the test to use our own format without relying on numpy.

effigies:
> we could also update the test to use our own format without relying on numpy.

Yeah. Apparently the details of the effective formatting change vary by what version of Python the wheel was compiled for (while `conda` packages seemed reasonably consistent). Added a print function that replaces the square brackets with spaces and letting the whitespace normalization do the rest.

effigies:
Hmm. Sphinx's `.. testsetup::` blocks are showing up in the docs:

* https://5032-791477-gh.circle-artifacts.com/0/home/circleci/work/docs/interfaces/generated/nipype.interfaces.image.html#examples
* https://nipype.readthedocs.io/en/latest/interfaces/generated/nipype.interfaces.dcm2nii.html#examples

effigies:
It's an indentation effect. `.. testsetup::` and `.. testcleanup::` are not treated correctly unless they start on the first character of the line. However, if we de-indent these, then the whole text of the docstring is treated as indented, changing appearance.

Is it acceptable to have a de-indented docstring?

satra:
@effigies - sorry didn't see this earlier. if it's necessary, it's necessary. looks darn ugly though :(

effigies:
Yeah. Could be possible to fix up the Sphinx extension, but that's beyond my current ambitions.

effigies:
Actually, I think all but one `testsetup` block can be removed.

PR #2603, opened 2018-05-29
CI: Add PyPI validation on rel/* branches

effigies:
Should at least catch one surprise we ran up against today.

- Add README render check to `rel/*` branches to avoid one source of PyPI rejection


effigies:
@mgxd Any objections to this one?

Issue #2606, opened 2018-06-04
Use geodesic distance to compute Frame Displacements for rigid registration of images


ninamiolane:
### Summary

`FramewiseDisplacement` computes distances between frames at time t and frames at time (t+1). A frame is characterized by 3 degrees of freedom of translations and 3 degrees of freedom of rotations, a frame is thus a 6D vector. 

The distance between two successive frames is currently computed as a L1 (weighted) norm of the element-wise difference between frames at (t+1) and frames at t. This does not respect the geometry of the 6D space of frames, which is the space of 3D rigid transformations, also called Special Euclidean group in 3D, or SE(3). 

We suggest to compute the distance between two frames using a (weighted) geodesic distance that respects the geometry of the 6D space of frames SE(3), i.e.  `SE3_GROUP = SpecialEuclideanGroup(n=3)`.

### Script/Workflow details

First convert the parameterization of frames given as input to a parameterization understandable by `geomstats`.
```
     se3pars = np.hstack(
           [mpars[:, DIM_TRANSLATIONS:], mpars[:, :DIM_TRANSLATIONS]])
```

Construct the metric on the space of frames `SE3_GROUP`:
```
     diag_rotations = self.inputs.radius * np.ones(DIM_ROTATIONS)
     diag_translations = np.ones(DIM_TRANSLATIONS)
     diag = np.concatenate([diag_rotations, diag_translations])
     inner_product = np.diag(diag)
     metric = InvariantMetric(
                       group=SE3_GROUP,
                       inner_product_mat_at_identity=inner_product,
                       left_or_right='left')
```
Use the metric to compute the geodesic distance between frames at t+1 and frames at t.

```
     fd_res = metric.dist(se3pars[:-1], se3pars[1:])
```

### Platform details:

Please paste the output of: `python -c "import nipype; print(nipype.get_info()); print(nipype.__version__)"`

[...]

PR #2607, opened 2018-06-04
ENH WIP: Use geodesic distance in FramewiseDisplacement

ninamiolane:
This is WIP fixing #2606.

Changes proposed in this PR:
- import the module `geomstats` to perform computations on manifolds, like the space of frame displacements in 3D which is the Special Euclidean group in 3D or SE(3)
- modify `FramewiseDisplacement` to compute the displacement of a frame as a geodesic distance on the space of frames SE(3).

TODO:
- investigate what is the current parameterization of the frames, i.e. how is the translation+rotation represented in the vectors of `mpars` @oesteban ?
- adapt unit tests + run make check-before-commit


satra:
@ninamiolane - this would be a nice fix. it's a little closer to the implementation in artifactdetect i think.

https://github.com/nipy/nipype/blob/master/nipype/algorithms/rapidart.py#L103

perhaps to keep compatibility with the original `FramewiseDisplacement`, we could create a new class here or add an option to the existing class. also since this would be a dependency on geomstats we would make that optional and imported inside the class rather than making it a nipype dependency.

i will have to play with `geomstats` but it looks like a really nice package.

nipype/algorithms/confounds.py
```diff
@@ -307,9 +314,20 @@ def _run_interface(self, runtime):
             axis=1,
             arr=mpars,
             source=self.inputs.parameter_source)
-        diff = mpars[:-1, :6] - mpars[1:, :6]
-        diff[:, 3:6] *= self.inputs.radius
-        fd_res = np.abs(diff).sum(axis=1)
```
effigies: This L1 norm is what's used in the cited paper, so I think we should not replace it.

What if we instead add a `metric` option to the input spec, with `L1` and `geodesic` (or is there a better name for this metric) as options?

chrisfilo:
Thanks for the PR @ninamiolane!

+1 for making this optional and keeping the default behavior (there is too much software depending on FD calculated in a particular way).

effigies:
Hi @ninamiolane, do you want to try to get this in for the next release (1.1.1)? I suspect we could get this done in a couple days.

oesteban:
Please keep me in the loop, I'll be more than happy to help fix the tests or whatever you may need from me :)

ninamiolane:
Yes, that'd be great, thanks! Let me have another look tomorrow and I'll ping you if needed.

ninamiolane:
@oesteban @effigies could maybe one of you tell me which parameterization is used for the frames, i.e. for each line of `mpars` here:
[https://github.com/nipy/nipype/blob/master/nipype/algorithms/confounds.py#L310](url)

Each line of `mpars` is 6D: 3 parameters for the translation and 3 parameters for the rotation. Could you tell me which parameter corresponds to what, i.e. which are for the translations, which are for the rotations and which parameterization is used for the rotations: axis-angle, euler angles (which euler angles?)..? I need to convert the `mpars` rotation parameterization to the axis-angle parameterization.

Thank you very much in advance.

effigies:
These have been normalized to SPM format:

https://github.com/nipy/nipype/blob/4b2a8213a1ec19959410f46ecd5fb77368d9cdda/nipype/utils/misc.py#L227-L238

I believe they're in Euler angles, but perhaps @satra knows for certain here...

effigies:
Hi @ninamiolane. Just a heads up that we'll try to make the next release on August 28, so it'd be great to try to get this in by August 20. Please let us know if you have any questions.

requirements.txt
```diff
@@ -1,19 +1,20 @@
-numpy>=1.9.0
-scipy>=0.14
+click>=6.6.0
+configparser
+funcsigs
+future>=0.16.0
+geomstats>=1.7
```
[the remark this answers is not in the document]
ninamiolane: I wanted to sort them in alphabetical order to improve readability. But sure, I can revert! Is there any specific reason why we don't use alphabetical order there? Thanks.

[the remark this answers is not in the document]
effigies: Mostly just to make it obvious during PR review what's changed.

ninamiolane:
SPM seems to be using Euler angles to representation rotations, however it is not clear to me which Euler angle convention is used. (There are -at least?- 12 Euler angles conventions.)

@satra Would you know if the Euler angles used by SPM:
- are in extrinsic (fixed reference frame) or intrinsic coordinates (reference frame is moving with each successive elementary rotation along x, y, and z), 
- are given in the order xyz or zyx (i.e. the order in which to perform the elementary rotations along x, y and z).

Thank you!

https://github.com/nipy/nipype/blob/4b2a8213a1ec19959410f46ecd5fb77368d9cdda/nipype/utils/misc.py#L227-L238

effigies:
Hi Nina, I merged master, which sometimes helps clean up the diff, so that we could look at only the changes you've made. It worked this time, fortunately. To sync your branch, just use `git pull` before making any further changes.

nipype/algorithms/confounds.py
```diff
@@ -236,6 +252,14 @@ class FramewiseDisplacementInputSpec(BaseInterfaceInputSpec):
         "NIPY",
         desc="Source of movement parameters",
         mandatory=True)
+    metric = traits.Enum(
+        'L1',
+        'riemannian',
+        usedefault=True,
+        desc='Distance metric to apply: '
+        'L1 = Manhattan distance (original definition),'
```
effigies: I think a semi-colon will make this slightly more readable than a comma. And when using multiple lines, put a space at the end of each line. This string is currently equivalent to `'Distance metric to apply: L1 = Manhattan distance (original definition),riemannian = Riemannian distance on theSpecial Euclidean group in 3D (geodesic)`.

nipype/algorithms/confounds.py
```diff
@@ -236,6 +252,14 @@ class FramewiseDisplacementInputSpec(BaseInterfaceInputSpec):
         "NIPY",
         desc="Source of movement parameters",
         mandatory=True)
+    metric = traits.Enum(
+        'L1',
+        'riemannian',
+        usedefault=True,
```
effigies: Can you re-add `mandatory=True`? Without it, it's possible for users to un-set the attribute by using `metric=Undefined`, which should be an error.

nipype/algorithms/confounds.py
```diff
@@ -300,16 +324,44 @@ class FramewiseDisplacement(BaseInterface):
...
+            inner_product = np.diag(diag)
+            metric = InvariantMetric(
+                       group=SE3_GROUP,
+                       inner_product_mat_at_identity=inner_product,
+                       left_or_right='left')
+            fd_res = metric.dist(se3pars[:-1], se3pars[1:])
+
+        else:
+            raise ValueError(
+                'The parameter \'metric\' should be \'L1\' or \'riemannian\'.')
```
effigies: By using the `Enum` trait with `mandatory=True` above, the only possible values are `'L1'` or `'riemannian'`, so we can drop this final `else`.

nipype/algorithms/tests/test_confounds.py
```diff
@@ -36,6 +37,21 @@ def test_fd(tmpdir):
         ground_truth, np.loadtxt(res.outputs.out_file, skiprows=1), atol=.16)
     assert np.abs(ground_truth.mean() - res.outputs.fd_average) < 1e-2
 
+    fdisplacement = FramewiseDisplacement(
```
effigies: Could you make this a new test, named something like `test_fd_riemannian`? Then you can preface it with `@pytest.mark.skipif(...)`, like so:

```Python
# Up near nonitime definition (line 13)
nogeomstats = True
try:
    import geomstats
    nogeomstats = False
except ImportError:
    pass

...

@pytest.mark.skipif(nogeomstats, reason="geomstats is not installed")
def test_fd_riemannian(tmpdir):
    ...
```

requirements.txt
```diff
@@ -17,3 +17,4 @@ mock
 pydotplus
 pydot>=1.2.3
 packaging
+geomstats>=1.10
```
effigies: It sounds like we're not prepared to make this a full dependency, so what we can do is add this to an `EXTRA_REQUIRES` in [`nipype.info`](https://github.com/nipy/nipype/blob/master/nipype/info.py):

https://github.com/nipy/nipype/blob/3dc550035eaf91afc2d932a7e7bafe4871c94602/nipype/info.py#L159-L170

Of the existing extras, it probably fits best in `nipy`, and we can put a constraint on it so we don't try to install it in Python versions that don't have it:

```Python
        'nipy': ['nitime', 'nilearn', 'dipy', 'nipy', 'matplotlib',
                 'geomstats>=1.1.0; python_version >= "3.5"'],
```

nipype/algorithms/tests/test_confounds.py
```diff
@@ -36,6 +37,21 @@ def test_fd(tmpdir):
...
+        metric='riemannian')
+    res = fdisplacement.run()
+
+    with open(res.outputs.out_file) as all_lines:
+        for line in all_lines:
+            assert 'FramewiseDisplacement' in line
+            break
+
+    assert np.allclose(
+        ground_truth, np.loadtxt(res.outputs.out_file, skiprows=1), atol=.16)
```
effigies: Going to guess we're going to need a different `ground_truth` file, but that's probably going to require figuring out the Euler angles first. So this is just a flag so we don't forget about it.

docker/generate_dockerfiles.sh
```diff
@@ -107,7 +107,7 @@ function generate_main_dockerfile() {
               pip_opts="-e" \
               pip_install="/src/nipype[all]" \
   --miniconda env_name=neuro \
-              pip_install="grabbit==0.1.2" \
+              pip_install="geomstats grabbit==0.1.2" \
```
effigies: Once we've added it to the `extras`, this should be unnecessary.

nipype/algorithms/confounds.py
```diff
@@ -13,6 +13,7 @@
 
 import nibabel as nb
 import numpy as np
+
```
effigies: Just for the sake of keeping the diff clean, can you remove this newline?

nipype/algorithms/tests/test_auto_FramewiseDisplacement.py
```diff
@@ -12,6 +12,7 @@ def test_FramewiseDisplacement_inputs():
         out_figure=dict(usedefault=True, ),
         out_file=dict(usedefault=True, ),
         parameter_source=dict(mandatory=True, ),
+        metric=dict(usedefault=True, ),
```
effigies: Can you re-run `make specs` and re-commit `test_auto_FramewiseDisplacement`?

nipype/algorithms/tests/test_confounds.py
```diff
@@ -37,6 +45,29 @@ def test_fd(tmpdir):
     assert np.abs(ground_truth.mean() - res.outputs.fd_average) < 1e-2
 
 
+@pytest.mark.skipif(nonitime, reason="nitime is not installed")
+def test_fd_riemannian(tmpdir):
```
effigies: (`no`)`geomstats`, not (`no`)`nitime`

effigies:
So we're looking to release next Monday, and tensorflow still hasn't had a full release support Python 3.7. Additionally, it's probably not great for nipype to start semi-requiring tensorflow or pytorch for such a small functionality increment. I've submitted an issue to geomstats/geomstats#136, and would be happy to submit a patch if you'd like.

Any thoughts on the ground-truth file?

effigies:
Hi @ninamiolane, @oesteban. Any chance of finishing this up this week?

ninamiolane:
@oesteban would you like to meet this week, for example thursday afternoon or friday afternoon? if not, next week?

effigies:
Just a heads up: We're looking to make the next release on Nov 26, if you want to try to get this in this month.

effigies:

PR #2608, opened 2018-06-05
 ENH: Allow BIDS-style slice timings to be passed directly to TShift 

effigies:
This is kind of an upstream from fMRIPrep. It replaces a [helper `Function` interface](https://github.com/poldracklab/fmriprep/blob/9871ccbf36e020114f047fd53e4e0e4eaf385002/fmriprep/workflows/bold/stc.py#L62-L76) we use before `TShift`.

This refines the `-tpattern` inputs to permit both verifying file existence and directly passing a list of floats, which will create a slice timing file to pass as an argument.

- Create `slice_timing` input, mutually exclusive with `tpattern`, which accepts a file
  - If a list of floats is passed, write a file and pass its name to `3dTshift`
- Change `tpattern` to an `Enum`, and list valid strings
  - A generic`Str` is kept as an option, to avoid breaking code that uses `@<fname>`, but warns user that they should use `slice_timing`


effigies:
Don't have time to do the proper tests on fmriprep today, so punting for this release.

effigies:
Just a note that I'm testing this on ds000030/sub-10159. Should be good to go today. Anybody up for a code review?

effigies:
Patching this branch and poldracklab/fmriprep#1160 into the `poldracklab/fmriprep:1.1.2` Docker image produced slice-timing corrected files that differed only in timestamp.

nipype/interfaces/afni/preprocess.py
```diff
@@ -2619,20 +2637,77 @@ class TShift(AFNICommand):
...
     _cmd = '3dTshift'
     input_spec = TShiftInputSpec
     output_spec = AFNICommandOutputSpec
 
+    def _format_arg(self, name, trait_spec, value):
+        if name == 'tpattern' and value.startswith('@'):
+            iflogger.warning('Passing a file prefixed by "@" will be deprecated'
+                             '; please use the `slice_timing` input')
+        elif name == 'slice_timing' and isinstance(value, list):
+            value = self._write_slice_timing()
```
mgxd: if we generate this file, perhaps we should include it in the interface's outputs

effigies: Seems reasonable.

PR #2621, opened 2018-06-19
ENH: S3 access using instance role

### LINKED ISSUE #2620: Boto do not get the chance to connect to S3 bucket using external authentication
### Summary
When using IAM roles on EC2 instances, the DataSink interface does not try to use given IAM role access for the instance to access S3 buckets. It goes straight to an anonymous connection. However, there is no indication in the environment that the role is present, so it can only be accessed by trying to connect to the bucket.

It is a scenario that I had to deal with when running workflows on AWS Batch, by change the Docker image / nipype code manually (yikes!)

### Actual behavior
If AWS Access Key / Secret are not defined via a file or environment variables, it tries to connect anonymously.

### Expected behavior
If AWS Access Key / Secret are not defined via a file or environment variables, it should give a try to IAM Roles before setting the anonymous connection.

https://gist.github.com/anibalsolon/f7586a85942792432e98470c512588b9

After creating this, you need to create an EC2 instance w/ this profile, and run a workflow w/ DataSink without informing Access Key & Secret. It will try to connect anonymously.

### Script/Workflow details
Just a DataSink node for S3 usage.

### Platform details:

```
{
  'pkg_path': '/home/anibalsolon/Documents/nipype/anibalsolon/nipype',
  'commit_hash': 'ad93735',
  'nipype_version': '1.1.0-dev+gad93735',
  'sys_version': '3.6.4 |Anaconda custom (64-bit)| (default, Jan 16 2018, 18:10:19) \n[GCC 7.2.0]',
  'sys_executable': '/opt/anaconda3/bin/python',
  'sys_platform': 'linux',
  'numpy_version': '1.14.0',
  'scipy_version': '1.0.0',
  'networkx_version': '2.1',
  'nibabel_version': '2.2.1',
  'traits_version': '4.6.0'
}
1.1.0-dev+gad93735
```

### Execution environment

Choose one
- Container [Tag: ce73e99f008e / py36]

anibalsolon:
Fixes #2620.

- Try to connect to the bucket non-anonymously when there is no AWS credential, before trying the anonymous config
- It will give the possibility to use instance roles w/ S3 access, that is safer than dealing with secrets (recommended by AWS)

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -123,7 +123,7 @@ def add_traits(base, names, trait_type=None):
     return base
 
 
-def get_head_bucket(s3_resource, bucket_name):
+def _get_head_bucket(s3_resource, bucket_name):
     """ Try to get the header info of a bucket, in order to 
     check if it exists and its permissions
     """
@@ -577,7 +577,7 @@ class DataSink(IOBase):
 
         # And try fetch the bucket with the name argument
         try:
-            get_head_bucket(s3_resource, bucket_name)
+            _get_head_bucket(s3_resource, bucket_name)
         except Exception as exc:
 
             # Try to connect anonymously
@@ -585,7 +585,7 @@ class DataSink(IOBase):
                 'choose-signer.s3.*', botocore.handlers.disable_signing)
 
             iflogger.info('Connecting to AWS: %s anonymously...', bucket_name)
-            get_head_bucket(s3_resource, bucket_name)
+            _get_head_bucket(s3_resource, bucket_name)
 
         # Explicitly declare a secure SSL connection for bucket object
         bucket = s3_resource.Bucket(bucket_name)
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -123,6 +123,35 @@ def add_traits(base, names, trait_type=None):
     return base
 
 
+def get_head_bucket(s3_resource, bucket_name):
+    """ Try to get the header info of a bucket, in order to 
+    check if it exists and its permissions
+    """
+
+    import botocore
+
+    # Try fetch the bucket with the name argument
+    try:
+        s3_resource.meta.client.head_bucket(Bucket=bucket_name)
+    except botocore.exceptions.ClientError as exc:
+        error_code = int(exc.response['Error']['Code'])
+        if error_code == 403:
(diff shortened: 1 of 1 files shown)


anibalsolon:
@mgxd isn't that because fakes3 is not installed?

https://github.com/nipy/nipype/blob/master/nipype/interfaces/tests/test_io.py#L319-L321

However, I am not sure about the appropriate way to test the IAM Role + S3. FakeS3 does not implement it.

nipype/interfaces/io.py
```diff
@@ -536,19 +536,39 @@ def _fetch_bucket(self, bucket_name):
...
+            self._get_head_bucket(s3_resource, bucket_name)
+
         # Explicitly declare a secure SSL connection for bucket object
         bucket = s3_resource.Bucket(bucket_name)
 
+        # Return the bucket
+        return bucket
+
+
+    def _get_head_bucket(self, s3_resource, bucket_name):
```
effigies: This should probably be a `@staticmethod` or a helper function outside the class, as it does not use `self` at all.

nipype/interfaces/io.py
```diff
@@ -592,10 +610,8 @@ def _upload_to_s3(self, bucket, src, dst):
         s3_prefix = s3_str + bucket.name
 
         # Explicitly lower-case the "s3"
-        if dst.lower().startswith(s3_str):
-            dst_sp = dst.split('/')
-            dst_sp[0] = dst_sp[0].lower()
-            dst = '/'.join(dst_sp)
+        if dst[:len(s3_str)].lower().startswith(s3_str):
```
effigies: Since you're already taking the substring, why not:

```Python
if dst[:len(s3_str)].lower() == s3_str:
```

nipype/interfaces/io.py
```diff
@@ -123,6 +123,35 @@ def add_traits(base, names, trait_type=None):
     return base
 
 
+def get_head_bucket(s3_resource, bucket_name):
```
effigies: Minor quibble: I would put a `_` prefix on the function still, just to mark it as a private function that external code shouldn't rely on.

anibalsolon:
Thank you for the reviews @effigies and @mgxd !

effigies:
Thanks. LGTM. @mgxd I'll let you merge if you have no more comments.

PR #2623, opened 2018-06-20
[ENH] add -dsort option to TProject

mattcieslak:
Adds a -dsort option to ``nipype.interfaces.afni.TProject``

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -2497,7 +2497,7 @@ class TProjectInputSpec(AFNICommandInputSpec):
         File(
             exists=True,
             copyfile=False),
-        argstr="%s",
+        argstr="-dsort %s...",
         desc="""Remove the 3D+time time series in dataset fset.
                 ++ That is, 'fset' contains a different nuisance time
                    series for each voxel (e.g., from AnatICOR).
@@ -2570,10 +2570,6 @@ class TProject(AFNICommand):
     input_spec = TProjectInputSpec
     output_spec = AFNICommandOutputSpec
     
-    def _format_arg(self, name, spec, value):
-        if name == 'dsort':
-            return ' '.join(["-dsort %s" % dsort_file for dsort_file in value])
-        return super(TProject, self)._format_arg(name, spec, value)    
 
 class TShiftInputSpec(AFNICommandInputSpec):
     in_file = File(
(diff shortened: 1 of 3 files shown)


nipype/interfaces/afni/preprocess.py
```diff
@@ -2492,6 +2492,14 @@ class TProjectInputSpec(AFNICommandInputSpec):
                    even if -ort contains constant terms, as all means are
                    removed.""",
         argstr="-polort %d")
+    dsort = File(
+        desc="""Remove the 3D+time time series in dataset fset.
+                ++ That is, 'fset' contains a different nuisance time
+                series for each voxel (e.g., from AnatICOR).
+                ++ Multiple -dsort options are allowed.""",
```
mgxd: Since multiple `-dsort` options are allowed, let's allow multiple files.

Here's an example, but you'll want to use `InputMultiObject` instead of `InputMultiPath`

https://github.com/nipy/nipype/blob/704b97dee7848283692bac38f04541c5af2a87b5/nipype/interfaces/dcm2nii.py#L342-L344

mattcieslak:
Thanks for pointing out that this can take multiple files! I changed it, checked that it produced a correct command line, and ran make specs. It looks like the test failed while trying to download something.

nipype/interfaces/afni/preprocess.py
```diff
@@ -2492,6 +2493,15 @@ class TProjectInputSpec(AFNICommandInputSpec):
                    even if -ort contains constant terms, as all means are
                    removed.""",
         argstr="-polort %d")
+    dsort = InputMultiObject(
+        File(
+            exists=True,
+            copyfile=False),
+        argstr="%s",
```
mgxd: you can change this to `argstr="-dsort %s..."` and it'll apply the same behavior you added in `_format_arg`

mgxd:
@mattcieslak this was very close! mind if I reopen?

mattcieslak:
Should it stay open until 1.1.0? Or should I remove the ``_format_arg`` and request another review? Sorry, I'm new to this

nipype/interfaces/afni/preprocess.py
mgxd: @mattcieslak no worries, we love getting new contributions! the only thing missing is changing dsort's argstr (see above) and removing the `_format_arg` method

PR #2626, opened 2018-06-28
ENH: Add versioning metadata to crash files

### LINKED ISSUE #2625: Difficult time to deal with crash files from different nipype versions
### Summary
When using CLI utility to open a crash file, if it was generated in a different nipype version, it may show strange/unrelated errors.

### Actual behavior
Exceptions from different reasons are shown when a crash file is opened with a different nipype version than the one that generated the crash.

### Expected behavior
Better deal with this version difference, at least showing that there is a problem to the user.

Try to open a crash file from a different nipype version, especially when the classes within the crash were changed.

### Script/Workflow details
N/A

### Platform details:

Please paste the output of: `python -c "import nipype; print(nipype.get_info()); print(nipype.__version__)"`
```
{
	'pkg_path': '/home/anibalsolon/Documents/nipype/anibalsolon/nipype',
	'commit_hash': 'b5d79b4',
	'nipype_version': '1.1.0-dev+gb5d79b4',
	'sys_version': '3.6.4 |Anaconda custom (64-bit)| (default, Jan 16 2018, 18:10:19) \n[GCC 7.2.0]',
	'sys_executable': '/opt/anaconda3/bin/python',
	'sys_platform': 'linux',
	'numpy_version': '1.14.0',
	'scipy_version': '1.0.0',
	'networkx_version': '2.1',
	'nibabel_version': '2.2.1',
	'traits_version': '4.6.0'
}
1.1.0-dev+gb5d79b4```

### Execution environment

Choose one

### PR DESCRIPTION
**Pull request by anibalsolon**
Fixes #2625.

- Add a JSON header as metadata for the crash files, containing the original nipype version.
- When opening a crash file with the cli utility, it can check for the nipype version and warn the user about problems that different versions of nipype may generate when using the utility.
- It shows the version from the header.
- It works seamlessly with old versions.

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -628,13 +628,13 @@ def load_json(filename):
 
 
 def loadcrash(infile, *args):
-    if '.pkl' in infile:
-        return loadpkl(infile)
+    if infile.endswith('pkl') or infile.endswith('pklz'):
+        return loadpkl(infile, versioning=True)
     else:
         raise ValueError('Only pickled crashfiles are supported')
 
 
-def loadpkl(infile):
+def loadpkl(infile, versioning=False):
     """Load a zipped or plain cPickled file
     """
     fmlogger.debug('Loading pkl: %s', infile)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: nipype/utils/filemanip.py [hunk 1]
```diff
@@ -682,11 +715,21 @@ def read_stream(stream, logger=None, encoding=None):
...
+def savepkl(filename, record, versioning=False):
     if filename.endswith('pklz'):
         pkl_file = gzip.open(filename, 'wb')
     else:
         pkl_file = open(filename, 'wb')
+
+    if versioning:
+        from nipype import __version__ as version
+        metadata = json.dumps({'version': version},
+                              ensure_ascii=True,
```
mgxd: why not use UTF-8 like we do throughout the package?

nipype/utils/filemanip.py
```diff
@@ -682,11 +715,21 @@ def read_stream(stream, logger=None, encoding=None):
...
+                              encoding='ascii')
+
+        pkl_file.write(metadata.encode('ascii'))
+        pkl_file.write('\n'.encode('ascii'))
```
mgxd: utf-8?

nipype/utils/filemanip.py
```diff
@@ -643,11 +643,44 @@ def loadpkl(infile):
     else:
         pkl_file = open(infile, 'rb')
 
+    if versioning:
+        pkl_metadata = {}
+
```
mgxd: many blank lines (personal preference, feel free to ignore)

nipype/utils/filemanip.py
anibalsolon: My initial preference around ASCII over UTF-8, in this case, is about Python 2 retro-compatibility.
If the error contains a path, for example, that uses an "é", Python 2 might not get so happy about it. Since JSON can be fully represented using ASCII, I thought that this config might avoid some headaches.

I've done some tests now, and using UTF-8 will not be a problem at all.

anibalsolon:
Thank you for the review @mgxd !

Issue #2539, opened 2018-04-11
Discuss: Nipype boilerplate simplification


satra:
### Summary

When one writes a nipype interface or workflow there is a fair bit of boilerplate. The intent of this discussion is to figure out how we reduce/simplify this, and to think about the v2 api.

should we:

1. just simplify syntax (allow more restricted forms only)/update our examples to use more compact pythonic forms. for example:

```python
# common form
a.inputs.one = 2
a.inputs.two = 3

# alternate allowed forms
a  = Interface(one=2, two=3)
or
a = Interface(...)
a.inputs.update(one=2, two=3)
```
except for special interfaces (e.g., Function)

2. add support for lazy evaluation in new api and make it more compact (my current preference):

```python
wf2=Workflow('wf12, ..., distribute=False, global_cache=False)
...
wf1=Workflow('wf1', ..., distribute=True)
wf1.add('a', BET(...))
wf1.add('wf2', wf2(..., inputs='node2.in1=a.out1'))
wf1.add('b', antsRegistration(..., in1='a.out1')).map('in1'))
wf1.add('x', SomeFunc(...)) # map automatically continues
wf1.add('c', ImageMaths(..., in1='wf2.node1.out1')).join('b.in1')
wf1.run()
```
or more compactly:

```python
wf2=Workflow('wf2', ..., distribute=False, global_cache=False)
...
wf1=Workflow('wf1', ..., distribute=True)
    .add('a', BET(...))
    .add('wf2', wf2(..., inputs='node2.in1=a.out1'))
    .add('b', antsRegistration(..., in1='a.out1')).map('in1'))
    .add('x', SomeFunc(...)) # map automatically continues
    .add('c', ImageMaths(..., in1='wf2.node1.out1')).join('b.in1')
wf1.run()
```

`.add` will return a composite reference to both the workflow and the interface added

3. create a DSL (domain specific language - like WDL/CWL) instead of using Python

@djarecka @Shotgunosine @effigies @chrisfilo @oesteban - open for discussion.

satra:
instead of 3, we should also be able to do:

`wf1 = Workflow(...).from_cwl('/path/to/cwl')` 

would use the cwl library to parse the file and convert to a nipype workflow.

and correspondingly `wf1.to_cwl()/boutiques()`

satra:
also for `map` and `join`, i'm thinking we could do something like:

1. map
[...]

PR #2629, opened 2018-07-02
WIP: New workflow syntax

effigies:
Work started during OHBM 2018 hackathon.

Closely following discussion in #2539. The goal is to start with @satra's syntax, and incorporate mine as a syntactic sugar. We'll see if they're equally implementable, or if there are inherent limits or affordances of each. Neither approach was fully consistent between comments, so this is somewhat an exegetical exercise.

Should merge with @djarecka's work on unifying nodes and workflows, and adding mapping and joining syntax.


nipype/pipeline/engine/workflows.py
```diff
@@ -1043,3 +1044,109 @@ def _get_dot(self,
...
+class MapState(object):
+    pass
+
+class NewNode(EngineBase):
+    def __init__(self, inputs={}, map_on=None, join_by=None,
+                 *args, **kwargs):
+        self._mappers = {}
+        self._joiners = {}
+
+    def map(self, field, values=None):
```
djarecka: I understand that `field` is a "mapper" itself (either list, tuple or string), is that right? Not sure about `values`, do you want to be a dictionary? From the Satra example I understand that you can provide values of some of the fields from the mapper (i.e. `field` in your syntax).

nipype/pipeline/engine/workflows.py
```diff
@@ -1043,3 +1044,109 @@ def _get_dot(self,
...
+            node = Node(runnable, name=name)
+        elif is_node(runnable):
+            node = runnable if runnable.name == name else runnable.clone(name=name)
+        else:
+            raise ValueError("Unknown workflow element: {!r}".format(runnable))
+        setattr(self, name, node)
+        self._nodes[name] = node
+        self._last_added = name
+
+    def map(self, field, node=None, values=None):
```
djarecka: not sure what is the meaning of `node`. it shouldn't be just `self._last_added`?

nipype/pipeline/engine/workflows.py
```diff
@@ -1043,3 +1044,109 @@ def _get_dot(self,
...
+        self.add_nodes([node])
+
+
+class Map(Node):
+    pass
+
+
+class Join(Node):
+    pass
+
```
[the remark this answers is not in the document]
effigies: This is just prep for my syntax suggestion from #2539, where `Join(node, ...)` is equivalent to `node.join(...)`.

nipype/pipeline/engine/workflows.py
effigies: I'm not actually positive of what I was doing here. Probably should have responded more promptly. :-) In any event, I clearly didn't finish this function (see L1076).

I believe what I was suggesting is that you could do:

```Python
node.map('a', [1, 2, 3, 4])

# or, where node.inputs.a takes the output of another node
node.map('a')
```

nipype/pipeline/engine/workflows.py
effigies: Here I'm almost certain the idea is that I could do:

```Python
wf.add(node).map(<something over node>)

# or
wf.add(node)
...
wf.map(<something over node>, node=node)
```

nipype/pipeline/engine/workflows.py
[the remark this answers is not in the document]
djarecka: ok

nipype/pipeline/engine/workflows.py
djarecka: oh, I see. Didn't think about this syntax.

djarecka:
@effigies fyi: I was fixing already some exceptions etc. https://github.com/djarecka/nipype/commit/682248a7f31e51a85541b8355e4c6f58a10a53cb

nipype/pipeline/engine/workflows.py
```diff
@@ -1043,3 +1044,109 @@ def _get_dot(self,
...
+        if node in self._mappers:
+            raise WorkflowError("Cannot assign two mappings to the same input")
+
+        self._mappers[node] = (field, values)
+
+    def join(self, field, node=None):
+        pass
+
+
+def is_function(obj):
```
kaczmarj: this is nitpicking, but the builtin `callable` can be used here instead of defining another function.

PR #2611, opened 2018-06-06
MAINT: fix interaction with external logging

mgxd:
Fixes #2569  .

Changes proposed in this pull request
- remove call to `logging.basicConfig`
- scope sub-loggers (interface, workflow, etc) within `nipype` ancestor

PR #2630, opened 2018-07-02
FIX: LegacyMultiProc workflow logger

effigies:
#2598 did not incorporate #2611's move to hierarchical loggers.

This is likely to be a breaking change for a lot of downstream users, so we should make a particular note of it in the release notes.


effigies:
Looks fixed. Merging to get master tests run.

Issue #2395, opened 2018-01-22
temporary directories in OS X


djarecka:
### Summary

There are several issues with the way OSX handles temporary dictionary, i.e. `/var -> private/var`, see discussion in #2380.
The problem with `test_nilearn` was fixed, but this might come back with various interfaces.

mgxd:
related: nipy/heudiconv#231

djarecka:
#2639 might solve the problem permanently, but let's observe a bit longer. 

If we notice any new problems, we probably should try using `realpath` in `File`, `Directory` and `ImageFile`.

satra:
@djarecka - cannot use realpaths in `File` as they would resolve symlinks to their original location, thereby removing any effects of `copyfile=False` in traits.

i was ok with output directory of a node using realpath in this PR, but this is something to keep an eye out for.

djarecka:
@satra ok, thanks for the comment, didn't think about `copyfile=False`. let's see if the problem come back

djarecka:
#2676  reporting problem on Gentoo might be related (I thought that this problems are only with OSX)

PR #2639, opened 2018-07-16
changing Node._output_dir to realpath

djarecka:
I changed `abspath` to `realpath` in `Node._output_dir` to fix problems with temporary directories on OSX. `abspath` gives `/var/...` and `realpath` returns `/private/var...`.  In [`clean_working_directory`](https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/utils.py#L1385) we had `needed_files` that used `/private/var/...` and `cwd` used `/var/...`, so many files were removed.

I'm not sure if this is the best place to change the path, I can do it also in `clean_working_directory`, if for any reason we should keep `abspath` in `_output_dir`.

There is a chance that this will solve the general issue from #2395. For now it fixed all tests from `heudiconv`, so should also solve [this issue](https://github.com/nipy/heudiconv/issues/231)

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -275,7 +275,7 @@ class Node(EngineBase):
                 params_str = [_parameterization_dir(p) for p in params_str]
             outputdir = op.join(outputdir, *params_str)
 
-        self._output_dir = op.abspath(op.join(outputdir, self.name))
+        self._output_dir = op.realpath(op.join(outputdir, self.name))
         return self._output_dir
 
     def set_input(self, parameter, val):


mgxd:
@djarecka are unused outputs still removed within a workflow (by default)? if so, i'm good with this change - nice catch

djarecka:
@mgxd - yes, unused output is still removed, you can check at the end that you don't have all dcm files at the end. My changes just ensure that we run `walk_files` using path that is a `realpath` and not an `abspath`.

`Dcm2niix` that we were testing yesterday uses `abspath` in [out_file](https://github.com/nipy/nipype/blob/master/nipype/interfaces/dcm2nii.py#L432), but `fnam` contains `.`, so `abspath` is equivalent to `realpath`.

I hope that we don't have interfaces that return  `abspath` that is different than `realpath` in output (this might lead to similar problems). If we notice any new problems, I'd suggest adding `realpath` in `File` class: [here](https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/traits_extension.py#L105) and [here](https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/traits_extension.py#L107) (and similar changes in `Directory` and `ImageFile`).

mgxd:
OK, let's merge this and see!

PR #2642, opened 2018-07-19
ENH: Allow transform to be saved from AFNI 3dWarp

salma1601:
I added an option to save the warp from the stdout because it can be useful. But the warp transform is printed only if  `verbose` is set to True, otherwise the saved file is empty. I don't know if there is something like the `xor` and `requires` that checks if a required trait is set to True.

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -2758,7 +2758,8 @@ class WarpInputSpec(AFNICommandInputSpec):
         name_template='%s_warp',
         desc='output image file name',
         argstr='-prefix %s',
-        name_source='in_file')
+        name_source='in_file',
+        keep_extension=True)
     tta2mni = traits.Bool(
         desc='transform dataset from Talairach to MNI152', argstr='-tta2mni')
     mni2tta = traits.Bool(
@@ -2789,6 +2790,13 @@ class WarpInputSpec(AFNICommandInputSpec):
         argstr='-zpad %d')
     verbose = traits.Bool(
         desc='Print out some information along the way.', argstr='-verb')
+    save_warp = traits.Bool(
+        False, usedefault=True, desc='save warp as .mat file')
+
+
+class WarpOutputSpec(TraitedSpec):
+    out_file = File(desc='Warped file.', exists=True)
+    warp_file = File(desc='warp transfrom .mat file')
 
 
 class Warp(AFNICommand):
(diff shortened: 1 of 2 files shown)


nipype/interfaces/afni/preprocess.py
```diff
@@ -2789,6 +2790,13 @@ class WarpInputSpec(AFNICommandInputSpec):
         argstr='-zpad %d')
     verbose = traits.Bool(
         desc='Print out some information along the way.', argstr='-verb')
+    save_warp = traits.Bool(
+        False, usedefault=True, desc='save warp as .mat file')
```
effigies: The default of a `Bool` is already false. You can just have `desc` here.

nipype/interfaces/afni/preprocess.py
```diff
@@ -2820,7 +2828,48 @@ class Warp(AFNICommand):
...
+            import numpy as np
+            if not self.inputs.out_file:
+                warp_file = fname_presuffix(self.inputs.in_file,
+                                            suffix='_warp_transform.mat',
+                                            use_ext=False)                        
+            else:
+                warp_file = fname_presuffix(self.inputs.out_file,
+                                            suffix='_transform.mat',
+                                            use_ext=False)
+            np.savetxt(warp_file, [runtime.stdout], fmt=str('%s'))
```
effigies: Rather than calculate the filename twice:

```Python
if self.inputs.save_warp:
    import numpy as np
    warp_file = self._list_outputs()['warp_file']
    np.savetxt(warp_file, [runtime.stdout], fmt='%s')
```

(Not sure why you have `str('%s')`. If that's important, feel free to re-add.)

nipype/interfaces/afni/preprocess.py
```diff
@@ -2789,6 +2790,13 @@ class WarpInputSpec(AFNICommandInputSpec):
         argstr='-zpad %d')
     verbose = traits.Bool(
         desc='Print out some information along the way.', argstr='-verb')
+    save_warp = traits.Bool(
+        False, usedefault=True, desc='save warp as .mat file')
+
+
+class WarpOutputSpec(TraitedSpec):
+    out_file = File(desc='Warped file.', exists=True)
+    warp_file = File(desc='warp transfrom .mat file')
```
effigies: s/transfrom/transform/

nipype/interfaces/afni/preprocess.py
```diff
@@ -2820,7 +2828,48 @@ class Warp(AFNICommand):
...
+                outputs['warp_file'] = os.path.abspath(warp_file)
+        else:
+            outputs['out_file'] = os.path.abspath(self.inputs.out_file)
+            if self.inputs.save_warp:
+                warp_file = fname_presuffix(self.inputs.out_file,
+                                            suffix='_transform.mat',
+                                            use_ext=False)
+                outputs['warp_file'] = os.path.abspath(warp_file)
+
+        return outputs
```
effigies: Would the following work:

```Python
def _list_outputs(self):
    outputs = super(Warp, self)._list_outputs()
    if self.inputs.save_warp:
        outputs['warp_file'] = fname_presuffix(outputs['out_file'],
                                               suffix='_transform.mat',
                                               use_ext=False)

    return outputs
```

nipype/interfaces/afni/preprocess.py
effigies: Oh, and RE your initial comment, I would set `requires=['verbose']`. I'm not sure what will happen if someone sets `verbose=False`, but at least it will catch when `verbose` is unset.

salma1601:
>Oh, and RE your initial comment, I would set requires=['verbose']. I'm not sure what will happen if someone sets verbose=False, but at least it will catch when verbose is unset.

Ok. I was putting the WIP because I was unsure there isn't a better solution to force `verbose` to True when `save_warp` is set to True. Then I will remove it

nipype/interfaces/afni/preprocess.py
salma1601: >(Not sure why you have str('%s'). If that's important, feel free to re-add.)

When I don't explicit the format, I have an error
TypeError:  Mismatch between array dtype ('<U209') and format specifier ('%.18e')

So we need to keep it I think.

effigies: Oh, I meant `str('%s')` vs `'%s'` (or is that also what you meant?). The `str()` seems like it should be a null-op, unless there's some weird Python2/3 hackery going on here.

nipype/interfaces/afni/preprocess.py
salma1601: what do you mean ?

effigies: Oh, sorry. That's regular expression. I just meant you misspelled "transform".

salma1601: Sorry ! I am correcting that

effigies:
Thanks. The test failures are related to #2650.

salma1601:
>The test failures are related to #2650.

Ok. Thank you for the review !

Issue #2645, opened 2018-07-22
Cannot start nipype.test()


shuiyiwensi:
### Summary
Run nipype.test(), shows
-c: error: unrecognized arguments: -n

### Actual behavior
usage: -c [options] [file_or_dir] [file_or_dir] [...]
-c: error: unrecognized arguments: -n
  inifile: /home/wensi/anaconda3/envs/python35/lib/python3.5/site-packages/nipype/pytest.ini
  rootdir: /home/wensi/anaconda3/envs/python35/lib/python3.5/site-packages/nipype

### Expected behavior
expect this will run

### How to replicate the behavior
just follow the two command lines on[ Anaconda nipype](https://anaconda.org/conda-forge/nipype)
then run 
python -c "import nipype; nipype.test()"

### Script/Workflow details

Please put URL to code or code here (if not too long).

### Platform details:

Please paste the output of: `python -c "import nipype; print(nipype.get_info()); print(nipype.__version__)"`

output is 
/home/wensi/anaconda3/envs/python35/lib/python3.5/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
{'pkg_path': '/home/wensi/anaconda3/envs/python35/lib/python3.5/site-packages/nipype', 'networkx_version': '2.1', 'sys_version': '3.5.5 |Anaconda custom (64-bit)| (default, May 13 2018, 21:12:35) \n[GCC 7.2.0]', 'commit_source': 'archive substitution', 'nibabel_version': '2.3.0', 'commit_hash': '%h', 'sys_executable': '/home/wensi/anaconda3/envs/python35/bin/python', 'sys_platform': 'linux', 'numpy_version': '1.14.3', 'nipype_version': '1.1.0', 'scipy_version': '1.1.0', 'traits_version': '4.6.0'}
1.1.0

### Execution environment
python 3.5

Choose one
- Container [Tag: ???]
- My python environment inside container [Base Tag: ???]
- My python environment outside container

djarecka:
it looks like you're missing one package [pytest-xdist](https://docs.pytest.org/en/3.0.0/xdist.html). You can `pip install` it (or skip the `-n`, as you did). And we should make sure that this is available in conda build. Thank you for letting us know!

PR #2649, opened 2018-07-25
MAINT: Make pytest-xdist a dependency

djarecka:
Since we have `pytest` in `REQUIRES` I added also `pytest-xdist`, since this is now required to run `nipype.test()` (motivated by #2645)

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -143,6 +143,7 @@ REQUIRES = [
     'click>=%s' % CLICK_MIN_VERSION,
     'funcsigs',
     'pytest>=%s' % PYTEST_MIN_VERSION,
+    'pytest-xdist',
     'mock',
     'pydotplus',
     'pydot>=%s' % PYDOT_MIN_VERSION,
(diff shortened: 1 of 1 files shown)


djarecka:
it's failing because of #2650

oesteban:
Why not moving the other way around? (pytest to TESTS_REQUIRES)

djarecka:
@oesteban - sure, can do other way around. However, if I remember correctly some discussion when I was changing tests, we wanted user to be able to run `nipype.test` even after the most basic installation. am I wrong?

oesteban:
You are right. I'm happy either way, just to make sure we are doing this.

effigies:
Either way is fine with me. Strictly speaking, xdist shouldn't be a requirement, but it does seem to be causing people issues. My preferred option would probably be to disable `-n auto` if xdist isn't present, but that's probably not trivial.

Given how tightly integrated pytest is at this point, with the fixtures and all, having tests as first class components means it is a dependency. So if we can't decouple xdist, we should probably have it as a normal `REQUIRES`.

djarecka:
@effigies I think you can't use `-n` flag at all if you don't have `xdist`. We can always check if `xdist` is available and based on this change the pytest command. On the other hand, I don't know if we have a good reason to avoid `xdist` in the requirements (I mean we're asking for many outside standard library packages anyway)

PR #2651, opened 2018-07-25
FIX: Various BIDSDataGrabber fixes

### LINKED ISSUE #2650: pybids 0.6.4 causes test failure in BIDSDataGrabber
### Summary

The error in https://travis-ci.org/nipy/nipype/builds/408021599:

```
______________________________ test_bids_grabber _______________________________
[gw0] linux -- Python 3.5.5 /home/travis/virtualenv/python3.5.5/bin/python
tmpdir = local('/tmp/pytest-of-travis/pytest-0/popen-gw0/test_bids_grabber0')
    @pytest.mark.skipif(not have_pybids,
                        reason="Pybids is not installed")
    @pytest.mark.skipif(sys.version_info < (3, 0),
                        reason="Pybids no longer supports Python 2")
    @pytest.mark.skipif(not dist_is_editable('pybids'),
                        reason="Pybids is not installed in editable mode")
    def test_bids_grabber(tmpdir):
        tmpdir.chdir()
        bg = nio.BIDSDataGrabber()
        bg.inputs.base_dir = os.path.join(datadir, 'ds005')
        bg.inputs.subject = '01'
        results = bg.run()
        assert os.path.basename(results.outputs.anat[0]) == 'sub-01_T1w.nii.gz'
>       assert os.path.basename(results.outputs.func[0]) == (
            'sub-01_task-mixedgamblestask_run-01_bold.nii.gz')
E       AssertionError: assert 'sub-01_task-...01_events.tsv' == 'sub-01_task-m...1_bold.nii.gz'
E         - sub-01_task-mixedgamblestask_run-01_events.tsv
E         ?                                     ^^^ ^^ ^^^
E         + sub-01_task-mixedgamblestask_run-01_bold.nii.gz
E         ?                                     ^^^^^ ^^ ^^
/home/travis/build/nipy/nipype/nipype/interfaces/tests/test_io.py:596: AssertionError
```

@adelavega @tyarkoni Flagging you in this in case the cause is obvious.

It's not immediately clear whether the correct fix will be in pybids or nipype.

### Execution environment

Travis CI

effigies:
## Summary

In future pybids versions, derivatives will not be automatically indexed, and test data has been added to help ensure that derivatives are properly handled.

This is a potentially breaking change, so we should consider if we would prefer some other approach.

Fixes #2650.

* Add `strict` input to `BIDSDataGrabber` that currently excludes `derivatives/`, `code/` and `sourcedata/` subdirectories from indexing.
* Only return images by default, per documentation.
* Make test more resilient by checking whether expected file is present, not first.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -2828,7 +2828,8 @@ class BIDSDataGrabber(IOBase):
         return runtime
 
     def _list_outputs(self):
-        layout = gb.BIDSLayout(self.inputs.base_dir)
+        layout = gb.BIDSLayout(self.inputs.base_dir,
+                               exclude=['derivatives/', 'code/', 'sourcedata/'])
 
         # If infield is not given nm input value, silently ignore
         filters = {}
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -2804,8 +2804,8 @@ class BIDSDataGrabber(IOBase):
 
         if not isdefined(self.inputs.output_query):
             self.inputs.output_query = {
-                "func": {"modality": "func", extensions=['nii', '.nii.gz']},
-                "anat": {"modality": "anat", extensions=['nii', '.nii.gz']},
+                "func": {"modality": "func", 'extensions': ['nii', '.nii.gz']},
+                "anat": {"modality": "anat", 'extensions': ['nii', '.nii.gz']},
                 }
 
         # If infields is empty, use all BIDS entities
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -2803,8 +2803,10 @@ class BIDSDataGrabber(IOBase):
         super(BIDSDataGrabber, self).__init__(**kwargs)
 
         if not isdefined(self.inputs.output_query):
-            self.inputs.output_query = {"func": {"modality": "func"},
-                                        "anat": {"modality": "anat"}}
+            self.inputs.output_query = {
+                "func": {"modality": "func", extensions=['nii', '.nii.gz']},
+                "anat": {"modality": "anat", extensions=['nii', '.nii.gz']},
+                }
 
         # If infields is empty, use all BIDS entities
         if infields is None and have_pybids:
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -2755,6 +2755,8 @@ class BIDSDataGrabberInputSpec(DynamicTraitedSpec):
                                  desc='Generate exception if list is empty '
                                  'for a given field')
     return_type = traits.Enum('file', 'namedtuple', usedefault=True)
+    strict = traits.Bool(desc='Return only BIDS "proper" files (e.g., '
+                              'ignore derivatives/, sourcedata/, etc.)')
 
 
 class BIDSDataGrabber(IOBase):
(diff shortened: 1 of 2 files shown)


nipype/interfaces/io.py
```diff
@@ -2828,7 +2828,8 @@ def _run_interface(self, runtime):
         return runtime
 
     def _list_outputs(self):
-        layout = gb.BIDSLayout(self.inputs.base_dir)
+        layout = gb.BIDSLayout(self.inputs.base_dir,
+                               exclude=['derivatives/', 'code/', 'sourcedata/'])
```
mgxd: This should be more elastic, I'd prefer adding an option to the inputspec:

```python
exclude = traits.List(traits.Str, value=['derivatives/', 'code/', 'sourcedata/'], usedefault=True, ...)
```
and then plugging that in here

adelavega:
See my comments on the issue, but I think the other part of the problem is that event files should not be indexed by the default behavior, only `bold` and `anat` files. So you need to add `type='bold'` to the default query.

nipype/interfaces/io.py
adelavega: Agreed. There are many instances in which you would want to index derivatives. By the way, if that's the case you probably also want to initate the `BIDSLayout` with 'bids' and 'derivatives' spec for the derivatives folder... So maybe we need to re-think how input folders are specified in this module (sort of how `fitlins` does it?)

effigies: There are weird compatibility issues here, such as whether the interface should have consistent behavior across some pretty wild API changes (some completed, some planned) in pybids, or just to call whatever version is installed, and have users pray.

For example, I don't think `exclude` was present in pybids when we added this interface. And once we stop indexing `derivatives/` by default, the appropriate way to modulate this will be to add derivatives, instead. So at that point, do we change the default `exclude` values, the effect of which will vary based on pybids version, or what?

mgxd: it looks like `BIDSLayout` added kwarg support in version `0.0.2`, and I doubt anyone is using an older version than that - I don't think it'll be breaking to add `exclude`, at worst it will just be ignored. 

However, if pybids will be setting this automatically in the future, perhaps we shouldn't set a default in nipype. To fix the tests, we can just make them exclude the necessary directories for now.

adelavega: Yeah, fair enough. Maybe for now we can just add the `exclude` trait (which should be fairly stable), and not worry about indexing derivatives for now. I image 99% of users only want to index core BIDS files. Then we can wait for pybids to mature a bit (tal said he might add a 'index_derivatives' argument).

effigies: Is the goal to have this provide all options that `BIDSLayout` provides? That seems dangerous, given how much of a moving target that is. I think more we should consider `BIDSDataGrabber` its own API and strive to make it behave consistently across pybids versions.

That's one reason I'm very hesitant to add `exclude`, because the interpretation of it changes based on what the default inclusions/exclusions are.

adelavega: I think the original goal was that it was a `BIDSLayout` wrapper, but I don't think it needs to mirror all the functionality. I'm fine leaving that off for now and reconsidering later once pybids has matured/stabilized a bit (or at least until BIDS Derivatives becomes more stable itself). I do think we're going to want to deal w/ how derivatives are indexed sooner or later (otherwise it limits the utility of `BIDSDataGrabber` quite a bit IMO).

effigies: Cool. I also think there's potential for multiple interfaces, if we get to the point where we have to keep cramming options into the one. So it's worth thinking about what are the essential pieces of functionality we want, and at what point is the complexity cost of more options greater than a new interface with a different target.

This is part of why I'm writing my own interfaces in FitLins... once I see what all I need, I can think about whether merging with `BIDSDataGrabber` makes sense or not.

PR #2652, opened 2018-07-25
FIX: replace the MRtrix3 command fs_parc_replace_sgm_first to labelsgmfix

SheridanQ:
The MRtrix3 has renamed the fs_parc_replace_sgm_first to labelsgmfix

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/interfaces/mrtrix3/preprocess.py
+++ b/nipype/interfaces/mrtrix3/preprocess.py
@@ -249,12 +249,12 @@ class ReplaceFSwithFIRST(CommandLine):
     >>> prep.inputs.in_t1w = 'T1.nii.gz'
     >>> prep.inputs.in_config = 'mrtrix3_labelconfig.txt'
     >>> prep.cmdline                               # doctest: +ELLIPSIS
-    'fs_parc_replace_sgm_first aparc+aseg.nii T1.nii.gz \
+    'labelsgmfix aparc+aseg.nii T1.nii.gz \
 mrtrix3_labelconfig.txt aparc+first.mif'
     >>> prep.run()                                 # doctest: +SKIP
     """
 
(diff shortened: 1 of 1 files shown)


effigies:
Was there a version where the old command was correct? If so, we should make the command depend on the version.

SheridanQ:
Hi effigies,
I am new to the github, sorry that I didn't include this in my PR. 
The last version where the old command was correct is version 0.3.13. After this version there was a [major upgrade](http://www.mrtrix.org/2016/03/12/major-update-to-mrtrix3/) released on Mar 12 of 2016 that changed the command.

effigies:
Okay, then we'll need to create a `nipype.interfaces.mrtrix3.base.Info` class that knows how to determine the version number. We can then select the correct command name based on whether the version is > 0.3.13.

@satra @mgxd Is there an existing example of an interface that has changed `_cmd` on version, or what would be the best model for this?

mgxd:
After making an `Info` class for `Mrtrix3` similar to the one below
https://github.com/nipy/nipype/blob/1ffe23d2b82b6fefa9fa92e59e86437591041081/nipype/interfaces/workbench/base.py#L25-L43
you can have a check in the interface's `__init__` that can compare versions and change the `_cmd`, something like we do here
https://github.com/nipy/nipype/blob/1ffe23d2b82b6fefa9fa92e59e86437591041081/nipype/interfaces/c3.py#L160-L164

@SheridanQ Lemme know if you have any questions/concerns!

effigies:
Hi @SheridanQ. Please let us know if you need any help to finish this off.

effigies:
Hi @SheridanQ, we're going to try to release 1.1.3 next Monday. Do you have time to finish this one up before then?

effigies:

PR #2648, opened 2018-07-24
[WIP] Add state/mappers to new workflow syntax

djarecka:
adding my mapping and submitter to #2629. Had problem with merging NewNode with my submitter without creating an extra class NewNodeBase and treating NewNode as a wrapper that runs nodes's interface.

NeeWorkflows inherits from NewNode.

For now, I only added things related mapper, ignored join and joindByKey.

PR #2654, opened 2018-07-26
MNT: Drop Python 2 support

effigies:
## Summary

This PR is a notification that I've started a `dev/2.0` branch, and perhaps a place to centralize discussion.

It's already basically unreadable, due to the changes to `make specs`, so if anybody wants to review, I'd recommend doing it commit-by-commit. Once we decide that this is okay, we then direct any further work to PRs against this branch (e.g. #2629 and #2648).

I'd also suggest periodically opening a PR to merge `master` into `dev/2.0`, perhaps at release time, so that 1.x improvements make it in.

* Remove Python 2 compatibility shims
* Remove Python 2 tests in CI
* Remove `interfaces.ants.legacy`
* Re-run `make specs`

## Acknowledgment


satra:
@effigies - since 2.0 is python 3 only, a lot of current code (e.g., use of future) can be removed/simplified. but this will also make bringing fixes in from 1.0 require some changes. we should perhaps not merge master, but cherry pick commits from master into a new PR against dev/2.0 (where a clean merge is possible that would be fine with me).

another thought is whether to break off the engine, interfaces into their own separate projects. we can start with the engine, since i think that is the most self-contained component.

.circleci/config.yml
```diff
@@ -192,16 +172,15 @@ jobs:
...
+                          nipype/nipype:py36 | gzip -1 > /tmp/docker/nipype-base-latest-py36.tar.gz \
+              && du -h /tmp/docker/nipype-base-latest-py36.tar.gz
             fi
       - persist_to_workspace:
           root: /tmp
           paths:
             - docker
 
-  test_py3_fmri_fsl_spm:
+  test_fmri_fsl_spm:
```
satra: should we consider breaking off the workflows and workflow tests to niflows, before we break off 2.0?

satra: i'm also fine with making all the breaks with 2.0.

.circleci/config.yml
```diff
@@ -339,21 +316,15 @@ jobs:
       - run:
           name: Check pypi preconditions
           command: |
+            pyenv local 3.5.2
             pip install --upgrade twine future wheel readme_renderer setuptools
```
satra: shouldn't we remove `future`?

.circleci/config.yml
effigies: I'm happy to start moving things to niflows. Have we figured out a working plan for that? Maybe could do a chat sometime this week to iron out a rough plan?

satra:
@effigies - current changes look good to me.

.circleci/config.yml
satra: i'll coordinate a time.

effigies:
> we should perhaps not merge master, but cherry pick commits from master into a new PR against dev/2.0

The reason I like merging master is that it marks changes as already included, which makes the next merge off master simpler. I am doing a `git merge --no-commit` and curating, at this point. In the future, we could split some branch off master that we then PR into `dev/2.0`, and can do any cleanups there. Maybe we can try that after 1.1.3?

effigies:
And thanks for the review. I'll make the `future` change you suggested, and then we can start going through the PR process for this branch instead.

satra:
@effigies - merging master is fine, as long as we are curating.

satra:
@effigies, @oesteban - should the following feature set be nipype `1.(next + 1)` or 2?

- py3 only
- separate out interface packages as different repos

i.e do we make this clean break with pydra-based nipype or earlier?

effigies:
I would be happy to drop official Py2 support before 2.0. This branch is almost entirely that, and...

Screen Shot 2019-08-20 at 22 04 32

My feeling is that if it's mostly done and would make 1.x maintenance easier overall (to whatever extent that will need to keep happening), it's worth doing.

Which purging Python 2 is. Separating out interface packages isn't really, so I'd lean toward making that part of 2.0.

effigies:
@satra I interpreted "nipype `1.(next + 1)`" as 1.4.0. Is this correct, and do you have a timeline? Perhaps early January, to match Python 2.7 EOL?

satra:
@effigies - let's say 1.4 will be released mid december, with master branch moving from supporting both to pure py3 starting after the next release 1.3 (which is scheduled soon, correct?) we should make a release branch for any specific hot fixes that should go in to a 1.3.x release

effigies:
@satra Sounds good to me. Including that, this would be the current schedule

| Release | Date | Note |
|--|--|--|
| 1.2.3 | September 23 | Final 3.4 release |
| 1.3.0 | October 28 | Final 2.7 series |
| 1.3.1 | November 25 | 1.3.0 + backported bug fixes only |
| 1.4.0 | December 16 | Python 3.5+, all new features post-1.3.0 |

oesteban:
Do you need a review on this?

effigies:
Not really, though you're welcome to. We'll do another sync after the final 1.3.0 release, and then we can merge in, so it might make more sense to wait on that.

effigies:
@satra Can you verify the latest [merge commit](https://github.com/nipy/nipype/commit/49cc0a75218b9f4aa7160560b3f24a1c8ea42c27)? If that's good, this should be good to merge.

satra:
looks good.

PR #2658, opened 2018-07-27
REL: 1.1.1

effigies:
Waiting on:

* [x] #2637


effigies:
@mgxd Should be ready for things like zenodo and mailmap updates.

djarecka:
@effigies - running the tests, will let you know

djarecka:
@effigies - looks like niype_tutorial is [fine](https://github.com/djarecka/nipype_tutorial/pull/16)

Also, I've just noticed that `xdist` have already a new release `1.22.5` :)  I don't remember which exactly travis job was giving you error on Friday, but it looks like `travis_retry` is not needed anymore.
https://github.com/djarecka/nipype/pull/5

effigies:
Sounds good. Since it's not anything that will end up in the distribution, we can fix that after release. Merging.

PR #2659, opened 2018-07-27
CI: Install pytest>=3.4 in Travis

effigies:
## Summary

pytest-xdist 1.22.4 bumped their requirement to pytest `>=3.4` (ours is `>=3.0`). It's unclear why this is failing to resolve. Creating this issue to see if there's a conflict.

See Travis errors in #2658.

Possibly related: #2649

## Acknowledgment


effigies:
That might be a better solution. Hopefully it's just a temporary resolution error and we don't actually have any conflicts that prevent us from using `>=3.4`, since the current pytest series is 3.6.x.

djarecka:
@effigies - I'll merge it to #2637, so the tests can run

djarecka:
or can we just merge to master? looks like everyone is happy with pytest>3.4

effigies:
Yeah, updating pip [wasn't sufficient](https://travis-ci.org/nipy/nipype/jobs/408987347) to get pip to correctly resolve the dependency. The only alternative I see is to add a `pip install --upgrade pytest>=3.4` to Travis.

@satra @oesteban Any thoughts on the best way to proceed? [3.4](https://github.com/pytest-dev/pytest/releases/tag/3.4.0) is from this year, so this is a fairly strong constraint for people who don't need the latest `pytest-xdist`.

djarecka:
@effigies - not sure why this is a problem. If they already have old `pytest-xdist` the older `pytest` will work. If they somehow got `pytest-xdist` that was released today morning, they should be ok with getting  `pytest` from Jan.

effigies:
Sure. But now *we're* requiring pytest 3.4. In general, it's good to be as flexible as possible to avoid putting users in a place where they have conflicting requirements. For instance, suppose somebody has a package that requires `pytest<=3.3`, and they have some `pytest-xdist` that is compatible with `pytest==3.3`. Now they want to install `nipype==1.1.1`, and suddenly they have an unresolvable dependency conflict. We don't actually need pytest 3.4, but the latest version of one of our dependencies causes tests to break on Travis, so we update it... It's not ideal.

djarecka:
@effigies - so one option would be to remove `xdist`, i.e. `pytest -n` from `nipype.test()` and keep it only for "test" installation, for people who want to test it regularly

effigies:
Eh. I decided to go the Travis hack route. It really is just a quirk of Travis' default python environment and a pip bug. Keeping a reasonable set of dependencies shouldn't affect typical users. If I'm wrong, and we need to update the dependency in the future, we can.

djarecka:
@effigies - but the standard pip installation of nipype will not lead to the error for `nipype.test()`?

effigies:
The case in which it does is if you have `pytest>=3.0,<3.4` already installed, but not `pytest-xdist`.

Sure there might be some setups like that, and we can walk people through some solution that fits their needs. Likely `pip install --upgrade pytest` is sufficient, but at least we won't be imposing an unresolvable dependency conflict.

djarecka:
ok, we can wait and see if anyone complains

PR #2669, opened 2018-08-02
FIX: Node __repr__ and detailed graph expansion

### LINKED ISSUE #2668: incorrect detailed graphs being generated
### Summary

the detailed graph is not listing names of nodes appropriately resulting in incorrect graphs.

scroll down to see the multiple arrows containing nodes in the figures of the following notebook.
https://miykael.github.io/nipype_tutorial/notebooks/introduction_quickstart.html

an example here:

image

i suspect this has something to do with how `__repr__` and `__str__` are being used. 

@oesteban, @effigies  - any quick ideas before i look into the code?

### Expected behavior

should have printed the old style with multiple nodes. 

see cell 14 in this notebook:
https://github.com/ReproNim/reproducible-imaging/blob/master/notebooks/introductory_dataflows.ipynb

run the current quickstart notebook in @miykael binder repo.

djarecka:
## Summary
fixing the detailed graph and printing `workflow.run().nodes() ` for workflows with iterables, the issue was due to the changes in the node `__str__`/`__repr__`

Fixes #2668 

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/base.py
+++ b/nipype/pipeline/engine/base.py
@@ -36,7 +36,7 @@ class EngineBase(object):
 
         """
         self._hierarchy = None
-        self._name = name
+        self.name = name
         self._id = self.name # for compatibility with node expansion using iterables
 
         self.base_dir = base_dir
--- a/nipype/pipeline/engine/base.py
+++ b/nipype/pipeline/engine/base.py
@@ -41,6 +41,7 @@ class EngineBase(object):
         self.base_dir = base_dir
         self.config = deepcopy(config._sections)
         self.name = name
+        self._id = self.name # for compatibility with node expansion using iterables
 
     @property
     def name(self):
@@ -65,6 +66,14 @@ class EngineBase(object):
     @property
     def outputs(self):
         raise NotImplementedError
+
+    @property
+    def itername(self):
+        """Name for expanded iterable"""
+        itername = self._id
+        if self._hierarchy:
+            itername = '%s.%s' % (self._hierarchy, self._id)
+        return itername
 
     def clone(self, name):
         """Clone an EngineBase object
--- a/nipype/pipeline/engine/base.py
+++ b/nipype/pipeline/engine/base.py
@@ -95,6 +95,9 @@ class EngineBase(object):
     def __str__(self):
         return self.fullname
 
+    def __repr__(self):
+        return self.itername
+
     def save(self, filename=None):
         if filename is None:
             filename = 'temp.pklz'
--- a/nipype/pipeline/engine/base.py
+++ b/nipype/pipeline/engine/base.py
@@ -36,12 +36,11 @@ class EngineBase(object):
 
         """
         self._hierarchy = None
-        self._name = None
+        self._name = name
+        self._id = self.name # for compatibility with node expansion using iterables
 
         self.base_dir = base_dir
         self.config = deepcopy(config._sections)
-        self.name = name
-        self._id = self.name # for compatibility with node expansion using iterables
 
     @property
     def name(self):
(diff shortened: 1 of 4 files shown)


djarecka:
@oesteban - yesterday with @satra we were fixing the graph plotting, but Travis is failing for `test_deep_nested_write_graph_runs`, since `Workflow` doesn't have `itername`. I've noticed that in #2376 you moved `itername`, `_id`,  from `EngineBase` to `Node`, so `Workflow` doesn't have it. Do you remember what was the motivation?

djarecka:
I'm trying to write tests for dot files, so reading the `write_graph` method and have problem with [`simple_form`](https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/workflows.py#L402). I can see that when this is `False` the `graph.dot` has longer names (e.g., "pipe.mod1.EngineTestInterface.engine" instead of "mod1 (engine)"), but graph_detailed has the same short names (so doesn't care about `simple_form`). Is this OK? Let me know if you think that this is a bug, otherwise I'll just keep it without any changes. 
(cc: @satra , @oesteban , @effigies , @mgxd )

effigies:
I think the detailed graph has never made a distinction wrt simple_form that I've been aware of, so it seems fine to leave it as is unless it was supposed to. I don't have strong opinions about the graphs.

effigies:
Is this ready for review?

djarecka:
yes

nipype/pipeline/engine/base.py
```diff
@@ -36,11 +36,11 @@ def __init__(self, name=None, base_dir=None):
 
         """
         self._hierarchy = None
-        self._name = None
+        self.name = name
```
oesteban: Doesn't this break the property definition below? (line 46)

djarecka: why? Before we had two lines: the first one  `self._name = None` and 3 lines later `self.name = name`. I thought that one was left there forgotten.

nipype/pipeline/engine/tests/test_engine.py
```diff
@@ -484,6 +485,154 @@ def test_deep_nested_write_graph_runs(tmpdir):
...
+    mod2 = pe.Node(interface=EngineTestInterface(), name='mod2')
+    pipe.connect([(mod1, mod2, [('output1', 'input1')])])
+    pipe.write_graph(
+        graph2use=graph_type, simple_form=simple, format='dot')
+
+    with open("graph.dot") as f:
+        graph_str = f.read()
+
+    if simple:
+        for str in eval("dotfile_{}".format(graph_type)):
```
effigies: Could we avoid `eval`s unless absolutely necessary? We could do this with a `dict`:

```Python
dotfiles = {
    'hierarchical': dotfile_hierarchical,
    ...
}


...

for contents in dotfiles[graph_type]:
    ...
```

Also, please do not use `str` as a variable name, as it is a built-in class. (I know we don't follow this rule in the entire code base, but I'd rather not introduce more cases of this.)

djarecka: sure, will remove `str`. Can you comment why you think I should not use `eval` here? Or in general you're against of using `eval` anywhere?

effigies: I'm against the use of `eval` at all, unless there is no other way to do what needs doing (and there is almost always another way).

`eval` makes it more difficult to reason about programs, including static analysis like `flake8` or `pylint`, because the code to be executed cannot in principle be known prior to execution. By design, it is a function that permits arbitrary code execution via string manipulation, which is also a security risk, even if it is not clear at the time how it could be exploited.

A decent StackOverflow thread: [Why is using 'eval' a bad practice?](https://stackoverflow.com/questions/1832940/why-is-using-eval-a-bad-practice)

djarecka: ok, will change

djarecka: thanks for comments

effigies:
@oesteban Any remaining concerns? Was your earlier concern addressed?

PR #2679, opened 2018-08-10
FIX: Read BIDS config.json under grabbids or layout

effigies:
## Summary
PyBIDS just renamed `bids.grabbids` to `bids.layout` (INCF/pybids#230). Althrough `grabbids` links back to `layout`, we look for files through the filesystem, which didn't survive the change.

This change should handle the new layout without breaking backwards compatibility.

@adelavega @tyarkoni I would appreciate reviews from you.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -42,9 +42,15 @@ from .base import (
 
 have_pybids = True
 try:
-    from bids import grabbids as gb
+    import bids
 except ImportError:
     have_pybids = False
+
+if have_pybids:
+    try:
+        from bids import layout as bidslayout
+    except ImportError:
+        from bids import grabbids as bidslayout
 
 try:
     import pyxnat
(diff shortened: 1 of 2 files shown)


adelavega:
Looks good to me but I'm not sure why the syslink doesn't work for you?

effigies:
What link?

effigies:
`bidslayout.__file__` works, but `bids.grabbids.__file__` no longer does.

adelavega:
Right, but why can't you just import `from bids import grabbids as bidslayout` and then check `bidslayout.__file__` Would that work after the update? If so that seems simpler.

Either way looks fine to me.

effigies:
> `from bids import grabbids as bidslayout` Would that work after the update?

No, because then `bidslayout.__file__` will be `bids.grabbids.__file__`, when what we need is `bids.layout.__file__`.

adelavega:
Ah I guess I was thinking you could do `realpath` which would resolve the symlink, but either way that adds complexity so the way you deal with it is fine.

effigies:
There is no symlink.

adelavega:
Ah okay. Ignore me then!

tyarkoni:
LGTM

PR #2682, opened 2018-08-13
CI: Test 3.7, resume testing nipy extras

effigies:

* Add Python 3.7-dev to test matrix
* Remove "fmri" extra removed in 3500213074f28def0f011ff59a1519fb2fd377b4

@satra Should `fmri` instead be replaced with `nipy` in Travis?

## Acknowledgment

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,8 +417,10 @@ workflows:
               only: /.*/
       - compare_base_dockerfiles:
           filters:
-            tags:
+            branches:
               ignore: /^(ci\/)?travis.*/
+            tags:
+              only: /.*/
       - test_pytest:
           filters:
             tags:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,34 +417,21 @@ workflows:
               only: /.*/
       - compare_base_dockerfiles:
           filters:
-            branches:
-              ignore: /^(ci\/)?travis.*/
             tags:
               only: /.*/
       - test_pytest:
           filters:
-            branches:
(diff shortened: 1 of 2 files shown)


djarecka:
@effigies - if you want you can also test py 3.7 from xenial, https://github.com/travis-ci/travis-ci/issues/9815

djarecka:
@effigies - this is the [syntax I used in pynidm](https://github.com/incf-nidash/PyNIDM/blob/master/.travis.yml). Probably can be done bettrer, but it works. Hopefully, this is only temporary, so either xenial or 3.7-dev works fine.

effigies:
That seems to only grab one entry from the env matrix...

effigies:
So it looks like only Xenial has 3.7, and only Trusty has 3.4. So not sure there's going to be a nice Travis matrix solution to get both without double-enumerating the environment.

effigies:
I think it probably makes the most sense to move to Xenial, and duplicate the matrix for 3.4, as that's going to be end-of-life soonest.

effigies:
Okay, this is set. @satra We haven't been testing interfaces that require `nipy` extras on Travis, so I haven't re-added it, but let me know if you think we should.

effigies:
(I just cancelled the Circle tests, since this is just futzing with Travis.)

effigies:
I'm interpreting that as `INSTALL_DEB_DEPENDENCIES=true`.

satra:
i think it's reasonable to ask, i don't know how much work it would be for them. @arokem - is there a practical obstacle to dipy wheels for 3.7 (now that it is available)?

Issue #2690, opened 2018-09-03
Incorrect exception handling in utils/spm_docs.py ?


kchawla-pi:
In `nipype/utils/spm_docs.py`, the function `_strip_header()` has the code:
```
    try:
        index = doc.index(hdr)
        ...
    except KeyError as e:
        raise_from(IOError('This docstring was not generated by Nipype!\n'), e)
```
However, `doc` doesn't seem to be a dict type so `KeyError` wouldn't apply here.
If `hdr` is not found in `doc`, that would raise a `ValueError`, which is not caught by the code here.

This was the error I got 
```
    index = doc.index(hdr)
ValueError: substring not found

```
Ubuntu 18

djarecka:
@kchawla-pi - yes, `doc` doesn't seem to be a dict, rather string. Thank you for reporting. Would you like to create a PR to fix it.

kchawla-pi:
Yup, on it.

effigies:
Fixed in #2691.

PR #2691, opened 2018-09-03
Fixed: Missing substring should raise ValueError not KeyError + try-except code cleanup

kchawla-pi:
Fix for issue #2690


djarecka:
@kchawla-pi - thank you for the PR!

nipype/utils/spm_docs.py
```diff
@@ -59,5 +59,5 @@ def _strip_header(doc):
         except ValueError:
             index = len(doc)
         return doc[:index]
-    except KeyError as e:
+    except ValueError as e:
```
effigies: I think this whole block might be better rewritten as

```Python
    try:
        index = doc.index(hdr)
    except ValueError as e:
        raise_from(IOError('This docstring was not generated by Nipype!\n'), e)
    index += len(hdr)
    index += 1
    doc = doc[index:]
    try:
        index = doc.index(cruft)
    except ValueError:
        index = len(doc)
    return doc[:index]
```

This helps make clear where the exception is expected, and avoids catching unexpected exceptions.

effigies:
Hi @kchawla-pi. Do you have a few minutes to finish this one off? I think we're almost ready to go.

We'll be releasing 1.1.3 next week, and this would be a great addition!

kchawla-pi:
Ok.

PR #2692, opened 2018-09-04
FIX: Provide informative error for interfaces that fail to return runtime object

effigies:
## Summary

If an interface has a `_run_interface` implementation that in some (or all) cases fails to return the passed `runtime` object, then the error you currently get is:

```
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 471, in run
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 555, in _run_interface
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_command
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base/core.py", line 544, in run
    runtime.endTime = dt.isoformat(dt.utcnow())
AttributeError: 'NoneType' object has no attribute 'endTime'
```

Arising from this chunk of code:

https://github.com/nipy/nipype/blob/37f3781a4f765a7793d17389cd959d2eb4fcd617/nipype/interfaces/base/core.py#L542-L553

This PR simply makes an absent `runtime` object raise a `RuntimeError` and give a likely cause.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -540,6 +540,9 @@ class BaseInterface(Interface):
             if not ignore_exception:
                 raise
         finally:
+            if runtime is None:
+                raise RuntimeError("{} interface failed to return runtime "
+                                   "object".format(interface.__class__.__name__))
             # This needs to be done always
             runtime.endTime = dt.isoformat(dt.utcnow())
             timediff = parseutc(runtime.endTime) - parseutc(runtime.startTime)


effigies:
Any objections/suggestions here? I'll merge tomorrow AM if nobody minds.

satra:
while we are at this, should we simply check for None, or actually check that the reserved keywords still exist.

effigies:
Sorry, could you be more explicit?

satra:
i.e. that these attributes are still present:

https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/core.py#L492

effigies:
The easiest way to do that without digging into private variables was to add a `Bunch.keys()` method. That alright?

satra:
@effigies - that's fine with me, but for the moment could we keep that api private. ideally we will move away from Bunch to more standard Python objects in Py 3.x.

`dict(bunch_var.items())` will return a dict that can be used as well.

so `set(dict(bunch_var.items())) == set(dict(orig_runtime.items()))` could also work as a check.

where `orig_runtime` is the pre-call runtime.

effigies:
Switched to dictcopy, which already exists. We use it elsewhere in the file, so a replace-all when moving to a standard object will catch these additions.

PR #2720, opened 2018-10-09
[FIX] Lock travis pybids 0.6.5

adelavega:
This is just a patch to attempt to pin pybids to 0.6.5, until we updates the interfaces that depend on pybids.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,8 +87,7 @@ before_install:
 
 - travis_retry pip install -r requirements.txt
 - travis_retry pip install grabbit==0.1.2
-- travis_retry git clone https://github.com/INCF/pybids.git ${HOME}/pybids &&
-  pip install -e ${HOME}/pybids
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -87,8 +87,7 @@ before_install:
 
 - travis_retry pip install -r requirements.txt
 - travis_retry pip install grabbit==0.1.2
-- travis_retry git clone https://github.com/INCF/pybids.git ${HOME}/pybids &&
-  pip install -e ${HOME}/pybids
+- travis_retry git clone -b 0.6.5 https://github.com/INCF/pybids.git ${HOME}/pybids && pip install -e ${HOME}/pybids
```
effigies: Shorter:

```Python
travis_retry pip install -e git+https://github.com/INCF/pybids.git@0.6.5#egg=pybids
```

oesteban:
Shouldn't we also pin the dependency [here](https://github.com/nipy/nipype/blob/15a8d6f57067494196fe639095253217a9235c3c/nipype/info.py#L167)?

effigies:
Merging despite all the scary red x's.

cc: @miykael.

Issue #2713, opened 2018-10-07
ensure docs build properly on readthedocs


satra:
nipype on readthedocs has not been building properly, since release 1.1.1:

https://readthedocs.org/projects/nipype/builds/7835987/

we should try to ensure that our own doc build returns appropriate errors on circle if they fail, which i thought it did. we should resolve what it is between our tests and readthedocs is altering the failure mode.

according to the error above it seems something about matplotlib sphinx extension

satra:
this extension exists at least in `matplotlib` release `2.2.2`, which is available inside the `nipype/nipype:py36` docker container.

so this is likely some setting in/for readthedocs that's causing issues.

miykael:
I tried to build the docs locally and got an error concerning `matplotlib.sphinxext.only_directives`. The easy solution was to just remove/comment this dependency in the `conf.py` file ([here](https://github.com/nipy/nipype/blob/master/doc/conf.py#L53)). But I'm not sure if it will lead to other errors down the line (or for what this extension is actually needed).

PR #2707, opened 2018-09-19
ENH: use neurodocker 0.4.1 + apt install afni

kaczmarj:
## Summary

This PR proposes updates to Dockerfile generation.

References https://github.com/nipy/nipype/pull/2706#issuecomment-422875107

## List of changes proposed in this PR (pull-request)
- upgrade to neurodocker 0.4.1
- install afni from neurodebian
- minor refactoring in `neurodocker generate` commands.

## Acknowledgment

- [x] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

PR #2723, opened 2018-10-09
FIX: corrects readthedocs build error

miykael:
Closes https://github.com/nipy/nipype/issues/2713.

Small PR that takes care of the readthedocs build error mentioned under https://github.com/nipy/nipype/issues/2713.

The problem was that `matplotlib` version 3.0.0 doesn't include the `sphinxext.only_directives` anymore.

### DIFF
--- a/docker/generate_dockerfiles.sh
+++ b/docker/generate_dockerfiles.sh
@@ -53,11 +53,11 @@ do
 done
 
 
-# neurodocker version 0.3.1-22-gb0ee069
-NEURODOCKER_IMAGE="kaczmarj/neurodocker@sha256:f15ca90803f4b89acfca55cd1c7269bf2ec2dfd95b3de1118b08afa34b87d9d1"
+# neurodocker:master pulled on October 9, 2018
+NEURODOCKER_IMAGE="kaczmarj/neurodocker@sha256:c5c10fc11cbd7efe18ac10d2370742b8ecb3588c3aa9b2c15ca7e18363bfa9dc"
 
-# neurodebian:stretch-non-free pulled on November 3, 2017
-BASE_IMAGE="neurodebian@sha256:7590552afd0e7a481a33314724ae27f76ccedd05ffd7ac06ec38638872427b9b"
+# neurodebian:stretch-non-free pulled on October 9, 2018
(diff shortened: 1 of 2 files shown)


miykael:
All the errors on circleci seem to be due to the spm version issue mentioned under https://github.com/kaczmarj/neurodocker/pull/227.

This PR updates pybids version to 0.6.5 with https://github.com/nipy/nipype/pull/2723/commits/46cbed655a4dcea4109a06b9e73845742a1a8c89, but I wasn't able to update neurodocker to 0.4.1 [here](https://github.com/nipy/nipype/blob/master/docker/generate_dockerfiles.sh#L57)? How do I get the new/corresponding `sha256`?

Also, can somebody help me to understand why the tests on travis were failing?

effigies:
The Travis tests are failing because of a coverage change in the pre-release. I think @adelavega can point to the issue, if needed.

effigies:
In any event, those pre-release checks shouldn't block merging, if that's the only thing holding you back right now. I think the solution is just to wait for a new coverage RC and Travis will fix itself.

miykael:
The main issue that I've tried to solve in this PR as well (I perhaps should rather create a new PR to do so), is to update `neurodocker` to 0.4.0. But this opens up questions, about `FreeSurfer` installation and license file, etc.

docker/generate_dockerfiles.sh
```diff
@@ -53,40 +53,40 @@ do
...
 function generate_main_dockerfile() {
-  docker run --rm "$NEURODOCKER_IMAGE" generate \
+  docker run --rm "$NEURODOCKER_IMAGE" generate docker \
   --base "$NIPYPE_BASE_IMAGE" --pkg-manager "$PKG_MANAGER" \
   --label maintainer="The nipype developers https://github.com/nipy/nipype" \
   --env MKL_NUM_THREADS=1 OMP_NUM_THREADS=1 \
   --user neuro \
-  --miniconda env_name=neuro \
+  --miniconda create_env=neuro \
+              conda_install='python=${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}' \
```
effigies: Looks like this is being used prior to the definition of `PYTHON_VERSION_*` (L102).

miykael:
I will revert all the commits concerning `docker/generate_dockerfiles.sh` as they will be taken care of with https://github.com/nipy/nipype/pull/2707.

PR #2724, opened 2018-10-10
[WIP] enh: turn interface into callable

satra:
turning interfaces into a callable to support dask based operations.

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -587,6 +587,11 @@ class BaseInterface(Interface):
 
         return results
 
+    def __call__(self, **kwargs):
+        """Make interface into a callable function"""
+        results = self.run(**kwargs)
+        return results.outputs
+
     def _list_outputs(self):
         """ List the expected outputs
         """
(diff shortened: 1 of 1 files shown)


oesteban:
I think this should work in nipype 1.0 without much trouble. Therefore, I see under 1.0 anything that makes the jump between 1.0 and 2.0 easier but does not mean to actually implement 2.0.

effigies:
Anything left to this one?

nipype/interfaces/base/core.py
```diff
@@ -587,6 +587,11 @@ def run(self, cwd=None, ignore_exception=None, **inputs):
 
         return results
 
+    def __call__(self, **kwargs):
+        """Make interface into a callable function"""
+        results = self.run(**kwargs)
+        return results.outputs
```
satra: what about this to match with the call to `run`?

```suggestion
        return results
```

effigies: I have zero expectations of behavior here, so sure?

oesteban: Dask might expect something like `results.outputs`.

effigies:

effigies:

Issue #2719, opened 2018-10-09
pytest never exits due to "config.enable_resource_monitor()" in test_resource_monitor.py


yarikoptic:
while building 1.1.3 (doesn't happen with 1.1.2) debian package, and showed to @effigies and @mgxd , pytest process never exits with some processes started within test_CommandLine_prefix still running in the background, not killed.
With some manual bisection it seems boiled down to `config.enable_resource_monitor()` call within `test_resource_monitor.py`.  Even if all tests are skipped in that file, that call is ran while pytest is just collecting the tests, so it is "in effect".

I wonder if that relates to 87ef62f1cd4d1de2cd554a6fae51de6de12c64ec (and may be some related changes)?

effigies:
cc @oesteban

yarikoptic:
probably there should be `disable_resource_monitor` as well, and enable/disable should happen in tests fixture, not at the test module import level

effigies:
:+1:

yarikoptic:
I will submit a PR shortly.... seems to "work" (those tests are sskipped atm anyways)

effigies:
Closed by #2725.

PR #2725, opened 2018-10-10
ENH: enable/disable resource monitor in the fixture per test

yarikoptic:
This is just a more "kosher" way to avoid side-effects across tests etc.
It was "inspired" to mitigate an issue of some stale processes being
kept in background preventing pytest run completion, see
https://github.com/nipy/nipype/issues/2719

So, even though it avoids that issue now (both tests are skipped anyways
unconditionally ATM), it does not solve it -- some issue causing that
was introduced somewhere between 1.1.2 and 1.1.3

### DIFF
--- a/nipype/interfaces/base/tests/test_resource_monitor.py
+++ b/nipype/interfaces/base/tests/test_resource_monitor.py
@@ -18,8 +18,14 @@ from ...base import traits, CommandLine, CommandLineInputSpec
 from ... import utility as niu
 
 # Try to enable the resource monitor
-config.enable_resource_monitor()
 run_profile = config.resource_monitor
+
+
+@pytest.fixture(scope="module")
+def use_resource_monitor():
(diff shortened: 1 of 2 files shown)


yarikoptic:
not sure what is up with those circleci fails - seems unrelated, waiting on travis.  Locally debian builds seems finally started to pass that hiccup

effigies:
Waiting on #2707.

effigies:
@yarikoptic Can you merge master? That may fix tests.

yarikoptic:
done

effigies:
Hoo boy, it sure didn't.

yarikoptic:
well -- travis is green
What confuses me also that there is no codecov report and chromium plugin says that the test file is not covered.  But otherwise it is "ready" afaik -- shouldn't make things worse and allows me to "pass" the tests

Issue #2571, opened 2018-05-09
Resource profiler not working?


atsuch:
### Summary
I have followed the instruction here (http://nipype.readthedocs.io/en/latest/users/resource_sched_profiler.html) to monitor the resource of the each node, but I do not seem to be getting the profiles I need, namely the runtime_memory_gb and runtime_num_threads. I get "N/A" for them in the run_stats.log, and these attributes are absent from results.pklz of each node as well as from the report.rst. 

Also, this is not directly related to this problem, but somehow my terminal outputs are written vertically in the report.rst, as can be seen in the attached file...

### Actual behavior
I attached an example report.rst from one of my nodes, as well as run_stats.log
[report.txt](https://github.com/nipy/nipype/files/1988437/report.txt)
[run_stats.log](https://github.com/nipy/nipype/files/1988442/run_stats.log)

### Expected behavior
I have psutil in my conda env, and it could be imported without any problem, so I expected to get runtime info.

### How to replicate the behavior
It might be specific version of things I have in my conda env, but I have not figured out which package is the culprit. 

### Platform details:

I'm using nipype 1.0.2 on python3.6 conda env with the default pickle, and psutil 5.4.3.

### Execution environment

Choose one
- Container [Tag: ???]
- My python environment inside container [Base Tag: ???]
- My python environment outside container

oesteban:
Hi @atsuch, yes this is an open problem. We had to rewrite the resource profiler and I am guessing that the final conversion from the new profiler to the old one is not working. Also the documentation is very outdated.

I'll try to address this problem ASAP.

Thanks for reporting. I might contact you for follow up information and to confirm that fixes work for you.

atsuch:
Thank you @oesteban for letting me know...!

I read a few issue threads on the new resource profiler but was not sure if this was an open issue. Meanwhile, could I perhaps try my pipeline in an older version of nipype to get the resource profile? Do you know from which version this is an issue?

oesteban:
I would rather prefer that you tried the new resource profiling framework and work out your own plotting. I've been meaning to work on this plotting myself for long. If you are up to trying, we can make it a concerted effort where I'd support you.

To enable the resource profiling, please check http://nipype.readthedocs.io/en/latest/users/config_file.html#resource-monitor

Then, it will generate a log file with all memory and cpu traces, so that you can build up the graphs you'd like on top.

The project here would be some D3 code to render these usage plots. WDYT?

oesteban:
[...]

PR #2726, opened 2018-10-10
DOC: Move users docs from nipype to nipype_tutorial

miykael:
## Summary

PR https://github.com/miykael/nipype_tutorial/pull/77 and https://github.com/miykael/nipype_tutorial/pull/128 moved the content under [User Guide](https://nipype.readthedocs.io/en/1.1.2/users/index.html) already to the `nipype_tutorial`, but we haven't removed this content from `nipype`, yet. This PR takes care of that.

## Deviation between the content

There are three sections that have been copied to the tutorial but either leads to errors or are still WIP:

#### [Saving Workflows and Nodes to a file (experimental)](https://miykael.github.io/nipype_tutorial/notebooks/wip_saving_workflows.html)
Running `wf.export('special_workflow.py')` in the notebook will lead to a `TypeError: a bytes-like object is required, not 'str'`. As the content of this section is still "experimental", I think it's ok if we leave it as WIP.

#### [Running Nipype Interfaces from the command line (nipype_cmd)](https://miykael.github.io/nipype_tutorial/notebooks/wip_nipype_cmd.html)
The content between `nipype` and `nipype_tutorial` is identical. But currently, we're not running any tests, as the command `nipype_cmd` doesn't work inside the docker image. I haven't used `nipype_cmd` yet myself, but it seems `nipype` needs to be installed via apt install python-nipype for this to work?

#### [Resource Scheduling and Profiling with Nipype](https://miykael.github.io/nipype_tutorial/notebooks/wip_resource_sched_profiler.html)
The content between `nipype` and `nipype_tutorial` is identical, but running the profiler within the docker image leads to the error `AttributeError: 'list' object has no attribute 'startTime'`. This is caused by the fact that the profiling information is not handed over as an object but as an object inside a list.

## To consider

Removing the "User Docs" from `nipype` leads to new issues that need to be considered:

1. `install.rst` wasn't removed from `doc/users` as it is also needed in `nipype`'s [Quickstart](https://nipype.readthedocs.io/en/1.1.2/quickstart.html) section.
2. `doc/users` contains a folder `examples` that contains the workflow examples under [Interfaces, Workflows and Examples](https://nipype.readthedocs.io/en/1.1.2/documentation.html). Once those workflows are included under `niflows`, this folder can be removed.
3. This PR inverts the order between the "Developer Guides" and the "Interfaces, Workflows and Examples" section, to this: 
new_documentation_layout
4. This PR puts the `API` and `Developer Guide` into two columns (see below). But the current setup has titles, as well as links to the corresponding sections. Should we take care of this?
developer_title


doc/users/install.rst
```diff
@@ -16,7 +16,8 @@ image from Docker hub::
     docker pull nipype/nipype
 
 You may also build custom docker containers with specific versions of software
-using Neurodocker_ (see the :doc:`neurodocker`).
+using Neurodocker_ (see the `Neurodocker tutorial
+<https://miykael.github.io/nipype_tutorial/notebooks/introduction_neurodocker.html>`_).
```
effigies: Can we move this to `nipype.github.io/tutorial`? Not sure if we already had discussion there...

miykael: You're talking about the whole `nipype_tutorial`? Fine with me, but we haven't discussed it yet, or decided what to do.

effigies: Yeah.

effigies:
This seems fine to me. Are there any outstanding issues? Can you merge master to get the latest version of testing in?

One quick note: `nipype_cmd` is now `nipypecli`. Not sure if that makes a difference for your notebook...

effigies:
Actually, would you mind pushing a build of the docs to your `gh-pages`, so we can check them out on https://miykael.github.io/nipype?

miykael:
I merged with master and renamed `nipype_cmd` to `nipypecli`. The newest version of the docs can be found on my `gh-pages` (https://miykael.github.io/nipype).

### 3 outstanding issues

#### Command line interface
If `nipype_cmd` is the same as `nipypecli`, does this mean that [this](https://nipype.readthedocs.io/en/latest/users/nipypecmd.html) content can be deleted?

#### Saving Workflows and Nodes to a file (experimental)
As mentioned above, [this content](https://nipype.readthedocs.io/en/latest/users/saving_workflows.html) is only in a [WIP notebook](https://github.com/miykael/nipype_tutorial/blob/master/notebooks/wip_saving_workflows.ipynb) and therefore not visible on the tutorial homepage. The problem is that `wf.export('special_workflow.py')` currently leads to an error in the notebook. Should we leave this as a WIP or remove the content?

#### Resource Scheduling and Profiling with Nipype
The [resource profiler content](https://nipype.readthedocs.io/en/latest/users/resource_sched_profiler.html) is only as a [WIP in the tutorial](https://github.com/miykael/nipype_tutorial/blob/master/notebooks/wip_resource_sched_profiler.ipynb) and doesn't work in the notebook, yet. I think this is addressed in https://github.com/nipy/nipype/issues/2571, but the proposed solution doesn't work in the container.

### Mishap
Also, I unfortunately first updated the `gh-pages` from `nipy/nipype`. I didn't had a gh-pages branch yet, and so `git checkout gh-pages` put me directly on the `upstream` one... I'm really sorry for that and am happy to correct it, if necessary (`gh-pages` doesn't seem to be used anymore?). How can I revert my changes without bringing github down (again?)?

effigies:
I reverted `gh-pages` to Satra's latest commit. Fortunately you didn't destroy the history.

Yes, `nipype_cmd` no longer exists. I don't know whether that information can be (usefully) updated for `nipypecli` or if it just needs deleting. I'm content to leave that decision to you.

I would leave the WIP notebooks as WIP at least until 2.0. Perhaps someone will fix the issues? I would go ahead and open issues for those notebooks here, as the fixes probably need to be done here.

Happy to be overruled by @satra on any and all points.

effigies:
@miykael Are you happy with this at least for this release? We can open issues for any remaining things to fix in 1.1.5.

miykael:
@effigies, yes. In my opinion it's good to go. Will keep track about any issues and open missing issues for the WIP on the tutrial homepage.

PR #2730, opened 2018-10-10
BF: provide "" into target graph dot depending on networkx version

yarikoptic:
Not sure if that started to happen with networkx `2` but comparing between networkx 1.11 and 2.1 this seems to work


yarikoptic:
travis is green, circle ci is known to misbehave (as in #2725)

PR #2731, opened 2018-10-10
FIX: Remove 'reg_term'/'regularisation' from dwi2tensor interface

ihnorton:
## Summary

- remove `regularisation` (`reg_term`) option from MRtrix3 dwi2tensor interface.

This option was removed as part of refactoring in:

  https://github.com/MRtrix3/mrtrix3/commit/9b886a94289a

Fixes

```
dwi2tensor: [ERROR] unknown option "-regularisation"
```

## Acknowledgment

### DIFF
--- a/nipype/interfaces/mrtrix3/base.py
+++ b/nipype/interfaces/mrtrix3/base.py
@@ -4,9 +4,37 @@
 from __future__ import (print_function, division, unicode_literals,
                         absolute_import)
 
-from ... import logging
-from ..base import (CommandLineInputSpec, CommandLine, traits, File, isdefined)
+from ... import logging, LooseVersion
+from ...utils.filemanip import which
+from ..base import (CommandLineInputSpec, CommandLine, traits, File, isdefined, PackageInfo)
 iflogger = logging.getLogger('nipype.interface')
+
+
+class Info(PackageInfo):
+    version_cmd = 'mrconvert --version'
+
+    @staticmethod
+    def parse_version(raw_info):
+        # info is like: "== mrconvert 0.3.15-githash"
+        for line in raw_info.splitlines():
+            if line.startswith('== mrconvert '):
+                v_string = line.split()[2]
+                break
+        else:
+            return None
+
+        # -githash may or may not be appended
+        v_string = v_string.split('-')[0]
+
+        return '.'.join(v_string.split('.')[:3])
+
+    @classmethod
+    def looseversion(cls):
+        """ Return a comparable version object
+
+        If no version found, use LooseVersion('0.0.0')
+        """
+        return LooseVersion(cls.version() or '0.0.0')
 
 
 class MRTrix3BaseInputSpec(CommandLineInputSpec):
@@ -78,3 +106,7 @@ class MRTrix3Base(CommandLine):
             pass
 
         return super(MRTrix3Base, self)._parse_inputs(skip=skip)
+
+    @property
+    def version(self):
+        return Info.version()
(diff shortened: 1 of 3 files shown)


effigies:
Since this is still valid for older versions of mrtrix3, we shouldn't remove it altogether. You can make the input spec depend on the version of the library. See, for example, `BBRegister`:

https://github.com/nipy/nipype/blob/ffd8cf8dc2e89dcec5b3a706f5d466ab192896c4/nipype/interfaces/freesurfer/preprocess.py#L1336-L1428

https://github.com/nipy/nipype/blob/ffd8cf8dc2e89dcec5b3a706f5d466ab192896c4/nipype/interfaces/freesurfer/preprocess.py#L1462-L1465

In this case, you'd probably want to go the other way around, with the newer version as the superclass and the older as the subclass.

satra:
doesn't `max_ver` metadata on the trait take care of this? or does `mrtrix` not provide versioning info?

effigies:
Ah, I'd forgotten about that one. My approach wouldn't work without version info, either.

ihnorton:
I added version info to the interface (changes pushed), which seems to work.
But `max_ver` doesn't do what the docs seem to imply -- I get an error:

```
    (name, self.__class__.__name__, version, max_ver))
Exception: Trait reg_term (FitTensor) (version 0.3.15 > required 0.3.13)
```

ihnorton:
Removing `usedefault` fixes the problem. I take it that needs to be mutually-exclusive with `max_ver`...

satra:
indeed. essentially its saying that if the environment conflicts with the use, raise an exception.

we should think about cleaner ways to handle versioning in nipype 2.0.

nipype/interfaces/mrtrix3/base.py
```diff
@@ -4,11 +4,45 @@
...
+from ... import logging, LooseVersion
+from ...utils.filemanip import which
+from ..base import (CommandLineInputSpec, CommandLine, traits, File, isdefined, PackageInfo)
 iflogger = logging.getLogger('nipype.interface')
 
 
+class Info(PackageInfo):
+    version_cmd = which('mrconvert')
+    if version_cmd:
+        version_cmd = version_cmd + ' --version'
```
effigies: If `mrconvert` is in your `PATH`, then you can simplify this as just:

```Python
version_cmd = 'mrconvert --version'
```

nipype/interfaces/mrtrix3/base.py
```diff
@@ -4,11 +4,45 @@
...
+    def looseversion(cls):
+        """ Return a comparable version object
+
+        If no version found, use LooseVersion('0.0.0')
+        """
+        ver = cls.version()
+        if ver is None:
+            return LooseVersion('0.0.0')
+
+        return cls.version())
```
effigies: I assume you want:

```Python
return LooseVersion(ver)
```

Or shorter:

```Python
@classmethod
def looseversion(cls):
    return LooseVersion(cls.version() or '0.0.0')
```

nipype/interfaces/mrtrix3/reconst.py
```diff
@@ -6,8 +6,9 @@
 
 import os.path as op
 
+from ... import LooseVersion
 from ..base import traits, TraitedSpec, File, Undefined
-from .base import MRTrix3BaseInputSpec, MRTrix3Base
+from .base import MRTrix3BaseInputSpec, MRTrix3Base, Info
```
effigies: No longer using the `Info` and `LooseVersion` imports.

nipype/interfaces/mrtrix3/reconst.py
```diff
@@ -39,8 +40,9 @@ class FitTensorInputSpec(MRTrix3BaseInputSpec):
         argstr='-method %s',
         desc=('select method used to perform the fitting'))
     reg_term = traits.Float(
-        5.e3, usedefault=True,
+        5.e3,
```
effigies: Could you go ahead and remove the `5.e3`? The reason `usedefault=True` was added was because it looked like a default was intended by the original interface author. If we're not going to apply defaults, we should remove it.

effigies:
Also please merge or rebase against master to fix tests.

PR #2735, opened 2018-10-12
PR: ENH add tab completion for node and interface inputs properties

leej3:
## Summary
Improves ease of use by giving more convenient tab completion for constructing graphs.

Fixes # .

- adds tab completion to the inputs property of both node and interface. This behavior is currently only observed in Ipython and if the user sets  "c.IPCompleter.limit_to__all__ = True" in their ipython config or executes the magic command "%config IPCompleter.limit_to__all__ = True"

## Acknowledgment

- I acknowledge that this contribution will be available under the Apache 2 license.


effigies:
Could you add a test for `TraitedSpec`s (e.g. `CommandLine` interface) and `DynamicTraitedSpec`s (e.g. `Function` interface)?

I'm guessing the reason for including this in `BaseInterface`, as opposed to `BaseTraitedSpec`, is that the spec should already be fleshed out by the time this is called. But I think `Function` at least calls `super` before adding inputs. Instead, I think maybe the better way to do this will be to update `self.__all__` in `BaseTraitedSpec` on `__init__` and when new traits are added.

But if the test passes for both static and dynamic specs, I'm okay with this, as it's pretty simple.

effigies:
Also, is this targeting 1.x or 2.0?

leej3:
Thanks. You pointers helped me a lot. It does indeed work better by modifying BaseTraitedSpec. I've added tests for my desired behavior.

I'm still struggling to implement the functionality for DynamicTraitedSpec. I'm not sure where to modify it. Do you have any thoughts?

effigies:
Perhaps override `trait_set` to update `__all__` after running `super().trait_set()`?

leej3:
This is almost working. I would expect the following test to pass but the second assert doesn't:

```
    def test_DynamicTraitedSpec_tab_completion():
        extract_func = lambda list_out: list_out[0]

        # Define node
        list_extract = Node(Function(
            input_names=["list_out"],output_names=["out_file"],
            function=extract_func), name="list_extract")

        # Check __all__ for node outputs
        expected_output = sorted(list_extract.outputs.editable_traits())
        assert(sorted(list_extract.outputs.__all__) == expected_output)

        # Add trait and retest
        list_extract.outputs.add_trait("added_out_trait","val")
        expected_output = sorted(['added_out_trait',*expected_output])
       assert(sorted(list_extract.outputs.__all__) == expected_output)
```
This doesn't pass on the commits prior to mine though.

leej3:
> Also, is this targeting 1.x or 2.0?

1.x. 2.0 will have to incorporate it following a pending pyhydra refactor.

effigies:
I think the problem may be that `Function` can't take a lambda. It requires a normal function.

leej3:
good. to know. i'll change that. It doesn't fix this problem though

nipype/interfaces/base/tests/test_specs.py
```diff
@@ -63,6 +81,36 @@ def test_TraitedSpec_dynamic():
...
+
+    # Check __all__ for node inputs
+    assert(sorted(list_extract.inputs.__all__) == expected_input)
+
+    # Check __all__ for node outputs
+    expected_output = sorted(list_extract.outputs.editable_traits())
+    assert(sorted(list_extract.outputs.__all__) == expected_output)
+
+    # Add trait and retest
+    list_extract.outputs.add_trait("added_out_trait","val")
```
effigies: The problem is that the output spec is actually instantiated in `IOBase._outputs()`, which calls `Function._add_output_traits`, which only will add things in `self._output_names`. So to do what you want here, you should do:

`list_extract._interface._output_names.append('added_out_trait')`

nipype/interfaces/base/specs.py
```diff
@@ -83,9 +83,11 @@ def __repr__(self):
         """ Return a well-formatted representation of the traits """
         outstr = []
         for name, value in sorted(self.trait_get().items()):
-            outstr.append('%s = %s' % (name, value))
+            if not name == '__all__':
+                outstr.append('%s = %s' % (name, value))
```
effigies: Is this still necessary, with `__all__` becoming a property?

nipype/interfaces/base/specs.py
```diff
@@ -83,9 +83,11 @@ def __repr__(self):
...
         return '\n{}\n'.format('\n'.join(outstr))
 
+
```
effigies: We should remove this line.

nipype/interfaces/base/specs.py
```diff
@@ -309,6 +311,11 @@ def _get_sorteddict(self,
                     out = objekt
         return out
 
+    @property
+    def __all__(self):
+        return [k for k,v in self.items() if not k == "__all__"]
+
+
```
effigies: Too many blank lines.

nipype/interfaces/base/specs.py
```diff
@@ -309,6 +311,11 @@ def _get_sorteddict(self,
                     out = objekt
         return out
 
+    @property
+    def __all__(self):
+        return [k for k,v in self.items() if not k == "__all__"]
```
effigies: Given that `items()` provides a dict keyed by `self.copyable_trait_names()`, and you only need keys, I think you can just make this:

```Python
@property
def __all__(self):
    return self.copyable_trait_names()
```

You'll need to change the BET tests to either sort or check set equality, but this should work.

leej3:
Thanks. I made all those changes

nipype/interfaces/base/tests/test_specs.py
```diff
@@ -63,6 +81,36 @@ def test_TraitedSpec_dynamic():
...
+    # Check __all__ for node inputs
+    assert(sorted(list_extract.inputs.__all__) == expected_input)
+
+    # Check __all__ for node outputs
+    expected_output = sorted(list_extract.outputs.editable_traits())
+    assert(sorted(list_extract.outputs.__all__) == expected_output)
+
+    # Add trait and retest
+    list_extract._interface._output_names.append('added_out_trait')
+    expected_output = sorted(['added_out_trait',*expected_output])
```
effigies: Python 2.7, 3.4 don't like the `[*x]` syntax. How about making `expected_output` a set above, moving to set comparisons, and changing this to:

```Python
expected_output.add('added_out_trait')
```

leej3: nice. Thanks for the tip. done

PR #2737, opened 2018-10-15
[ENH] Update BIDSDataGrabber for pybids 0.7

adelavega:
This PR updates `BIDSDataGrabber` for pybids 0.7.

- Updated defunct bids entities (e.g. modality)
- Changed default query to be more specific (`T1w` vs `anat`; `bold` vs `func`)
- Added derivatives handling (two inputs traits, boolean `derivatives` and option to add other directories)
- Updated tests
- Minor: Removed `outfields` from examples, as they are dynamically generated. 

Need to wait for pybids to officially release 0.7 (and pin version), before merging.


nipype/interfaces/io.py
```diff
@@ -2810,8 +2812,10 @@ def __init__(self, infields=None, **kwargs):
 
         if not isdefined(self.inputs.output_query):
             self.inputs.output_query = {
-                "func": {"modality": "func", 'extensions': ['nii', '.nii.gz']},
-                "anat": {"modality": "anat", 'extensions': ['nii', '.nii.gz']},
+                "bold": {"datatype": "func", "suffix": "bold",
+                         "extensions": ["nii", ".nii.gz"]},
+                "T1w": {"datatype": "anat", "suffix": "T1w",
+                         "extensions": ["nii", ".nii.gz"]},
```
effigies: Not a problem for this PR, but we may want to think about making it a much more generally usable default.

adelavega: Sure. Such as? you mean returning more things?

effigies: Yeah. Ideally something that will contain most, if not all, files, and can just be queried sensibly. I feel like specifying the query is more of a power-user move, so making sure there's a reasonable way to ask for the files most people will want would be good.

adelavega:
Hopefully the tests will pass now. Otherwise, any other reviews? Either way, waiting for `pybids 0.7` but aside from that I'm done.

nipype/interfaces/io.py
adelavega: That sounds good to me, although I worry about about performance if a bunch of unnecessary queries are made. Although the indexing itself is probably the slowest part, so it should probably be fine.

adelavega: While we're at it, other fields you'd want? `brain_mask` is another one I can think of for fmri. I would like to hear from ppl doing preprocessing and and DWI as I'm unfamiliar with what is typically expected.

effigies:
Put on 1.1.8 milestone, as I expect pybids 0.7 will be out by then. Also, we should coordinate with @jdkent and #2815 on a mutually satisfactory version of this interface.

adelavega:
Waiting for pybids 0.7 release. Need to re-test and update the pin when that happens.

adelavega:
Looks like tests are passing, but building docs have failed. Not my fault, promise!

effigies:
Please merge master.

@jdkent Can you review this in light of your thoughts in #2815?

.travis.yml
```diff
@@ -86,8 +86,8 @@ before_install:
   fi;
 
 - travis_retry pip install -r requirements.txt
-- travis_retry pip install grabbit==0.1.2
-- travis_retry git clone -b 0.6.5 https://github.com/INCF/pybids.git ${HOME}/pybids && pip install -e ${HOME}/pybids
+- travis_retry pip install grabbit==0.2.6
+- travis_retry pip install git+https://github.com/bids-standard/pybids.git@0.7.0#egg=pybids
```
effigies: Tests [skip](https://travis-ci.org/nipy/nipype/jobs/478502837#L6273-L6279) unless it's installed in editable mode.

```
interfaces/tests/test_io.py::test_bids_grabber 
[gw0] [ 83%] SKIPPED interfaces/tests/test_io.py::test_bids_grabber 
interfaces/tests/test_io.py::test_bids_fields 
[gw0] [ 84%] SKIPPED interfaces/tests/test_io.py::test_bids_fields 
interfaces/tests/test_io.py::test_bids_infields_outfields 
[gw0] [ 84%] SKIPPED interfaces/tests/test_io.py::test_bids_infields_outfields 
```

nipype/info.py
```diff
@@ -169,7 +169,7 @@ def get_nipype_gitversion():
     'profiler': ['psutil>=5.0'],
     'duecredit': ['duecredit'],
     'xvfbwrapper': ['xvfbwrapper'],
-    'pybids': ['pybids==0.6.5'],
+    'pybids': ['pybids==0.7.0'],
```
effigies: Do we want to make this such a hard pin? Admittedly, strict pins in extras don't have much force.

nipype/interfaces/io.py
```diff
@@ -2721,18 +2721,25 @@ def _list_outputs(self):
...
+        mandatory=True)
+    output_query = traits.Dict(
+        key_trait=Str,
+        value_trait=traits.Dict,
+        desc='Queries for outfield outputs')
+    raise_on_empty = traits.Bool(
+        True, usedefault=True,
+        desc='Generate exception if list is empty for a given field')
+    return_type = traits.Enum(
+        'file', 'namedtuple', usedefault=True)
```
effigies: Have we tested `return_type == 'namedtuple'` anywhere? I think this went away in 0.7.0. If we're going to drop backwards compatibility, it might be worth removing this one.

nipype/interfaces/io.py
```diff
@@ -2781,18 +2788,17 @@ def __init__(self, infields=None, **kwargs):
...
-            try:
-                from bids import layout as bidslayout
-            except ImportError:
-                from bids import grabbids as bidslayout
-            bids_config = join(dirname(bidslayout.__file__), 'config', 'bids.json')
+            from bids import layout as bidslayout
+            bids_config = join(
+                dirname(bidslayout.__file__), 'config', 'bids.json')
             bids_config = json.load(open(bids_config, 'r'))
             infields = [i['name'] for i in bids_config['entities']]
```
effigies: Can we replace all of this with `BIDSLayout().entities`? I'm not a fan of digging through the directory structure of a package unless there are no other options.

nipype/interfaces/io.py
```diff
@@ -2721,18 +2721,25 @@ def _list_outputs(self):
...
+        value_trait=traits.Dict,
+        desc='Queries for outfield outputs')
+    raise_on_empty = traits.Bool(
+        True, usedefault=True,
+        desc='Generate exception if list is empty for a given field')
+    return_type = traits.Enum(
+        'file', 'namedtuple', usedefault=True)
+    index_derivatives = traits.Bool(
+        False, usedefault=True,
+        desc='Index derivatives/ sub-directory')
```
effigies: This should be `mandatory=True`, since you don't use an `isdefined()` check on it below. It's possible to pass an `Undefined` to things with defaults to disable the input, unless `mandatory` is set `True`.

nipype/interfaces/tests/test_io.py
```diff
@@ -591,9 +591,9 @@ def test_bids_grabber(tmpdir):
...
     bg.inputs.subject = '01'
     results = bg.run()
-    assert 'sub-01_T1w.nii.gz' in map(os.path.basename, results.outputs.anat)
+    assert 'sub-01_T1w.nii.gz' in map(os.path.basename, results.outputs.T1w)
     assert 'sub-01_task-mixedgamblestask_run-01_bold.nii.gz' in \
-        map(os.path.basename, results.outputs.func)
+        map(os.path.basename, results.outputs.bold)
 
 
 @pytest.mark.skipif(not have_pybids,
```
effigies: Can we remove the `@pytest.mark.skipif(sys.version_info < (3, 0),` lines a couple below, and for the other tests, too? Pybids does still support Python 2, and we should make sure that it works, as long as it does.

nipype/interfaces/io.py
effigies: Oh, sorry, I meant that in a pseudocode sense of the entities attribute of an instantiated object, not that that snippet would work.

.travis.yml
adelavega: That is by design because when you install in editable mode you get the data that is needed for the tests

nipype/info.py
adelavega: No I guess we can do `>`

nipype/interfaces/io.py
adelavega: I'll remove that

nipype/interfaces/io.py
adelavega: Yeah, maybe I can reorganize this to set up the layout first, before setting the output traits. 

Is there any problem with instantiating the layout in `__init__`?

nipype/interfaces/io.py
adelavega: ok!

.travis.yml
effigies: Right. So we need to switch back to editable.

adelavega:
Well, I merged master and now some other unrelated tests are failing.. Any clues?

adelavega:
Also, it looks like @jdkent's issue is a more general pybids issue: https://github.com/bids-standard/pybids/issues/343

Also, some derivatives were not being validated because `dataset_description.json` was missing.

The other issue was deprecated derivatives keys. In the meantime, we can allow some custom keys to be passed in, as you can already do in pybids. I'll look into this.

nipype/interfaces/tests/test_io.py
adelavega: Yeah I guess...

.travis.yml
adelavega: Oh, I thought it was editable, not sure why that changed.

nipype/interfaces/io.py
effigies: The biggest problem is that inputs can change between instantiation and running.

adelavega:
Cool, looks like tests pass in Python 2 (aside from those failing due to another issue)

nipype/interfaces/io.py
adelavega: That's a pretty big problem! 

So I think for now instantiating a layout just to get entities does not work. Adding a helper function to pybids and then changing this later is a better option.

effigies:
Yeah, I'm starting to think it's nipype 1.16 that's breaking things somehow.

adelavega:
Hey, it works! Do you think it's ready to merge @effigies?

effigies:
Sorry, I still need to get around to a final review.

adelavega:
OK, I addressed all changes except `extra_derivatives = InputMultiObject`

What's that about? You're saying I should or shouldn't do it in this PR?

adelavega:
Regarding backwards compatibility, I think supporting pybids < 0.7 is more trouble than its worth. 

Turns out its just a single key that is causing problems, so perhaps instead we should allow custom keys to be passed in (which pybids allows), and that can be used to read in older fmripep outputs.

This is what @jdkent and I agreed on in Florida. I just wasn't necessarily going to put that into this PR, but I can.

effigies:
> What's that about? You're saying I should or shouldn't do it in this PR?

We should allow multiple paths, but changing all `MultiPath`s in the file to `MultiObject`s is out-of-scope. I probably shouldn't have mentioned it; it's just that we're moving from one to the other, so new things should be `Object` instead of `Path`.

effigies:
> Regarding backwards compatibility, I think supporting pybids < 0.7 is more trouble than its worth.

Okay. Hopefully nobody else is using this.

adelavega:
How about if we leave this PR as is (assuming I addressed your others comments and tests pass), and we can address the custom keys argument in another PR? That way @jdkent can test it out (since he has an old fmriprep input he is trying to read in.

adelavega:
Something else unrelated seems to be failing now... 

`tmpdir = local('/tmp/pytest-of-neuro/pytest-0/popen-gw1/test_split_and_merge0')`

effigies:
Merge master.

effigies:
(See #2855.)

adelavega:
It says I'm already up to date on master. I merged master very recently and that PR was 6 days ago.

adelavega:
There was one auto test that didnt get added (new file) so hopefully that's the issue.

effigies:
Nope, it's #2862. Not sure if that will work. We'll see.

adelavega:
Code maintenance is super fun!

PR #2738, opened 2018-10-18
[ENH] add option to not motion correct in create_featreg_preproc + tests

AdinaWagner:
## Summary

Fixes # .

- create_featreg_preproc gets additional option None for parameters whichvol and whichrun. If set to None, no motion correction will be performed.
- add a bit of rudimentary test

## Acknowledgment

### DIFF
--- a/nipype/workflows/fmri/fsl/preprocess.py
+++ b/nipype/workflows/fmri/fsl/preprocess.py
@@ -415,7 +415,11 @@ def create_featreg_preproc(name='featpreproc',
                            highpass=True,
                            whichvol='middle',
                            whichrun=0):
-    """Create a FEAT preprocessing workflow with registration to one volume of the first run
+    """Create a FEAT preprocessing workflow with registration to one volume of
+    the first run. If whichvol = None and/or whichrun = None, no motion
+    correction will be performed and the FEAT preprocessing workflow will not
+    return outputspec.reference, outputspec.realigned_files, and
+    outputspec.motion_plots.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


effigies:
Hi @AdinaWagner, we're no longer supporting `nipype.workflows`. We're going to be moving to the [niflows](https://github.com/niflows/) organization for organizing new workflows. The reason for this is that the current workflow structure makes it appear that the workflows are maintained by the nipype developers, which is rarely true. For the most part, these workflows have been contributed, and perhaps maintained by a short period, by a single person. When errors are found, it is not necessarily the case that the original author, who has the most expertise to assess any proposed fix, will be involved in the process. Further, for the purposes of provenance and reproducibility, the lack of versioning in the workflows, and any implicit dependence on the version of the nipype engine or the underlying interfaces, becomes problematic. Therefore, the existing nipype workflows will become `niflows/nipype1-workflows` and be included in nipype 1.x for historical reference, but will not accept modifications.

Niflows will be organized by `<author/lab/group>-<collectionname>`. For example, @poldracklab members have begun creating a nipype reimplementation of the `antsBrainExtraction.sh` script, so we've called the niflow [`poldracklab-antsbrainextraction`](https://github.com/niflows/poldracklab-antsbrainextraction). We will thus be responsible for the maintenance of this workflow, and, if it is abandoned, it will be evident to users.

If you're interested in contributing this as a niflow, the [specification](https://docs.google.com/document/d/16eWuOy7rIRKtlCBtnGO5dw4u8o6A0y3ieD0DpduS_zQ/edit#) is still in draft state, but we'd be happy to work with you (and your feedback would help us flesh it out).

cc @satra to ensure I'm accurately representing the latest consensus.

AdinaWagner:
Hi @effigies, 
Thanks for the comment and explanation! Good to hear about these future plans.

> If you're interested in contributing this as a niflow, the specification is still in draft state, but we'd be 
> happy to work with you (and your feedback would help us flesh it out).

Once I'm a bit more proficient in these procedures I'm happy to help anywhere I can. Currently though, I'm still learning with smaller scaled things. Maybe @yarikoptic will elbow me into it in a few months.

PR #2739, opened 2018-10-19
FIX: Ensure outputs can be listed in camino.ProcStreamlines by defining instance variable 

cni-md:
## Summary
Camino procstreamlines can be used without -outputroot to filter streamlines. The function  _list_outputs insits on a definition of outputroot_files in contrast to _run_interface.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/camino/convert.py
+++ b/nipype/interfaces/camino/convert.py
@@ -441,6 +441,10 @@ class ProcStreamlines(StdOutCommandLine):
             return spec.argstr % self._get_actual_outputroot(value)
         return super(ProcStreamlines, self)._format_arg(name, spec, value)
 
+    def __init__(self, *args, **kwargs):
+            super(ProcStreamlines, self).__init__(*args, **kwargs)
+            self.outputroot_files = []
+
     def _run_interface(self, runtime):
         outputroot = self.inputs.outputroot
         if isdefined(outputroot):


nipype/interfaces/camino/convert.py
```diff
@@ -463,7 +463,8 @@ def _get_actual_outputroot(self, outputroot):
     def _list_outputs(self):
         outputs = self.output_spec().get()
         outputs['proc'] = os.path.abspath(self._gen_outfilename())
-        outputs['outputroot_files'] = self.outputroot_files
+        if isdefined(self.inputs.outputroot):
+            outputs['outputroot_files'] = self.outputroot_files
```
effigies: I think there's a slightly bigger bug, which is that `self.outputroot_files` is not defined until the interface is run, and then only sometimes. I would instead make sure that it is defined at `__init__`:

```Python
def __init__(self, *args, **kwargs):
    super(ProcStreamlines, self).__init__(*args, **kwargs)
    self.outputroot_files = []
```

This way `_list_outputs()` will not raise an exception either before or after the interface is run, and whether `self.inputs.outputroot` is defined or not. For `OutputMultiPath`s, `[]` is equivalent to `Undefined`, so this change can be reverted, as well.

effigies:
Hi @cni-md. Do you have a few minutes to finish this one off?

cni-md:
@effigies not this month.

effigies:
When you do get back to this, please merge master to fix tests.

effigies:
Just a heads up: We're looking to make the next release on Nov 26, if you want to try to get this in this month.

effigies:

cni-md:
@effigies Ups, I had to work on other projects first. I am not able to finish it til monday 28th. Let's keep it for 1.1.9

effigies:
@cni-md Thanks for the response! I'll push to 1.1.9.

effigies:
LGTM, thanks.

Issue #2745, opened 2018-10-23
Apparent multiprocessing API change in Python 3.7


effigies:
### Summary

There's a Python 3.7 failure that just began exhibiting on Travis.

```Python
[gw0] [ 99%] FAILED pipeline/plugins/test/test_legacymultiproc_nondaemon.py::test_run_multiproc_nondaemon_true
...
=================================== FAILURES ===================================
______________________ test_run_multiproc_nondaemon_true _______________________
[gw0] linux -- Python 3.7.1 /home/travis/virtualenv/python3.7.1/bin/python
    def test_run_multiproc_nondaemon_true():
        # with nondaemon_flag = True, the execution should succeed
>       result = run_multiproc_nondaemon_with_flag(True)
/home/travis/build/nipy/nipype/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py:165: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/travis/build/nipy/nipype/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py:132: in run_multiproc_nondaemon_with_flag
    'non_daemon': nondaemon_flag
/home/travis/build/nipy/nipype/nipype/pipeline/engine/workflows.py:579: in run
    runner = plugin_mod(plugin_args=plugin_args)
/home/travis/build/nipy/nipype/nipype/pipeline/plugins/legacymultiproc.py:162: in __init__
    initargs=(self._cwd,)
/opt/python/3.7.1/lib/python3.7/multiprocessing/pool.py:177: in __init__
    self._repopulate_pool()
/opt/python/3.7.1/lib/python3.7/multiprocessing/pool.py:238: in _repopulate_pool
    self._wrap_exception)
/opt/python/3.7.1/lib/python3.7/multiprocessing/pool.py:252: in _repopulate_pool_static
    wrap_exception)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <[AttributeError("'NonDaemonProcess' object has no attribute '_closed'") raised in repr()] NonDaemonProcess object at 0x7fa0ad3d24a8>
group = <multiprocessing.context.ForkContext object at 0x7fa0d7aeda90>
[...]

PR #2746, opened 2018-10-23
FIX: Refine FSL.split output identification

miykael:
The current implementation of FSL's split selects all files that start with `vol*` (or the specified prefix) as output.

This caused a problem in my setup, because I had additional files in the same folder, that started with `vol_*`. This small PR corrects this, by forcing the output to start with a number.


effigies:
Thanks, @miykael!

Failures are related to #2707 and #2745. Merging.

PR #2749, opened 2018-10-24
ENH: Add 'sse' output to FSL DTIFit interface

### LINKED ISSUE #2743: FSL dtifit add sse to outputspec
### Summary
nipype's interface for FSL dtifit currently has a parameter "sse" which produces the sum of squared errors. This output does not appear in the OutputSpec.

### Actual behavior
You can use the parameter and the image is produced correctly, but there is no way to further connect the output to other nodes, since it is not part of dtifit's official outputs.

### Expected behavior
Similar to the tensor file the sse-image should be made available for further processing.

    >>> from nipype.interfaces import fsl
    >>> dti = fsl.DTIFit()
    >>> dti.inputs.dwi = 'diffusion.nii'
    >>> dti.inputs.bvecs = 'bvecs'
    >>> dti.inputs.bvals = 'bvals'
    >>> dti.inputs.base_name = 'TP'
    >>> dti.inputs.mask = 'mask.nii'
    >>> dti.inputs.sse = True
    >>> dti.save_tensor = True
    >>> dti.cmdline
    'dtifit -k diffusion.nii -o TP -m mask.nii -r bvecs -b bvals --save_tensor --sse'
    >>> out = dti.run()
    >>> out.outputs
    'FA = private/dtifit__FA.nii.gz
     L1 = private/dtifit__L1.nii.gz
     L2 = private/dtifit__L2.nii.gz
     L3 = private/dtifit__L3.nii.gz
     MD = private/dtifit__MD.nii.gz
     MO = private/dtifit__MO.nii.gz
     S0 = private/dtifit__S0.nii.gz
     V1 = private/dtifit__V1.nii.gz
     V2 = private/dtifit__V2.nii.gz
     V3 = private/dtifit__V3.nii.gz
     tensor = private/dtifit__tensor.nii.gz'

-> sse is missing

### Platform details:
This is not platform dependant!
{'commit_hash': '3ec8065',
 'networkx_version': '2.1',
 'nibabel_version': '2.1.0',
 'nipype_version': '1.1.4-dev+g3ec8065',
 'numpy_version': '1.15.0',
 'pkg_path': 'private/nipype/nipype',
 'scipy_version': '1.1.0',
 'sys_executable': '/usr/bin/python',
 'sys_platform': 'linux2',
 'sys_version': '2.7.9 (default, Jun 29 2016, 13:08:31) \n[GCC 4.9.2]',
 'traits_version': '4.6.0'}

I have written a fix based on the example of 'save_tensor', but the guidelines say i should open an issue before making a pull request, so here you go ;)
Thanks for the great work so far!

divetea:
## Summary
Add Sum-Squared-Error to DTIFit output.

Fixes #2743  .

* add 'sse' to DTIFitOutputSpec
* add logic to ignore this output if the corresponding input-flag was not set

I was not able to run the tests since i cannot install all requirements on my work system.
My changes are very similar to the already existing 'save_tensor' input with 'tensor' output.
The logic to only add the output if the corresponding input-flag was defined and set to True, was only tested manually with an example dataset, since it depends on the actual output/file creation!
I did not find any tests for the 'save tensor' -> 'tensor' input-output pair, so i had no orientation how to approach this for my new output.
In fact i did not find any tests that go further than the correct assembly of the cmdline.

I think the approach to define a dictionary which maps an optional output to the corresponding input-flag is quite generic and might be useful in similar use-cases. In the old code 'save_tensor' was handled as a special case with some duplication.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/fsl/dti.py
+++ b/nipype/interfaces/fsl/dti.py
@@ -85,6 +85,8 @@ class DTIFitOutputSpec(TraitedSpec):
               'diffusion weighting'))
     tensor = File(
         exists=True, desc='path/name of file with the 4D tensor volume')
+    sse = File(
+        exists=True, desc='path/name of file with the summed squared error')
 
 
 class DTIFit(FSLCommand):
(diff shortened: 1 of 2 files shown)


effigies:
LGTM. Thanks.

PR #2753, opened 2018-10-25
[ENH] added slice_encoding_direction input to TShift

### LINKED ISSUE #2752: adding SliceEncodingDirection to afni.TShift
### Summary
BIDS allows to specify `SliceEncodingDirection`. How about adding a  parameter to `afni.TShift` to use this information and `SliceTiming` from BIDS sidecar files directly.

This comes via [fmriprep](https://github.com/poldracklab/fmriprep/issues/1345). @effigies suggested to add this here.

BIDS allows to declare `SliceEncodingDirection` to specify the direction of `SliceTiming`. What do you think about adding a `SliceEncodingDirection` input (`{k, k-}`) to `afni.TShift`. If this is `k-`, reverse the slice timing list before running the correction.

Does it make sense to also add `i, j, i-, j-`? Or does it make sense to always assume that slices have been acquired along the third axis?

What do you think?

fliem:
## Summary

Fixes #2752 

* adds slice_encoding_direction input to TShift
* if `slice_encoding_direction == k-`, reverse `slice_timing`

## Acknowledgment

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -2665,6 +2665,17 @@ class TShift(AFNICommand):
     >>> tshift._list_outputs()['timing_file']  # doctest: +ELLIPSIS
     '.../slice_timing.1D'
 
+    >>> np.loadtxt(tshift._list_outputs()['timing_file'])[:5]
+    [0.0, 0.4, 0.8, 1.2, 1.6]
+
+    If ``slice_encoding_direction`` is set to ``'k-'``, the slice timing is reversed:
+
+    >>> tshift.inputs.slice_encoding_direction = 'k-'
+    >>> tshift.cmdline
+    '3dTshift -prefix functional_tshift -tpattern @slice_timing.1D -TR 2.5s -tzero 0.0 functional.nii'
+    >>> np.loadtxt(tshift._list_outputs()['timing_file'])[:5]
+    [15.6, 15.2, 14.8, 14.4, 14.0]
+
     This method creates a ``slice_timing.1D`` file to be passed to ``3dTshift``.
     A pre-existing slice-timing file may be used in the same way:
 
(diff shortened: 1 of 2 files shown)


fliem:
@effigies is there a simple way to add a test for that?

nipype/interfaces/afni/preprocess.py
```diff
@@ -2618,6 +2618,10 @@ class TShiftInputSpec(AFNICommandInputSpec):
         desc='time offsets from the volume acquisition onset for each slice',
         argstr='-tpattern @%s',
         xor=['tpattern'])
+    slice_encoding_direction = traits.Enum(
+        ('k', 'k-'),
+        desc='Direction in which slice_timing is specified (default: k). If negative,'
+             'slice_timing is defined in reverse order -- see BIDS specification for details.',)
```
effigies: We can have `k` be default as follows:

```Python
traits.Enum('k', 'k-', usedefault=True, desc=...)
```

Also, I would not refer to the BIDS spec here, but just explain that `k` means the third spatial dimension of the data block. While it's nice to be consistent with BIDS, when possible, nipype should stand on its own. (You can copy-paste from the spec, if you like, however.) And I would note that it only has an effect if `slice_timing` is a list; we will not reorder files.

nipype/interfaces/afni/preprocess.py
```diff
@@ -2723,9 +2727,17 @@ def _format_arg(self, name, trait_spec, value):
         return super(TShift, self)._format_arg(name, trait_spec, value)
 
     def _write_slice_timing(self):
+        slice_timing = self.inputs.slice_timing.copy()
+        if not self.inputs.slice_encoding_direction:
+            slice_encoding_direction = "k"
+        else:
+            slice_encoding_direction = self.inputs.slice_encoding_direction
```
effigies: With `usedefault` above, no need for this `if`/`else` block.

fliem:
That all sounds reasonable. If there should be any other changes, I can take a look in a couple of days.

effigies:
This LGTM. If it passes the Travis tests for Python < 3.7, I'll merge. Thanks!

PR #2769, opened 2018-11-05
FIX: Consistent treatment of timestamps in Gantt chart visualization

dPys:
Timestamps in run_stats.log are saved as strings, but previously these functions directly treated the timestamps as datetime objects. This has now been corrected across all functions.

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/docker/generate_dockerfiles.sh
+++ b/docker/generate_dockerfiles.sh
@@ -94,7 +94,7 @@ function generate_main_dockerfile() {
                 conda_install='python=${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}
                                libxml2 libxslt matplotlib mkl "numpy!=1.16.0" paramiko
                                pandas psutil scikit-learn scipy traits rdflib' \
-                pip_install="pytest-xdist niflow-nipype1-workflows" \
+                pip_install="pytest-xdist" \
                 activate=true \
     --copy docker/files/run_builddocs.sh docker/files/run_examples.sh \
            docker/files/run_pytests.sh nipype/external/fsl_imglob.py /usr/bin/ \
@@ -110,6 +110,8 @@ function generate_main_dockerfile() {
     --miniconda use_env=neuro \
                 pip_opts="-e" \
                 pip_install="/src/nipype[all] https://github.com/bids-standard/pybids/tarball/0.7.0" \
+    --miniconda use_env=neuro \
+                pip_install='"niflow-nipype1-workflows>=0.4.0"' \
     --workdir /work \
     --label org.label-schema.build-date='$BUILD_DATE' \
             org.label-schema.name="NIPYPE" \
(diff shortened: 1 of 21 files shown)


nipype/utils/draw_gantt_chart.py
```diff
@@ -291,8 +303,8 @@ def draw_nodes(start, nodes_list, cores, minute_scale, space_between_minutes,
             'color': color,
             'node_name': node['name'],
             'node_dur': node['duration'] / 60.0,
-            'node_start': node_start.strftime("%Y-%m-%d %H:%M:%S"),
-            'node_finish': node_finish.strftime("%Y-%m-%d %H:%M:%S")
+            'node_start': node_start,
+            'node_finish': node_finish
```
effigies: This seems to be replacing a string with a `datetime` object. Is that intended? It will now be formatted with `str(node_start)` and `str(node_finish)` in the format string below:

```Python
new_node = "<div class='node' style='left:%(left)spx;top:%(offset)spx;"\
    ...
    "start:%(node_start)s\nend:%(node_finish)s'></div>" %
    node_dict
```

I'm not sure that's desirable. `strftime` is insensitive to changes in `datetime.__str__` across versions.

nipype/utils/draw_gantt_chart.py
```diff
@@ -304,6 +316,9 @@ def draw_nodes(start, nodes_list, cores, minute_scale, space_between_minutes,
         # Append to output result
         result += new_node
 
+        del node_start
+        del node_finish
```
effigies: Again, no real need to delete these that I can see.

nipype/utils/draw_gantt_chart.py
```diff
@@ -360,8 +377,8 @@ def draw_resource_bar(start_time, finish_time, time_series,
             'left': left,
             'label': label,
             'duration': duration_mins,
-            'start': ts_start.strftime('%Y-%m-%d %H:%M:%S'),
-            'finish': ts_end.strftime('%Y-%m-%d %H:%M:%S')
+            'start': ts_start,
+            'finish': ts_end
```
effigies: Same as above.

nipype/utils/profiler.py
```diff
@@ -138,14 +137,20 @@ def log_nodes_cb(node, status):
     import logging
     import json
 
+    # Address nested case
+    if isinstance(node.result.runtime, (list,)):
+        resrunlist = node.result.runtime[0]
```
effigies: `resrunlist` is a single runtime, not a list. Let's just go with `result_runtime` as variable name.

nipype/utils/profiler.py
```diff
@@ -138,14 +137,20 @@ def log_nodes_cb(node, status):
...
+    except:
```
effigies:
```suggestion
    else:
```

effigies:

Issue #2700, opened 2018-09-11
Concurrent futures MultiProc backend failing to constrain resources


effigies:
### Summary

We've recently had 3 separate reports that fMRIPrep is running into memory allocation errors, since moving to the new `MultiProc` backend, based on `concurrent.futures`. In at least two of the cases, setting the plugin to `LegacyMultiProc` resolved the issue.

This suggests that the measures put into place to reduce the memory footprint of subprocesses no longer work for `concurrent.futures`.

Related: poldracklab/fmriprep#1259

dPys:
Hi @effigies , I can also confirm that these errors do not appear to be specific to fmriprep either:

When running pynets on HPC using a forkserver and restricted to the available resources on a single compute node:
```
exception calling callback for <Future at 0x2adf42c4d7f0 state=finished raised BrokenProcessPool>
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/concurrent/futures/_base.py", line 324, in _invoke_callbacks
    callback(self)
  File "/opt/conda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 143, in _async_callback
    result = args.result()
  File "/opt/conda/lib/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/opt/conda/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```
This tends to occur at varying points during the workflow depending on the resources allocated to MultiProc at runtime.

As you noted, using ```LegacyMultiProc``` appears to resolve the issue.

-Derek

PR #2773, opened 2018-11-09
[FIX] LegacyMultiProc hangs up indefinitely

### LINKED ISSUE #2548: MultiProc goes into infinite loop (resource management related)

oesteban:
This PR relates to #2700, and should fix the problem
underlying #2548.

I first considered adding a control thread that monitors
the `Pool` of workers, but that would require a large overhead
keeping track of PIDs and polling very often.

Just adding the core file of [bpo-22393](https://github.com/python/cpython/pull/10441)
should fix #2548

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/external/patchedpool.py
+++ b/nipype/external/patchedpool.py
@@ -1,0 +1,912 @@
+#
+# Module providing the `Pool` class for managing a process pool
+#
+# multiprocessing/pool.py
+#
+# Copyright (c) 2006-2008, R Oudkerk
+# Licensed to PSF under a Contributor Agreement.
+#
+
+__all__ = ['Pool', 'ThreadPool']
+
+#
+# Imports
+#
+
+import threading
+import queue
+import itertools
+import collections
+import os
+import time
+import traceback
+
+# If threading is available then ThreadPool should be provided.  Therefore
+# we avoid top-level imports which are liable to fail on some systems.
+from multiprocessing import util
+from multiprocessing import get_context, TimeoutError
+
+#
+# Constants representing the state of a pool
+#
+
+RUN = 0
+CLOSE = 1
+TERMINATE = 2
+BROKEN = 3
+
+#
+# Miscellaneous
+#
+
+job_counter = itertools.count()
+
+
+def mapstar(args):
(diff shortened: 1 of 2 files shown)


oesteban:
(I haven't tested this yet, though)

effigies:
As the Python 3.8 version of this, I'm not sure how well this will play with 2.7 and 3.4, at the least.

oesteban:
Yep, unfortunately, the fix is python +3.7. I'd say we start recommending switching to python 3.7 if the workflow is memory-hungry.

effigies:
Given the recent [numpy bug](https://www.opensourceanswers.com/blog/you-shouldnt-use-python-37-for-data-science-right-now.html), I'm hesitant to suggest moving scientific applications to 3.7 until upstream projects have had a bit more time to find and fix critical bugs.

How hard would this be to port to 3.6? I suspect if we get it working there, 3.5 and 3.4 will fall into place quickly.

I'm okay with abandoning 2.7, as long as we don't break current usage further.

oesteban:
Some thoughts:

- As per your link, anaconda fixed the bug in numpy 1.15.2, and it is fixed with 1.15.3 with [Pypi](https://pypi.org/project/numpy/). I guess we should update the pinned versions if python == 3.7.
- I'm not proposing to discontinue python <3.7, I would just make this fix available for python  +3.7.
- If nothing of the above works, then I'd say it should not be too hard to backport to 3.6 or even 3.5 (prior to 3.5 it would be harder). Although I'm not sure it is worthwhile.

PR #2775, opened 2018-11-12
[PIN] Force numpy>=1.15.4 when Python>=3.7

oesteban:
Addresses the [numpy
bug](https://www.opensourceanswers.com/blog/you-shouldnt-use-python-37-for-data-science-right-now.html).

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -101,7 +101,9 @@ existing pipeline systems.
 # versions
 NIBABEL_MIN_VERSION = '2.1.0'
 NETWORKX_MIN_VERSION = '1.9'
-NUMPY_MIN_VERSION = '1.9.0'
+# Numpy bug in python 3.7:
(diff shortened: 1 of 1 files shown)


satra:
@mgxd - just a note that we should adjust the conda feedstock for this as well.

effigies:
I think this will work for `sdist`, but not `bdist`. To encode this requirement in the actual package metadata, I think we'll need to use [PEP 508](https://www.python.org/dev/peps/pep-0508/) syntax. e.g.

```
numpy >= 1.9.1
numpy >= 1.15.3 ; python_version >= "3.7"
```

(That may not work.)

@satra Conda doesn't seem to work well with 3.7 yet. There's an open PR (conda-forge/nipype-feedstock#28) that tries to update to support 3.7 that hasn't yet passed tests.

PR #2785, opened 2018-11-18
ENH: CompositeTransformUtil, new ANTs interface

TheChymera:
Haven't pushed with `make specs`, since it edited exclusively files unrelated to this PR.

## Acknowledgment


nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,59 @@ def _list_outputs(self):
...
+    Examples
+    --------
+
+    >>> from nipype.interfaces.ants import CompositeTransformUtil
+    >>> tran = CompositeTransformUtil()
+    >>> tran.inputs.process = 'disassemble'
+    >>> tran.inputs.in_file = 'output_Composite.h5'
+    >>> reg.cmdline
+    'CompositeTransformUtil --disassemble output_Composite.h5 transform'
+    >>> reg.run()  # doctest: +SKIP
```
effigies:
```Python
>>> tran.cmdline
>>> tran.run()
```

TheChymera:
@effigies sorry, I somehow missed the notification :-/

nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,59 @@ def _list_outputs(self):
...
+
+class CompositeTransformUtil(ANTSCommand):
+    """
+    ANTs utility which can combine or break apart transform files into their individual
+    constituent components.
+
+    Examples
+    --------
+
+    >>> from nipype.interfaces.ants import CompositeTransformUtil
```
effigies: This is failing. For it to work, you'll need to import this interface into [`nipype/interfaces/ants/__init__.py`](https://github.com/nipy/nipype/blob/dcc013b0187bf5312db9174f8cb8932991c601cb/nipype/interfaces/ants/__init__.py#L7-L8):

https://github.com/nipy/nipype/blob/dcc013b0187bf5312db9174f8cb8932991c601cb/nipype/interfaces/ants/__init__.py#L7-L8

nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,59 @@ def _list_outputs(self):
...
+
+    def _format_arg(self, name, spec, value):
+        if name == 'output_prefix' and self.inputs.process == 'assemble':
+            value = ''
+        return super(CompositeTransformUtil, self)._format_arg(name, spec, value)
+
+    def _list_outputs(self):
+        outputs = self.output_spec().get()
+        outputs['affine_transform'] = os.path.abspath(
+            '00_'+self.inputs.output_prefix+'_AffineTransform.mat')
```
effigies: Need spaces around operators, or you can use format strings, e.g.:

```suggestion
            '00_' + self.inputs.output_prefix + '_AffineTransform.mat')
```

or

```suggestion
            '00_%s_AffineTransform.mat' % self.inputs.output_prefix)
```

nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,59 @@ def _list_outputs(self):
...
+        outputs['displacement_field'] = os.path.abspath(
+            '01_'+self.inputs.output_prefix+'_DisplacementFieldTransform.nii.gz')
```
effigies: And:

```suggestion
            '01_' + self.inputs.output_prefix + '_DisplacementFieldTransform.nii.gz')
```

or

```suggestion
            '01_%s_DisplacementFieldTransform.nii.gz' % self.inputs.output_prefix)
```

nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,59 @@ def _list_outputs(self):
...
+    --------
+
+    >>> from nipype.interfaces.ants import CompositeTransformUtil
+    >>> tran = CompositeTransformUtil()
+    >>> tran.inputs.process = 'disassemble'
+    >>> tran.inputs.in_file = 'output_Composite.h5'
+    >>> tran.cmdline
+    'CompositeTransformUtil --disassemble output_Composite.h5 transform'
+    >>> tran.run()  # doctest: +SKIP
+    """
```
effigies: Maybe add an example with `'assemble'`, as well, since that changes the use of `process`. That will both make clearer how the interface works and make sure we exercise that bit of code when running tests.

effigies:
@TheChymera No worries. Tagging you so this hits your inbox with higher probability.

nipype/interfaces/ants/registration.py
TheChymera: how about `.format()`?

TheChymera:
@effigies done. Did not include changes introduced by `make specs` as they are unrelated to this commit. If you want I can follow up with a separate PR on that.

TheChymera:
@effigies The function is generally used to transition between the newer .h5 format and the old .mat + .nii.gz format in ANTs. So generally these assumptions hold. I agree that a general-purpose solution would be even better, but in that case there's no way to know what transforms went into the file. Is there any way for nipype to just monitor what files were produce and return that list for the outputs?

In any case I don't think splitting this fairly simple function further up would help. At most we can just disable the `--disassemble` mode until such a time when we can figure out a solution for the output file names which are undefined until you run it.

effigies:
> At most we can just disable the `--disassemble` mode until such a time when we can figure out a solution for the output file names

I don't think it works for `--assemble` at all, though. Just given that the inputs and outputs are so different in the two modes is what made me consider that this might be better as two separate interfaces, called however you like. But in the meantime, maybe we should just make this a `--disassemble` only interface. And if somebody runs into the case where outputs are different from expected, we can fix it up then.

effigies:
> with the new PR --assemble is the one which works

The outputs seem targeted towards disassemble only, with no mention of an assembled output file.

> `--disassemble` also works, but only in the restricted use case

This seems fine, for now. The common case seems to be affine+warp.

> Is there any way to record the names of whatever files the command line interface creates?

If you know the pattern, you can use `glob`. It's not ideal, but when the names/number of outputs depend on the contents of inputs, there's no real getting around it.

TheChymera:
> The outputs seem targeted towards disassemble only, with no mention of an assembled output file.

You're right, I've been editing this in parallel to some other things and lost sight of what I had already done.
 I've tested this now again with my common inputs, and it works with for both assembly and disassembly.

nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,77 @@ def _list_outputs(self):
...
+    def _num_threads_update(self):
+        """
+        CompositeTransformUtil ignores environment variables,
+        so override environment update from ANTSCommand class
+        """
+        pass
+
+    def _format_arg(self, name, spec, value):
+        if name == 'output_prefix' and self.inputs.process == 'assemble':
+            value = ''
```
effigies:
```suggestion
            return ''
```

nipype/interfaces/ants/registration.py
```diff
@@ -1596,3 +1596,77 @@ def _list_outputs(self):
...
+        if name == 'out_file' and self.inputs.process == 'disassemble':
+            value = ''
```
effigies:
```suggestion
            return ''
```

TheChymera:
`make specs` again just adds things unrelated to the interface:

```
diff --git a/nipype/algorithms/tests/test_auto_ACompCor.py b/nipype/algorithms/tests/test_auto_ACompCor.py
index eadbf3e12..235d15da9 100644
--- a/nipype/algorithms/tests/test_auto_ACompCor.py
+++ b/nipype/algorithms/tests/test_auto_ACompCor.py
@@ -6,6 +6,7 @@ from ..confounds import ACompCor
 def test_ACompCor_inputs():
     input_map = dict(
         components_file=dict(usedefault=True, ),
+        failure_mode=dict(usedefault=True, ),
         header_prefix=dict(),
         high_pass_cutoff=dict(usedefault=True, ),
         ignore_initial_volumes=dict(usedefault=True, ),
diff --git a/nipype/algorithms/tests/test_auto_TCompCor.py b/nipype/algorithms/tests/test_auto_TCompCor.py
index 44b01b297..59a5b84f7 100644
--- a/nipype/algorithms/tests/test_auto_TCompCor.py
+++ b/nipype/algorithms/tests/test_auto_TCompCor.py
@@ -6,6 +6,7 @@ from ..confounds import TCompCor
 def test_TCompCor_inputs():
     input_map = dict(
         components_file=dict(usedefault=True, ),
+        failure_mode=dict(usedefault=True, ),
         header_prefix=dict(),
         high_pass_cutoff=dict(usedefault=True, ),
         ignore_initial_volumes=dict(usedefault=True, ),
```

I can submit a separate PR for that.

PR #2786, opened 2018-11-21
[FIX] MapNodes fail when ``MultiProcPlugin`` passed by instance

oesteban:
When calling ``workflow.run(plugin=MultiProcPlugin())``, MapNodes are set the ``use_plugin`` member that is not pickleable if ``plugin`` is a ``MultiProcPlugin`` object.

## Summary

Fixes # .

## Acknowledgment


oesteban:
Given the very low risk of this patch, I'll go ahead and merge it.

Issue #2776, opened 2018-11-13
Idea to minimize memory fingerprint


oesteban:
### Summary
Large workflows are prone to get killed by the OOM Killer in linux and similar devices in other systems. The common denominator is that the RSS memory fingerprint is sensible, not too big, but the VM skyrockets when the workflow has many nodes.

### Actual behavior

Except for pure python interfaces, all nipype interfaces fork new processes via `subprocess.Popen`. That way, all new processes are started using `fork` and allocate for twice as much virtual memory as it is allocated before forking. This, in addition with python inefficiency to garbage collect after finishing the process, leads to overcommitting memory (and processes being killed on systems that would not allow that).

### Expected behavior
Less memory consumption

### How to replicate the behavior
Run fmriprep on a large dataset, disallowing overcommitting and with a low memory limit (e.g. 8GB)

### Idea:

Patch `subprocess.Popen` with `multiprocessing.context.Popen`. Doing that, in theory, all these processes should be forked on the server process (which should have a constant, small memory fingerprint).

WDYT @satra @effigies @chrisfilo ?

BTW, this solution might be only possible with Python>=3.4. As per https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods:

> _Changed in version 3.4_: spawn added on all unix platforms, and forkserver added for some unix platforms. Child processes no longer inherit all of the parents inheritable handles on Windows.

chrisfilo:
Sounds reasonable. By "Patch" you mean replace?

effigies:
Seems worth trying.

satra:
let's try it out. 

are we the only ones struggling with memory? how does dask handle this?

oesteban:
> @chrisfilo: By "Patch" you mean replace?
> @effigies: Seems worth trying.
> @satra let's try it out.

Yes, although I'm checking and such replacement is not as easy as I first thought. The interface of the `Popen` in `multiprocessing` is not the same as that in `subprocess`.

> @satra are we the only ones struggling with memory? how does dask handle this?
I think this is heavily related to this: https://bugs.python.org/issue8713#msg105719, which gave rise to adding the forkserver mode in multiprocessing.

[...]

PR #2787, opened 2018-11-21
[RF] Detachable ``run_command`` process

oesteban:
## Summary

**Introduction**: the current implementation of ``run_command`` makes use of ``Stream``s to redirect output in a sort of inflexible way, and does not allow running the processes in the background.

Ref #2776.

  * modifies the behavior of ``run_command`` opening a new self-control thread, that is in charge of printing to the terminal if the ``terminal_output`` is set to ``'stream'``
  * output streams are redirected to files by default, trying to reduce the amount of data buffered in memory and also to ensure that there is a trace of nipype's execution.
  * a new argument ``background`` can be set to ``True``, detaching from the process. This lends itself to important improvements on the ``MultiProc`` variants, as all these processes can be launched directly from the main thread and run in the background. I hope this will drastically reduce the memory fingerprint of workers (if we actually need to maintain a ``Pool``, which is probably not the case anymore!).

**Tests**: I've added a couple of tests, but I'm happy to add more as I am aware that this may be too big of a change to the nipype interfaces.

**Extra**: removed a couple of deprecated objects that we promised to trash past nipype-1.1.0.

## Acknowledgment


effigies:

oesteban:
I think I will start over, with less broad, smaller PRs to take on this.

PR #2788, opened 2018-11-22
[MAINT] Stop using deprecated ``logger.warn()``

oesteban:
Trying to avoid:

```
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```


oesteban:
Tests should pass now with master fixed. Merging.

PR #2790, opened 2018-11-22
Revert "[FIX] MapNodes fail when ``MultiProcPlugin`` passed by instance"

oesteban:
Reverts nipy/nipype#2786


oesteban:
nilearn-0.5.0 was released over the last week. That is the issue.

PR #2800, opened 2018-11-26
REL: 1.1.6

effigies:
## Summary

Prep for 1.1.6 release, targeting Monday, November 26.

## Release checklist

* [x] Update changelog
* [x] Update .mailmap
* [x] Update .zenodo.json
* [x] Set release number in `nipype/info.py` and `doc/conf.py`
* [x] Update `doc/documentation.rst` with previous releases
* [x] Tutorial tests (@djarecka)

## Acknowledgment

- [x] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

PR #2801, opened 2018-11-26
[STY] Combine split import

oesteban:
## Summary

Fixes # .

## Acknowledgment


oesteban:
I think this is safe to include in #2800

PR #2802, opened 2018-11-26
[TST] Improve py.test configuration of doctests

oesteban:
This PR ensures that doctests are run OUTSIDE nipype's source directory
in all the cases. Previously, doctests would chdir into the testing/data
folder and some interfaces would generate results.

Also, deletes ``nipype/testing/data/slice_timing.1D``, which seems to
have sneaked into the tree for these very reasons in some past PR.

This can be held out until next release after 1.1.6.

## Summary

Fixes # .

## Acknowledgment


oesteban:
Tests are passing and coverage decreased because the #smoketests target was not received by codecov. I've checked and both travis and circle are giving exactly the same number of tests (passed, xfailed, skipped) before and after these changes. Merge?

PR #2805, opened 2018-11-27
[WIP] Lightweight implementation of ``InterfaceResult``

oesteban:
## Summary

(Ref #2776)
This PR replaces the good old ``Bunch``. ``Bunch`` has been moved to the module ``nipype.utils.misc``, although it is still loadable from ``nipype.interfaces.base``.

The memory fingerprint of runtime objects improve by some 20%, not that we will see a huge difference but it justifies itself as an optimization:

```Python
>>> from nipype.interfaces.base.support import InterfaceRuntime 
>>> from pympler import asizeof 
>>> rt = InterfaceRuntime(cmdline='/bin/echo', cwd='/home/oesteban', stdout='some text that is actually somewhat longer than other parameters') 
>>> asizeof.asizeof(rt)
472

>>> from nipype.interfaces.base import Bunch
>>> from pympler import asizeof
>>> rt = Bunch(cmdline='/bin/echo', cwd='/home/oesteban', stdout='some text that is actually somewhat longer than other parameters')
>>> asizeof.asizeof(rt)
584
```
** ``pympler`` measures the whole size of the data structure, including the full tree of references.

  - A new ``nipype.interfaces.base.support.InterfaceRuntime`` object with a reduce memory footprint that the previous ``Bunch``.
  - ``Bunch`` has been moved to ``nipype.utils.misc`` and some unused code has been removed.

## Acknowledgment


nipype/interfaces/base/support.py
```diff
@@ -9,185 +9,166 @@
...
+from ... import __version__
 
-from ... import logging
-from ...utils.misc import is_container
-from ...utils.filemanip import md5, to_str, hash_infile
-iflogger = logging.getLogger('nipype.interface')
 
-
-class NipypeInterfaceError(Exception):
+class NipypeInterfaceError(RuntimeError):
```
effigies: I would prefer not to make this API change, as previously an `except RuntimeError` would not catch this, so somebody who is counting on this to make it past that exception will be surprised. I think we should move to a proper `NipypeError` hierarchy in 2.0, though.

nipype/interfaces/base/support.py
```diff
@@ -9,185 +9,166 @@
...
+    can be pickled/unpickled:
 
-    def items(self):
-        """iterates over bunch attributes as key, value pairs"""
-        return list(self.__dict__.items())
+    >>> import pickle
+    >>> pickleds = pickle.dumps(rt)
+    >>> newrt = pickle.loads(pickleds)
+    >>> newrt
+    Bunch(cmdline='/bin/echo', cwd='/scratch/workflow', returncode=0)
```
effigies: Why `Bunch` and not `InterfaceRuntime`?

nipype/interfaces/base/support.py
```diff
@@ -215,34 +196,68 @@ class InterfaceResult(object):
...
+    >>> import pickle
+    >>> pickleds = pickle.dumps(result)
+    >>> newresult = pickle.loads(pickleds)
+    >>> newresult  # doctest: +ELLIPSIS
+    InterfaceResult(interface='CommandLine', runtime=\
+Bunch(cmdline='/bin/echo', returncode=0), version=...)
+
     """
+    __slots__ = ['interface', 'runtime', 'inputs', 'outputs', 'provenance']
+    version = __version__
```
effigies: I think the goal of the `version` attribute was to invalidate results loaded from an incompatible format, not record the nipype version.

nipype/interfaces/base/support.py
```diff
@@ -215,34 +196,68 @@ class InterfaceResult(object):
...
         self.provenance = provenance
 
-    @property
-    def version(self):
-        return self._version
 
+    def __setstate__(self, state):
+        """Necessary for un-pickling"""
+        for key in self.__class__.__slots__:
+            setattr(self, key, state.get(key, None))
```
effigies: Should you check that `state.get('version') == self.version`?

nipype/pipeline/engine/utils.py
```diff
@@ -192,37 +192,37 @@ def write_report(node, report_type=None, is_mapnode=False):
         'prev_wd': getattr(result.runtime, 'prevcwd', '<not-set>'),
     }
 
-    if hasattr(result.runtime, 'cmdline'):
+    if result.runtime.cmdline:
```
effigies: Should these all be `is not None`, or are empty strings/dicts/zeros meant to be treated the same?

nipype/utils/draw_gantt_chart.py
```diff
@@ -322,10 +322,7 @@ def draw_resource_bar(start_time, finish_time, time_series,
     space_between_minutes = space_between_minutes / scale
 
     # Iterate through time series
-    if PY3:
-        ts_items = time_series.items()
-    else:
-        ts_items = time_series.iteritems()
+    ts_items = list(time_series.items())
```
effigies: This seems to be specifically undoing the iterable-ness of `ts_items`. Why?

nipype/interfaces/base/support.py
oesteban: That is for backwards compatibility. If I change Bunch I'm afraid I'll need to update all the tests. But I can make that extra mile.

nipype/interfaces/base/support.py
oesteban: MMM so this is why we had 2.0 here? was that the pickle protocol version?

nipype/interfaces/base/support.py
oesteban: Throw a warning at least would be appropriate. Since I was trying to do a minimal proof of concept, I passed. But yes, I'll include the warning.

nipype/utils/draw_gantt_chart.py
oesteban: it runs `len` a couple of lines below. I can roll back this change if you want (tried to remove all the `iteritems` that were remaining).

nipype/pipeline/engine/utils.py
oesteban: I'll change the `cmdline`, `mem_peak_gb`, and `cpu_percent` 👍

nipype/utils/draw_gantt_chart.py
oesteban: You were right on this one, the `len` is calculated on a different variable. Fixed.

nipype/interfaces/base/support.py
oesteban: Warning added :+1:

nipype/interfaces/base/support.py
effigies: @satra Perhaps you can clarify? I interpreted as the `InterfaceResult` API version.

oesteban:
Do not merge yet, still under testing.

effigies:

oesteban:
We can push this one to 1.1.9

effigies:
@oesteban Just a bump to note that the merge window closes next Friday, Feb 22.

oesteban:
I'm going to close this one for now. I am confident this is a good direction, but I need to mull over the details of the implementation. I'll reopen when ready.

effigies:
I'm going to mark it orphaned. Let's start a new PR when you're ready.

effigies:

Issue #2764, opened 2018-10-29
[BUG] Relative path in ResourceMonitor


oesteban:
### Summary

This https://github.com/nipy/nipype/blob/502fc63488764434374a41e7456ae72f0ff9ae51/nipype/utils/profiler.py#L43 should be an absolute path (it isn't).

### Actual behavior

It doesn't seem to break things, but this may be the origin for weird readings in the last line of the monitor files. The directory may change before the file descriptor is open.

PR #2816, opened 2018-12-09
[FIX] Indices were swapped for memory and cpu profile data

satra:
## Summary

Fixes what looks like a swapped index.

change indices

## Acknowledgment


oesteban:
OMG - Can we also address #2764 with this PR? (just adding the os.path.abspath).

PR #2817, opened 2018-12-11
FIX: Order of SPM.NewSegment channel_info boolean tuple is (Field, Corrected)

servoz:
## Summary

* In SPM12, for Segment module (the NewSegment in Nipype), the spm.spatial.preproc.channel.write parameter (SPM's Save Bias Corrected parameter) takes [Save Bias Field = 0 or 1, Save Bias Corrected = 0 or 1]. In nypipe the initialisation of the outputs bias_field_images and bias_corrected_images are done with the inverse order. 

* This commit fix this inconsistency.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -1487,7 +1487,7 @@ class NewSegmentInputSpec(SPMCommandInputSpec):
         desc="""A tuple with the following fields:
             - bias reguralisation (0-10)
             - FWHM of Gaussian smoothness of bias
-            - which maps to save (Corrected, Field) - a tuple of two boolean values""",
+            - which maps to save (Field, Corrected) - a tuple of two boolean values""",
         field='channel')
     tissues = traits.List(
         traits.Tuple(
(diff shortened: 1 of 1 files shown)


effigies:
If I'm understanding you correctly, users were submitting valid `(bool, bool)` pairs that were being correctly interpreted by SPM, but Nipype was misinterpreting them, and specifying the wrong output?

It looks like this was written for SPM8 and adapted to support SPM12. Is it possible that the order switched in SPM12? If so, the correct response might be to reverse that tuple if SPM12 is detected, and make a note of it in the docstring.

servoz:
@elfigies,
i just checked:
In SPM8:
New Segment module:
spm.tools.preproc8.channel.write = [0 1] #  Save Bias Corrected (only)
spm.tools.preproc8.channel.write = [1 0] #  Save Bias Field (only)

As fas as I can check, in SPM8 the first element of the write list is concerning to the Bias Field (BiasField_aName.nii) and the second the Bias Corrected  image (maName.nii) ... It is exactly the same in SPM12 (my first post). You  invert in nipype! We currently overload the corresponding nipype class in our project. It will be cleaner if you fix it by merging my PR !

chrisfilo:
I don't see anything wrong here, but I have not used SPM in ages.

PR #2823, opened 2018-12-13
ENH: Support tckgen -select in MRtrix3 v3+

### LINKED ISSUE #2821: Renamed MRtrix 3.0 `tckgen` option(s)
https://github.com/nipy/nipype/blob/704b97dee7848283692bac38f04541c5af2a87b5/nipype/interfaces/mrtrix3/tracking.py#L78

Hi, this option (and maybe others) is no longer up-to-date according to [MRtrix 3.0 documentation](https://mrtrix.readthedocs.io/en/latest/reference/commands/tckgen.html#streamlines-tractography-options). It has changed to `-select`.

jguillon:
## Summary

Fixes #2821 .

* Switch `tckgen` option from `-number` to `-select`

## Acknowledgment


effigies:
I guess the question is whether we should support the 0.3.x series or just 3.0. Changing this will break anybody that uses old MRtrix3 but new Nipype.

Perhaps the better option would be to put a `max_ver=0.4` on `n_tracks` and add a `select` input with a `min_ver=3`.

nipype/interfaces/mrtrix3/tracking.py
```diff
@@ -76,6 +76,13 @@ class TractographyInputSpec(MRTrix3BaseInputSpec):
               'is 90deg x stepsize / voxelsize)'))
     n_tracks = traits.Int(
         argstr='-number %d',
+        max_ver=0.4,
```
effigies:
```suggestion
        max_ver='0.4',
```

nipype/interfaces/mrtrix3/tracking.py
```diff
@@ -76,6 +76,13 @@ class TractographyInputSpec(MRTrix3BaseInputSpec):
...
+        desc=('set the desired number of tracks. The program will continue'
+              ' to generate tracks until this number of tracks have been '
+              'selected and written to the output file'))
+    select = traits.Int(
+        argstr='-select %d',
+        min_ver=3,
```
effigies:
```suggestion
        min_ver='3',
```

effigies:
@jguillon Could you merge current master. I think that should fix the tests.

PR #2830, opened 2018-12-14
[ENH] Add interfaces wrapping DIPY worflows

skoudoro:
Hi Nipype Devs,

The main goal of this PR is to show how to convert a DIPY command line (named workflow) to a Nipype interface. This is a follow up from our discussion with @satra, @arokem, @Garyfallidis and me. 

This PR adds : 

- DKI diffusion Model
- CSA diffusion Model
- MAPMRI diffusion Model
- DTI Model
- Streamline Registration
- Reslicing
- Recobundles
- LabelsBundles
- Deterministic tracking

TODO:
--------
- [x] catch and set up default values for the optional arguments and outputs.
- [x] Add unit Test (I do not know the convention here since they are auto-generated)

### DIFF
--- a/nipype/interfaces/dipy/tracks.py
+++ b/nipype/interfaces/dipy/tracks.py
@@ -30,7 +30,7 @@ if HAVE_DIPY and LooseVersion(dipy_version()) >= LooseVersion('0.15'):
 
 else:
     IFLOGGER.info("We advise you to upgrade DIPY version. This upgrade will"
-                  " activate RecoBundles, LabelsBundles, DetTrackPAMFlow.")
+                  " activate RecoBundles, LabelsBundles, DeterministicTracking.")
 
 
 class TrackDensityMapInputSpec(BaseInterfaceInputSpec):
--- a/nipype/interfaces/dipy/tracks.py
+++ b/nipype/interfaces/dipy/tracks.py
@@ -6,12 +6,31 @@ import os.path as op
 import numpy as np
 import nibabel as nb
 import nibabel.trackvis as nbt
+from distutils.version import LooseVersion
 
 from ... import logging
 from ..base import (TraitedSpec, BaseInterfaceInputSpec, File, isdefined,
                     traits)
-from .base import DipyBaseInterface
+from .base import (DipyBaseInterface, HAVE_DIPY, dipy_version,
+                   dipy_to_nipype_interface)
+
 IFLOGGER = logging.getLogger('nipype.interface')
+
+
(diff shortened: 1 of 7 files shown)


skoudoro:
Hi @satra @effigies, 

Any thoughts about this PR ? Any chances to have this PR in  for the next release?

effigies:
Hi Serge, I think we can shoot for that. I'm not sure how much energy everyone has right now, but we've got a little over a week before we should freeze new merges.

skoudoro:
Ok, Thank you for your feedback Chris!

nipype/interfaces/dipy/tracks.py
```diff
@@ -6,14 +6,33 @@
...
+
+    RecoBundles = dipy_to_nipype_interface("RecoBundles", RecoBundlesFlow)
+    LabelsBundles = dipy_to_nipype_interface("LabelsBundles",
+                                             LabelsBundlesFlow)
+    DeterministicTracking = dipy_to_nipype_interface("DeterministicTracking",
+                                                     DetTrackPAMFlow)
+
+else:
+    IFLOGGER.info("We advise you to upgrade DIPY version. This upgrade will"
+                  " activate RecoBundles, LabelsBundles, DetTrackPAMFlow.")
```
effigies:
```suggestion
                  " activate RecoBundles, LabelsBundles, DeterministicTracking.")
```

effigies:
@skoudoro Just a quick bump. I think this is basically ready to go if you want to look over my suggestions.

skoudoro:
Thank you for the review and sorry for the late answer, I was travelling.

I applied your suggestion, it was a good catch.

@arokem and I were wondering if it is possible to automate this task for each DIPY release, or during nipype installation. Indeed, It should be easy to catch all workflows from DIPY and then, apply `dipy_to_nipype` function.

Any suggestion is welcomed!

arokem:
Even at import time! Nipype would import dipy and traverse the workflows
namespace, finding names that conform with a particular pattern and making
the interfaces from these names.

I'd love to hear whether you think this is plausible, but I don't
it's necessary to do now. I think that can be a refactor of the work here.

effigies:
@arokem I agree with that approach. It might also be useful to have a registry, though, that associates interfaces with minimum required dipy versions, so that we can give a useful error when somebody tries to use a more recent workflow with an older dipy installed.

I also agree that that's a problem for another day.

effigies:
Oh, and yes, that seems completely plausible.

Issue #948, opened 2014-10-10
TOPUP and ApplyTOPUP crash with images with odd number of pixels


oesteban:
When any of the dimensions of the images has odd number of pixels, topup crashes:

```
Topup: msg=topup_clp::topup_clp: Subsampling levels incompatible with image data
```

It'd be great if the interfaces checked this and transparently added the slices as FSL experts suggest here: https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;67dcb45c.1209

akeshavan:
I'm reopening this! I'm still having this problem. I've changed oscar's original commit by adding a flag to turn the resizing on/off. 

For some reason, the applytopup interface isn't working. Any ideas?

nqueder:
I noticed that this issue is a bit old, but I am running into the same error when I'm trying to run topup on my dmri dataset. 
Has anyone been able to find a way around it?

DISAPPEARED13:
I solved this problem with adding some slices in the front and back of the volume...

tasciseda:
How did you add slice to your image?
Thank you.

PR #2832, opened 2018-12-19
Enh/wip/fix/brk topup for odd dimensions. help needed

akeshavan:
## Summary

Attempts to fix https://github.com/nipy/nipype/issues/948 -- its 4 years later and we are still having this problem! I slightly modified @oesteban 's original commit with a flag called `checksize`, which, if set to `True` will make the size even before running TOPUP and ApplyTOPUP. It's set to `False` by default.

Problem is, the ApplyTOPUP interface doesn't work!! The process just gets killed. **Someone help please!**

- [ ] fix the ApplyTOPUP interface

## Acknowledgment


akeshavan:
Nevermind about the applytopup, it turns out I was testing this on a pretty large file :) thanks @oesteban for the help!

nipype/interfaces/fsl/epi.py
```diff
@@ -124,11 +124,118 @@ def _run_interface(self, runtime):
...
+                except KeyError:
+                    pass
+        return outputs
+
+    def _run_interface(self, runtime):
+        check = dict(checksize=lambda t: t is not None)
+        for name, spec in sorted(self.inputs.traits(**check).items()):
+            if spec.checksize and self.inputs.checksize:
+                value = self._checksize(name)
+                setattr(self.inputs, name, value)
```
effigies: Is this going to cause it to be rerun every time? I would verify.

mattcieslak:
Would you consider adding a file dependency? There is a second cnf that works for volumes with odd numbers of slices. 

here's `b02b0_1.cnf` from the fsl mailing list

```
# Resolution (knot-spacing) of warps in mm
--warpres=20,16,14,12,10,6,4,4,4
# Subsampling level (a value of 2 indicates that a 2x2x2 neighbourhood is collapsed to 1 voxel)
--subsamp=1,1,1,1,1,1,1,1,1
# FWHM of gaussian smoothing
--fwhm=8,6,4,3,3,2,1,0,0
# Maximum number of iterations
--miter=5,5,5,5,5,10,10,20,20
# Relative weight of regularisation
--lambda=0.0005,0.0001,0.00001,0.0000015,0.0000005,0.0000005,0.00000005,0.0000000005,0.00000000001
# If set to 1 lambda is multiplied by the current average squared difference
--ssqlambda=1
# Regularisation model
--regmod=bending_energy
# If set to 1 movements are estimated along with the field
--estmov=1,1,1,1,1,0,0,0,0
# 0=Levenberg-Marquardt, 1=Scaled Conjugate Gradient
--minmet=0,0,0,0,0,1,1,1,1
# Quadratic or cubic splines
--splineorder=3
# Precision for calculation and storage of Hessian
--numprec=double
# Linear or spline interpolation
--interp=spline
# If set to 1 the images are individually scaled to a common mean intensity 
--scale=1
```

effigies:
@mattcieslak If you want to open a PR, that would be fine. It shouldn't be a problem to bundle a file.

Apologies for the thread necromancy, but I'm closing dead PRs.

PR #2838, opened 2018-12-21
ENH/FIX: Resolve LinAlgError during SVD

feilong:
The default `lapack_driver` uses divide and conquer SVD (`gesdd`), which occasionally will cause `LinAlgError` even if the data are completely fine. Switching `lapack_driver` to `gesvd` when it fails can solve this issue.

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -14,7 +14,6 @@ import os.path as op
 import nibabel as nb
 import numpy as np
 from numpy.polynomial import Legendre
-from scipy import linalg
 
 from .. import config, logging
 from ..external.due import BibTeX
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -14,6 +14,7 @@ import os.path as op
 import nibabel as nb
 import numpy as np
 from numpy.polynomial import Legendre
+from scipy import linalg
 
 from .. import config, logging
 from ..external.due import BibTeX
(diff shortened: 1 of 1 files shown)


nipype/algorithms/confounds.py
```diff
@@ -1193,9 +1194,12 @@ def compute_noise_components(imgseries, mask_images, num_components,
         try:
             u, _, _ = np.linalg.svd(M, full_matrices=False)
```
effigies: Nested `try`/`except` blocks get a little hard to reason about. How about a helper function that will do the fall-back:

```Python
# Use the numpy.linalg.svd protocol
def fallback_svd(a, full_matrices=True, compute_uv=True):
    try:
        return np.linalg.svd(a, full_matrices=full_matrices, compute_uv=compute_uv)
    except np.linalg.LinAlgError:
        pass

    from scipy.linalg import svd
    return svd(a, full_matrices=full_matrices, compute_uv=compute_uv, lapack_driver='gesvd')
```

And then we just replace the above with:
```suggestion
            u, _, _ = fallback_svd(M, full_matrices=False)
```

effigies: And to be clear, I'm suggesting we place the above `fallback_svd` function at the file level.

chrisfilo:
Great contribution @feilong! Thanks for getting to the bottom of this!

nipype/algorithms/confounds.py
feilong: Great idea! By putting it at the file level do you mean putting it at the top of the file?

effigies: Yes. Anywhere, really, but not nested within a class or another function. The top is fine.

nipype/algorithms/confounds.py
```diff
@@ -14,6 +14,7 @@
 import nibabel as nb
 import numpy as np
 from numpy.polynomial import Legendre
+from scipy import linalg
```
effigies: Unneeded import.

feilong: Thanks for pointing that out!

feilong:
Great!  Thanks for making these great softwares.  Looking forward to more commits!

PR #2846, opened 2019-01-11
BF: ANTS LaplacianThickness cmdline opts fixed up

yarikoptic:
That interface had a chance to work only when no any optional options specified since cmdline specs.
I also stripped  opt_ prefix from tollerance since they are all optional.

Some related discussions/changes on ANTs:
- https://github.com/ANTsX/ANTs/issues/688
- https://github.com/ANTsX/ANTs/pull/690

### DIFF
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -207,13 +207,27 @@ class LaplacianThicknessInputSpec(ANTSCommandInputSpec):
         name_template='%s_thickness',
         keep_extension=True,
         hash_files=False)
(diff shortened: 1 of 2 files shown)


effigies:
This is a change based on a new version of ANTs? We should then set minimum/maximum versions or in some way account for version differences.

yarikoptic:
no, I think it is for the "forever version of ANTs" which was actually never tested to work correctly via nipype.  ... I even double checked by looking at 98188c3927a69c775cd1ac3f1ec6e41a0fd5a9d4 of ANTS from 2009 -- the same logic in handling the cmdline args

effigies:
Please merge master to fix tests.

PR #2853, opened 2019-01-14
Revert "RF: Move pytest and pytest-xdist from general requirement into tests_required"

effigies:
Testing whether this somehow introduced the current test failure.

Reverts nipy/nipype#2850

PR #2850, opened 2019-01-12
RF: Move pytest and pytest-xdist from general requirement into tests_required

yarikoptic:
not really sure why this didn't give me troubles before but it does now:

```shell
reproin@59965c6fd52f:/$ /usr/bin/heudiconv
Traceback (most recent call last):
  File "/usr/bin/heudiconv", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3019, in <module>
    @_call_aside
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3003, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3032, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 657, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 670, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 849, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'pytest>=3.6' distribution was not found and is required by nipype
```
- I do not think pytest must be a runtime requirement
- I do not know if there is any negative side-effect (e.g. some unguarded import of nipype at runtime)

PR #2827, opened 2018-12-14
[FIX] Update pytest req'd version to 3.6

oesteban:
Fixes #2826.

## Summary

Fixes # .

## Acknowledgment

- [x] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

PR #2854, opened 2019-01-14
RF: do not require testing harnesses for regular usage

yarikoptic:
continuation to #2850 , I guess awaiting on verdict from #2853  as well


yarikoptic:
additional note/motivation is that pytest >= 3.6 cannot even be satisfied on stable Debian/Ubuntus (e.g. debian stretch/stable has 3.0.6).

yarikoptic:
@oesteban do you remember "why?" (ref: a4dfe7e6a2ec78ba9cea20005799001f16ddcb99):
```
commit a4dfe7e6a2ec78ba9cea20005799001f16ddcb99 (origin/oesteban-patch-2)
Author: Oscar Esteban <code@oscaresteban.es>
Date:   Thu Dec 13 16:41:55 2018 -0800

    [FIX] Update pytest req'd version to 3.6
    
    Fixes #2826.

diff --git a/nipype/info.py b/nipype/info.py
index 1cf361c40..be6edb713 100644
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -108,7 +108,7 @@ NUMPY_MIN_VERSION_37 = '1.15.3'
 SCIPY_MIN_VERSION = '0.14'
 TRAITS_MIN_VERSION = '4.6'
 DATEUTIL_MIN_VERSION = '2.2'
-PYTEST_MIN_VERSION = '3.0'
+PYTEST_MIN_VERSION = '3.6'
 FUTURE_MIN_VERSION = '0.16.0'
 SIMPLEJSON_MIN_VERSION = '3.8.0'
 PROV_VERSION = '1.5.2'
```
seems to run tests just fine on debian stretch with 3.0.6-1 .

yarikoptic:
ah, @oesteban - please disregard my above questsion -- did not spot https://github.com/nipy/nipype/issues/2826  which somewhat explains the reason.  I think for the Debian package I would just eliminate any versioning of the dependency since it would install most recent available, and fail if fails to build/test

effigies:
Okay. So we may wish to drop that requirement, but that means either setting a max version on xdist, which itself set a minimum version of 3.6 (see #2826, #2827) or dropping the requirement for xdist, which will in turn require removing the `-n auto` from the [`pytest.ini`](https://github.com/nipy/nipype/blob/master/nipype/pytest.ini). At which point, we'd want to instruct Travis and Circle to install `pytest-xdist` separately, and add `-n auto` for the sake of speed.

I'm fine with all of this. It's just a little rabbit hole.

yarikoptic:
Thank you @effigies !  whatever you feel works ok for you is ok with me  -- at least I can workaround.

PR #2855, opened 2019-01-14
MAINT: Disable numpy 1.16.0 for Py2.7

effigies:
## Summary
Another attempt to fix Python 2.7 test failures. Numpy released 1.16.0 yesterday.

## Acknowledgment


effigies:
Yup, this seems to be it. I thought our `--pre` checks should have caught this, but IDK what all has changed in numpy.

effigies:
Looks like numpy/numpy#12157 is the culprit, which replaced checks that used `isinstance(..., basestring)` with something that doesn't accept `unicode` objects. I'm not sure if we can get them to re-allow unicode, so perhaps we'll just need to make sure that the test in question sends a `str` object.

effigies:
Upstreamed to numpy/numpy#12749, but we should probably go ahead and just fix up the test to not pass a `unicode` path in 2.7.

Issue #2844, opened 2019-01-10
2 bugs in SpecifySPMModel with concatenate_runs = True?


hstojic:
### Summary

I get an error defining SPM model level 1, with multiple runs and concatenate_runs set to True. I made sure that I have the same regressors in each run. The culprit seems to be the nipype code that adds run specific dummy variables [(code)](https://github.com/nipy/nipype/blob/dcc013b0187bf5312db9174f8cb8932991c601cb/nipype/algorithms/modelgen.py#565) - regressor_names are not updated with these added variables and then error is raised at this [line](https://github.com/nipy/nipype/blob/dcc013b0187bf5312db9174f8cb8932991c601cb/nipype/algorithms/modelgen.py#L366) as of course it cannot find the regressor name. 

Here is the error part from stdout:
```python
[Node] Error on "stimuli_LSA_test_pipeline.model" (/media/hstojic/dataneuro/fnclearning_fmri/dProcessed/nipype_work/stimuli_LSA_test_pipeline/_subject_s057/model)
[Node] Error on "stimuli_LSA_test_pipeline.model" (/media/hstojic/dataneuro/fnclearning_fmri/dProcessed/nipype_work/stimuli_LSA_test_pipeline/_subject_s057/model)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-a5adcf3b88b5> in <module>()
      2 pipeline.run(
      3     'MultiProc',
----> 4     plugin_args = {'n_procs': pars_gen['resources']['n_cores']}
      5 )

/home/hstojic/.pyenv/nipy/local/lib/python2.7/site-packages/nipype/pipeline/engine/workflows.pyc in run(self, plugin, plugin_args, updatehash)
    593         if str2bool(self.config['execution']['create_report']):
    594             self._write_report_info(self.base_dir, self.name, execgraph)
--> 595         runner.run(execgraph, updatehash=updatehash, config=self.config)
    596         datestr = datetime.utcnow().strftime('%Y%m%dT%H%M%S')
    597         if str2bool(self.config['execution']['write_provenance']):

/home/hstojic/.pyenv/nipy/local/lib/python2.7/site-packages/nipype/pipeline/plugins/base.pyc in run(self, graph, config, updatehash)
    160                         if result['traceback']:
    161                             notrun.append(
--> 162                                 self._clean_queue(jobid, graph, result=result))
    163                         else:
    164                             self._task_finished_cb(jobid)

[...]

PR #2857, opened 2019-01-18
FIX: correct bug in _concatenate_info function

hstojic:
This is related to issue #2844 that I raised, more details is provided there. 

In summary, using `concatenate_runs = True` results in an error as regressor names are missing for session dummy variables, this one-line corrects it by adding a session regressor name to the regressor_names variable..

If my interpretation is correct, its an easy fix with a single line I think:
`infoout.regressor_names.extend(['run' + str(i + 1)])`


effigies (dismissed):
This looks sensible. I think it could be written slightly more cleanly. And can you verify that this produces the results you expect?

And do you want to add yourself to the [`.zenodo.json`](https://github.com/nipy/nipype/blob/master/.zenodo.json) file?

hstojic:
Yes, this produces the results I expect, I already use it in my local code.

I will add myself to the zenodo file, thanks!

effigies:
Just noticed the failing tests. Looks like `infoout.regressor_names` doesn't exist to extend. If you look at lines 563-564, you need to create the list first. I'm a little curious why it works for you, then...

I'll try to have a deeper look tomorrow.

effigies:
Also, can you go ahead and merge/cherry-pick 4158583 into this PR instead of having a separate one?

hstojic:
Ok, silly mistake, `regressors` is an optional argument and so is `regressor_names` then. In my case I had nuisance regressors and their names so it never complained. Tests for sure have cases without any nuisance covariate. should be Ok by just adding creating the list few lines above.

satra:
@hstojic  and @effigies - since this creates a patch without a broken test, it would be great to have a broken test as well. i just asked @hstojic to add that in the original issue.

hstojic:
@satra and @effigies I have been a bit tight on time these few days to add a test and resolve the issue before the new release, it will have to wait for the new one, I'll try to do it soonish

effigies:
No worries. When you get some time.

effigies:
Hi @hstojic. Just a bump to let you know we'll be aiming to have everything in for the 1.1.9 release by next Friday, Feb 22. Let us know if you need any help on this.

PR #2859, opened 2019-01-19
[ENH] CompCor enhancement

rciric:
## Summary

As a step toward implementing poldracklab/fmriprep#1458, this PR updates the CompCor algorithm to allow greater flexibility.

Given that I would like to incorporate features enabled by these changes into `fmriprep`, does it make sense to propose this separately as a patch in `niworkflows`?

* As before, `num_components` encodes the number of components to return if it takes an integer value greater than 1.
* If `num_components` is `all`, then CompCor returns all component time series.
* Alternatively, a mutually exclusive interface option, `variance_threshold`, allows for automatic selection of components on the basis of their capacity to explain variance in the data. `variance_threshold` is mutually exclusive with `num_components` and takes values between 0 and 1. If `variance_threshold` is set, CompCor returns the minimum number of components necessary to explain the indicated variance in the data, computed as the cumulative sum of (singular_value^2)/sum(singular_values^2).
* For backward compatibility, if neither `num_components` nor `variance_threshold` is defined, then CompCor executes as though `num_components` were set to 6 (the previous default behaviour).
* CompCor optionally writes out a metadata table, which includes for each component the source mask, the singular value, the explained variance, and the cumulative explained variance.

## Acknowledgment

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -271,7 +271,7 @@ class FramewiseDisplacement(BaseInterface):
     .. [Power2012] Power et al., Spurious but systematic correlations in functional
          connectivity MRI networks arise from subject motion, NeuroImage 59(3),
          2012. doi:`10.1016/j.neuroimage.2011.10.018
-         <http://dx.doi.org/10.1016/j.neuroimage.2011.10.018>`_.
+         <https://doi.org/10.1016/j.neuroimage.2011.10.018>`_.
 
 
     """
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -215,7 +215,7 @@ this way you can use the initial paper published in 2011 (see below).
 	month            = "08",
 	doi              = "10.3389/fninf.2011.00013",
 	pubmed           = "21897815",
-	url              = "http://dx.doi.org/10.3389/fninf.2011.00013",
+	url              = "https://doi.org/10.3389/fninf.2011.00013",
 	issn             = "1662-5196"}
 
 
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -1587,7 +1587,7 @@ class KellyKapowski(ANTSCommand):
             "year={2009},"
             "issn={1053-8119},"
             "url={http://www.sciencedirect.com/science/article/pii/S1053811908012780},"
-            "doi={http://dx.doi.org/10.1016/j.neuroimage.2008.12.016}"
+            "doi={https://doi.org/10.1016/j.neuroimage.2008.12.016}"
             "}"),
         'description':
         'The details on the implementation of DiReCT.',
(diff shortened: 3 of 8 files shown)


effigies:
I think it's reasonable to put this here. I have a couple overall comments about the approach, though:

> If num_components is valued between 0 and 1, then CompCor instead returns the minimum number of components necessary to explain the indicated variance in the data, computed as the cumulative sum of (singular_value^2)/sum(singular_values^2).

I would rather not overload this, as it's not an intuitive interpretation of "num_components". I would rather have a separate input like `variance_threshold` or some other descriptive name to enable this behavior. You can add `xor=['num_components']` to the trait to ensure that people don't specify both.

> If num_components is -1, then CompCor returns all component time series.

What about `num_components = 'all'`? You can make the trait:

```Python
num_components = traits.Either('all', traits.Int, ...)
```

rciric:
Thank you for the review — I’ve now updated the PR. In the update, I automatically instantiate the selection criterion equivalently to `num_components = 6` if neither `num_components` nor `variance_threshold` is defined (the current default behaviour, for backward compatibility) and raise a warning (perhaps this should be a `FutureWarning` or `DeprecationWarning`); I’m not sure what the best way of handling this is.

nipype/algorithms/confounds.py
```diff
@@ -613,13 +663,19 @@ def _list_outputs(self):
                 save_pre_filter = os.path.abspath('pre_filter.tsv')
             outputs['pre_filter_file'] = save_pre_filter
 
+        save_metadata = self.inputs.save_metadata
+        if save_metadata:
```
oesteban:
```suggestion
        if save_metadata is True:
```

nipype/algorithms/confounds.py
```diff
@@ -613,13 +663,19 @@ def _list_outputs(self):
...
+            if isinstance(save_metadata, bool):
```
oesteban: remove this line

rciric: Hmm, looking at it now, I'm not sure that change would work because the conditional branch should be entered if `save_metadata` is either `True` or a string (the metadata file path) so as to set the output path.
```python
        if save_metadata:
            if isinstance(save_metadata, bool):
                save_metadata = os.path.abspath('component_metadata.tsv')
            outputs['metadata_file'] = save_metadata
```
Of course, we could circumvent this entirely with a `SimpleInterface` as per your earlier suggestion. I'm not sure if there's a good reason for not using `SimpleInterface` or `fname_presuffix` here (perhaps this function predates their integration into `nipype`?), but I mirrored this block structure from the `save_pre_filter` block above it. I could definitely try and see if I can refactor into a `SimpleInterface` with `fname_presuffix` if you think it would be worthwhile.

rciric:
There are a couple of potential changes here that I think could be worthwhile, both in general and downstream in `fmriprep`:
* Save all component metadata regardless of the number of components that are retained.
* Automatically remove zero-variance/zero-SV columns from the decomposition.

PR #2860, opened 2019-01-20
Update .zenodo.json

hstojic:
Related to the fix in #2857


hstojic:
moved the update of the file to the original pull request in #2857

PR #2845, opened 2019-01-10
ENH: Support for BIDS event files

TheChymera:
SpecifyModel now supports BIDS event files as inputs.

## Acknowledgment

- [x] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

PR #2867, opened 2019-01-30
FIX: Sort conditions in bids_gen_info to ensure consistent order

TheChymera:
@effigies 
If this is not done, condition order for orthogonalizarion is
unpredictable.

## Acknowledgment


nipype/algorithms/modelgen.py
```diff
@@ -176,6 +176,7 @@ def bids_gen_info(bids_event_files,
             f_events = csv.DictReader(f, skipinitialspace=True, delimiter='\t')
             events = [{k: v for k, v in row.items()} for row in f_events]
         conditions = list(set([i[condition_column] for i in events]))
```
effigies: Just to condense this into one line:

```suggestion
        conditions = sorted(set([i[condition_column] for i in events]))
```

satra:
@effigies  and @TheChymera - yes, it may be useful to see a point of failure where this will not work. i'm mostly concerned whether this worked and stopped working or did not work properly ever.

also whether this works across both FSL and SPM.

effigies:
This was recently added (#2845) as a BIDS-y alternative to the existing events files, so it's not a failure of a historical construct.

effigies:
@TheChymera Can you review the suggestion? And is there any chance of a small regression test?

effigies:
@TheChymera Just a note that the 1.1.9 release is coming up on Monday, and it would be good to have any PRs we want to include in on Friday.

effigies:
In this case, a simple regression test would be a BIDS-style event file and a function that compares the output of `bids_gen_info` to an expected output:

```Python
def test_bids_gen_info():
    fname = ...
    res = bids_gen_info(fname)
    # Tests of properties
    assert len(res) == 1
    assert res[0].onsets == ...
    assert res[0].durations == ...
    assert res[0].amplitudes == ...
    assert res[0].conditions == ...
```

TheChymera:
@effigies what directory should I put the example file into? I can't find other example files, e.g. `test1.nii` anywhere.

satra:
@TheChymera - if you create a test with:

```python
def test_bids_info(tmpdir):
    filename = tmpdir / 'testfile.nii'
    import nibabel as nb
    # create a local file here and use it for a test.
    ....
    img.to_nifti(filename)
    # use filename for test
```

TheChymera:
@satra so I should create the .tsv file as part of the test? that's a lot more and a lot less legible text than if I tracked it somewhere. But where? `examples/`? In the test dir?

satra:
the general place where test data go is `testing/data` but most files there are empty. so you can put small files, but not large ones.

TheChymera:
@satra @effigies works over here, hope this is good now :)

TheChymera:
@satra @effigies it works on my box locally, but I can't get the path to work on circleci. Any advice?

nipype/algorithms/tests/test_modelgen.py
```diff
@@ -12,10 +12,18 @@
...
 import numpy.testing as npt
 from nipype.interfaces.base import Bunch, TraitError
-from nipype.algorithms.modelgen import (SpecifyModel, SpecifySparseModel,
-                                        SpecifySPMModel)
+from nipype.algorithms.modelgen import (bids_gen_info, SpecifyModel,
+                                        SpecifySparseModel, SpecifySPMModel)
 
 
+def test_bids_gen_info():
+    fname = os.path.realpath('../../testing/data/events.tsv')
```
satra:
```suggestion
    fname = os.path.realpath('data/events.tsv')
```

TheChymera: ah, I see I had been running the tests from another directory..

TheChymera: I still get
```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-neuro/pytest-0/popen-gw0/test_warning0/data/events.tsv'
```

nipype/algorithms/tests/test_modelgen.py
```diff
@@ -12,10 +12,18 @@
...
 import numpy.testing as npt
 from nipype.interfaces.base import Bunch, TraitError
-from nipype.algorithms.modelgen import (SpecifyModel, SpecifySparseModel,
-                                        SpecifySPMModel)
+from nipype.algorithms.modelgen import (bids_gen_info, SpecifyModel,
+                                        SpecifySparseModel, SpecifySPMModel)
 
 
+def test_bids_gen_info():
+    fname = os.path.realpath('data/events.tsv')
```
satra:
```suggestion
    fname = npt.example_data('events.tsv')
```

satra: sorry i forgot that location is not a default in the regular tests.

nipype/algorithms/modelgen.py
```diff
@@ -175,7 +175,11 @@ def bids_gen_info(bids_event_files,
         with open(bids_event_file) as f:
             f_events = csv.DictReader(f, skipinitialspace=True, delimiter='\t')
             events = [{k: v for k, v in row.items()} for row in f_events]
-        conditions = list(set([i[condition_column] for i in events]))
+        if not condition_column:
+            condition_column = '_trial_type'
+            for i in events:
+                i.update({condition_column:'ev0'})
```
effigies:
```suggestion
                i.update({condition_column: 'ev0'})
```

nipype/algorithms/modelgen.py
```diff
@@ -199,7 +200,6 @@ def bids_gen_info(bids_event_files,
         info.append(runinfo)
     return info
 
-
```
effigies: Style guides want two newlines between functions.

nipype/algorithms/tests/test_modelgen.py
TheChymera: @satra still not working :-/

TheChymera:
Added separate import because @satra's suggestion failed with

```
def test_bids_gen_info():
>       fname = npt.example_data('events.tsv')
E       AttributeError: module 'numpy.testing' has no attribute 'example_data'

/src/nipype/nipype/algorithms/tests/test_modelgen.py:20: AttributeError
```

TheChymera:
Ok, can't reproduce this new failure locally, sounds like some unicode support is missing from the testing environment:

```
def test_bids_gen_info():
        fname = example_data('events.tsv')
>       res = bids_gen_info([fname])

/src/nipype/nipype/algorithms/tests/test_modelgen.py:22: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/src/nipype/nipype/algorithms/modelgen.py:177: in bids_gen_info
    f_events = csv.DictReader(f, skipinitialspace=True, delimiter='\t')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <csv.DictReader instance at 0x7f4a4d694d88>
f = <closed file '/src/nipype/nipype/testing/data/events.tsv', mode 'r' at 0x7f4a4d943300>
fieldnames = None, restkey = None, restval = None, dialect = 'excel', args = ()
kwds = {'delimiter': '	', 'skipinitialspace': True}

    def __init__(self, f, fieldnames=None, restkey=None, restval=None,
                 dialect="excel", *args, **kwds):
        self._fieldnames = fieldnames   # list of keys for the dict
        self.restkey = restkey          # key to catch long rows
        self.restval = restval          # default value for short rows
>       self.reader = reader(f, dialect, *args, **kwds)
E       TypeError: "delimiter" must be string, not unicode

/opt/miniconda-latest/envs/neuro/lib/python2.7/csv.py:79: TypeError
```

nipype/algorithms/modelgen.py
```diff
@@ -175,7 +176,11 @@ def bids_gen_info(bids_event_files,
         with open(bids_event_file) as f:
             f_events = csv.DictReader(f, skipinitialspace=True, delimiter='\t')
```
effigies: This should resolve the Python 2/Unicode issue.

```suggestion
            f_events = csv.DictReader(f, skipinitialspace=True, delimiter=str('\t'))
```

TheChymera:
@effigies 

ok, so using the str type from builtins gives me this:

```
E       TypeError: "delimiter" must be string, not newstr
```

The only way I was able to fix it is by re-assigning the base type before the import, since removing the import breaks another function.

effigies:
Going to go ahead and merge. Thanks for your patience!

PR #2871, opened 2019-01-31
ENH: Add workbench.CiftiSmooth interface

mgxd:

* Adds `-cifti-smoothing` command from `wb_command`
## Acknowledgment


nipype/interfaces/workbench/cifti.py
```diff
@@ -0,0 +1,143 @@
...
+        desc="The output CIFTI")
+    left_surf = File(
+        exists=True,
+        mandatory=True,
+        position=5,
+        argstr="-left-surface %s",
+        desc="Specify the left surface to use")
+    left_corrected_areas = File(
+        exists=True,
+        position=6,
```
effigies: Are all these flagged arguments truly positional?

mgxd: i think not, but was staying true to how it appears to be written up
https://github.com/Washington-University/workbench/blob/683ce2e9658ad55a926f0dda947d4537c0360d78/src/Algorithms/AlgorithmCiftiSmoothing.cxx#L54

PR #2878, opened 2019-02-09
[ENH] CompCor enhancement

rciric:
## Summary

Continuing #2859 

As a step toward implementing poldracklab/fmriprep#1458, this PR updates the CompCor algorithm to allow greater flexibility.

* As before, `num_components` encodes the number of components to return if it takes an integer value greater than 1.
* If `num_components` is `all`, then CompCor returns all component time series.
* Alternatively, a mutually exclusive interface option, `variance_threshold`, allows for automatic selection of components on the basis of their capacity to explain variance in the data. `variance_threshold` is mutually exclusive with `num_components` and takes values between 0 and 1. 
* If `variance_threshold` is set, CompCor returns the minimum number of components necessary to explain the indicated variance in the data, computed as the cumulative sum of (singular_value^2)/sum(singular_values^2).
* For backward compatibility, if neither `num_components` nor `variance_threshold` is defined, then CompCor executes as though `num_components` were set to 6 (the previous default behaviour).
* CompCor optionally writes out a metadata table, which includes for each component the source mask, the singular value, the explained variance, and the cumulative explained variance.

## Acknowledgment


nipype/algorithms/confounds.py
```diff
@@ -8,6 +8,11 @@
                         absolute_import)
 from builtins import range
 
+# Py2 compat: http://python-future.org/compatible_idioms.html#collections-counter-and-ordereddict
```
oesteban: We don't support python 2.6, lines 11-13 are not necessary

nipype/algorithms/confounds.py
```diff
@@ -561,16 +602,25 @@ def _run_interface(self, runtime):
...
-            header=self._make_headers(components.shape[1]),
+            header='\t'.join(components_header),
             comments='')
+        self._results['components_file'] = os.path.abspath(
+            self.inputs.components_file)
 
-        if self.inputs.pre_filter and self.inputs.save_pre_filter:
-            pre_filter_file = self._list_outputs()['pre_filter_file']
+        save_pre_filter = self.inputs.save_pre_filter
+        if save_pre_filter:
```
oesteban: lines 617-648 can be simpler:

```Python
        if save_pre_filter:
            self._results['pre_filter_file'] = save_pre_filter
            if save_pre_filter is True:
                self._results['pre_filter_file'] = os.path.abspath('pre_filter.tsv')
            <lines 624-648 at this level of indentation>
```

Also replace that `pre_filter_file` with `self._results['pre_filter_file']` in line 643.

nipype/algorithms/confounds.py
```diff
@@ -561,16 +602,25 @@ def _run_interface(self, runtime):
...
             delimiter='\t',
-            header=self._make_headers(components.shape[1]),
+            header='\t'.join(components_header),
             comments='')
+        self._results['components_file'] = os.path.abspath(
+            self.inputs.components_file)
 
-        if self.inputs.pre_filter and self.inputs.save_pre_filter:
-            pre_filter_file = self._list_outputs()['pre_filter_file']
+        save_pre_filter = self.inputs.save_pre_filter
```
oesteban: I can't follow very well the block starting in line 623, wouldn't this be equivalent?
```suggestion
        save_pre_filter = False
        if self.inputs.pre_filter in ['polynomial', 'cosine']:
            save_pre_filter = self.inputs.save_pre_filter
```

nipype/algorithms/confounds.py
```diff
@@ -597,29 +647,37 @@ def _run_interface(self, runtime):
                 header='\t'.join(header),
                 comments='')
 
+        save_metadata = self.inputs.save_metadata
```
oesteban: Lines 650-655:

```Python
        metadata_file = self.inputs.save_metadata
        if metadata_file:
            if metadata_file is True:
                metadata_file = os.path.join(runtime.cwd, 'component_metadata.tsv')
```

nipype/algorithms/confounds.py
```diff
@@ -597,29 +647,37 @@ def _run_interface(self, runtime):
...
+        if save_metadata:
+            if isinstance(save_metadata, bool):
+                metadata_file = os.path.abspath('component_metadata.tsv')
+            else:
+                metadata_file = save_metadata
+            components_names = np.empty(len(metadata['mask']),
+                dtype='object_')
+            retained = np.where(metadata['retained'])
```
oesteban: I think we would save a lot of trouble using pandas here. Maybe @effigies or @satra correct me, but I'm a bit scared of all this massaging here.

nipype/algorithms/confounds.py
```diff
@@ -1186,35 +1267,72 @@ def compute_noise_components(imgseries, mask_images, num_components,
                 voxel_timecourses, repetition_time, period_cut)
         elif filter_type in ('polynomial', False):
             # from paper:
-            # "The constant and linear trends of the columns in the matrix M were
-            # removed [prior to ...]"
+            # "The constant and linear trends of the columns in the matrix M
```
oesteban: Please roll back these stylistic changes, they divert the attention during review.

nipype/algorithms/confounds.py
```diff
@@ -1139,15 +1197,30 @@ def combine_mask_files(mask_files, mask_method=None, mask_index=None):
         return [img]
 
 
-def compute_noise_components(imgseries, mask_images, num_components,
-                             filter_type, degree, period_cut, repetition_time):
+def compute_noise_components(imgseries, mask_images, components_criterion=0.5,
+                             filter_type=False, degree=0, period_cut=128,
+                             repetition_time=None, failure_mode='error',
+                             mask_names=''):
```
oesteban:
```suggestion
                             mask_names=None):
```

nipype/algorithms/confounds.py
```diff
@@ -1158,16 +1231,24 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
+        Dictionary of eigenvalues, fractional explained variances, and
+        cumulative explained variances.
     """
     components = None
     basis = np.array([])
-    for img in mask_images:
-        mask = img.get_data().astype(np.bool)
+    if components_criterion == 'all':
+        components_criterion = -1
+    if not mask_names:
```
oesteban: These two lines:
```Python
    if not mask_names:
        mask_names = range(len(mask_images))
```
Are more readable like this:
```Python
    mask_names = mask_names or range(len(mask_images))
```

nipype/algorithms/confounds.py
```diff
@@ -1158,16 +1231,24 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
+        mask_names = range(len(mask_images))
+    for i, img in zip(mask_names, mask_images):
```
oesteban:
```suggestion
    for name, img in zip(mask_names, mask_images):
```
Then fix subsequent appearances of ``i`` later on. You overload it in the end, I wonder if that is what is confusing Python 2.7.

nipype/algorithms/confounds.py
```diff
@@ -1186,35 +1267,72 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
+        elif components_criterion == -1:
+            num_components = len(s)
+
+        num_components = int(num_components)
+        if num_components == 0:
+            break
         if components is None:
             components = u[:, :num_components]
+            metadata = OrderedDict()
+            metadata['mask'] = [i] * len(s)
```
oesteban:
```suggestion
            metadata['mask'] = [name] * len(s)
```

nipype/algorithms/confounds.py
```diff
@@ -1186,35 +1267,72 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
+            metadata['singular_value'] = s
+            metadata['variance_explained'] = variance_explained
+            metadata['cumulative_variance_explained'] = (
+                cumulative_variance_explained)
+            metadata['retained'] = [i < num_components for i in range(len(s))]
         else:
             components = np.hstack((components, u[:, :num_components]))
-    if components is None and num_components > 0:
-        if self.inputs.failure_mode == 'error':
+            metadata['mask'] = metadata['mask'] + [i] * len(s)
```
oesteban:
```suggestion
            metadata['mask'] = metadata['mask'] + [name] * len(s)
```

nipype/algorithms/confounds.py
```diff
@@ -418,6 +439,8 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
         'unspecified')
     save_pre_filter = traits.Either(
```
oesteban: These two should have a default value of `False`

nipype/algorithms/confounds.py
```diff
@@ -418,6 +439,8 @@ class CompCorInputSpec(BaseInterfaceInputSpec):
...
         traits.Bool, File, desc='Save pre-filter basis as text file')
+    save_metadata = traits.Either(
```
oesteban: default

effigies:
@oesteban @rciric Is this going in for 1.1.9? Do you need another set of eyes, or are you comfortable merging as is?

oesteban:
This would be great to have in 1.1.9. Depending on your bandwidth, another pair of eyes might preempt undesired side effects.

effigies:
Do we have an open branch of fMRIPrep where we've checked the outputs look sensible and don't crash?

oesteban:
We have one branch now - https://circleci.com/workflow-run/83b5965a-0f30-470b-8329-1bc5f66f3a13

Let's see how that looks

effigies:
```
190221-05:46:26,246 nipype.workflow ERROR:
	 Node acompcor failed to run on host 77ead578c84d.
190221-05:46:26,252 nipype.workflow ERROR:
	 Saving crash info to /out/fmriprep/sub-01/log/20190221-045816_251af727-76d1-4842-963c-fc64b6657462/crash-20190221-054626-root-acompcor-4c4d676f-be11-4d3d-a0d6-fea6fce37920.txt
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/legacymultiproc.py", line 69, in run_node
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 473, in run
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 557, in _run_interface
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 637, in _run_command
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 369, in run
  File "/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/patches.py", line 28, in _run_interface
    runtime = super(RobustACompCor, self)._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/algorithms/confounds.py", line 594, in _run_interface
    self.inputs.failure_mode, self.inputs.mask_names)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/algorithms/confounds.py", line 1266, in compute_noise_components
    voxel_timecourses, repetition_time, period_cut)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/algorithms/confounds.py", line 1094, in cosine_filter
    betas = np.linalg.lstsq(X, data.T)[0]
  File "/usr/local/miniconda/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 2125, in lstsq
    _assertNoEmpty2d(a, b)  # TODO: relax this constraint
  File "/usr/local/miniconda/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 229, in _assertNoEmpty2d
    raise LinAlgError("Arrays cannot be empty")
numpy.linalg.linalg.LinAlgError: Arrays cannot be empty
```

nipype/algorithms/confounds.py
```diff
@@ -1158,16 +1231,23 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
     """
     components = None
     basis = np.array([])
-    for img in mask_images:
-        mask = img.get_data().astype(np.bool)
+    if components_criterion == 'all':
+        components_criterion = -1
+    mask_names = mask_names or range(len(mask_images))
+    for name, img in zip(mask_names, mask_images):
+        mask = img.get_data().astype(np.bool).squeeze()
```
effigies: The main thing I can see that might produce this issue is the addition of `.squeeze()` here. What's the justification?

rciric: Thanks for catching this -- just ran through my test examples with the `squeeze` removed, and none of them gave any hiccups. Reverted.

nipype/algorithms/confounds.py
```diff
@@ -1139,15 +1197,30 @@ def combine_mask_files(mask_files, mask_method=None, mask_index=None):
         return [img]
 
 
-def compute_noise_components(imgseries, mask_images, num_components,
-                             filter_type, degree, period_cut, repetition_time):
+def compute_noise_components(imgseries, mask_images, components_criterion=0.5,
+                             filter_type=False, degree=0, period_cut=128,
+                             repetition_time=None, failure_mode='error',
+                             mask_names=None):
```
effigies: `failure_mode` and `mask_names` are not documented in the docstring.

nipype/algorithms/confounds.py
```diff
@@ -1158,15 +1231,22 @@ def compute_noise_components(imgseries, mask_images, num_components,
     period_cut: minimum period (in sec) for DCT high-pass filter
     repetition_time: time (in sec) between volume acquisitions
```
effigies: Maybe note that repetition_time must be defined if `filter_type` is `'cosine'`?

Similarly, it would be good to have a check for that and raise an informative exception.

nipype/algorithms/confounds.py
```diff
@@ -1201,20 +1281,56 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
-            u, _, _ = fallback_svd(M, full_matrices=False)
+            u, s, _ = fallback_svd(M, full_matrices=False)
         except np.linalg.LinAlgError:
-            if self.inputs.failure_mode == 'error':
+            if failure_mode == 'error':
                 raise
-            u = np.ones((M.shape[0], num_components), dtype=np.float32) * np.nan
+            if components_criterion >= 1:
+                u = np.empty((M.shape[0], components_criterion),
+                             dtype=np.float32) * np.nan
```
effigies: To accommodate above suggestion:

```suggestion
```

Also `s` is not defined in this case.

nipype/algorithms/confounds.py
```diff
@@ -1201,20 +1281,56 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
+                np.hstack((metadata['cumulative_variance_explained'],
+                           cumulative_variance_explained)))
+            metadata['retained'] = (
+                metadata['retained'] + [i < num_components for i in range(len(s))])
+    if components is None:
+        if failure_mode == 'error':
             raise ValueError('No components found')
-        components = np.ones((M.shape[0], num_components), dtype=np.float32) * np.nan
-    return components, basis
+        components = np.full((M.shape[0], num_components), np.NaN)
```
effigies:
```suggestion
        components = np.full((M.shape[0], num_components), np.nan)
```

See [np.NaN](https://docs.scipy.org/doc/numpy/reference/constants.html?highlight=nan#numpy.NaN).

nipype/algorithms/confounds.py
```diff
@@ -1201,20 +1281,56 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
         try:
-            u, _, _ = fallback_svd(M, full_matrices=False)
+            u, s, _ = fallback_svd(M, full_matrices=False)
         except np.linalg.LinAlgError:
-            if self.inputs.failure_mode == 'error':
+            if failure_mode == 'error':
                 raise
-            u = np.ones((M.shape[0], num_components), dtype=np.float32) * np.nan
+            if components_criterion >= 1:
+                u = np.empty((M.shape[0], components_criterion),
```
effigies:
```suggestion
                u = np.full((M.shape[0], components_criterion), np.nan, dtype=np.float32)
```

nipype/workflows/rsfmri/fsl/tests/test_resting.py
```diff
@@ -89,16 +89,19 @@ def test_create_resting_preproc(self, mock_node, mock_realign_wf):
...
+        if not (num_got_components == self.num_noise_components or
+                num_got_components == self.fake_data.shape[3]):
+            raise AssertionError()
+        first_two = [row[:2] for row in components_data[1:]]
+        if first_two != [['-0.5172356654', '-0.6973053243'],
+                         ['0.2574722644', '0.1645270737'],
+                         ['-0.0806469590', '0.5156853779'],
+                         ['0.7187176051', '-0.3235820287'],
+                         ['-0.3783072450', '0.3406749013']]:
+            raise AssertionError()
```
effigies: I think all of this can be reverted. What's the motivation here? It makes sense to be keep only the `components_data` line inside the `with` block, but all of the rest of these changes don't make much sense to me.

nipype/workflows/rsfmri/fsl/resting.py
```diff
@@ -3,7 +3,6 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 from __future__ import (print_function, division, unicode_literals,
                         absolute_import)
-from builtins import str
```
effigies: Why?

nipype/algorithms/confounds.py
effigies: Out-of-order. ["Above suggestion"](https://github.com/nipy/nipype/pull/2878#discussion_r259085346).

nipype/workflows/rsfmri/fsl/resting.py
oesteban: That was an unnecessary import, I think.

nipype/workflows/rsfmri/fsl/tests/test_resting.py
rciric: The motivation was that [codacy complained about the use of `assert` -- see 'fixed issues' tab](https://app.codacy.com/app/nipype/nipype/pullRequest?prid=3061026). I'll go ahead and revert and see if it raises the same concern again.

effigies: Yeah, codacy is not turning out to be very helpful, IMO. Whatever general policy one might have about assertions, they're completely reasonable to have in tests.

nipype/workflows/rsfmri/fsl/tests/test_resting.py
```diff
@@ -89,16 +89,17 @@ def test_create_resting_preproc(self, mock_node, mock_realign_wf):
...
-            assert (num_got_components == self.num_noise_components
-                    or num_got_components == self.fake_data.shape[3])
-            first_two = [row[:2] for row in components_data[1:]]
-            assert first_two == [['-0.5172356654', '-0.6973053243'], [
-                '0.2574722644', '0.1645270737'
-            ], ['-0.0806469590',
-                '0.5156853779'], ['0.7187176051', '-0.3235820287'],
-                                 ['-0.3783072450', '0.3406749013']]
+            components_data = [line.split()
+                               for line in components_file.read().splitlines()]
```
effigies: Sorry, why is this not just `components_data = [line.split() for line in components_file]`? Iterating over a file should give you lines. If the problem is trailing newlines:

```Python
components_data = [line.rstrip().split() for line in components_file]
```

nipype/algorithms/confounds.py
```diff
@@ -1201,20 +1298,56 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
         except np.linalg.LinAlgError:
-            if self.inputs.failure_mode == 'error':
+            if failure_mode == 'error':
                 raise
-            u = np.ones((M.shape[0], num_components), dtype=np.float32) * np.nan
+            if components_criterion >= 1:
+                u = np.full((M.shape[0], components_criterion),
+                            np.nan, dtype=np.float32)
+            else:
+                continue
```
effigies: If `fallback_svd` fails, `s` will not be defined, which will fail in L1311. If, additionally, `components_criterion < 1`, `u` will not be defined, either. What's the expected behavior in these cases?

nipype/algorithms/confounds.py
```diff
@@ -1201,20 +1298,56 @@ def compute_noise_components(imgseries, mask_images, num_components,
...
+            if failure_mode == 'error':
                 raise
-            u = np.ones((M.shape[0], num_components), dtype=np.float32) * np.nan
+            if components_criterion >= 1:
+                u = np.full((M.shape[0], components_criterion),
+                            np.nan, dtype=np.float32)
+            else:
+                u = np.full((M.shape[0], 1), np.nan, dtype=np.float32)
+
+        variance_explained = (s ** 2) / np.sum(s ** 2)
```
effigies: This line will still fail if `fallback_svd` fails. `s` must be defined by this point.

rciric:
Interestingly, I'm now getting
```python
ValueError: Inputs for CompCor, timeseries and mask, do not have matching spatial dimensions ((64, 64, 46) and (64, 64, 46, 1), respectively)
```
for one of my test masks. Probably an oddity on my end -- I didn't add the `squeeze` back in, since it seemed to be causing trouble earlier.

rciric:
Looks like the empty mask is now breaking `cosine_filter` -- I'll try and fix that next.

effigies:
@rciric Check out [`nibabel.squeeze_image`](https://nipy.org/nibabel/reference/nibabel.funcs.html#nibabel.funcs.squeeze_image).

mgxd:
hi @rciric - do you have a moment to finish this up sometime this week? We hope to release next week and would love to get this in.

rciric:
Happy to do what it takes to get this over the finish line -- it looks like I'm getting errors related to diffusion processing now -- specifically an import failure from `dipy.workflows.tracking` for `DetTrackPAMFlow`. [The only workflows in `dipy.workflows.tracking` are `PFTrackingPAMFlow` and `LocalFiberTrackingPAMFlow`.](https://github.com/nipy/dipy/blob/master/dipy/workflows/tracking.py) I guess it's now [`LocalFiberTrackingPAMFlow` with `sh_strategy="deterministic"`](https://github.com/nipy/dipy/commit/5833a1ee9199fcbeb4f2749f647ec047217f2f65#diff-0fc780d3255e45d8d9e28710ad4c24f7)?

mgxd:
@rciric yeah that is unrelated to this PR - I just merged #2903 which should fix it so if you pull the latest master it should be good.

oesteban:
@rciric what is the status of this PR? /cc  @mgxd @effigies

rciric:
I tested it with my `fmriprep` install before the latest push, so I think we just need verification as to whether it works from reviewers. Additional feedback is also always welcome.

oesteban:
@rciric are there any public tests (i.e. CircleCI) where we can check how this is playing along with fMRIPrep?

oesteban:
Okay, so it seems this PR is working out as expected (https://circleci.com/workflow-run/3c1ae1ca-27db-43e6-897e-54cc5d323dcb, the red light on ds005 is unrelated to this PR).

If @rciric promises that the correlation between ``a_comp_cor_xx`` regressors does not pertain to this PR and should be addressed in niworkflows/fmriprep context then we can go ahead with this. Could you double check whether we get the same outputs if you force the CompCor interfaces in fmriprep to behave like the former implementation (ie. leave both neither `num_components` and `variance_threshold` undefined)?

Do you want to have a final review @effigies, @mgxd ?

@satra, it would be great if you could take a quick look, we don't want to regret having this PR merged in the future.

satra:
@rciric  and @oesteban - this looks good to me!

rciric:
Hello, and thanks for the comments! I think they've cleaned up the code considerably -- I've updated the code to reflect those comments with the following exceptions:
* I wasn't able to get `md_retained` working as a generator on my local test machine, so I ended up using a list.
* I am using an `OrderedDict` because the call that prints the metadata to TSV currently requires its entries to be in a particular order.

nipype/info.py
```diff
@@ -141,7 +141,8 @@ def get_nipype_gitversion():
     'numpy>=%s ; python_version >= "3.7"' % NUMPY_MIN_VERSION_37,
     'python-dateutil>=%s' % DATEUTIL_MIN_VERSION,
     'scipy>=%s' % SCIPY_MIN_VERSION,
-    'traits>=%s' % TRAITS_MIN_VERSION,
+    'traits>=%s,<%s ; python_version == "2.7"' % (TRAITS_MIN_VERSION, '5.0.0'),
```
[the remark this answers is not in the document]
mgxd: I'm fine with this, though we should probably only restrict 5.0 if the user is running py27 - to my knowledge everything was working fine with py3.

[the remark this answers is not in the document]
effigies: I don't have a strong opinion here (this seems simpler than splitting out by Python version, but yes there may be someone who really needs Py3 + traits 5.0), but I think maybe let's make that change after this is merged?

effigies:
Oh, looks like tests are failing. Sorry if I've led you astray.

rciric:
No, this is totally me not updating my own unit tests.

oesteban:
yay! @rciric

PR #2885, opened 2019-02-22
ENH: io interface accepts template strings with modern format ordered arguments

polosecki:
## Summary
I modified all instances in nipype/interfaces/io.py where the input template string is formatted.
I.e. I replaced all occurrences of:
template % (argtuple)
by: 
template.format(*argtuple)

With this enhancement it is possible to have templates that assume ordered arguments:
such as: "/{0}/{1}/visit_{2}_subj_{1}" which was not possible before but is often convenient.

- I acknowledge that this contribution will be available under the Apache 2 license.

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -975,7 +975,7 @@ class S3DataGrabber(LibraryBaseInterface, IOBase):
                     filledtemplate = template
                     if argtuple:
                         try:
-                            filledtemplate = template % tuple(argtuple)
+                            filledtemplate = template.format(*argtuple)
                         except TypeError as e:
                             raise TypeError(
                                 e.message +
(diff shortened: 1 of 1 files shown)


effigies:
I don't think format strings are interchangeable with `%` and `.format` syntax. This would need to be either a new interface or behavior controlled by a parameter.

polosecki:
I have used this change successfully on my local install for some time, just wanted to share it. But if you see an issue, no problem.

effigies:
It will break every node where people currently use [printf-style formatting](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting), so we can't change it without some way of ensuring that existing workflows continue to work.

polosecki:
I see your point. I can't think of the top of my head of a head of a good way to tell when they will work or not, since the exceptions probably would not be raised at the time of the format statement but further downstream. Too bad.

effigies:
I'm going to close this for now. Feel free to reopen if you come back to this.

PR #2894, opened 2019-02-28
REF: Update nipype2boutiques script

erinb90:
## Summary

boutiques/boutiques#135

Refactored and improved the `nipype2boutiques` script that converts Nipype interfaces into Boutiques descriptors. It's still not perfect, and doesn't cover every edge case (might even crash for some interfaces), but it does a lot more than it did before. May update this or make another PR in the future if I find ways to improve it more.

[Examples of script output](https://github.com/erinb90/boutiques-descriptors/tree/master/nipype)

@glatard @gkiar 

## Acknowledgment


glatard:
Thanks for doing this @erinb90, it looks great to me!

nipype/utils/nipype2boutiques.py
```diff
@@ -55,53 +57,133 @@ def generate_boutiques_descriptor(
...
+                if not isinstance(desc_tags[k], list):
+                    desc_tags[k] = [desc_tags[k]]
+                desc_tags[k].append(v)
+            else:
+                desc_tags[k] = v
+
+    tool_desc['tags'] = desc_tags
+
+    # Check for positional arguments and reorder command line args if necessary
+    tool_desc['command-line'] = reorder_cmd_line_args(tool_desc['command-line'], interface, ignore_inputs)
```
mgxd: rf: line length

nipype/utils/nipype2boutiques.py
```diff
@@ -55,53 +57,133 @@ def generate_boutiques_descriptor(
...
+
+    # Save descriptor to a file
+    if save:
+        path = save_path if save_path is not None else os.path.join(os.getcwd(), interface_name + '.json')
+        with open(path, 'w') as outfile:
+            json.dump(tool_desc, outfile, indent=4, separators=(',', ': '))
+        if verbose:
+            print("-> Descriptor saved to file " + outfile.name)
+
+    print("NOTE: Descriptors produced by this script may not entirely conform to the Nipype interface "
```
mgxd: rf: line length

nipype/utils/nipype2boutiques.py
```diff
@@ -110,62 +192,163 @@ def get_boutiques_input(inputs, interface, input_name, spec,
...
+
+    if handler_type == "File" or handler_type == "Directory":
+        inp['type'] = "File"
+    elif handler_type == "Int":
+        inp['type'] = "Number"
+        inp['integer'] = True
+    elif handler_type == "Float":
+        inp['type'] = "Number"
+    elif handler_type == "Bool":
+        if spec.argstr and len(spec.argstr.split("=")) > 1 and (spec.argstr.split("=")[1] == '0' or spec.argstr.split("=")[1] == '1'):
```
mgxd: rf: line length

nipype/utils/nipype2boutiques.py
```diff
@@ -177,116 +360,179 @@ def get_boutiques_output(name, interface, tool_inputs, verbose=False):
...
+    if output_value:
+        output['path-template'] = os.path.relpath(output_value)
+    else:
+        for inp_name, inp_spec in sorted(interface.inputs.traits(transient=None).items()):
+            if inp_name == name and inp_spec.name_source and inp_spec.name_template:
+                if isinstance(inp_spec.name_source, list):
+                    source = inp_spec.name_source[0]
+                else:
+                    source = inp_spec.name_source
+                output['path-template'] = inp_spec.name_template.replace("%s", "[" + source.upper() + "]")
```
mgxd: rf: line length

nipype/utils/nipype2boutiques.py
```diff
@@ -177,116 +360,179 @@ def get_boutiques_output(name, interface, tool_inputs, verbose=False):
...
+                    output['command-line-flag'] = inp_spec.argstr.split("%")[0].strip()
+                break
+
     return output
 
 
-def get_type_from_spec_info(spec_info):
+def get_boutiques_groups(input_traits):
+    """
+    Returns a list of dictionaries containing Boutiques groups for the mutually exclusive and all-or-none Nipype inputs.
```
mgxd: rf: line length

nipype/utils/nipype2boutiques.py
```diff
@@ -177,116 +360,179 @@ def get_boutiques_output(name, interface, tool_inputs, verbose=False):
...
+            non_positional_args.append(value_key)
+
+    last_arg = None
+    for item in sorted(positional_arg_dict.items()):
+        if item[0] == -1:
+            last_arg = item[1]
+            continue
+        positional_args.append(item[1])
+
+    return interface_name + " " + " ".join(positional_args) + " " + ((last_arg + " ") if last_arg else "") + " ".join(non_positional_args)
```
mgxd: rf: line length

nipype/utils/nipype2boutiques.py
```diff
@@ -55,53 +57,133 @@ def generate_boutiques_descriptor(
...
-    tool_desc['outputs'] = []
-    tool_desc['tool-version'] = interface.version
-    tool_desc['schema-version'] = '0.2-snapshot'
-    if docker_image:
-        tool_desc['docker-image'] = docker_image
-    if docker_index:
-        tool_desc['docker-index'] = docker_index
+    tool_desc['output-files'] = []
+    tool_desc['groups'] = []
+    tool_desc['tool-version'] = interface.version if interface.version is not None else "1.0.0"
```
mgxd: Wouldn't it be better to state no version was provided rather than assume this?

nipype/utils/nipype2boutiques.py
```diff
@@ -55,53 +57,133 @@ def generate_boutiques_descriptor(
...
+
+    # Check for positional arguments and reorder command line args if necessary
+    tool_desc['command-line'] = reorder_cmd_line_args(tool_desc['command-line'], interface, ignore_inputs)
+
+    # Remove the extra space at the end of the command line
+    tool_desc['command-line'] = tool_desc['command-line'].strip()
+
+    # Save descriptor to a file
+    if save:
+        path = save_path if save_path is not None else os.path.join(os.getcwd(), interface_name + '.json')
```
mgxd:
```suggestion
        path = save_path or os.path.join(os.getcwd(), interface_name + '.json')
```

nipype/utils/nipype2boutiques.py
```diff
@@ -110,62 +192,163 @@ def get_boutiques_input(inputs, interface, input_name, spec,
...
-    input['command-line-key'] = "[" + input_name.upper(
+
+    # If spec has a name source, means it's an output, so skip it here.
+    # Also skip any ignored inputs
+    if spec.name_source or ignore_inputs is not None and input_name in ignore_inputs:
+        return None
+
+    inp = {}
+
+    if input_number is not None and input_number != 0:  # No need to append a number to the first of a list of compound inputs
```
mgxd:
```suggestion
    if input_number:  # No need to append a number to the first of a list of compound inputs
```

nipype/utils/nipype2boutiques.py
erinb90: That's what I thought at first, but then I found it looked weird on Zenodo to not have a version number here:
image

nipype/utils/nipype2boutiques.py
erinb90: @mgxd What should be the max line length?

mgxd: We should follow [pep8](https://www.python.org/dev/peps/pep-0008/) conventions, so 79 characters

nipype/utils/tests/test_nipype2boutiques.py
```diff
@@ -2,16 +2,40 @@
...
+                                                          'docker-fsl:latest'),
+                                         container_index='index.docker.io',
+                                         container_type='docker',
+                                         verbose=False,
+                                         save=False,
+                                         ignore_inputs=ignored_inputs,
+                                         author=("Oxford Centre for Functional"
+                                                 " MRI of the Brain (FMRIB)"))
+
+    with open('utils/nipype2boutiques_example.json', 'r',
```
erinb90: Hi @mgxd, I want to use this example output file in my test case but am not sure how to reference it (what I have above doesn't work). What should the path be?

nipype/utils/nipype2boutiques_example.json
```diff
@@ -0,0 +1,549 @@
+{
```
mgxd: this file should go in [`nipype/testing/data`](https://github.com/nipy/nipype/tree/master/nipype/testing/data). You can then import the data with

```python
from nipype.testing import example_data
example_data('relative/path/from/testing/directory')
```

nipype/utils/tests/test_nipype2boutiques.py
```diff
@@ -2,16 +2,40 @@
...
+              encoding='utf-8') as desc_file:
```
mgxd: `encoding` will not work on Py2 - though if replace `open` with `from io import open`, it should fix it.

erinb90: I'll just remove `encoding`, it's not necessary here.

erinb90:
So my test is failing because on CircleCI the descriptor produced has interface version 5.0.9, while the example file has version 1.0.0 (placeholder when version is not found). When I run `nipype2boutiques` locally I always see `interface.version` as `None`, so how did CircleCI get that version number?

mgxd:
@erinb90 the Circle build has FSL 5.0.9 installed, so the good thing is it's outputting what's expected. To avoid having to constantly reconfigure this test though, I would recommend checking only a subset of fields that will be consistent across environments.

mgxd:
@erinb90  if you sync with current master it'll fix the travis tests

nipype/utils/nipype2boutiques.py
```diff
@@ -55,119 +67,321 @@ def generate_boutiques_descriptor(
...
+                    inp['type'] = "Number"
+                    inp['integer'] = True
+                elif all(isinstance(n, float) for n in value_choices):
+                    inp['type'] = "Number"
+                inp['value-choices'] = value_choices
+        else:
+            inp['type'] = "String"
+        if trait_handler.minlen != 0:
+            inp['min-list-entries'] = trait_handler.minlen
+        if trait_handler.maxlen != six.MAXSIZE:
```
satra:
```suggestion
        if trait_handler.maxlen != sys.maxsize:
```

nipype/utils/nipype2boutiques.py
```diff
@@ -3,37 +3,49 @@
...
+#     be verified and corrected.
+#   * Still need to add some fields to the descriptor manually, e.g. url,
+#     descriptor-url, path-template-stripped-extensions, etc.
 
 import os
-import argparse
 import sys
-import tempfile
 import simplejson as json
+import six
```
satra: we don't use six as a dependency of nipype. this bit should use the relevant pieces from `future`
```suggestion
```

erinb90: I seem to have broken something in the build in removing this import. It fails with:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - matplotlib
Use "conda search <package> --info" to see the dependencies for each package.
```
Do you know where this could come from?

mgxd: @erinb90 the error is unrelated, and should be fixed if you fetch current master

Sorry for all these CI build inconveniences - it seems everything waited on your PR to break

PR #2896, opened 2019-03-06
doc: add warning when using max_jobs

mgxd:
## Summary

Limiting the amount of jobs is a bit confusing currently - some plugins rely on `max_jobs`, while others rely on `n_procs`.

- Adds warnings to MultiProc and LegacyMultiProc when user is relying on `max_jobs` to limit submissions.
## Acknowledgment

### DIFF
--- a/nipype/pipeline/plugins/legacymultiproc.py
+++ b/nipype/pipeline/plugins/legacymultiproc.py
@@ -184,6 +184,8 @@ class LegacyMultiProcPlugin(DistributedPluginBase):
         non_daemon = self.plugin_args.get('non_daemon', True)
         maxtasks = self.plugin_args.get('maxtasksperchild', 10)
         self.processors = self.plugin_args.get('n_procs', cpu_count())
+        if self.plugin_args.get('max_jobs') < self.processors:
+            logger.warning("To limit processes, use n_procs")
         self.memory_gb = self.plugin_args.get(
             'memory_gb',  # Allocate 90% of system memory
             get_system_total_memory_gb() * 0.9)
--- a/nipype/pipeline/plugins/legacymultiproc.py
+++ b/nipype/pipeline/plugins/legacymultiproc.py
@@ -184,7 +184,7 @@ class LegacyMultiProcPlugin(DistributedPluginBase):
         non_daemon = self.plugin_args.get('non_daemon', True)
         maxtasks = self.plugin_args.get('maxtasksperchild', 10)
         self.processors = self.plugin_args.get('n_procs', cpu_count())
-        if self.plugin_args.get('max_jobs') < self.processors:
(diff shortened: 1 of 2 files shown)


nipype/pipeline/plugins/legacymultiproc.py
```diff
@@ -184,6 +184,8 @@ def __init__(self, plugin_args=None):
         non_daemon = self.plugin_args.get('non_daemon', True)
         maxtasks = self.plugin_args.get('maxtasksperchild', 10)
         self.processors = self.plugin_args.get('n_procs', cpu_count())
+        if self.plugin_args.get('max_jobs', np.inf) < self.processors:
```
effigies: Maybe just:
```suggestion
        if self.plugin_args.get('max_jobs) is not None:
```

nipype/pipeline/plugins/multiproc.py
```diff
@@ -121,6 +121,8 @@ def __init__(self, plugin_args=None):
 
         # Read in options or set defaults.
         self.processors = self.plugin_args.get('n_procs', mp.cpu_count())
+        if self.plugin_args.get('max_jobs', np.inf) < self.processors:
```
effigies:
```suggestion
        if self.plugin_args.get('max_jobs') is not None:
```

nipype/pipeline/plugins/legacymultiproc.py
```diff
@@ -184,6 +184,8 @@ def __init__(self, plugin_args=None):
...
+            logger.warning("To limit processes, use n_procs")
```
effigies: Also, shouldn't passing `'max_jobs'` be an error, not a warning? 

```suggestion
            raise ValueError("Unknown plugin argument 'max_jobs' - "
                             "to limit processes, use 'n_procs'")
```

nipype/pipeline/plugins/multiproc.py
```diff
@@ -121,6 +121,8 @@ def __init__(self, plugin_args=None):
...
+            logger.warning("To limit processes, use n_procs")
```
effigies:
```suggestion
            raise ValueError("Unknown plugin argument 'max_jobs' - "
                             "to limit processes, use 'n_procs'")
```

nipype/pipeline/plugins/legacymultiproc.py
mgxd: in some cases, this may already be set?https://github.com/nipy/nipype/blob/232ac23bfd48240906e7e629e98eec854bfdcb1c/nipype/pipeline/plugins/base.py#L104

nipype/pipeline/plugins/legacymultiproc.py
mgxd: I'm fine with raising an error.

PR #2905, opened 2019-03-28
[ENH] Add all DIPY workflows dynamically  

skoudoro:
This PR is a follow up of #2830. It permits to add automatically all DIPY workflows following your DIPY version.  This should fix [this comment](https://github.com/nipy/nipype/pull/2830#issuecomment-456883290) from @arokem and @chrisfilo.

Originally, the goal was to fix [this issue](https://github.com/nipy/dipy/issues/1794) but thanks to @mgxd who already did it!

I tag @Garyfallidis and @satra just to keep them updated.

### DIFF
--- a/nipype/interfaces/dipy/base.py
+++ b/nipype/interfaces/dipy/base.py
@@ -229,7 +229,7 @@ def get_dipy_workflows(module):
 
     Examples
     --------
-    >>> from dipy.workflows import align
+    >>> from dipy.workflows import align  # doctest: +SKIP
     >>> get_dipy_workflows(align)  # doctest: +SKIP
 
     """
--- a/nipype/interfaces/dipy/base.py
+++ b/nipype/interfaces/dipy/base.py
@@ -211,3 +211,29 @@ def dipy_to_nipype_interface(cls_name, dipy_flow, BaseClass=DipyBaseInterface):
                      "_run_interface": _run_interface,
                      "_list_outputs:": _list_outputs})
     return newclass
+
+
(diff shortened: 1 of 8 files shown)


nipype/interfaces/dipy/base.py
```diff
@@ -211,3 +211,29 @@ def _list_outputs(self):
...
+    Examples
+    --------
+    >>> from dipy.workflows import align  # doctest: +SKIP
+    >>> get_dipy_workflows(align)  # doctest: +SKIP
+
+    """
+    return [(m, obj) for m, obj in inspect.getmembers(module)
+            if inspect.isclass(obj) and
+            issubclass(obj, module.Workflow) and
+            m not in ['Workflow', 'CombinedWorkflow']]
```
arokem: As in https://github.com/nipy/dipy/pull/1800, I suggest that you set this list to be a module-level constant. 

BTW: the fact that you are using this same code there suggests that this might be a pattern worth including in the dipy library code, instead of here.

nipype/interfaces/dipy/reconstruction.py
```diff
@@ -33,9 +34,19 @@
     DTIModel = dipy_to_nipype_interface("DTIModel", ReconstDtiFlow)
```
arokem: Don't you want to remove these from here?

satra:
@skoudoro - this looks great. do you want to address @arokem comments?

skoudoro:
No problem @satra, I just did it!

PR #2909, opened 2019-04-05
FIX: parsing of filename in AlignEpiAnatPy when filename does not have +

gpiantoni:
## Summary

AlignEpiAnatPy can take an AFNI file or a NIFTI file as input. If it's an AFNI file, it correctly removes the part after the +. However, a NIFTI file does not need to have a + in the filename, so the part `''.join(anat_prefix.split('+')[:-1])` returns `""`. This fix only removes the AFNI suffix if the filename contains a +.

Do not split + if it's not in the filename.

## Acknowledgment

- [x ] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -168,10 +168,12 @@ class AlignEpiAnatPy(AFNIPythonCommand):
 
     def _list_outputs(self):
         outputs = self.output_spec().get()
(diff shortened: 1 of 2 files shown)


mgxd:
@gpiantoni thanks for the patch - if you sync with current master it should get tests passing

gpiantoni:
It should be good now. thanks

PR #2919, opened 2019-04-18
ENH: Do not override caught exceptions with FileNotFoundError from unfinished hashfile

oesteban:
This PR downgrades the situation to a debug trace. The rationale is that going through crash reports of fMRIPrep is really hard because this error shows up in a large portion of them.

This error (trying to remove the unfinished hashfile after some run time error has occurred) is likely to happen when two twin nipype workflows are competing in the same filesystem. Downgrading this to a debug trace will help split errors derived from race conditions and other errors.

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -440,7 +440,6 @@ class Node(EngineBase):
             for outdatedhash in glob(op.join(self.output_dir(), '_0x*.json')):
                 os.remove(outdatedhash)
 
-
         # Hashfile while running
         hashfile_unfinished = op.join(
             outdir, '_0x%s_unfinished.json' % self._hashvalue)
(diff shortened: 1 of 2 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -474,7 +474,9 @@ def run(self, updatehash=False):
         except Exception:
             logger.warning('[Node] Error on "%s" (%s)', self.fullname, outdir)
             # Tear-up after error
-            os.remove(hashfile_unfinished)
+            if not silentrm(hashfile_unfinished):
+                logger.debug('Unfinished hashfile %s does not exist',
+                             hashfile_unfinished)
```
effigies: If this is likely to result from a race, would it be a good idea to flag that?

oesteban: Indeed, and maybe make a warning instead?

oesteban:
Thumbs-up?

effigies:
Lgtm.

PR #2940, opened 2019-05-31
FIX: Allow ``max_sh``not to be set (auto mode)

garikoitz:
For further information and why it can be left unset see http://community.mrtrix.org/t/lmax-for-dwi2response-and-dwi2fod/1099/4

cc / @oesteban

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/interfaces/mrtrix3/preprocess.py
+++ b/nipype/interfaces/mrtrix3/preprocess.py
@@ -276,8 +276,6 @@ class ResponseSDInputSpec(MRTrix3BaseInputSpec):
         exists=True, argstr='-mask %s', desc='provide initial mask image')
     max_sh = InputMultiObject(
         traits.Int,
-        value=[8],
-        usedefault=True,
         argstr='-lmax %s',
         sep=',',
         desc=('maximum harmonic degree of response function - single value for '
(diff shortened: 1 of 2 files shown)


oesteban:
@satra, @effigies, @mgxd anything against merging this one?

effigies:
The idea of setting the defaults is for provenance and to it encodes behavior so that the same interface invocation should get basically the same behavior, even if the underlying software changes its default in a future version. Is there some invocation that actually gets default behavior?

oesteban:
I don't know the original default behavior* but at some point, the default has become some sort of "auto" mode, triggered when the flag is not present. Therefore, after this PR the interface will actually implement their default.

* I think it would crash if none given, and I defined 8 as a typically used value.

effigies:
Yes. To put a finer point on my questions:

1) Is it possible to pass a non-blank value that is equivalent to leaving it blank?
2) If so, does it make sense to set that to the default, so that the interface can maintain consistent behavior in case the default changes again?

We have interfaces that go both ways on question 2, but I thought the preference was to use `usedefault` when the defaults correspond to specific arguments, for the reasons I mentioned above.

Not trying to hold this up. Feel free to merge, if the answer to either question is no.

oesteban:
> 1. Is it possible to pass a non-blank value that is equivalent to leaving it blank?

I think that is not possible, after interpreting this: https://mrtrix.readthedocs.io/en/latest/concepts/sh_basis_lmax.html

> For any command or script operating on data in the spherical harmonic basis, it should be possible to manually set the maximum harmonic degree of the output using the -lmax command-line option. If this is not provided, then an appropriate value will be determined automatically.

I haven't found better documentation [elsewhere](https://mrtrix.readthedocs.io/en/latest/reference/scripts/dwi2response.html).

> 2. If so, does it make sense to set that to the default, so that the interface can maintain consistent behavior in case the default changes again?

I don't think this is the case.

Issue #2935, opened 2019-05-21
EddyQuad input specification -> should omit "="


fBeyer89:
Dear community,
eddy_quad from FSL 6.0.1 doesn't seem to accept input defined by the nipype input spec.  It only runs if input is formatted like  -bvals "bvals.scheme" compared to -bvals="bvals.scheme" (input spec from nipype version 0.1.2)

(on my machine with FSL 6.0.1, this command worked:
eddy_quad /data/pt_life_dti_followup/diffusion/LI0026893X/eddy_corrected --eddyIdx /home/raid1/fbeyer/Documents/Scripts/LIFE_followup/preprocessing/nipy1.4/diffusion/index.txt --eddyParams /home/raid1/fbeyer/Documents/Scripts/LIFE_followup/preprocessing/nipy1.4/diffusion/acqparams_dwi.txt --mask /data/pt_life_dti_followup/diffusion/LI0026893X/dwi_appa_corrected_maths_brain_mask.nii.gz --bvals /data/pt_life_dti_followup/diffusion/LI0026893X/cmrrmbep2ddiffs009a001.bval

but this one generated by nipype does not:
eddy_quad /data/pt_life_dti_followup/diffusion/LI0026893X/eddy_corrected --eddyIdx=/home/raid1/fbeyer/Documents/Scripts/LIFE_followup/preprocessing/nipy1.4/diffusion/index.txt --eddyParams=/home/raid1/fbeyer/Documents/Scripts/LIFE_followup/preprocessing/nipy1.4/diffusion/acqparams_dwi.txt --mask=/data/pt_life_dti_followup/diffusion/LI0026893X/dwi_appa_corrected_maths_brain_mask.nii.gz --bvals=/data/pt_life_dti_followup/diffusion/LI0026893X/cmrrmbep2ddiffs009a001.bval

Maybe the input spec should be updated?

### Platform details:
{'commit_hash': '%h',
 'commit_source': 'archive substitution',
 'networkx_version': '2.2',
 'nibabel_version': '2.2.0',
 'nipype_version': '1.2.0',
 'numpy_version': '1.13.1',
 'pkg_path': '/home/raid1/fbeyer/local_pt_life/miniconda2/envs/agewell_nip1.2/lib/python2.7/site-packages/nipype',
 'scipy_version': '0.19.1',
 'sys_executable': '/home/raid1/fbeyer/local_pt_life/miniconda2/envs/agewell_nip1.2/bin/python',
[...]

PR #2941, opened 2019-06-06
[FIX] Remove '=' signs from EddyQuad argument specifications

josephmje:
Addresses #2935


effigies:
You'll need to run `make specs`.

josephmje:
Thanks @effigies . I ran `make specs` and updated the PR.

josephmje:
Just wanted to revisit this. Is this ok the merge now?

effigies:
Ah, thanks for the bump. I think this looks reasonable. To be clear, this won't break older versions of `eddy_quad`?

josephmje:
`eddy_quad` was only introduced in FSL 6.0. I've tried it on both v6.0 and v6.0.1 and it works fine

effigies:
Sounds good. Thanks for the contribution.

PR #2943, opened 2019-06-10
[FIX] modified afni's cat_matvec to accept empty string opposed to opkey

ostanley:
## Summary

AFNI's cat_matvec allows for transforms to be concatenated together. The opkey field after each transform allows for things like inversions but is non-mandatory. Currently, if no opkey is specified the - character is added resulting in an error when the interface is run. 
Example:
cat_matvec a.1D - b.1D - > c.1D

Reference documentation:
https://afni.nimh.nih.gov/pub/dist/doc/program_help/cat_matvec.html

## Fixes

* The interface is not able to perform if no opkey is specified aka a list of tuples which are ('file.1D',''). I have fixed this at the format args level. 
* New behaviour: if an empty opkey is found then the - character is not added.

## Acknowledgment


nipype/interfaces/afni/utils.py
```diff
@@ -617,11 +617,15 @@ class CatMatvec(AFNICommand):
...
         if name == 'in_file':
-            return spec.argstr % (' '.join([i[0] + ' -' + i[1]
-                                            for i in value]))
+            xfm_args=''
+            for v in value:
+                if len(v[1])>0:
+                    xfm_args += ' ' +  v[0] + ' -' + v[1] + ' '
+                else:
+                    xfm_args += ' ' + v[0] + ' '
+            return spec.argstr % (xfm_args)
```
effigies: I think it would be clearer to continue writing this as a comprehension, using the `x if y else z` idiom, but we can use more verbose variable names to make the logic clearer. I also think it would be useful to include a comment:

```Python
# Concatenate a series of filenames, with optional opkeys
return ' '.join('%s -%s' % (mfile, opkey) if opkey else mfile
                for mfile, opkey in value)
```

I dropped the initial `spec.argstr %`, just because `spec.argstr` is `"%s"`, so there's no real point, but you could keep it if you wanted.

nipype/interfaces/afni/utils.py
```diff
@@ -617,11 +617,15 @@ class CatMatvec(AFNICommand):
...
         return super(CatMatvec, self)._format_arg(name, spec, value)
 
-
```
effigies: Can you re-add this? Python style guidelines suggest 2 lines between classes.

ostanley: Done

nipype/interfaces/afni/utils.py
ostanley: I agree this is more elegant, I changed it and confirmed it passed my test for formatting.

nipype/interfaces/afni/utils.py
```diff
@@ -617,8 +617,9 @@ class CatMatvec(AFNICommand):
 
     def _format_arg(self, name, spec, value):
         if name == 'in_file':
-            return spec.argstr % (' '.join([i[0] + ' -' + i[1]
-                                            for i in value]))
+            # Concatenate a series of filenames, with optional opkeys
+            return ' '.join('%s -%s' % (mfile, opkey) if opkey else mfile
+                for mfile, opkey in value)
```
effigies: The style convention is to align new lines within parentheses, so all the contents of `.join()` are to the right of the left paren.

```suggestion
                            for mfile, opkey in value)
```

effigies:
The failing tests are due to an issue fixed in #2963. Can you merge master or rebase onto master?

nipype/interfaces/afni/utils.py
```diff
@@ -619,8 +619,9 @@ class CatMatvec(AFNICommand):
 
     def _format_arg(self, name, spec, value):
         if name == 'in_file':
-            return spec.argstr % (' '.join([i[0] + ' -' + i[1]
-                                            for i in value]))
+            # Concatenate a series of filenames, with optional opkeys
+            return ' '.join('%s -%s' % (mfile, opkey) if opkey else mfile
+                           for mfile, opkey in value)
```
effigies:
```suggestion
                            for mfile, opkey in value)
```

effigies:
Thanks! Merging.

PR #2945, opened 2019-06-13
PR: FIX Check that file exists in basedir, not current dir

### LINKED ISSUE #2944: Workflow crashing with file not found due to strange nipype.pipeline.engine.utils.modify_path behavior
### Summary

### Actual behavior

### Expected behavior

See below

### Script/Workflow details

```python


    x = traits.String()


    y = traits.String()


class Concat2(BaseInterface):
    input_spec = Concat2InputSpec

        return runtime

    def _list_outputs(self):
        outputs['y'] = self._outval
        return outputs


if __name__ == '__main__':
    node.run()
```

### Platform details:

```
{'commit_hash': '7fbc7cbac',
 'networkx_version': '2.3',
 'nibabel_version': '2.4.1',
 'numpy_version': '1.16.4',
 'scipy_version': '1.3.0',
 'sys_platform': 'linux',
 'traits_version': '5.1.1'}

```

### Execution environment

stilley2:
## Summary

Fixes #2944 

- modify_paths function checks if argument is a file in the basedir, not the current directory.

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -486,7 +486,7 @@ def modify_paths(object, relative=True, basedir=None):
             out = tuple(out)
     else:
         if isdefined(object):
-            if isinstance(object, (str, bytes)) and os.path.isfile(object):
+            if isinstance(object, (str, bytes)) and os.path.isfile(os.path.join(basedir, object)):
                 if relative:
                     if config.getboolean('execution', 'use_relative_paths'):
                         out = relpath(object, start=basedir)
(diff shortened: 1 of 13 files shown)


nipype/pipeline/engine/utils.py
```diff
@@ -486,7 +486,7 @@ def modify_paths(object, relative=True, basedir=None):
             out = tuple(out)
     else:
         if isdefined(object):
-            if isinstance(object, (str, bytes)) and os.path.isfile(object):
+            if isinstance(object, (str, bytes)) and os.path.isfile(os.path.join(basedir, object)):
```
oesteban: Please check whether this:

```suggestion
            if isinstance(object, (str, bytes)) and os.path.isfile(os.path.join(basedir, os.path.basename(object))):
```

Does not bring us back to the original problem. My impression is that it will. I agree with @effigies this is a result of loosing track of the traits.

PR #2947, opened 2019-06-21
FIX: Set input model to bedpostx for camino.TrackBedpostxProba

mick-d:
## Summary

Fix command line generated to run probability tracking on bedpostx output with Camino

## Acknowledgment

- [x ] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

### DIFF
--- a/nipype/interfaces/camino/dti.py
+++ b/nipype/interfaces/camino/dti.py
@@ -995,7 +995,7 @@ class TrackBedpostxProba(Track):
     input_spec = TrackBedpostxProbaInputSpec
 
     def __init__(self, command=None, **inputs):
-        inputs["inputmodel"] = "bedpostx_dyad"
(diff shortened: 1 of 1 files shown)


effigies:
What changed in the last four years? Has this always been a bug, or has there been an update in the underlying tool, and this is moving to be consistent with the latest version?

mick-d:
> What changed in the last four years? 

I used for the first time bedpostx + camino's probabilistic tractography for a project. Camino's code itself did not change in the last four years.

> Has this always been a bug, or has there been an update in the underlying tool, and this is moving to be consistent with the latest version?

It has always been a bug (results were the same as deterministic tractography, and not much different from probabilistic tractography with a high connectivty threshold in terms of streamline count or the metric of interest). It is now consistent with the latest version of Camino (and Camino's version from 4 years ago).

PR #2949, opened 2019-06-26
FIX: Use traits to provide type hints when modifying paths

### LINKED ISSUE #2944: Workflow crashing with file not found due to strange nipype.pipeline.engine.utils.modify_path behavior

effigies:
## Summary
This may not be the best way to do things, but in all cases where we're calling `modify_paths`, we have a traited spec that we can use to provide type hints, if we're careful about inspecting a fairly inconsistent API. Hopefully this falls back to the old behavior.

Additionally, I created a function `resolve_path` which should provide a consistent convention for returning absolute or relative paths, given an absolute or relative path, with respect to a base directory. This allowed some of the logic of `modify_paths` to be simplified, although that hardly makes up for the total muddle that is the type hints.

Let me know what you think, @oesteban, @satra, @stilley2. If this seems like the way to go, I can finish it up.

TODO:

* [ ] Test cases for `modify_paths`
* [ ] May need to comment on type hints bits

Fixes #2944.

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -25,7 +25,7 @@ from future import standard_library
 
 from ... import logging, config, LooseVersion
 from ...utils.filemanip import (
-    relpath,
+    resolve_path,
     makedirs,
     fname_presuffix,
     to_str,
@@ -41,7 +41,7 @@ from ...utils.filemanip import (
 from ...utils.misc import str2bool
 from ...utils.functions import create_function_from_source
 from ...interfaces.base import (Bunch, CommandLine, isdefined, Undefined,
-                                InterfaceResult, traits)
+                                InterfaceResult, traits, File)
 from ...interfaces.utility import IdentityInterface
 from ...utils.provenance import ProvStore, pm, nipype_ns, get_id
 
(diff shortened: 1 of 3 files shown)


nipype/pipeline/engine/utils.py
```diff
@@ -455,7 +466,7 @@ def format_node(node, format='python', include_config=False):
     return lines
 
 
-def modify_paths(object, relative=True, basedir=None):
+def modify_paths(object, relative=True, basedir=None, type_hints=None):
```
satra: @effigies - you may want to take a peek at the recursive inputs hash function in core specs. since the hints tell whether to hash a path (nested or not) or not, i think that function may be relevant in this scenario.

stilley2:
Using traits is a much better solution than my proposed fix. My knowledge of the traits system isn't so great, so I don't have any comments on the implementation itself.

PR #2959, opened 2019-07-16
ENH: Modify ``Directory`` and ``File`` traits to get along with pathlib

oesteban:
Ref #2944

## Summary

Fixes # .

## Acknowledgment


satra:
looks like a reasonable PR. the File trait should take `hash_files` into account. 

the `Directory` trait is a weird one since `hash_files` doesn't apply to it as we don't hash the contents of a directory.

satra:
@oesteban - do we really need to enable `usedefault=True` for all `File` traits?

oesteban:
I'm thinking it should be enabled only if the default value is given.

effigies:
@oesteban That's counter to how all of the other traits work, which require manual enabling of `usedefault=True`. While it's a little annoying to have to add when you set a default, this change breaks consistency.

oesteban:
The current implementation of interfaces setting a default value on ``File``s behave as if they had set `usedefault=True`.

effigies:
Oh, fun.

satra:
> The current implementation of interfaces setting a default value on `Files` behave as if they had set `usedefault=True.`

i don't think this is true. 

```
In [1]: from nipype.interfaces.base import TraitedSpec, File

In [2]: class A(TraitedSpec):
   ...:     foo = File()

In [3]: a = A()

In [4]: a
Out[4]: 

foo = <undefined>

In [5]: class A(TraitedSpec):
   ...:     foo = File('/foo')

In [6]: a = A()

In [7]: a
Out[7]: 

foo = <undefined>

In [12]: class A(TraitedSpec):
    ...:     foo = File(value='/foo')

In [13]: a = A()

In [14]: a
Out[14]: 

foo = <undefined>
```

and when `usedefault=True` then:

```
In [15]: class A(TraitedSpec):
    ...:     foo = File(value='/foo', usedefault=True)

In [16]: a = A()

In [17]: a
Out[17]: 

foo = /foo
```

oesteban:
Okay, yes, I double-checked and realized I was misinterpreting one test. Reverting.

effigies:
It's pretty unclear what are the necessary changes and which ones were done while you were at it. If the new trait is a drop-in replacement, then all the SPM/AFNI changes are superfluous. If not, then are we breaking downstream code?

oesteban:
Closed in favor of #2962

PR #2961, opened 2019-07-16
MAINT: Sort dependencies alphabetically

oesteban:
Splitting up #2959 in digestible bits.

This would be great to merge quickly if possible.

## Summary

Fixes # .

## Acknowledgment


effigies:
This is fine by me, but I think the idea was to give a rough order of perceived importance of the dependency. Is the change just to make it easier to search and see if a dependency is or isn't in the list?

nipype/info.py
```diff
@@ -135,47 +135,47 @@ def get_nipype_gitversion():
...
-    'networkx>=%s,<=%s ; python_version < "3.0"' % (NETWORKX_MIN_VERSION, NETWORKX_MAX_VERSION_27),
+    'click>=%s' % CLICK_MIN_VERSION,
+    'configparser; python_version <= "3.4"',
+    'funcsigs',
+    'future>=%s' % FUTURE_MIN_VERSION,
+    'futures; python_version == "2.7"',
     'networkx>=%s ; python_version >= "3.0"' % NETWORKX_MIN_VERSION,
-    'numpy>=%s,!=%s ; python_version == "2.7"' % (NUMPY_MIN_VERSION, NUMPY_BAD_VERSION_27),
+    'networkx>=%s,<=%s ; python_version < "3.0"' % (NETWORKX_MIN_VERSION, NETWORKX_MAX_VERSION_27),
+    'neurdflib',
```
satra: this one has to come at the end. we are using pip install order hack to install a same named library but with additions that remain to be merged into a release.

requirements.txt
```diff
@@ -1,16 +1,16 @@
-numpy>=1.9.0
-scipy>=0.14
+click>=6.6.0
+configparser
+funcsigs
+future>=0.16.0
 networkx>=1.9
-traits>=4.6
-python-dateutil>=2.2
+neurdflib
```
satra: same here. this has to come over prov.

oesteban:
> Is the change just to make it easier to search and see if a dependency is or isn't in the list?

Yes, that is the idea.

@satra, I'd say to remove those dependencies in a separate PR, just in case tests start failing...

@kaczmarj any fast way to instruct neurodocker to run all pip installs with ``--no-cache-dir``?

satra:
> I'd say to remove those dependencies in a separate PR, just in case tests start failing

unfortunately i don't think we have a test for this. which i'll try to add. however this is not about removing, but the order of listing the dependencies. so `neurdflib` **has to come after** `prov`. 

right now the tests are failing for other reasons.

effigies:
This might be fixed by using a newer neurodocker.

kaczmarj:
@oesteban - neurodocker should include some `conda clean` call after `conda install`. at least version 0.5.0 does. is the cache present in the container?

oesteban:
@kaczmarj 0.5.0 is failing too

kaczmarj:
@oesteban - i will try to look into this tomorrow

effigies:
The issue is caused by https://github.com/conda/conda-package-handling/pull/34. We need to set the working directory to something owned by the `neuro` user.

requirements.txt
effigies:
```suggestion
```

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
 nibabel>=2.1.0
-future>=0.16.0
-simplejson>=3.8.0
+numpy>=1.9.0
+packaging
 prov>=1.5.2
```
effigies:
```suggestion
prov>=1.5.2
neurdflib
```

rtd_requirements.txt
```diff
@@ -1,19 +1,19 @@
...
 networkx>=1.9
-traits>=4.6
-python-dateutil>=2.2
 nibabel>=2.1.0
-future>=0.16.0
-simplejson>=3.8.0
+numpy>=1.9.0
+numpydoc
+packaging
 prov==1.5.0
```
effigies:
```suggestion
prov>=1.5.2
neurdflib
```

oesteban:
@effigies, I added your suggestions within the merge upstream/master commit. I hope that's fine.

rtd_requirements.txt
```diff
@@ -1,19 +1,19 @@
...
+future>=0.16.0
 matplotlib
-packaging
+mock
+networkx>=1.9
+nibabel>=2.1.0
+numpy>=1.9.0
 numpydoc
+packaging
+prov>=1.5.2
```
effigies:
```suggestion
prov>=1.5.2
neurdflib
```

satra:
@effigies - looks good to me. the one thing about rtd is that it was limited to the dependencies required for doc building, but it's fine to have everything. we should definitely check, once this is merged that rtd does build.

effigies:
Docs built fine. https://readthedocs.org/projects/nipype/builds/9387007/

Somewhat unclear why they aren't built with every commit to master. I had to manually trigger the build.

satra:
i clicked on a few buttons (specifically enabled github integration - which i thought was already enabled). let's see if that makes a difference.

PR #2962, opened 2019-07-16
ENH: Modify ``Directory`` and ``File`` traits to get along with pathlib

oesteban:
Closes #2959

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/specs.py
+++ b/nipype/interfaces/base/specs.py
@@ -21,7 +21,8 @@ from packaging.version import Version
 
 from traits.trait_errors import TraitError
 from traits.trait_handlers import TraitDictObject, TraitListObject
-from ...utils.filemanip import md5, hash_infile, hash_timestamp, to_str
+from ...utils.filemanip import (
+    md5, hash_infile, hash_timestamp, to_str, USING_PATHLIB2)
 from .traits_extension import (
     traits,
     Undefined,
@@ -31,13 +32,6 @@ from .traits_extension import (
 
 from ... import config, __version__
 
-
-USING_PATHLIB2 = False
-try:
-    from pathlib import Path
-except ImportError:
-    from pathlib2 import Path  # noqa
-    USING_PATHLIB2 = True
 
 FLOAT_FORMAT = '{:.10f}'.format
 nipype_version = Version(__version__)
--- a/nipype/interfaces/base/specs.py
+++ b/nipype/interfaces/base/specs.py
@@ -19,18 +19,25 @@ from warnings import warn
 from builtins import str, bytes
 from packaging.version import Version
 
+from traits.trait_errors import TraitError
+from traits.trait_handlers import TraitDictObject, TraitListObject
 from ...utils.filemanip import md5, hash_infile, hash_timestamp, to_str
 from .traits_extension import (
     traits,
     Undefined,
     isdefined,
-    TraitError,
-    TraitDictObject,
-    TraitListObject,
     has_metadata,
 )
 
 from ... import config, __version__
+
+
+USING_PATHLIB2 = False
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path  # noqa
+    USING_PATHLIB2 = True
 
 FLOAT_FORMAT = '{:.10f}'.format
 nipype_version = Version(__version__)
@@ -314,6 +321,39 @@ class BaseTraitedSpec(traits.HasTraits):
         return self.copyable_trait_names()
 
 
+def _deepcopypatch(self, memo):
+    """
+    Replace the ``__deepcopy__`` member with a traits-friendly implementation.
+
+    A bug in ``__deepcopy__`` for ``HasTraits`` results in weird cloning behaviors.
+    Occurs for all specs in Python<3 and only for DynamicTraitedSpec in Python>2.
+
+    """
+    id_self = id(self)
+    if id_self in memo:
+        return memo[id_self]
+    dup_dict = deepcopy(self.trait_get(), memo)
+    # access all keys
+    for key in self.copyable_trait_names():
+        if key in self.__dict__.keys():
+            _ = getattr(self, key)
+    # clone once
+    dup = self.clone_traits(memo=memo)
+    for key in self.copyable_trait_names():
+        try:
+            _ = getattr(dup, key)
+        except:
+            pass
+    # clone twice
+    dup = self.clone_traits(memo=memo)
+    dup.trait_set(**dup_dict)
+    return dup
+
+
+if USING_PATHLIB2:
+    BaseTraitedSpec.__deepcopy__ = _deepcopypatch
+
+
 class TraitedSpec(BaseTraitedSpec):
     """ Create a subclass with strict traits.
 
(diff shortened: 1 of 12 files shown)


oesteban:
Okay, getting green lights with my user's Travis builds.

effigies:
Cool. Will look tonight if I get a chance, otherwise first thing in the morning. Thanks for taking this one over.

effigies:
Incidentally, we should try to have a release this month...

oesteban:
> Thanks for taking this one over

This doesn't address #2944, but sets the ground to tackle it.

nipype/interfaces/spm/base.py
```diff
@@ -597,30 +598,11 @@ def _make_matlab_command(self, contents, postscript=None):
...
-        super(ImageFileSPM,
-              self).__init__(value, filter, auto_set, entries, exists, types,
-                             allow_compressed, **metadata)
+    """Defines a trait whose value must be a NIfTI file."""
+
+    def __init__(self, value=NoDefaultSpecified, exists=False,
+                 pathlike=False, resolve=False, **metadata):
+        """Create an ImageFileSPM trait."""
+        super(ImageFileSPM, self).__init__(
+            value=value, exists=exists, types=['nifti1', 'nifti2'],
```
satra:
```suggestion
            value=value, exists=exists, types=['nifti1', 'nifti2'], allow_compressed=False,
```

nipype/interfaces/base/specs.py
```diff
@@ -314,6 +315,39 @@ def __all__(self):
         return self.copyable_trait_names()
 
 
+def _deepcopypatch(self, memo):
+    """
+    Replace the ``__deepcopy__`` member with a traits-friendly implementation.
+
+    A bug in ``__deepcopy__`` for ``HasTraits`` results in weird cloning behaviors.
+    Occurs for all specs in Python<3 and only for DynamicTraitedSpec in Python>2.
```
effigies: These both mean Python 2.7...

nipype/interfaces/base/tests/test_specs.py
```diff
@@ -420,18 +420,17 @@ def test_ImageFile():
...
+        x.add_trait('newtype', nib.ImageFile(types=['nifti10']))
     x.add_trait('nocompress',
                 nib.ImageFile(types=['mgh'], allow_compressed=False))
 
     with pytest.raises(nib.TraitError):
         x.nifti = 'test.mgz'
     x.nifti = 'test.nii'
     x.anytype = 'test.xml'
-    with pytest.raises(AttributeError):
-        x.newtype = 'test.nii'
```
effigies: Should this not still produce an `AttributeError`?

nipype/interfaces/base/tests/test_specs.py
```diff
@@ -420,18 +420,17 @@ def test_ImageFile():
...
     with pytest.raises(nib.TraitError):
-        x.nocompress = 'test.nii.gz'
```
effigies: Should this not still produce a `TraitError`?

nipype/interfaces/base/traits_extension.py
```diff
@@ -54,244 +95,239 @@ class Str(Unicode):
...
-        if exists:
-            self.info_text = 'an existing file name'
+        super(BasePath, self).__init__(value, **metadata)
 
-        super(File, self).__init__(value, **metadata)
+    def validate(self, objekt, name, value, return_pathlike=False):
+        """Validate a value change."""
+        try:
+            if USING_PATHLIB2 and isinstance(value, newstr):
+                value = '%s' % value  # pathlib2 doesn't like newstr
```
effigies: Is this different from:

```suggestion
                value = str(value)  # pathlib2 doesn't like newstr
```

Feels weirdly indirect.

nipype/interfaces/base/traits_extension.py
```diff
@@ -54,244 +95,239 @@ class Str(Unicode):
...
-            return validated_value
-        elif os.path.isfile(value):
-            return validated_value
-        else:
-            raise TraitError(
-                args='The trait \'{}\' of {} instance is {}, but the path '
-                ' \'{}\' does not exist.'.format(name, class_of(object),
-                                                 self.info_text, value))
+        if self.exists:
+            if self.exists and not value.exists():
```
effigies: Already covered by previous line.

```suggestion
            if not value.exists():
```

nipype/utils/filemanip.py
```diff
@@ -40,8 +39,52 @@
...
+    """Defines the exception for Python 2."""
+
+    def __init__(self, path):
+        """Initialize the exception."""
+        super(FileNotFoundError, self).__init__(
+            2, 'No such file or directory', '%s' % path)
+
+
+USING_PATHLIB2 = False
+USING_PATHLIB_PATCHED = False
```
effigies: This global variable doesn't seem to be used anywhere.

nipype/interfaces/base/traits_extension.py
```diff
@@ -54,244 +95,239 @@ class Str(Unicode):
...
+        if any((exists, self._is_file, self._is_dir)):
+            self.info_text += ' representing a'
+            if exists:
+                self.info_text += 'n existing'
+            if self._is_file:
+                self.info_text += ' file'
+            elif self._is_dir:
+                self.info_text += ' directory'
+            else:
+                self.info_text += ' file or directory'
```
effigies: Does it make sense to make `info_text` a property, instead? I realize basically nobody plays with traits directly, but:

```Python
>>> x = File(exists=True)
>>> x.info_text
'a pathlike object or string representing an existing file'
>>> x.exists = False
>>> x.info_text
'a pathlike object or string representing an existing file'
```

This probably doesn't matter.

nipype/interfaces/base/traits_extension.py
oesteban: Yes, after importing `str` from futures `str(value)` will produce a `newstr` object causing `Path` to choke - tested it: using `str(value)` breaks tests.

nipype/interfaces/base/specs.py
oesteban: Not really - for Python 2.7 now this bug arises also with static specs. For Python 3 it keeps happening only with `DynamicTraitedSpec`

nipype/interfaces/base/tests/test_specs.py
oesteban: Yes, I just made the test consistent with the `types=['mgh']` given for the input `nocompress`.

nipype/interfaces/base/specs.py
effigies: Ah, you're right. I guess I was reading too quickly.

nipype/interfaces/base/tests/test_specs.py
oesteban: This was a bad test - it has been replaced by lines 423-424. When defining an `Image File` with an invalid `types` metadata should have raised a `ValueError`. Now it does.

In other words `x.newtype` is not created (nor added) because of the bad value passed to `types`.

nipype/interfaces/base/traits_extension.py
oesteban: I believe we don't change the `exists` metadata of inputs/outputs but I'll give it a try.

effigies: Yeah, we don't. It's just a general feeling that it's still a Python object, and people could modify it, and maybe we should behave "correctly" in that case.

Again, it probably doesn't matter because almost certainly nobody does use our API that way. Just couldn't help myself commenting.

nipype/interfaces/io.py
```diff
@@ -1024,15 +1024,18 @@ def _list_outputs(self):
...
-            self.inputs.bucket_path += '/'
-        if self.inputs.template[0] == '/':
-            self.inputs.template = self.inputs.template[1:]
-
-        localpath = s3path.replace(self.inputs.bucket_path,
-                                   self.inputs.local_directory)
+        local_directory = str(self.inputs.local_directory)
+        bucket_path = str(self.inputs.bucket_path)
+        template = str(self.inputs.template)
+        if not os.path.split(local_directory)[1] == '':
```
mgxd: minor refactor for added clarity
```suggestion
        if not os.path.basename(local_directory) == '':
```

nipype/interfaces/io.py
```diff
@@ -1024,15 +1024,18 @@ def _list_outputs(self):
...
+            local_directory += '/'
+        if not os.path.split(bucket_path)[1] == '':
```
mgxd:
```suggestion
        if not os.path.basename(bucket_path) == '':
```

oesteban:
@mgxd I'd rather apply your suggestions in a different PR, with the AFNI and other minor fixtures I detached from #2959.

Would that be okay by you?

oesteban:
Everything seems to be checking out - I'll merge, cross your fingers this doesn't break everything.

effigies:
🤞

PR #2966, opened 2019-07-18
 FIX: Resolving absolute to relative paths in output

oesteban:
Building a solution to #2944, starting from a refactor of
``aggregate_outputs`` to be robuster and perform the referrencing when
requested via the new arguiment ``rebase_cwd``.


satra:
@oesteban - just a quick thing if this new release is going to not work with previous versions of pickled files, we should provide an appropriate message. this is mostly an fyi at this point.

oesteban:
Another issue is really implementing the `use_relative_paths` config option, and its scope. What is its extent?

- Mandates whether paths in the results and inputs pickles should not be relative? - I'd say no to this one.
- Does this affect only workflows or the setting is also meant for interfaces? (we could resolve/rebase in ``aggregate_outputs``, and actually I've been playing around it).

w.r.t. the old results files, I could add support to read legacy results files too.

oesteban:
@satra, I've split this PR into more digestible pieces, much less involved:

  - #2970: Implements some general functions and their tests extending some compound traits to make it possible to check the trait type before any path manipulation.
  - #2971: Addresses the original issue #2944 in the most straightforward way - replaces the old `modify_paths` utility by the ones added in the other PR, which are far more selective. The original test replicating the bug by @effigies is included in that PR. The way it is done, we do not need to change the way we were pickling (and thus, it should be fully compatible with existing result files).
  - #2969: a refactor of the ``aggregate_outputs`` to make it more readable.

We should also address the questions of how to really implement `use_relative_paths`:

```
$git grep use_relative_paths
nipype/pipeline/engine/utils.py:                    if config.getboolean('execution', 'use_relative_paths'):
nipype/utils/config.py:use_relative_paths = false
```

It seems that the original intent was indeed to prescribe whether results files contain relative or absolute paths.

IMHO results files should have relative paths for the outputs, and we could try to have so for the inputs (although that would require a smarter rebasing function that can identify parents of the base directory).

Then, we could have `use_relative_paths` to define user-accessible behavior - i.e., whether interfaces can resolve relative paths and they return absolute or relative paths.

Please let me know if you see something else that can be rescued from this closed PR - in particular, whether there is an appetite to change how results are pickled (without all the unroll/combine process we have right now, which was there to make the data structure amenable to `modify_paths`, and this is not needed anymore).

WDYT?

satra:
@oesteban - thanks for all this work and this very descriptive comment. since this requires some extra gray cells, i will read this tonight and make any comments. in general all of this sounds reasonable. 

one quick note. with relative paths, we would have to consider symlinks on different filesystems/mount points, which is why it was not fully realized. it was primarily created for moving the working directory, which only works with content hashing and not timestamp hashing.

PR #2969, opened 2019-07-19
MAINT: Refactor ``aggregate_outputs`` for readability

oesteban:
Just a rewrite for improved readability.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -465,7 +465,8 @@ class BaseInterface(Interface):
             _na_outputs = self._check_version_requirements(outputs)
             na_names = aggregate_names.intersection(_na_outputs)
             if na_names:
-                raise TypeError("""\
+                # XXX Change to TypeError in Nipype 2.0
+                raise KeyError("""\
 Output trait(s) %s not available in version %s of interface %s.\
 """ % (', '.join(na_names), self.version, self.__class__.__name__))
 
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -204,10 +204,10 @@ class BaseInterface(Interface):
             ]
             if any(values) and isdefined(value):
                 if len(values) > 1:
-                    fmt = ("%s requires values for inputs %s because '%s' is set. " 
+                    fmt = ("%s requires values for inputs %s because '%s' is set. "
                            "For a list of required inputs, see %s.help()")
                 else:
-                    fmt = ("%s requires a value for input %s because '%s' is set. " 
+                    fmt = ("%s requires a value for input %s because '%s' is set. "
                            "For a list of required inputs, see %s.help()")
                 msg = fmt % (self.__class__.__name__,
                              ', '.join("'%s'" % req for req in spec.requires),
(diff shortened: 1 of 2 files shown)


nipype/interfaces/base/core.py
```diff
@@ -450,34 +450,35 @@ def _list_outputs(self):
...
+        # Precalculate the list of output trait names that should be aggregated
+        aggregate_names = set(predicted_outputs.keys())
+        if needed_outputs is not None:
+            aggregate_names = set(needed_outputs).intersection(aggregate_names)
+
+        if aggregate_names:  # Make sure outputs are compatible
+            _na_outputs = self._check_version_requirements(outputs)
+            na_names = aggregate_names.intersection(set(_na_outputs))
+            if na_names:
+                raise TypeError("""\
```
effigies: This is an API change. Probably doesn't make sense at this late stage of nipype 1.x.

```suggestion
                raise KeyError("""\
```

nipype/interfaces/base/core.py
```diff
@@ -450,34 +450,35 @@ def _list_outputs(self):
...
+            return outputs
 
+        # Precalculate the list of output trait names that should be aggregated
+        aggregate_names = set(predicted_outputs.keys())
+        if needed_outputs is not None:
+            aggregate_names = set(needed_outputs).intersection(aggregate_names)
+
+        if aggregate_names:  # Make sure outputs are compatible
+            _na_outputs = self._check_version_requirements(outputs)
+            na_names = aggregate_names.intersection(set(_na_outputs))
```
effigies: Implicit.

```suggestion
            na_names = aggregate_names.intersection(_na_outputs)
```

nipype/interfaces/base/core.py
```diff
@@ -450,34 +450,35 @@ def _list_outputs(self):
...
-                        raise FileNotFoundError(msg)
-                    raise error
+        """Collate expected outputs and apply output traits validation."""
+        outputs = self._outputs()  # Generate an empty output spec object
+        predicted_outputs = self._list_outputs()  # Predictions from _list_outputs
+        if not predicted_outputs:
+            return outputs
 
+        # Precalculate the list of output trait names that should be aggregated
+        aggregate_names = set(predicted_outputs.keys())
```
effigies: Implicit.

```suggestion
        aggregate_names = set(predicted_outputs)
```

nipype/interfaces/base/tests/test_core.py
```diff
@@ -289,7 +289,7 @@ def _list_outputs(self):
             return {'foo': 1}
 
     obj = DerivedInterface1()
-    with pytest.raises(KeyError):
+    with pytest.raises(TypeError):
```
effigies:
```suggestion
    with pytest.raises(KeyError):
```

nipype/interfaces/base/core.py
oesteban: IMO there is a high chance for us to propagate these minor details into codebase related or in nipype 2.

In addition, raising a `KeyError` here is really confusing to the user - by definition of Python exceptions, this is a `TypeError`. Since the error message is still the same (only removed the request of forwarding the error to the developers), I would prefer to keep this change.

effigies: I strongly disagree with this move on the grounds that people who have been catching this exception will no longer be on upgrade. If the concern is to keep it out of nipype 2, how about adding a comment:

```suggestion
                # XXX Change to TypeError in Nipype 2.0
                raise KeyError("""\
```

PR #2970, opened 2019-07-19
ENH: Add resolve/rebase ``BasePath`` traits methods & tests

oesteban:

Two new methods ``resolve_path_traits`` and ``rebase_path_traits`` are being included.

They take trait instances from a spec (selected via ``spec.trait('traitname')``), the value and a base path. Returns the modified value (if necessary).

## Summary

These two functions will be useful to progress towards #2944.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/tests/test_traits_extension.py
+++ b/nipype/interfaces/base/tests/test_traits_extension.py
@@ -8,14 +8,13 @@ from ..traits_extension import rebase_path_traits, resolve_path_traits, Path
 
 
 class _test_spec(nib.TraitedSpec):
-    a = nib.traits.File()
-    b = nib.traits.Tuple(nib.File(),
-                         nib.File())
+    a = nib.File()
+    b = nib.Tuple(nib.File(), nib.File())
     c = nib.traits.List(nib.File())
-    d = nib.traits.Either(nib.File(), nib.traits.Float())
+    d = nib.Either(nib.File(), nib.traits.Float())
     e = nib.OutputMultiObject(nib.File())
     f = nib.traits.Dict(nib.Str, nib.File())
-    g = nib.traits.Either(nib.File, nib.Str)
+    g = nib.Either(nib.File, nib.Str)
     h = nib.Str
     ee = nib.OutputMultiObject(nib.Str)
 
@@ -27,6 +26,10 @@ def test_rebase_path_traits():
     a = rebase_path_traits(
         spec.trait('a'), '/some/path/f1.txt', '/some/path')
     assert a == Path('f1.txt')
+
+    a = rebase_path_traits(
+        spec.trait('a'), '/some/path/f1.txt', '/some/other/path')
+    assert a == Path('/some/path/f1.txt')
 
     b = rebase_path_traits(
         spec.trait('b'), ('/some/path/f1.txt', '/some/path/f2.txt'), '/some/path')
@@ -90,6 +93,10 @@ def test_resolve_path_traits():
         spec.trait('a'), 'f1.txt', '/some/path')
     assert a == Path('/some/path/f1.txt')
 
+    a = resolve_path_traits(
+        spec.trait('a'), '/already/absolute/f1.txt', '/some/path')
+    assert a == Path('/already/absolute/f1.txt')
+
     b = resolve_path_traits(
         spec.trait('b'), ('f1.txt', 'f2.txt'), '/some/path')
     assert b == (Path('/some/path/f1.txt'), Path('/some/path/f2.txt'))
@@ -128,10 +135,10 @@ def test_resolve_path_traits():
 
     # This is a problematic case, it is impossible to know whether this
     # was meant to be a string or a file.
-    # Commented out because in this implementation, strings take precedence
-    # g = resolve_path_traits(
-    #     spec.trait('g'), 'path/either.txt', '/some')
-    # assert g == Path('/some/path/either.txt')
+    # In this implementation, strings take precedence
+    g = resolve_path_traits(
+        spec.trait('g'), 'path/either.txt', '/some')
+    assert g == 'path/either.txt'
 
     # This is a problematic case, it is impossible to know whether this
     # was meant to be a string or a file.
--- a/nipype/interfaces/base/tests/test_traits_extension.py
+++ b/nipype/interfaces/base/tests/test_traits_extension.py
@@ -9,17 +9,17 @@ from ..traits_extension import rebase_path_traits, resolve_path_traits, Path
 
 class _test_spec(nib.TraitedSpec):
     a = nib.File()
-    b = nib.Tuple(nib.File(), nib.File())
+    b = nib.traits.Tuple(nib.File(), nib.File())
     c = nib.traits.List(nib.File())
-    d = nib.Either(nib.File(), nib.traits.Float())
+    d = nib.traits.Either(nib.File(), nib.traits.Float())
     e = nib.OutputMultiObject(nib.File())
     ee = nib.OutputMultiObject(nib.Str)
     f = nib.traits.Dict(nib.Str, nib.File())
-    g = nib.Either(nib.File, nib.Str)
+    g = nib.traits.Either(nib.File, nib.Str)
     h = nib.Str
-    i = nib.Either(nib.File, nib.Tuple(nib.File, nib.traits.Int))
-    j = nib.Either(nib.File, nib.Tuple(nib.File, nib.traits.Int),
-                   nib.traits.Dict(nib.Str, nib.File()))
+    i = nib.traits.Either(nib.File, nib.traits.Tuple(nib.File, nib.traits.Int))
+    j = nib.traits.Either(nib.File, nib.traits.Tuple(nib.File, nib.traits.Int),
+                          nib.traits.Dict(nib.Str, nib.File()))
 
 
 def test_rebase_path_traits():
--- a/nipype/interfaces/base/tests/test_traits_extension.py
+++ b/nipype/interfaces/base/tests/test_traits_extension.py
@@ -1,6 +1,6 @@
-# -*- coding: utf-8 -*-
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
+"""Check the resolving/rebasing feature of ``BasePath``s."""
 from __future__ import print_function, unicode_literals
 
 from ... import base as nib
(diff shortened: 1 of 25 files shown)


nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +471,120 @@ class InputMultiObject(MultiObject):
...
+    elif thistrait.is_trait_type(Tuple):
+        value = tuple([rebase_path_traits(subtrait, v, cwd)
+                       for subtrait, v in zip(thistrait.inner_traits, value)])
+    elif thistrait.alternatives:
+        is_str = [f.is_trait_type((traits.String, traits.BaseStr, traits.BaseBytes, Str))
+                  for f in thistrait.alternatives]
+        if any(is_str) and isinstance(value, (bytes, str)) and not value.startswith('/'):
+            return value
+        for subtrait in thistrait.alternatives:
+            value = rebase_path_traits(subtrait, value, cwd)
```
effigies: This loop does not seem right. Should there be a stop condition, rather than assigning `value` N times and returning the last value?

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +471,120 @@ class InputMultiObject(MultiObject):
...
+
+class Tuple(traits.BaseTuple):
+    """Defines a new type of Tuple trait that reports inner types."""
+
+    def init_fast_validator(self, *args):
+        """Set up the C-level fast validator."""
+        super(Tuple, self).init_fast_validator(*args)
+        self.fast_validate = args
+
+    def inner_traits(self):
```
effigies: Below, `inner_traits` is accessed as an attribute. Should this be:
```suggestion
    @property
    def inner_traits(self):
```

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +471,120 @@ class InputMultiObject(MultiObject):
...
+            pass
+    return value
+
+
+def rebase_path_traits(thistrait, value, cwd):
+    """Rebase a BasePath-derived trait given an interface spec."""
+    if thistrait.is_trait_type(BasePath):
+        value = _rebase_path(value, cwd)
+    elif thistrait.is_trait_type(traits.List):
+        innertrait, = thistrait.inner_traits
```
effigies: Above, `inner_traits` is a method. Should this be
```suggestion
        innertrait, = thistrait.inner_traits()
```

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +471,120 @@ class InputMultiObject(MultiObject):
...
+    if isinstance(value, list):
+        return [_rebase_path(v, cwd) for v in value]
+
+    try:
+        value = Path(value)
+    except TypeError:
+        pass
+    else:
+        try:
+            value = Path(value).relative_to(cwd)
```
effigies: Already coerced in L512:

```suggestion
            value = value.relative_to(cwd)
```

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +471,120 @@ class InputMultiObject(MultiObject):
...
+        value = Path(value)
+    except TypeError:
+        pass
+    else:
+        if not value.is_absolute():
+            value = Path(cwd) / value
+    return value
+
+
+def resolve_path_traits(thistrait, value, cwd):
```
effigies: Are there meaningful differences besides the `_resolve_path` vs `_rebase_path`? If they're supposed to be identical except for the root function, I think it'll be safer to make it a higher-order function like:

```Python
def _recurse_on_path_traits(func, thistrait, value, cwd):
    if thistrait.is_trait_type(BasePath):
        value = func(value, cwd)
    ...

def rebase_path_traits(thistrait, value, cwd):
    return _recurse_on_path_traits(_rebase_path, thistrait, value, cwd)

def resolve_path_traits(thistrait, value, cwd):
    return _recurse_on_path_traits(_resolve_path, thistrait, value, cwd)
```

nipype/interfaces/base/tests/test_traits_extension.py
```diff
@@ -0,0 +1,152 @@
...
+    ee = nib.OutputMultiObject(nib.Str)
+
+
+def test_rebase_path_traits():
+    """Check rebase_path_traits."""
+    spec = _test_spec()
+
+    a = rebase_path_traits(
+        spec.trait('a'), '/some/path/f1.txt', '/some/path')
+    assert '%s' % a == 'f1.txt'
```
effigies: You use `Path`s elsewhere:

```suggestion
    assert a == Path('f1.txt')
```

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +471,120 @@ class InputMultiObject(MultiObject):
...
+    if isinstance(value, list):
+        return [_resolve_path(v, cwd) for v in value]
+
+    try:
+        value = Path(value)
+    except TypeError:
+        pass
+    else:
+        if not value.is_absolute():
+            value = Path(cwd) / value
```
effigies: Is there any possibility that `cwd` is itself not absolute?

```suggestion
            value = Path(cwd).absolute() / value
```

nipype/interfaces/base/traits_extension.py
```diff
@@ -304,6 +305,11 @@ def validate(self, objekt, name, value, return_pathlike=False):
         return value
 
 
+# Patch in traits these two new
+traits.File = File
+traits.Directory = Directory
```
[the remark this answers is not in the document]
oesteban: I think there's a trade-off between breaking compatibility with other software (via traits) and the likelihood of a user mistakenly using `traits.File` instead of `File`.

Happy not to patch, although I don't see it as risky.

nipype/interfaces/base/traits_extension.py
oesteban: The condition is checked in `rebase_path_traits`, but I will optimize this to avoid entering/exiting the function just to check it.

nipype/interfaces/base/traits_extension.py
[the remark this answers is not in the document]
effigies: I would prefer not to patch, as we've been using these traits without doing so for years now.

nipype/interfaces/base/traits_extension.py
oesteban: traits makes it look like a property without decorator: https://github.com/enthought/traits/blob/cf5f04b22eaf751f2bb8143a7b5945ea915b704a/traits/trait_types.py#L2601

nipype/interfaces/base/traits_extension.py
oesteban: this is a traits particularity. it is accessed as an attribute.

nipype/interfaces/base/traits_extension.py
oesteban: I'll have a look :+1:

nipype/interfaces/base/traits_extension.py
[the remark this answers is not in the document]
oesteban: true for File, not always the case for Directory. I actually patched because I fell into this myself when writing a test.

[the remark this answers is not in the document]
oesteban: however, since this is controversial I'll leave the patching out of this PR and we may consider it again in the future

effigies:
Sorry for taking so long to get back to this one. I'm a little torn on the `Tuple/Either` issue. I see that the point of adding the `inner_traits` option is to have a uniform access method, but it does mean that our methods won't work with standard `traits.Tuple` and `traits.Either`, which is evidenced by the need to mass-replace those with the `traits_extension` variants.

Does it not make more sense to have an accessor function that doesn't require either monkey-patching or replacing? This way we're not breaking downstream tools that define their own interface.

If not, then I think monkey-patching the pre-existing traits will be necessary.

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +466,100 @@ class InputMultiObject(MultiObject):
...
+        value = tuple([_recurse_on_path_traits(func, subtrait, v, cwd)
+                       for subtrait, v in zip(thistrait.inner_traits, value)])
+    elif thistrait.alternatives:
+        is_str = [f.is_trait_type((traits.String, traits.BaseStr, traits.BaseBytes, Str))
+                  for f in thistrait.alternatives]
+        if any(is_str) and isinstance(value, (bytes, str)) and not value.startswith('/'):
+            return value
+        is_basepath = [f.is_trait_type(BasePath) for f in thistrait.alternatives]
+        if any(is_basepath):
+            subtrait = thistrait.alternatives[is_basepath.index(True)]
```
effigies: Is this guaranteed to only have one?

Perhaps we can add:

```suggestion
            subtrait = thistrait.alternatives[is_basepath.index(True)]
            if sum(is_basepath) > 1:
                logger.debug('Multiple BasePath inner traits: %r' % thistrait.alternatives)
```

This would at least be a possible flag we could look for if something goes wrong.

effigies: Also, what happens if one of the inner traits is a list or dict?

oesteban: Question 1: having an `Either(BasePath, BasePath)` doesn't make much sense. You can have it, but being an `Either` only one will suffice to trigger the path replacement. The only use-case I couldn't figure out is `Either(Str, BasePath)` where you cannot tell whether a value is meant to be a string or a path. I decided here `Str` is prevalent.

Question 2: I'll add a test case, something like `Either(Dict, Tuple)` and see what is happening. I bet I'm not recursing into those.

effigies: What if I did:

```
Either(BasePath, Tuple(BasePath, Int))
```

This would be relatively common for AFNI, where you can specify a volume in a file.

oesteban:
> Sorry for taking so long to get back to this one. I'm a little torn on the `Tuple/Either` issue. I see that the point of adding the `inner_traits` option is to have a uniform access method, but it does mean that our methods won't work with standard `traits.Tuple` and `traits.Either`, which is evidenced by the need to mass-replace those with the `traits_extension` variants.
Correct, the original Tuple and Either won't work.

> Does it not make more sense to have an accessor function that doesn't require either monkey-patching or replacing? This way we're not breaking downstream tools that define their own interface.

I tried, but accessing traits without their prescribed methods has a huge boilerplate. I finally opted for this solution of least changes. We should consider streaming them up to traits - at least for the `Tuple` (not so convinced about `Either`). Reading through their code, it seems to me that the `inner_traits` was left there to implement this behavior but they never finally did.

> If not, then I think monkey-patching the pre-existing traits will be necessary.

[That was my initial solution](https://github.com/nipy/nipype/pull/2970/commits/689e5212ad2c0afd3ff8bbd52900edc3c7f858c7#diff-07934241447dd5e65f51347e36fdd1ffL498), but after reading your position about `File`, I went ahead and changed everything for these other two. Sorry if I misunderstood your previous comment.

effigies:
> That was my initial solution, but after reading your position about File, I went ahead and changed everything for these other two. Sorry if I misunderstood your previous comment.

Well, with File and Directory, we're creating them and throwing them into that namespace, which feels strange. Especially because we haven't done it before, so everybody who has been using them has been importing them from nipype.

While I would rather just create an accessor method to contain the boilerplate to one place, e.g.:

```Python
def get_inner_traits(trait_type):
    try:
        return trait_type.inner_traits
    except:
        pass
    # Tuple hack
    # Either hack
    # Other hacks?
```

If that's out of the question, and we're producing otherwise-identical updates to existing traits, I think monkey-patching is a cleaner solution than forcing all downstream projects to change their import targets.

oesteban:
> Well, with File and Directory, we're creating them and throwing them into that namespace, which feels strange.

https://github.com/enthought/traits/blob/master/traits/trait_types.py#L1659
https://github.com/enthought/traits/blob/master/traits/trait_types.py#L1762

I don't see the difference with Either and Tuple.

> Especially because we haven't done it before, so everybody who has been using them has been importing them from nipype.

I disagree - https://github.com/nipy/nipype/pull/2970/commits/f1f4cf2e1a0a05831d04ef0f46ad1b881a433b1e

effigies:
Huh. I didn't realize there already were `File`/`Directory` traits. Sorry about that.

So I don't know. It's annoying to push these changes downstream, but we haven't been monkey-patching `traits`, so this is a pretty big change to do that.

oesteban:
I would bring monkeypatching back, it is very likely that people have their interfaces defined using `traits.(File|Directory|Tuple|Either)`, but keep all those conversions to nipype's definitions for the codebase. WDYT?

effigies:
That seems reasonable, but I feel like I want to get @satra's opinion on monkeypatching before I ask you to do anything big again...

satra:
@oesteban  and @effigies - `File`/`Directory` were overwritten to provide better error messages. nothing about the functionality of `File`/`Directory` was changed. and `ImageFile` is a new trait, so that is unique to nipype itself.

the current change that we are making is to change the behavior of `File` to not only accept/provide Pathlib objects but also to create additional semantics. if we were to forget backwards compatibility we would simply consider two new nipype traits:

1. `PathX`, string representing input/output files which need to be content hashed
2. `StrX`, strings representing Paths/FileNames (typically outputs), which need to be string hashed (these would have the metadata `hash_files=False`.

all places where we use `File` could be replaced by one of these two traits.

the functionality of the new `File` trait that @oesteban implemented would mostly correspond to `PathX`. the names `PathX` and `StrX` can be decided to be something different. 

this way we are not monkeypatching traits, and the introduced functionality is within the nipype context and people are not confused.

oesteban:
The problem I see @satra is that people having interfaces not contributed to nipype are potentially using `traits.File` and co. If we don't want to break compatibility with those users, we'll need to monkeypatch.

One thing you did not mention is what we do with the new Tuple/Either traits (so that they return their inner trait types).

oesteban:
That said, I would not revert back all those `traits.File` (at least we now have all of them from nipype, regardless we monkeypatch or not).

nipype/interfaces/base/traits_extension.py
oesteban: Added your use case and one more with a Dict.

oesteban:
1. I believe I've found the way so that not to shadow ``traits.Either`` and ``traits.Tuple`` - I'll revert back those changes (maybe opening a new PR?)

2. Should we patch `File` and `Directory` back into traits? We were using `traits.File`, so it is likely users are using it themselves.

3. The `hash_files` and the config `use_relative_paths` are tangential to this PR and #2971, but could be easily addressed with the new features implemented here.

effigies:
@oesteban I think you rebased on an out-of-date master.

effigies:
Sorry, can you remove the commits with @josephmje? I don't think we want them and they add a `min_ver` back in that had been removed from his PR.

nipype/interfaces/dipy/base.py
```diff
@@ -100,7 +100,7 @@ def convert_to_traits_type(dipy_type, is_file=False):
     elif "string" in dipy_type and not is_file:
         return traits.Str, is_mandatory
```
effigies: While we're switching from `traits.File`:

```suggestion
        return Str, is_mandatory
```

This will need an import.

@skoudoro Could you verify that this is fine and that you didn't explicitly intend to use `traits.Str` over `nipype.interfaces.base.Str` for some reason?

nipype/interfaces/base/traits_extension.py
```diff
@@ -465,3 +463,73 @@ class InputMultiObject(MultiObject):
...
+        value = tuple([_recurse_on_path_traits(func, subtrait, v, cwd)
+                       for subtrait, v in zip(thistrait.handler.types, value)])
+    elif thistrait.is_trait_type(traits.TraitCompound):
+        is_str = [isinstance(f, (traits.String, traits.BaseStr, traits.BaseBytes, Str))
+                  for f in thistrait.handler.handlers]
+        if any(is_str) and isinstance(value, (bytes, str)) and not value.startswith('/'):
+            return value
+
+        for subtrait in thistrait.handler.handlers:
+            value = _recurse_on_path_traits(func, subtrait(), value, cwd)
```
effigies: Oh, I see. We're updating `value` each time, so as long as one is a Path-like, then it gets rebased/resolved.

Do we show idempotence? e.g.,

```Python
for func in _rebase_path, _resolve_path:
    newval = _recurse_on_path_traits(func, thistrait, value, cwd)
    assert newval == _recurse_on_path_traits(func, thistrait, newval, cwd)
```

I looked for it in the tests, but you may have had this in mind already but written it in a way I didn't recognize.

nipype/interfaces/dipy/base.py
skoudoro: I just check it and everything is fine.

nipype/interfaces/base/traits_extension.py
oesteban: Please let me know if the latest commit addresses this issue

oesteban:
Three questions remain open (only 1 is relevant to this PR):

1. Whether to patch `File` and `Directory` back in traits - I guess @effigies' would be against (?)
2. Whether we want to simplify/refactor the implementation of `hash_files` (https://github.com/nipy/nipype/pull/2970#issuecomment-515538213): this PR is a drop-in replacement of traits that should allow #2971 to remain strictly scoped into reading/writing of results files (i.e., no need to rework `hash_files`).
3. Whether we want to implement the config `use_relative_paths`, which I think fully pertains to the Interface level (i.e., it would be independent of the internal format of results files).

cc @satra

nipype/interfaces/base/traits_extension.py
effigies: Hard to tell... it's not the most readable test, so I'm working at trying to figure out if there's an idempotence test in there somewhere.

effigies: It doesn't look like it. I see a lot of inverse tests, e.g., f(f^-1(x)) = x and f^-1(f(y)) = y. In case it wasn't clear, by idempotence, I mean f(f(x)) = f(x). In case it was clear, could you clearly label those tests?

oesteban: Example:

https://github.com/nipy/nipype/pull/2970/files#diff-2f70a0b8533ee2d429dcdfd7d1bcec87R39

It uses the output from the rebase function. It is true that I'm not testing idempotence bidirectionally [I'm testing x = f'(f(x)) but I'm not testing y = f(f'(y))]

oesteban: Okay, I believe now I made idempotence tests :)

nipype/interfaces/dipy/base.py
effigies: @oesteban Did we not want to change this?

oesteban:
Ready to go :)

PR #2971, opened 2019-07-19
FIX: Resolving/rebasing paths from/to results files

### LINKED ISSUE #2944: Workflow crashing with file not found due to strange nipype.pipeline.engine.utils.modify_path behavior

oesteban:
Based on functionalities added in #2970.

## Summary

Searches for paths in outputs of the result object to resolve/rebase them when loading/saving the pickle.

Includes a minor variation of the test @effigies proposed.

Fixes #2944 .

## Side effects

#2944 and posterior efforts to address it uncovered a couple of problems we may want to address some time (from https://github.com/nipy/nipype/pull/2970#issuecomment-516979640):

1. Whether we want to simplify/refactor the implementation of `hash_files` (https://github.com/nipy/nipype/pull/2970#issuecomment-515538213): this PR is a drop-in replacement of traits that should allow #2971 to remain strictly scoped into reading/writing of results files (i.e., no need to rework `hash_files`).
2. Whether we want to implement the config `use_relative_paths`, which I think fully pertains to the Interface level (i.e., it would be independent of the internal format of results files).

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -1260,7 +1260,7 @@ class MapNode(Node):
                 stop_first=str2bool(
                     self.config['execution']['stop_on_first_crash'])))
         # And store results
-        _save_resultfile(result, cwd, self.name)
+        _save_resultfile(result, cwd, self.name, rebase=False)
         # remove any node directories no longer required
         dirs2remove = []
         for path in glob(op.join(cwd, 'mapflow', '*')):
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -587,7 +587,9 @@ Error populating the input "%s" of node "%s": the results file of the source nod
                     runtime=runtime,
                     inputs=self._interface.inputs.get_traitsfree(),
                     outputs=aggouts)
-                _save_resultfile(result, cwd, self.name)
+                _save_resultfile(
+                    result, cwd, self.name,
+                    rebase=str2bool(self.config['execution']['use_relative_paths']))
             else:
                 logger.debug('aggregating mapnode results')
                 result = self._run_interface()
@@ -634,7 +636,9 @@ Error populating the input "%s" of node "%s": the results file of the source nod
             except Exception as msg:
                 result.runtime.stderr = '{}\n\n{}'.format(
                     getattr(result.runtime, 'stderr', ''), msg)
-                _save_resultfile(result, outdir, self.name)
+                _save_resultfile(
+                    result, outdir, self.name,
+                    rebase=str2bool(self.config['execution']['use_relative_paths']))
                 raise
             cmdfile = op.join(outdir, 'command.txt')
             with open(cmdfile, 'wt') as fd:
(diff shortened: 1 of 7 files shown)


oesteban:
This contains the test from and closes #2949

oesteban:
This feels ready to merge. The diff will be easier after #2985 is merged. Hopefully, this ends my burst of refactors. Thanks for your patience!

nipype/pipeline/engine/nodes.py
```diff
@@ -1260,7 +1260,7 @@ def _run_interface(self, execute=True, updatehash=False):
                 stop_first=str2bool(
                     self.config['execution']['stop_on_first_crash'])))
         # And store results
-        _save_resultfile(result, cwd, self.name)
+        _save_resultfile(result, cwd, self.name, rebase=False)
```
effigies: Should this be `self.config.getboolean('execution', 'use_relative_paths')`?

```suggestion
        _save_resultfile(result, cwd, self.name,
                         rebase=self.config.getboolean('execution', 'use_relative_paths'))
```

oesteban: This is the discussion point 2 I wanted to bring up atop. On the one hand, should we expose the inner workings of the engine - after all, adding that here does not change anything to the user (just that they won't be able to move the work directory anymore).

On the other hand, for this to be completely consistent across the node implementation, we need to also rebase/resolve the inputs pickle (actually I'll open an issue because this should be addressed).

I'm under the impression that `use_relative_paths` could be useful for the user at the interface level, forcing interfaces to return relative paths if desired.

oesteban: @satra can you share your views as regards `use_relative_paths`

@effigies WDYT about not exposing the inner workings (paths) of the results file to the users?

effigies: My overall perspective is "Let's not change the API if we don't absolutely have to to fix the bug." Which includes config file options.

I may not be understanding your position though, so we might be talking past each other.

oesteban: Okay, [from that standpoint](https://nipype.readthedocs.io/en/1.1.1/users/config_file.html):

> **use_relative_paths**
> Should the paths stored in results (and used to look for inputs) be relative or absolute. Relative paths allow moving the whole working directory around but may cause problems with symlinks. (possible values: true and false; default value: false)

The option is clearly defined specifically for the results file. Under that perspective, yes, any rebasing should be done only if `use_relative_paths` is true.

oesteban: But actually not at the point you suggested, because that one happens at MapNode aggregation.

effigies: Oh, okay. Sorry, my attention is split a lot of ways right now.

oesteban: Done, however, we may want to add parameterized tests with ``use_relative_paths`` on and off.

oesteban:
@satra @mgxd @djarecka any voices against merging this PR?

PR #2980, opened 2019-07-31
FIX: Python 3.4 is not compatible with the recent release of lxml-4.4.0

oesteban:
We are starting to see failed jobs in Travis.

## Summary

Fixes # .

## Acknowledgment


effigies:
The cancelled test was me, because we don't need to double-up our Travis builds when the queue is slammed.

oesteban:
It's not passing. Is there a way to create `requirements.txt` filtering versions?

oesteban:
Travis worked [on my branch](https://travis-ci.org/oesteban/nipype/builds/566293522). Merging in.

PR #2983, opened 2019-08-01
FIX: Correctly pickle ``OuputMulti{Object,Path}`` traits

### LINKED ISSUE #2968: Obscure behavior of ``OutputMultiObject`` traits
### Summary

The ``OutputMultiObject`` trait automatically flattens lists of length one without options to prevent such behavior. The behavior is obscure because there's no way to know whether flattening happened or not.

oesteban:
It seems that #2944 has uncovered a rats-nest hidden in the engine. In
resolving that issue, I found out that a great deal of boilerplate was
set in place when loading/saving results to deal with
``OutputMulti{Object,Path}`` traits. The reason being that these traits
flatten single-element-list values.

This PR fixes the pickling behavior of traited specs containing these
types of traits.

Additionally, this PR also avoids the ``modify_paths`` function that was
causing problems originally in #2944. Therefore, this PR effectively
make results files static, meaning: caching if the ``base_dir`` of the
workflow is changed will not work anymore.

I plan to re-insert this feature (results file mobility) with #2971.
This PR is just to split that one in more digestible bits.

All the boilerplate mentioned above has been cleaned up.

Fixes #2968 

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/specs.py
+++ b/nipype/interfaces/base/specs.py
@@ -31,7 +31,6 @@ from .traits_extension import (
     isdefined,
     has_metadata,
     OutputMultiObject,
-    OutputMultiPath,
 )
 
 from ... import config, __version__
--- a/nipype/interfaces/base/specs.py
+++ b/nipype/interfaces/base/specs.py
@@ -30,6 +30,8 @@ from .traits_extension import (
     Undefined,
     isdefined,
     has_metadata,
+    OutputMultiObject,
+    OutputMultiPath,
 )
 
 from ... import config, __version__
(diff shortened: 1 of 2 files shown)


effigies:
Love to see it
Screen Shot 2019-08-01 at 16 23 30

oesteban:
And most of the green are doctests :1st_place_medal:

oesteban:
``make tests`` works locally for python 2.7 and 3.7 - we'll see what happens with workflows on circle

nipype/interfaces/base/specs.py
```diff
@@ -316,6 +318,32 @@ def _get_sorteddict(self,
...
+
+        >>> spec.__setstate__(spec.__getstate__())
+        >>> spec.out
+        [4]
+
+        """
+        state = super(BaseTraitedSpec, self).__getstate__()
+        for key in self.__all__:
+            _trait_spec = self.trait(key)
+            if _trait_spec.is_trait_type((OutputMultiObject, OutputMultiPath)):
```
effigies: Aren't these the same thing?

```suggestion
            if _trait_spec.is_trait_type(OutputMultiObject):
```

effigies: https://github.com/nipy/nipype/blob/7262b2425943e0b777b94066975b6606a2303020/nipype/interfaces/base/traits_extension.py#L465

oesteban: Yes, but I'm unsure the test will come positive either way. I preferred to be explicit (although I don't mind checking)

oesteban: Okay:

```Python
>>> import nipype.interfaces.base as nib
>>> class OutputSpec(nib.TraitedSpec):
...     out = nib.OutputMultiObject(nib.traits.Int)
>>> spec = OutputSpec()
>>> spec.trait('out').is_trait_type(nib.OutputMultiPath)
True
```

nipype/interfaces/base/specs.py
```diff
@@ -30,6 +30,8 @@
     Undefined,
     isdefined,
     has_metadata,
+    OutputMultiObject,
+    OutputMultiPath,
```
effigies: Sorry...

```suggestion
```

oesteban:
@satra this is something you may want to have a quick look before it gets in.

@mgxd, @djarecka your feedback would be very welcome too!

oesteban:
All green (https://github.com/nipy/nipype/pull/2983/commits/c8a37979456b75db0c6e0c1c65bba559a59131cd)

PR #2985, opened 2019-08-02
FIX: Use ``load_resultfile`` when loading a results pickle

oesteban:
## Summary
Some sections of the code were using ``filemanip.utils.loadpkl``, which does not resolve paths (probably dodging more frequent errors from ``modify_paths`` akin to #2944).

These changes are spun off of #2971.

- A revision of ``loadpkl`` for readability and reliability.
- Replacement of instances where ``loadpkl`` was used for reading result files instead of ``pipeline.engine.utils.load_resultfile``.

## Acknowledgment


nipype/utils/filemanip.py
```diff
@@ -676,52 +676,50 @@ def loadcrash(infile, *args):
...
-    if infile.endswith('pklz'):
-        pkl_file = gzip.open(infile, 'rb')
-    else:
-        pkl_file = open(infile, 'rb')
+    pklopen = gzip.open if infile.suffix == '.pklz' else open
+    pkl_metadata = None
 
-    if versioning:
-        pkl_metadata = {}
+    with indirectory(str(infile.parent)):
```
effigies: Maybe let's make `indirectory` handle `Path`s?

oesteban: I'll open a separate PR :+1:

PR #2998, opened 2019-08-13
[FIX] check for existance of NULL file

stilley2:
## Summary

The antsRegistration interfaces uses "NULL" to indicate no mask file. This will fail if there actually is a file named NULL. This PR checks for the existence of such a file and raises an exception if it exists.

- Check whether a file named NULL exists when formatting antsRegistration call, if NULL is used to indicate no mask

## Acknowledgment

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -676,6 +676,11 @@
       "affiliation": "MIT, HMS",
       "name": "Ghosh, Satrajit",
       "orcid": "0000-0002-5312-6729"
+    },
+    {
+      "affiliation": "Holland Bloorview Kids Rehabilitation Hospital",
+      "name": "Tilley II, Steven",
+      "orcid": "0000-0003-4853-5082"
     }
   ],
   "keywords": [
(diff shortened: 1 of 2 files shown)


oesteban:
I believe this shouldn't be a problem after #2971 was merged. Could you perhaps write a test?

stilley2:
I don't think it's related to #2971. I tried to write a test case, which led me to the conclusion that it's probably not an issue anyway. Even if a file named 'NULL' exists, ants will fail to read it because it has no suffix. While I dislike using a fake file name to indicate no mask, that's really a problem with ANTS and not nipype.

effigies:
@stilley2 I'm not sure whether to interpret your last post as saying this change is not necessary. If so please close it. Otherwise, are you ready for a review or still working on the failing tests?

stilley2:
Sorry for the ambiguity. Yes, I don't think this change is necessary.

effigies:
Great, thanks for the confirmation.

PR #3004, opened 2019-08-19
[FIX] Mrtrix3 usedefault issue

### LINKED ISSUE #2822: MRtrix 3.0 `dwi2fod` single output error
Hi, I got this error when executing `mrtrix3.EstimateFOD` interface with the latest MRtrix version (`3.0_RC2-117-gf098f097`) installed on macOS Mojave 10.14.2 through a `brew install mrtrix3`:

```
dwi2fod: [ERROR] CSD algorithm expects a single input response function and single output FOD image
```

It seems that MRtrix now expects "pairs of input tissue response and output ODF images" according to [the documentation](https://mrtrix.readthedocs.io/en/latest/reference/commands/dwi2fod.html#usage). So I guess that the following lines need to be updated:

https://github.com/nipy/nipype/blob/35b58cdb59a900430f4b9639e220c8408341805d/nipype/interfaces/mrtrix3/reconst.py#L95-L109

matteomancini:
## Summary

Fixes #2822 .

* Set `useDefault=False` for `gm_odf` and `csf_odf` in order to avoid issues with the `csd` algorithm;
* Removed `useDefault=True` for the `-lmax` argument in order to avoid issues in case of multi-shell data.

## Acknowledgment


nipype/interfaces/mrtrix3/reconst.py
```diff
@@ -102,10 +102,10 @@ class EstimateFODInputSpec(MRTrix3BaseInputSpec):
         mandatory=True,
         desc='output WM ODF')
     gm_txt = File(argstr='%s', position=-4, desc='GM response text file')
-    gm_odf = File('gm.mif', usedefault=True, argstr='%s',
+    gm_odf = File('gm.mif', usedefault=False, argstr='%s',
```
effigies:
```suggestion
    gm_odf = File(argstr='%s',
```

nipype/interfaces/mrtrix3/reconst.py
```diff
@@ -102,10 +102,10 @@ class EstimateFODInputSpec(MRTrix3BaseInputSpec):
...
                   position=-3, desc='output GM ODF')
     csf_txt = File(argstr='%s', position=-2, desc='CSF response text file')
-    csf_odf = File('csf.mif', usedefault=True, argstr='%s',
+    csf_odf = File('csf.mif', usedefault=False, argstr='%s',
```
effigies:
```suggestion
    csf_odf = File(argstr='%s',
```

matteomancini:
@effigies thank you for the suggestions, I will implement them later today. The issue is that in the current form the interface is broken for anyone who tries to use the `csd` algorithm. If the fix I proposed it is accepted, anyone who was using the `msmt_csd` algorithm and was not explicitly defining `gm_odf` and `csd_odf` will need to do so. The difference is that in the latter case it easy for an user to make things work again, while in the former and current scenario the only ways to fix things are either to set `csf_odf` and `gm_odf` to `Undefined` or to actually change the interface in Nipype (or not use it at all).

If you look at the step before the FOD computation, the response estimation, is implemented in `ResponseSD` in a similar manner to the proposed change: if you are using the `msmt_5tt` algorithm, you need to specify the `gm_file` and the `csf_file`.

Since it is just a matter of defining additional output names for one of the two algorithms, I think that two different interfaces are quite redundant and the result would be even more disruptive than the change here proposed. If you have any idea on how to automatically set `gm_odf` and `csf_odf` in case the algorithm is `msmt_csd` that would be great, and should be probably implemented also in `ResponseSD`.

effigies:
> If the fix I proposed it is accepted, anyone who was using the msmt_csd algorithm and was not explicitly defining gm_odf and csd_odf will need to do so.

Right. This is the problem. I agree that it's not a heavy fix for an actively maintained workflow, but the concern is for workflows that are not actively maintained and don't depend on a specific version of nipype. On the whole, we try not to break backwards compatibility under the current setup, unless it's clear that the feature never worked for anybody. Here it's apparently worked for people using msmt_csd. (In the future of nipype, interfaces will themselves be versioned, and design decisions can be made with at least a bit less regard for workflows in the wild.)

My suggestion for an alternative interface could be done with very little code:

```Python
class EstimateCSDInputSpec(EstimateFODInputSpec):
    algorithm = traits.Enum('csd', ...)
    gm_odf = Undefined
    csf_odf = Undefined

class EstimateCSD(EstimateFOD):
    input_spec = EstimateCSDInputSpec
```

I haven't tested this, in particular whether you can get away with the `Undefined` trick, but the idea is to leave the current interface as is, for people who use `msmt_csd`, and just override the bits that need overriding. (If Undefined doesn't work, you can just use `File(desc='DO NOT SET')` or something.

To redirect CSD users to this interface, I'd update the documentation of the original to indicate that `algorithm='csd'` doesn't work, and this alternative should be used. I would then switch the default `algorithm` to `'msmt_csd'`, since apparently everybody's been running that, and raise an error pointing users to the new interface if they want to try to use `csd`.

There's actually another alternative, which is that you can add an `EstimateFOD._format_arg` to disable these for `csd`:

```Python
    if name in ("gm_odf", "csf_odf") and self.inputs.algorithm == "csd":
        return None
    super(EstimateFOD, self)._format_arg(name, trait_spec, value)
```

And you'd also need to update `_list_outputs`:

```Diff
 def _list_outputs(self):
         outputs['wm_odf'] = op.abspath(self.inputs.wm_odf)
-        if self.inputs.gm_odf != Undefined:
+        if self.inputs.gm_odf != Undefined and self.inputs.algorithm != 'csd':
             outputs['gm_odf'] = op.abspath(self.inputs.gm_odf)
-        if self.inputs.csf_odf != Undefined:
+        if self.inputs.csf_odf != Undefined and self.inputs.algorithm != 'csd':
             outputs['csf_odf'] = op.abspath(self.inputs.csf_odf)
         return outputs
```

I don't know which is more appealing to you, but either would be fine with me. I think the former is a bit more explicit, but the latter would probably be less work.

In either case, it would be good to update the doctests to show the command lines produced for both algorithms.

matteomancini:
> On the whole, we try not to break backwards compatibility under the current setup, unless it's clear that the feature never worked for anybody.

I can see your point here, but the issue is that at the moment for keeping backwards compatibility the interface to the csd algorithm is broken. I don't know if there are actual numbers on which of the algorithms is more used, but if people agree to separate the two algorithms (and if it doesn't break pipelines based on the csd algorithm), I'm happy to implement that. `Undefined` actually does the trick, I already tried it. I would be probably happier with the `_format_arg` solution (which I didn't know about), but I would like to have feedback from people who actually use this interface.

effigies:
Right, I'm not saying we should preserve backwards compatibility for the broken part, just for the parts that are already working. If `_format_arg` appeals to you, let's go in that direction. It will allow those two options to be ignored for `csd` and not break them for `msmt_csd`.

effigies:
Hi @matteomancini, we're going to be aiming for a release in about a week. Please let me know if I can be any help.

matteomancini:
Hi @effigies,
I have been travelling in the last two weeks so I missed the deadline for the release, but next week I'll be able to work on It.

effigies:
Hi @matteomancini, sorry to do this to you, but we just merged some pretty huge changes. Could you re-apply these changes on the latest `master`? You'll want to re-run `make specs` entirely, and we're now using the [Black](https://black.readthedocs.io/en/stable/installation_and_usage.html) formatter. You can run it manually, or use [pre-commit](https://pre-commit.com/) to have it automatically run every time you commit.

Please let me know if you need any help or clarifications.

matteomancini:
Hi @effigies thank you for the message. I have been swamped in the last days, but this is still on my to-do list. So should I pull the latest changes, then run make specs and push to the branch on my repo clone?

effigies:
Hi @matteomancini. Looks like your response got drowned in a flood of notifications...

Anyway, I would do the following:

```Shell
git fetch upstream
# Merge, but pause for fixes
git merge upstream/master --no-commit
# Get this branch's version of the changed file
git checkout be919a5 nipype/interfaces/mrtrix3/reconst.py
black nipype/interfaces/mrtrix3/reconst.py
git add nipype/interfaces/mrtrix3/reconst.py
# Inspect the differences to make sure nothing else was reverted.
git diff --cached upstream/master
# Assuming everything's good
git commit
git push
```

If there are other issues, you can use `git checkout -p upstream/master nipype/interfaces/mrtrix3/reconst.py` to select specific chunks to revert.

If you need to install `black`, you can use `pip install black`.

effigies:
@matteomancini I added a `ConstrainedSphericalDeconvolution` interface with your fixes. If you have further issues, let's fix those in that interface, and leave `EstimateFOD`.

PR #3008, opened 2019-08-28
ENH: Updates interface for FSL's ``eddy`` command

josephmje:
## Summary

Adds a new interface for Mrtrix3's ``mrresize`` command and updates inputs and outputs for FSL's ``eddy`` command

- MRResize added
- Mrtrix3 tests updated after `MRTrix3BaseInputSpec` modified to take either `grad_file` or `grad_fsl`
- Eddy
  - `field` now requires a file instead of string
  - add additional options to inputspec for outlier replacement, intra-volume motion correction and susceptibility-by-movement correction
  - add additional output files to outputspec (EddyQuad fails to provide outlier metrics if outlier maps are missing)

## Acknowledgment


josephmje:
This is ready to be reviewed. But I'm not sure if my change to the `field` option in Eddy should be kept since it removes backwards compatibility. Before, you'd have to pass in the fieldmap without the '.nii.gz' extension. But now, the fieldmap is passed as is and the extension is removed by `_format_arg`.

oesteban:
Hi @josephmje sorry for duplicating requests, hopefully git is making it easy for you.

Could you split this PR in two? One for the changes in eddy and the other for the new interface? That will make review and investigation of regressions much easier.

PR #2996, opened 2019-08-12
Additional arguments to ANTs N4BiasFieldCorrection

rciric:
## Summary

This PR adds some missing arguments to the `N4BiasFieldCorrection` command. The ability to toggle the `--rescale_intensities` argument is required for full implementation of a template construction workflow.

## List of changes proposed in this PR (pull-request)
* Adds `rescale_intensities` argument to `N4BiasFieldCorrection`.
* Enables nondefault parameters for `--histogram_sharpening`.

## Acknowledgment

- [x] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

PR #3011, opened 2019-08-29
ENH: Add ``--rescale-intensities`` and name_source to N4BiasFieldCorrection

oesteban:
## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -378,9 +378,9 @@ class N4BiasFieldCorrection(ANTSCommand):
     >>> n4_4.inputs.input_image = 'structural.nii'
     >>> n4_4.inputs.save_bias = True
     >>> n4_4.inputs.dimension = 3
-    >>> n4_4.cmdline
+    >>> n4_4.cmdline  # doctest: +ELLIPSIS
     'N4BiasFieldCorrection -d 3 --input-image structural.nii \
---output [ structural_corrected.nii, structural_bias.nii ]'
+--output [ structural_corrected.nii, ...structural_bias.nii ]'
     """
 
     _cmd = 'N4BiasFieldCorrection'
--- a/nipype/interfaces/ants/segmentation.py
+++ b/nipype/interfaces/ants/segmentation.py
@@ -386,9 +386,13 @@ class N4BiasFieldCorrection(ANTSCommand):
     input_spec = N4BiasFieldCorrectionInputSpec
     output_spec = N4BiasFieldCorrectionOutputSpec
 
+    def __init__(self, *args, **kwargs):
+        self._out_bias_file = None
+        super(N4BiasFieldCorrection, self).__init__(*args, **kwargs)
+
     def _format_arg(self, name, trait_spec, value):
-        if name == 'output_image' and getattr(self, '_out_bias_file', None):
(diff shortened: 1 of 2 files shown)


effigies:
I wonder if it makes sense to deprecate `save_bias`, and just always output the bias image, using `name_template` (unless someone specifically undefines it). Presumably it's already calculated, and so the cost is minimal, and it would save the contortions of `_out_bias_file`.

Failing that, I would probably make `_out_bias_file` a property that reconstructs the file as called for or a hidden attribute that's `None` by default. Either of which will avoid the explicit `getattr`, `setattr`, `delattr` calls, which would make it more readable, IMO.

oesteban:
> I wonder if it makes sense to deprecate `save_bias`, and just always output the bias image, using `name_template` (unless someone specifically undefines it). Presumably it's already calculated, and so the cost is minimal, and it would save the contortions of `_out_bias_file`.

As much as I like this idea, I think the `name_source` mechanism does not fully cover this behavior. In cases like this, it would be fine just generating the file anyways. For the general case, it would be best to allow some kind of metadata (`name_enable`) pointing to some other `traits.Bool` (in this case `save_bias`) that is used to check whether the name substitution should be executed.


> Failing that, I would probably make `_out_bias_file` a property that reconstructs the file as called for or a hidden attribute that's `None` by default. Either of which will avoid the explicit `getattr`, `setattr`, `delattr` calls, which would make it more readable, IMO.

Agreed, done.

effigies:
Have you tested? I'm not positive that you'll get an absolute path as the output.

oesteban:
> Have you tested? I'm not positive that you'll get an absolute path as the output.

0b695ae ensures the absolute path of the bias output and I have tested locally.

nipype/interfaces/ants/segmentation.py
```diff
@@ -306,6 +308,15 @@ class N4BiasFieldCorrectionInputSpec(ANTSCommandInputSpec):
         usedefault=True,
         desc='copy headers of the original image into the '
         'output (corrected) file')
+    rescale_intensities = traits.Bool(
+        False, usedefault=True, argstr='-r',
```
effigies: min_ver?

nipype/interfaces/ants/segmentation.py
```diff
@@ -418,34 +417,35 @@ def _format_arg(self, name, trait_spec, value):
...
-            skip = []
-        skip += ['save_bias', 'bias_image']
+        skip = (skip or []) + ['save_bias', 'bias_image']
+        self._out_bias_file = None
+        if self.inputs.save_bias or isdefined(self.inputs.bias_image):
+            bias_image = self.inputs.bias_image
+            if not isdefined(bias_image):
+                bias_image = fname_presuffix(os.path.basename(self.inputs.input_image),
+                                             suffix='_bias', newpath=os.getcwd())
+            self._out_bias_file = bias_image
```
effigies: I'd put the abspath here, rather than in fname_presuffix, so the bias_image case is covered as well.

oesteban:
I just realized that this overlaps with #2996 on the ``--rescale-intensities`` parameter.

rciric:
I closed mine for now -- I'll open a new one with just the other N4 parameters so that it doesn't conflict.

PR #3017, opened 2019-09-05
FIX: loadpkl failed when pklz file contained versioning info

stilley2:
## Summary

loadpkl called pickle.load at file position 0, which would fail if the first line of the file was json  versioning info (as when using savepkl with versioning=True). This was fixed by only seeking to position 0 if no versioning information is found.

## Acknowledgment


nipype/tests/test_utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+def test_pickle(tmp_path):
+    testobj = 'iamateststr'
+    pickle_fname = str(tmp_path / 'testpickle.pklz')
+    utils.filemanip.savepkl(pickle_fname, testobj)
+    outobj = utils.filemanip.loadpkl(pickle_fname)
+    assert outobj == testobj
+
+
+def test_pickle_versioning(tmp_path):
```
effigies: What about parameterizing this into a single test?

```Python
@pytest.mark.parametrize("save_ver", (True, False))
@pytest.mark.parametrize("load_ver", (True, False))
def test_pickle(tmp_path, save_ver, load_ver):
    testobj = 'iamateststr'
    pickle_fname = str(tmp_path / 'testpickle.pklz')
    utils.filemanip.savepkl(pickle_fname, testobj, versioning=save_ver)
    outobj = utils.filemanip.loadpkl(pickle_fname, versioning=load_ver)
    assert outobj == testobj
```

effigies:
@oesteban I think this issue was introduced in the previous version (#2985). I would recommend this to go into the 1.2.2 release, if it's ready in time.

nipype/tests/test_utils.py
stilley2: Good idea.

stilley2:
I actually just removed the versioning argument from loadpkl. All it was doing was altering the error reporting, and it seemed unnecessary (although maybe it was there for a reason?).

effigies:
Looks like it was added in #2626, and what it seems to be doing is guarding against accessing the undefined `pkl_metadata` in the next couple lines. I would leave it for now, just because there doesn't seem to be a big gain in removing these lines, and there will be a semantic change.

We can open an issue to review that, though, if you think it will help with clarity.

stilley2:
Ok I reverted it. However, I don't think it guards against that any more. I'll open up an issue.

nipype/tests/test_utils.py
```diff
@@ -0,0 +1,12 @@
+from nipype import utils
+import pytest
+
+
+@pytest.mark.parametrize("load_versioning", [True, False])
+@pytest.mark.parametrize("save_versioning", [True, False])
+def test_pickle(tmp_path, save_versioning, load_versioning):
```
effigies: Oh, this test should go in `nipype/utils/tests/test_filemanip.py`. Sorry, I just caught this.

stilley2: Oops I didn't see that file.

PR #3019, opened 2019-09-06
Remove versioning from loadpkl

### LINKED ISSUE #3018: Propose removing "versioning" argument from loadpkl
The "versioning" argument to loadpkl no longer serves much purpose. I propose it be deleted. Thoughts?

https://github.com/nipy/nipype/blob/508df5e0531fb691b8b052e81a3bc03d1300d594/nipype/utils/filemanip.py#L678

stilley2:
## Summary

This flag is no longer need, so this PR removes it to simplify the code.

Fixes #3018

## Acknowledgment

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -670,12 +670,12 @@ def load_json(filename):
 
 def loadcrash(infile, *args):
     if infile.endswith('pkl') or infile.endswith('pklz'):
-        return loadpkl(infile, versioning=True)
+        return loadpkl(infile)
     else:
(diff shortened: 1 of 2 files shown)


satra:
the PR looks good. 

iirc (and that's asking a lot) the pickle issue that gave rise to this may have stemmed from trying to do things across python 2 and 3. i.e., run a workflow with python 2 and then rerun using the same cache with python 3. i don't think we have a multi-environment test for this. given our upcoming plans to ditch python 2, i'm not sure it's worth spending the effort at this point.

effigies:
Sounds reasonable. Thanks for this, @stilley2.

Issue #3014, opened 2019-09-04
Error checking MapNode hash with fMRIPrep


oesteban:
### Summary

```
190904-01:20:18,695 nipype.workflow ERROR:
	 Node t1_conform failed to run on host 890949213e2e.
190904-01:20:18,696 nipype.workflow ERROR:
	 Saving crash info to /tmp/ds054/derivatives/smriprep/sub-100185/log/20190904-011955_7143d03c-6f71-442e-b03b-868465d5d7f5/crash-20190904-012018-UID1001-t1_conform-79d9866d-bc88-4848-9512-c54ff701db1b.txt
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 269, in _send_procs_to_workers
    num_subnodes = self.procs[jobid].num_subnodes()
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 1204, in num_subnodes
    self._get_inputs()
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 1219, in _get_inputs
    super(MapNode, self)._get_inputs()
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 521, in _get_inputs
    outputs = _load_resultfile(results_file)[0].outputs
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/utils.py", line 311, in load_resultfile
    if resolve and result.outputs:
AttributeError: 'NoneType' object has no attribute 'outputs'
```
https://circleci.com/gh/oesteban/smriprep/502

Related to #3009 (in that case, it occurred for a regular Node).

oesteban:
Cleaning up the cache solved the problem.

@satra - I believe this (and #3009) are an artifact of working directory degradation, but I'm worried that they could have some other source since they happen while `_get_inputs` is run. Gathering inputs from previous nodes *should* work, since their result file has been already checked (and should exist, for the case of #3009. In this case of MapNodes, I think that the outputs of the prior node got aggregated for some obscure reason.

Before I go ahead and start some refactor that can end up being a rats' nest, we wanted to touch base with you. Do you want to allocate some time to chat (cc/ @effigies )?

satra:
[...]

PR #3020, opened 2019-09-06
FIX: Ensure ``loadpkl`` returns a not None value

oesteban:
Ref.: https://github.com/nipy/nipype/issues/3014#issuecomment-528408995

## Acknowledgment

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -682,6 +682,7 @@ def loadpkl(infile):
     pklopen = gzip.open if infile.suffix == '.pklz' else open
     pkl_metadata = None
 
+    unpkl = None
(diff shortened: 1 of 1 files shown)


effigies:
Can you rebase on master, to ensure this interacts safely with #3017?

effigies:
A question: Should we be able to pickle `None`? Or is that just never a useful thing to do?

satra:
> Should we be able to pickle None? Or is that just never a useful thing to do?

yes. custom dictionary inputs to custom functions. or custom outputs from custom functions.

satra:
but never as a stand alone unit in nipype (results, nodes, and crashfiles are always not None)

oesteban:
@effigies, rebased

Issue #3009, opened 2019-08-28
fMRIprep / nipype error while checking node hash


jeroenvanbaar:
### Summary

When running fmriprep 1.5.0rc2 in a singularity container, I get the following nipype.workflow warning. It asked for opening an issue here. I don't get this error when I run fmriprep 1.4.1rc1.

### The warning/error:

```
190828-18:05:54,262 nipype.workflow WARNING:
	 Error while checking node hash, forcing re-run. Although this error may not prevent the workflow from running, it could indicate a major problem. Please report a new issue at https://github.com/nipy/nipype/issues adding the following information:

	Node: fmriprep_wf.single_subject_001_wf.func_preproc_ses_1_task_wordReading_run_4_wf.bold_reference_wf.enhance_and_skullstrip_bold_wf.map_brainmask
	Interface: niworkflows.interfaces.fixes.FixHeaderApplyTransforms
	Traceback:
Traceback (most recent call last):

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/base.py", line 337, in _local_hash_check
    cached, updated = self.procs[jobid].is_cached()

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 297, in is_cached
    hashed_inputs, hashvalue = self._get_hashval()

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 494, in _get_hashval
    self._get_inputs()

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 521, in _get_inputs
    outputs = _load_resultfile(results_file)[0].outputs

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/utils.py", line 300, in load_resultfile
    result = loadpkl(results_file)

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/utils/filemanip.py", line 686, in loadpkl
    pkl_file = pklopen(infile.name, 'rb')

  File "/usr/local/miniconda/lib/python3.7/gzip.py", line 53, in open
    binary_file = GzipFile(filename, gz_mode, compresslevel)

  File "/usr/local/miniconda/lib/python3.7/gzip.py", line 163, in __init__
    fileobj = self.myfileobj = builtins.open(filename, mode or 'rb')

[...]

PR #3023, opened 2019-09-10
FIX: Disallow returning ``None`` in ``pipeline.utils.load_resultfile``

oesteban:
## Summary
Purpose: make the handling of result files more robust.

Prevents #3009 and #3014 from happening - although this might not solve those issues,
this patch will help find their origin by making ``load_resultfile`` more strict (and
letting it raise exceptions).

The try .. except structure is moved to the only place is was being used within the Node code.

## Acknowledgment


nipype/pipeline/plugins/tools.py
```diff
@@ -26,17 +26,29 @@ def report_crash(node, traceback=None, hostname=None):
     """
     name = node._id
     host = None
-    if node.result and getattr(node.result, 'runtime'):
```
satra: we seem to have lost the ability to check if `runtime` is part of result. i think at this point, `node.result` can be `None`

oesteban: Should be fine now

oesteban:
I cached one more bug (rebase False not being honored).

This PR should be ready to merge.

Please note that I'm marking one test condition as xfail with python 2 - not sure it is worth digging up the pickling of Files with Jan 2020 this close.

PR #3024, opened 2019-09-10
ENH: Avoid loading result from file when writing reports

oesteban:
## Summary
Minimize the access to the ``result`` property when writing
pre/post-execution reports.

This modification should particularly preempt https://github.com/nipy/nipype/issues/3009#issuecomment-529799338

## Acknowledgment

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -57,11 +57,9 @@ except ImportError:
     from pathlib2 import Path
     USING_PATHLIB2 = True
 
-try:
+try: # PY35 - strict mode was added in 3.6
     Path('/invented/file/path').resolve(strict=True)
 except TypeError:
-    from tempfile import gettempdir
-
     def _patch_resolve(self, strict=False):
         """Add the argument strict to signature in Python>3,<3.6."""
         resolved = Path().old_resolve() / self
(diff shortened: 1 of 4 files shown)


oesteban:
@effigies, if you could spare one minute with this I'd really appreciate it. Especially, the new additions to make Path more consistent across 3.x versions - https://github.com/nipy/nipype/pull/3024/files#diff-7d84fc0957a520f25bb02479beb87c51R75 (see also some tests that I added).

nipype/utils/tests/test_filemanip.py
```diff
@@ -596,3 +596,21 @@ def test_pickle(tmp_path, save_versioning):
     savepkl(pickle_fname, testobj, versioning=save_versioning)
     outobj = loadpkl(pickle_fname)
     assert outobj == testobj
+
+
+def test_Path(tmpdir):
+    tmp_path = Path(tmpdir.strpath)
+
+    assert hasattr(tmp_path, 'write_text')
```
effigies: Redundant check with next line.

nipype/utils/filemanip.py
```diff
@@ -69,6 +71,24 @@ def _patch_resolve(self, strict=False):
...
     Path.resolve = _patch_resolve
+
+    if not hasattr(Path, 'write_text'):
+        def _write_text(self, text):
+            with open(str(self), 'w') as f:
+                f.write(text)
+        Path.write_text = _write_text
+
+    try:
+        (Path(gettempdir()) / 'exist_ok_test').mkdir(exist_ok=True)
```
effigies: This is just going to pollute the temporary directory on every nipype import. Might as well clean up after ourselves:

```suggestion
        with tempfile.TemporaryDirectory() as tmpdir:
            (Path(tmpdir) / 'exist_ok_test').mkdir(exist_ok=True)
```

PR #3026, opened 2019-09-11
ENH: Lightweight node cache checking

oesteban:
Generating the hashvalue when outputs are not ready at cache check stage
when the node's directory does not exist (or no results file is in
there) leads to #3014.

This PR preempts those problems by delaying the hashval calculation.

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -293,26 +293,28 @@ class Node(EngineBase):
         """
         outdir = self.output_dir()
 
-        # Update hash
-        hashed_inputs, hashvalue = self._get_hashval()
-
         # The output folder does not exist: not cached
-        if not op.exists(outdir):
-            logger.debug('[Node] Directory not found "%s".', outdir)
+        if not op.exists(outdir) or \
+                not op.exists(op.join(outdir, 'result_%s.pklz' % self.name)):
+            logger.debug('[Node] Not cached "%s".', outdir)
             return False, False
 
-        hashfile = op.join(outdir, '_0x%s.json' % hashvalue)
-        cached = op.exists(hashfile)
-
-        # Check if updated
+        # Check if there are hashfiles
         globhashes = glob(op.join(outdir, '_0x*.json'))
         unfinished = [
             path for path in globhashes
             if path.endswith('_unfinished.json')
         ]
         hashfiles = list(set(globhashes) - set(unfinished))
+
+        # Update hash
+        hashed_inputs, hashvalue = self._get_hashval()
+
+        hashfile = op.join(outdir, '_0x%s.json' % hashvalue)
         logger.debug('[Node] Hashes: %s, %s, %s, %s',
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -441,6 +443,7 @@ def run(self, updatehash=False):
             for outdatedhash in glob(op.join(self.output_dir(), '_0x*.json')):
                 os.remove(outdatedhash)
 
+        self._get_hashval()
```
satra: @oesteban - is there a specific reason for calling this here?

nipype/pipeline/engine/nodes.py
```diff
@@ -293,27 +293,29 @@ def is_cached(self, rm_outdated=False):
...
         hashfiles = list(set(globhashes) - set(unfinished))
+
+        # Update hash
+        hashed_inputs, hashvalue = self._get_hashval()
+
+        hashfile = op.join(outdir, '_0x%s.json' % hashvalue)
         logger.debug('[Node] Hashes: %s, %s, %s, %s',
                      hashed_inputs, hashvalue, hashfile, hashfiles)
 
+        cached = op.exists(hashfile)
```
effigies: Can `hashfile` be created by `self._get_hashval()`? If not, we can avoid hitting the filesystem again:

```suggestion
        cached = hashfile in hashfiles
```

nipype/pipeline/engine/nodes.py
oesteban: Yes, it was being called within `is_cached`. But now, `is_cached` exists fast if the result file is not there, etc. So, if the node's folder or the result file did not exist, flow reached here without generating the hashval.

nipype/pipeline/engine/nodes.py
oesteban: Creating the `hashfile` within `self._get_hashval` seems a good option. I'll send a different PR, so that this one involves the minimal set of changes.

effigies: Is there no case in which we might want to get the hash without generating a file?

satra: yes, often we may want to do this. so i wouldn't create the file within that call.

nipype/pipeline/engine/nodes.py
satra: thank you.

effigies: I think it'd be good to add a comment to this effect.

oesteban: Added

nipype/pipeline/engine/nodes.py
effigies: Looks like this got reverted in the force push.

oesteban:
Travis seems a bit under the weather today. I'll go ahead and merge, making sure master passes all tests.

PR #3027, opened 2019-09-12
enh: add telemetry to nipype

satra:
## Summary
Adds the ability to check for the most recent released version.

- add etelemetry as a dependency

## Acknowledgment

### DIFF
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -67,7 +67,7 @@ if config.getboolean('execution', 'check_version'):
     try:
         latest = etelemetry.get_project("nipy/nipype")
     except Exception as e:
-        logger.warning("Could not check for version updates: ", e)
+        logger.warning("Could not check for version updates: \n%s", e)
     finally:
         logger.info(INIT_MSG(packname='nipype',
                              version=__version__,
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -22,7 +22,9 @@ except (ImportError, IOError) as e:
 
 config = NipypeConfig()
 logging = Logging(config)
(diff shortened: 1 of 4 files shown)


effigies:
Is this all set?

mgxd:
@effigies let's wait to merge this until I cut a new release of etelemetry

effigies:
@mgxd I'll leave it to you to merge when you're ready.

nipype/__init__.py
```diff
@@ -56,3 +58,17 @@ def get_info():
...
 from .interfaces import (DataGrabber, DataSink, SelectFiles, IdentityInterface,
                          Rename, Function, Select, Merge)
+
+
+if config.getboolean('execution', 'check_version'):
+    import etelemetry
+
+    latest = {"version": 'Unknown'}
+    try:
+        latest = etelemetry.get_project("nipy/nipype")
```
effigies: @mgxd @satra It looks like etelemetry lets us send arbitrary kwargs. Should we add version information, so we can get a sense of which versions are still in use? e.g.,

```Python
latest = etelemetry.get_project("nipy/nipype", version=__version__)
```

mgxd: that was intended to allow altering any `requests` specific arguments (such as timeout), but we could leverage `params` to do this.

satra:
or the `.et` file in a project repo.

the plan was to store the contents of the file (a json/ld dictionary) and send that back alongside version.

effigies:
Happy to do it any way that works for you. IMO it would be nice to get do this even if etelemetry doesn't support it yet, so that 1.2.3 can start reporting as soon as etelemetry gets upgraded.

PR #3031, opened 2019-09-16
ENH: Add ``mrrtrix3.MRResize`` interface

josephmje:
## Summary
Adds a new interface for MRtrix's `mrresize` command

Splits #3008 into 2 PRs.

- MRResize added

## Acknowledgment


nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        (traits.Int, traits.Int, traits.Int),
+        argstr='-size %d,%d,%d',
+        mandatory=True,
+        desc='define the new image size for the output image. This should be '
+        'specified as a comma-separated list.',
+        xor=['voxel_size', 'scale_factor'],
+    )
+    voxel_size = traits.Tuple(
+        (traits.Float, traits.Float, traits.Float),
+        argstr='-voxel %d,%d,%d',
```
effigies: If these can be floats, we need to use `%f` or `%g`:

```suggestion
        argstr='-voxel %g,%g,%g',
```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        argstr='%s',
+        position=-2,
+        mandatory=True,
+        desc='input DWI image'
+    )
+    image_size = traits.Tuple(
+        (traits.Int, traits.Int, traits.Int),
+        argstr='-size %d,%d,%d',
+        mandatory=True,
+        desc='define the new image size for the output image. This should be '
```
effigies: I think it will read more naturally to describe the trait, rather than what the user should do. e.g.,

```suggestion
        desc='Number of voxels in each dimension of output image'
```

(I assume that's the correct interpretation of "size".)

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        'specified as a comma-separated list.',
```
effigies: This is a CLI detail not really needed for the Python user.

```suggestion
```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        xor=['voxel_size', 'scale_factor'],
+    )
+    voxel_size = traits.Tuple(
+        (traits.Float, traits.Float, traits.Float),
+        argstr='-voxel %d,%d,%d',
+        mandatory=True,
+        desc='define the new voxel size for the output image. This can be '
+        'specified either as a single value to be used for all '
+        'dimensions, or as a comma-separated list of the size for each '
+        'voxel dimension.',
```
effigies: I'd do something like: "Voxel size in mm". You're not giving an option to use a single value for isotropic (which seems fine, IMO), so most of this discussion is unnecessary.

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        xor=['image_size', 'scale_factor'],
+    )
+    scale_factor = traits.Tuple(
+        (traits.Float, traits.Float, traits.Float),
+        argstr='-scale %d,%d,%d',
```
effigies:
```suggestion
        argstr='-scale %g,%g,%g',
```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        mandatory=True,
+        desc='scale the image resolution by the supplied factor. This can be '
+        'specified either as a single value to be used for all '
+        'dimensions, or as a comma-separated list of scale factors for '
+        'each dimension.',
```
effigies: Again, something like

```Python
desc='Scale factors to rescale the image by in each dimension'
```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        xor=['image_size', 'voxel_size'],
+    )
+    interp = traits.Enum(
+        'cubic',
+        'nearest',
+        'linear',
+        'sinc',
+        argstr='-interp %s',
```
effigies: If you're going to specify a default in the description, it's good to specify it in the trait. This makes the documentation apply correctly to the interface, even if for some reason the underlying tool changes its defaults in the future.

```suggestion
        argstr='-interp %s',
        usedefault=True,
```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+        (traits.Float, traits.Float, traits.Float),
+        argstr='-scale %d,%d,%d',
+        mandatory=True,
+        desc='scale the image resolution by the supplied factor. This can be '
+        'specified either as a single value to be used for all '
+        'dimensions, or as a comma-separated list of scale factors for '
+        'each dimension.',
+        xor=['image_size', 'voxel_size'],
+    )
+    interp = traits.Enum(
```
effigies: I might call this `interpolation`. There's not a big advantage to shortening these variables, and it makes the help text more readable to use full words.

```suggestion
    interpolation = traits.Enum(
```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+
+
+class MRResize(MRTrix3Base):
+    """
+    Resize an image by defining the new image resolution, voxel size or a
+    scale factor. If the image is 4D, then only the first 3 dimensions can be
+    resized. Also, if the image is down-sampled, the appropriate smoothing is
+    automatically applied using Gaussian smoothing.
+    For more information, see
+    <https://mrtrix.readthedocs.io/en/latest/reference/commands/mrresize.html>
```
effigies: New line before the example:

```suggestion
    <https://mrtrix.readthedocs.io/en/latest/reference/commands/mrresize.html>

```

nipype/interfaces/mrtrix3/utils.py
```diff
@@ -677,3 +677,86 @@ def _list_outputs(self):
...
+    Example
+    -------
+    >>> import nipype.interfaces.mrtrix3 as mrt
+    >>> resize = mrt.MRResize()
+    >>> resize.inputs.in_file = 'dwi.mif'
+    >>> resize.inputs.voxel_size = (1, 1, 1)
+    >>> resize.cmdline                               # doctest: +ELLIPSIS
+    'mrresize -voxel 1,1,1 dwi.mif dwi_resized.mif'
+    >>> resize.run()                                 # doctest: +SKIP
+    """
```
effigies: Would you mind adding examples for `scale_factor` and `image_size`, and making `scale_factor` a non-integer? It will help ensure that the formatting works as expected.

josephmje: Will do! Thanks for the feedback

effigies:
Just as a heads up, you can test your doctests locally with:

```
pytest --doctest-modules nipype/interfaces/mrtrix3/utils.py
```

I think you're going to need to create new interfaces for each of the three examples, or you're going to get bitten by the `xor`.

josephmje:
> Just as a heads up, you can test your doctests locally with:
> 
> ```
> pytest --doctest-modules nipype/interfaces/mrtrix3/utils.py
> ```
> 
> I think you're going to need to create new interfaces for each of the three examples, or you're going to get bitten by the `xor`.

Great catch! I was also missing a coma

effigies:
LGTM. Thanks!

PR #3033, opened 2019-09-16
MAINT: Set minimum yapf for `checkspecs` to 0.27

effigies:
## Summary

`make specs` currently flails a bit from user to user, apparently because `yapf` has changed its formatting rules.

Additionally, we've been ignoring errors reported by `make specs`, probably because they weren't really clearly errors.

* Set a new minimum version of `yapf >=0.27`. I've verified that 0.27 and 0.28 both give the same result, while 0.26 gives the old result.
* Add `extensions` metadata to list of accepted metadata to reduce noise
* Remove `exists` metadata from non-file traits as well as `mandatory=False` metadata

## Acknowledgment

- [x] (Mandatory) I acknowledge that this contribution will be available under the Apache 2 license.


effigies:
Any objections?

PR #3037, opened 2019-09-18
FIX: Patch Path.mkdir for Python 2

effigies:
## Summary
This should be a blessedly short-lived hack, but it is needed to resolve nipy/heudiconv#374.

* `Path.mkdir()` was failing for Python 2, due to a PY3 guard
* Had to import `TemporaryDirectory` from `nipype.utils.tmpdirs` instead of `tempfile`
* Updated our existing `makedirs` hack to include `mode`, for full replacement of `os.makedirs`

## Acknowledgment

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -95,19 +95,18 @@ if not hasattr(Path, 'write_text'):
             f.write(text)
     Path.write_text = _write_text
 
-if PY3:
-    try:  # PY34 - mkdir does not have exist_ok
-        from tempfile import TemporaryDirectory
-        with TemporaryDirectory() as tmpdir:
-            (Path(tmpdir) / 'exist_ok_test').mkdir(exist_ok=True)
-    except TypeError:
-        def _mkdir(self, mode=0o777, parents=False, exist_ok=False):
-            if parents:
-                os.makedirs(str(self), mode=mode, exist_ok=exist_ok)
-            elif not exist_ok or not self.exists():
-                os.mkdir(str(self), mode=mode)
+try:  # PY27/PY34 - mkdir does not have exist_ok
+    from .tmpdirs import TemporaryDirectory
+    with TemporaryDirectory() as tmpdir:
+        (Path(tmpdir) / 'exist_ok_test').mkdir(exist_ok=True)
+except TypeError:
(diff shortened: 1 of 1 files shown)


nipype/utils/filemanip.py
```diff
@@ -828,7 +827,7 @@ def dist_is_editable(dist):
     return False
 
 
-def makedirs(path, exist_ok=False):
+def makedirs(path, mode=0o777, exist_ok=False):
```
mgxd: Just to be safe, I'd add it to the end of the signature to avoid any wrong assignments. I can't wait til we move to py3 only to support keyword only args

effigies: This keeps the protocol consistent with `os.makedirs` in all versions:

```Shell
$ pydoc2.7 os.makedirs | cat
Help on function makedirs in os:

os.makedirs = makedirs(name, mode=511)
    makedirs(path [, mode=0777])

$ pydoc os.makedirs | cat 
Help on function makedirs in os:

os.makedirs = makedirs(name, mode=511, exist_ok=False)
    makedirs(name [, mode=0o777][, exist_ok=False])
```

We only use this function with the keyword:

```
$ grep -rI makedirs\( nipype | grep -v os\.makedirs
nipype/pipeline/plugins/tools.py:    makedirs(crashdir, exist_ok=True)
nipype/pipeline/engine/utils.py:    makedirs(base_dir, exist_ok=True)
nipype/pipeline/engine/workflows.py:        base_dir = makedirs(base_dir, exist_ok=True)
nipype/pipeline/engine/workflows.py:        makedirs(report_dir, exist_ok=True)
nipype/pipeline/engine/nodes.py:        makedirs(outdir, exist_ok=True)
nipype/pipeline/engine/nodes.py:            makedirs(outdir, exist_ok=True)
nipype/utils/filemanip.py:            makedirs(str(self), mode=mode, exist_ok=exist_ok)
nipype/utils/filemanip.py:def makedirs(path, mode=0o777, exist_ok=False):
nipype/utils/filemanip.py:    makedirs(path)
```

mgxd: works for me - looks good

oesteban:
Not sure this will solve https://github.com/nipy/heudiconv/issues/374 though.

effigies:
It did.

PR #3038, opened 2019-09-19
REL: 1.2.3

effigies:
## Summary

Prep for 1.2.3 release, targeting Monday, September 23.

This is a regularly scheduled release, and the last to support Python 3.4.

Please see the [1.2.3 milestone](https://github.com/nipy/nipype/milestone/35) for issues/PRs to resolve or postpone. If there's a blocking issue please comment.

For issues/PRs that will require additional time, please move them to the [1.3.0 milesteone](https://github.com/nipy/nipype/milestone/34), scheduled for Monday, October 25.

## Release checklist

* [x] Update changelog
* [x] Update .mailmap
* [x] Update .zenodo.json
* [x] Set release number in `nipype/info.py` and `doc/conf.py`
* [x] Update `doc/documentation.rst` with previous releases
* [ ] Check conda-forge feedstock build (conda-forge/nipype-feedstock#51)
* [ ] Tutorial tests (@djarecka)

## Uncredited authors

The following authors have contributed, but not added themselves to the [`.zenodo.json`](https://github.com/nipy/nipype/blob/master/.zenodo.json) file. If you would like to be an author on Zenodo releases, please add yourself or comment with your preferred publication name, affiliation and [ORCID](https://orcid.org/). If you would like to stop being spammed whenever I'm the one doing releases, let me know, and I'll add you to a blacklist.

No entry to sort: Gio Piantoni (@gpiantoni)
No entry to sort: Victor Férat (@vferat)
No entry to sort: Isaiah Norton (@ihnorton)
No entry to sort: Niklas Förster (@niklasfoe)
No entry to sort: daniel glen (@afni-dglen) 
No entry to sort: AKSoo (@AKSoo)
No entry to sort: Kirstie Whitaker (@KirstieJane)
No entry to sort: hstojic (@hstojic)
No entry to sort: Jonathan R. Williford (@williford)
No entry to sort: Hao (@anbai106)
No entry to sort: Olivia Stanley (@ostanley)
No entry to sort: Fred Mertz
No entry to sort: Kevin Sitek (@sitek)
No entry to sort: Ami Tsuchida
No entry to sort: Abel González (@abelalez)
No entry to sort: Michiel Cottaar (@MichielCottaar)
No entry to sort: Daniel Brenner (@brennerd11)

I don't think Fred Mertz is a real name, and I can't quickly find an Ami Tsuchida on GitHub, but I think I've mapped everybody else to their GitHub users.

## Acknowledgment

[...]

PR #3040, opened 2019-09-19
add name to .zenodo.json

AKSoo:
## Acknowledgment


effigies:
Cherry-picked into #3038. I also updated the .mailmap so your name will appear as `Kim, Sin` in `git log` outputs (this is what we use to sort Zenodo entries). If you would prefer `Sin Kim`, please let me know.

PR #3043, opened 2019-09-20
MAINT: Revision of Zenodo update script

oesteban:
Unlocks the first position as per private conversation with @chrisgorgo.


oesteban:
This should do :)

oesteban:
Canceling CircleCI not to overload it

effigies:
@oesteban Do you want to target the `rel/1.2.3` branch and run the updated script to reorder? That would be a good demonstration that it works.

effigies:
Included in #3038.

Issue #3046, opened 2019-09-23
test_dcm2niix_dwi failing intermittently


effigies:
### Summary

Merged #3038 and released 1.2.3 this morning, and new errors have popped up.

* Python 3.5 on Travis [now exhibits](https://travis-ci.org/nipy/nipype/builds/588507288):

```
[gw0] [ 96%] PASSED utils/tests/test_misc.py::test_rgetcwd INTERNALERROR> Traceback (most recent call last):

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/_pytest/main.py", line 191, in wrap_session

INTERNALERROR>     session.exitstatus = doit(config, session) or 0

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/_pytest/main.py", line 235, in _main

INTERNALERROR>     config.hook.pytest_runtestloop(session=session)

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/hooks.py", line 286, in __call__

INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/manager.py", line 92, in _hookexec

INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/manager.py", line 86, in <lambda>

INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/callers.py", line 208, in _multicall

INTERNALERROR>     return outcome.get_result()

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/callers.py", line 80, in get_result

INTERNALERROR>     raise ex[1].with_traceback(ex[2])

INTERNALERROR>   File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/callers.py", line 187, in _multicall

INTERNALERROR>     res = hook_impl.function(*args)

[...]

PR #3049, opened 2019-09-24
ENH: Delay etelemetry for non-interactive sessions, report bad versions

satra:
## Summary
Improves version checking

this will require the server to support the .et file in a project directory

- check for return from server and do not print anything if there is no change in version
- checking is it's own function and raises exception if asked for bad versions

## Acknowledgment


nipype/__init__.py
```diff
@@ -60,15 +60,55 @@ def get_info():
                          Rename, Function, Select, Merge)
 
 
-if config.getboolean('execution', 'check_version'):
+def sys_based_cache(condition):
+    def decorator(func):
+        if condition:
+            return func
```
effigies: This means that if somebody's using Python 2, we're going to ping on every version. Since we're only worrying about one function with one boolean argument for a couple more releases, we can do a very quick-and-dirty memoization:

```Python
class _memoize(object):
    def __init__(self, func):
        self.cache = {}
        self.func = func
    def __call__(self, *args, **kwargs):
        argtuple = (args,) + tuple(kwargs.items())
        if argtuple not in self.cache:
            self.cache[argtuple] = self.func(*args, **kwargs)
        return self.cache[argtuple]
```

Another option is to manage a global state ourselves

```Python
etelemetry_results = {}

def check_latest_version(raise_exception=False):
    if raise_exception not in etelemetry_results:
        ...  # Current function
        etelemetry_results[raise_exception] = latest
    return etelemetry_results[raise_exception]
```

Or we could use something like [memoization](https://github.com/lonelyenvoy/python-memoization), which is surely a big hammer, but won't add code that will be most sensibly removed in two months.

satra:
the failing test is #3046 

we should merge this in before a release - and also fix up the bad_versions keys

nipype/__init__.py
effigies: Sorry, thinko. That should be "ping on every interface/node/workflow instantiation", in case your brain didn't make the leap. I think it's a pretty major drawback, although I don't know how many Python 2 users we still have.

nipype/__init__.py
```diff
@@ -5,6 +5,8 @@
                         absolute_import)
 
 import os
+import functools
+import sys
```
effigies: Unused imports.

nipype/__init__.py
```diff
@@ -59,16 +59,51 @@ def get_info():
...
+                    logger.critical(message)
+    etelemetry_results[raise_exception] = latest
+    return latest
+
+# Run telemetry on import for interactive sessions, such as IPython, Jupyter notebooks, Python REPL
+if config.getboolean('execution', 'check_version'):
+    import __main__
+    if not hasattr(__main__, '__file__'):
+        from .interfaces.base import BaseInterface
+        BaseInterface.check_version = check_latest_version()
```
effigies: We're kind of doubling-up on memoization now by including it in `etelemetry_results[False]` and `BaseInterface.check_version`. What's the advantage?

nipype/interfaces/base/core.py
```diff
@@ -166,9 +166,15 @@ class BaseInterface(Interface):
     _redirect_x = False
     references_ = []
     resource_monitor = True  # Enabled for this interface IFF enabled in the config
+    check_version = None
```
effigies: Might choose a different name, as `version` attached to an `Interface` feels like it should apply to the interface itself, not nipype. Maybe:

```suggestion
    _etelemetry_version_data = None  # Dictionary to be populated on first interface init
```

nipype/__init__.py
satra: none - just lazily did not remove :)

nipype/__init__.py
```diff
@@ -60,15 +58,48 @@ def get_info():
...
+                    any([LooseVersion(__version__) == LooseVersion(ver)
+                         for ver in latest["bad_versions"]]):
+                message = ('You are using a version of Nipype with a critical '
+                           'bug. Please use a different version.')
+                if raise_exception:
+                    raise RuntimeError(message)
+                else:
+                    logger.critical(message)
+        else:
+            latest = None
```
effigies: This is going to cause repeated warnings on systems lacking network access.

.et
```diff
@@ -0,0 +1,3 @@
+{ "bad_versions" : [ "1.2.1",
+                     "1.2.3"]
```
satra: this should be checked and updated before release.

effigies: I think these are legitimately buggy. We can really only plausibly check back to 1.3.0, since this functionality will be missing before that. Is it worth looking for critical bugs in earlier [release logs](https://github.com/nipy/nipype/releases)?

satra: this was more about whether even these two were correct. from a usage perspective, nothing will change till the server starts reading these `.et` files and returning the info. i'll update the server soon.

satra:
@mgxd @effigies - fyi

```
$ git co 88d479b
Note: checking out '88d479b'.

HEAD is now at 88d479b69 Merge pull request #3093 from nipy/rel/1.3.0

$ ipython
Python 3.7.5 (default, Oct 25 2019, 10:52:18) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import nipype                                                                                                                      
191119-17:14:19,512 nipype.utils INFO:
	 Running nipype version 1.3.0 (latest: 1.3.1)
191119-17:14:19,512 nipype.utils CRITICAL:
	 You are using a version of Nipype with a critical bug. Please use a different version.

```

PR #3054, opened 2019-09-26
ENH: Add "ExportFile" interface as simple alternative to "DataSink"

stilley2:
## Acknowledgment


effigies:
Hi @stilley2, could you clarify the reasoning here? Is the idea to use this as a kind of cheap `DataSink`? In general, symlinking should be a safe operation within the workflow.

stilley2:
Hi @effigies , yeah that's exactly what I was thinking. Specifically it could be used when output file name generation is handled by an external program (e.g. pybids). Perhaps a new interface would be clearer than hijacking Rename for this purpose?

effigies:
> Perhaps a new interface would be clearer than hijacking Rename for this purpose?

I think so. It may be that `DataSink` will actually be enough if it's a single file, but you could easily write a simple `ExportFile` (or whatever name seems appropriate) interface that really does just copy one file with a given name.

Given that you're looking at using pybids, you could also check out [`niworkflows.interfaces.bids.DerivativesDataSink`](https://github.com/poldracklab/niworkflows/blob/6118af2d8ef59abfa79f665da7caad646c40bc5c/niworkflows/interfaces/bids.py#L233-L506) or [`fitlins.interfaces.bids.BIDSDataSink`](https://github.com/poldracklab/fitlins/blob/d3d0d983b18ae155c91a7345579d7c94019c3fdd/fitlins/interfaces/bids.py#L494-L544), to see if those meet your needs. It might be time to consider upstreaming these to nipype, if they're usable outside our specific contexts.

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
...
+class ExportFileOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='Output file name')
+
+
+class ExportFile(BaseInterface):
+    input_spec = ExportFileInputSpec
+    output_spec = ExportFileOutputSpec
+
+    def _run_interface(self, runtime):
+        if not self.inputs.clobber and op.exists(self.inputs.out_file):
```
effigies: I suspect you're going to want to require that `out_file` be an absolute path?

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
 
     def _add_output_traits(self, base):
         return add_traits(base, list(self.inputs.output_query.keys()))
+
+
+class ExportFileInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, desc='Input file name')
+    out_file = File(exists=False, desc='Output file name')
+    check_extension = traits.Bool(False, desc='Ensure that the input and output file extensions match')
```
effigies: Somewhat your call, but I would probably set the default for this to `True`, and require users to turn it off to show they understand that the extension is changing. Otherwise you might get situations where a `.nii` file is copied to ` .nii.gz` file, and most software will get confused.

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
...
+    clobber = traits.Bool(False, desc='Permit overwriting existing files')
```
effigies: The default is already `False`.

```suggestion
    clobber = traits.Bool(desc='Permit overwriting existing files')
```

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
 
     def _add_output_traits(self, base):
         return add_traits(base, list(self.inputs.output_query.keys()))
+
+
+class ExportFileInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, desc='Input file name')
+    out_file = File(exists=False, desc='Output file name')
```
effigies: The default is `False`.

```suggestion
    out_file = File(mandatory=True, desc='Output file name')
```

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
 
     def _add_output_traits(self, base):
         return add_traits(base, list(self.inputs.output_query.keys()))
+
+
+class ExportFileInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, desc='Input file name')
```
effigies:
```suggestion
    in_file = File(exists=True, mandatory=True, desc='Input file name')
```

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
...
+    out_file = File(exists=True, desc='Output file name')
+
+
+class ExportFile(BaseInterface):
+    input_spec = ExportFileInputSpec
+    output_spec = ExportFileOutputSpec
+
+    def _run_interface(self, runtime):
+        if not self.inputs.clobber and op.exists(self.inputs.out_file):
+            raise FileExistsError(errno.EEXIST, f'File {self.inputs.out_file} exists')
```
effigies: f-strings are 3.6+, we support down to 2.7 and 3.4 for now, and will move to 3.5+ in a couple releases.

```suggestion
            raise FileExistsError(errno.EEXIST, 'File %s exists' % self.inputs.out_file)
```

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
...
+        if (self.inputs.check_extension and
+                op.splitext(self.inputs.in_file)[1] != op.splitext(self.inputs.out_file)[1]):
```
effigies: Please use [`nipype.utils.filemanip.split_filename`](https://github.com/nipy/nipype/blob/0fa28d5b6000ddaa513301d98322d48e6fe3c642/nipype/utils/filemanip.py#L107-L155) over `os.path.splitext`.

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,33 @@ def _list_outputs(self):
...
+            raise RuntimeError(f'{self.inputs.in_file} and {self.inputs.out_file} have different extensions')
```
effigies: f-string

effigies:
Ah, you have a `FileExistsError`, which didn't exist prior to Python 3.5. If you want to use it, could you add it to `utils.filemanip`, along with `FileNotFoundError`?

https://github.com/nipy/nipype/blob/1ccd70bd124975c936fe1bf9e2cef6e03a0d980b/nipype/utils/filemanip.py#L43-L52

There's also a Python 3.5 issue, which I *suspect* will be fixed if you merge/rebase master. I would certainly try that before we start digging deeper.

nipype/interfaces/tests/test_io.py
```diff
@@ -680,3 +680,26 @@ def _mock_get_ssh_client(self):
             .check(file=True, exists=True)) # exists?
 
     old_cwd.chdir()
+
+
+def test_ExportFile(tmp_path):
+    testin = tmp_path / 'in.txt'
+    testin.write_text('test string')
+    i = nio.ExportFile()
+    i.inputs.in_file = testin
```
effigies:
```suggestion
    i.inputs.in_file = str(testin)
```

nipype/interfaces/tests/test_io.py
```diff
@@ -680,3 +680,26 @@ def _mock_get_ssh_client(self):
...
+    i.inputs.out_file = tmp_path / 'out.tsv'
```
effigies:
```suggestion
    i.inputs.out_file = str(tmp_path / 'out.tsv')
```

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,31 @@ def _list_outputs(self):
...
+    out_file = File(mandatory=True, desc='Output file name')
+    check_extension = traits.Bool(True, desc='Ensure that the input and output file extensions match')
+    clobber = traits.Bool(desc='Permit overwriting existing files')
+
+
+class ExportFileOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='Output file name')
+
+
+class ExportFile(SimpleInterface):
```
effigies: Could you make a docstring that explains what you're doing here and gives an example or two? The other interfaces in this file are good examples, if a little inconsistently formatted. I would go with the formatting in:

https://github.com/nipy/nipype/blob/47bb8e8376d1847fc377d41ddf82a1911227ed79/nipype/interfaces/io.py#L1297-L1327

I'm sorry to draw this process out... I'm not sure why I didn't bring this up before.

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,52 @@ def _list_outputs(self):
...
+    This interface copies an input file to a named output file.
+    This is useful to save files to a specified location
+    (as opposed to using DataSink).
+
+    Examples
+    --------
+
+    A trivial example that copies temporary_file.nii.gz
+    to sub1_out.nii.gz. (A more realistic example would set
+    in_file as the output of another Node.)
```
effigies: This feels like an unnecessary qualifier. I think you could probably just skip this text, unless you want to have multiple tests showing off `check_extension` and `clobber`. If so, you could say "By default, file extensions need to match, and the output file should not exist:" and something similar above the other examples.

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,52 @@ def _list_outputs(self):
...
+
+    >>> from nipype.interfaces.io import ExportFile
+    >>> import os.path as op
+    >>> ef = Node(ExportFile(), "export")
+    >>> ef.inputs.in_file = "temporary_file.nii.gz"
+    >>> ef.inputs.out_file = op.abspath("output_folder/sub1_out.nii.gz")
```
effigies: Your interface doesn't make directories that don't exist, so you're going to either need to target an existing directory or `mkdir` one. You can just `os.mkdir('output_folder')` to continue using this example.

stilley2:
Thanks @effigies . Sorry for the delay. Busy week.

nipype/interfaces/io.py
```diff
@@ -2863,3 +2864,49 @@ def _list_outputs(self):
...
+    Examples
+    --------
+
+    >>> from nipype.interfaces.io import ExportFile
+    >>> import os.path as op
+    >>> ef = ExportFile()
+    >>> ef.inputs.in_file = "T1.nii.gz"
+    >>> os.mkdir("output_folder")
+    >>> ef.inputs.out_file = op.abspath("output_folder/sub1_out.nii.gz")
+    >>> ef.run()
```
effigies: This returns an `InterfaceResult` object.

```suggestion
    >>> res = ef.run()
```

And maybe you can follow it up with an:

```Python
>>> os.path.exists(res.outputs.out_file)
True
```

nipype/interfaces/tests/test_io.py
```diff
@@ -680,3 +680,26 @@ def _mock_get_ssh_client(self):
...
+    i = nio.ExportFile()
+    i.inputs.in_file = str(testin)
+    i.inputs.out_file = str(tmp_path / 'out.tsv')
+    i.inputs.check_extension = True
+    with pytest.raises(RuntimeError):
+        i.run()
+    i.inputs.check_extension = False
+    i.run()
+    assert (tmp_path / 'out.tsv').read_text() == 'test string'
+    i.inputs.out_file = tmp_path / 'out.txt'
```
effigies:
```suggestion
    i.inputs.out_file = str(tmp_path / 'out.txt')
```

effigies:
No worries. This looks good. Two minor issues should fix up the tests, I think.

effigies:
Can you run `make specs` and commit `nipype/interfaces/tests/test_auto_ExportFile.py`?

PR #3059, opened 2019-10-01
tst: use pytest fixture

mgxd:
Trying again after #3058 - otherwise we should xfail it for the time being.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/tests/test_extra_dcm2nii.py
+++ b/nipype/interfaces/tests/test_extra_dcm2nii.py
@@ -14,42 +14,45 @@ except ImportError:
 DICOM_DIR = 'http://datasets-tests.datalad.org/dicoms/dcm2niix-tests'
 
 
-def fetch_data(datadir, dicoms):
-    """Fetches some test DICOMs using datalad"""
-    api.install(path=datadir, source=DICOM_DIR)
-    data = os.path.join(datadir, dicoms)
-    api.get(path=data)
-    return data
+@pytest.fixture
+def fetch_data():
+    def _fetch_data(datadir, dicoms):
+        try:
+            """Fetches some test DICOMs using datalad"""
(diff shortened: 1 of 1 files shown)


effigies:
Should we retry, to improve the odds of completion? If it's really hit-or-miss, this is likely to be a constant source of noise in coverage reports.

nipype/interfaces/tests/test_extra_dcm2nii.py
```diff
@@ -14,42 +14,45 @@
...
 
-def fetch_data(datadir, dicoms):
-    """Fetches some test DICOMs using datalad"""
-    api.install(path=datadir, source=DICOM_DIR)
-    data = os.path.join(datadir, dicoms)
-    api.get(path=data)
-    return data
+@pytest.fixture
+def fetch_data():
+    def _fetch_data(datadir, dicoms):
```
effigies:
```suggestion
    def _fetch_data(datadir, dicoms, retry=3):
```

nipype/interfaces/tests/test_extra_dcm2nii.py
```diff
@@ -14,42 +14,45 @@
...
+        try:
+            """Fetches some test DICOMs using datalad"""
+            api.install(path=datadir, source=DICOM_DIR)
+            data = os.path.join(datadir, dicoms)
+            api.get(path=data)
+        except IncompleteResultsError as exc:
+            pytest.skip("Failed to fetch test data: %s" % str(exc))
```
effigies:
```suggestion
            if retry:
                return _fetch_data(datadir, dicoms, retry=retry-1)
            pytest.skip("Failed to fetch test data: %s" % str(exc))
```

effigies:
Merging. If the codecov thing becomes an issue, we can add in the retry, but the stochastic failures are a bigger problem right now.

PR #3060, opened 2019-10-01
FIX: Remove asynchronous chdir callback

stilley2:
## Summary

Remove the asynchronous chdir callback in the multiproc plugin

## Acknowledgment

### DIFF
--- a/nipype/pipeline/plugins/multiproc.py
+++ b/nipype/pipeline/plugins/multiproc.py
@@ -146,8 +146,6 @@ class MultiProcPlugin(DistributedPluginBase):
         self._stats = None
 
     def _async_callback(self, args):
-        # Make sure runtime is not left at a dubious working directory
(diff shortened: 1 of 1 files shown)


effigies:
The problem that led to this, though, was something calling `getcwd()` from a deleted directory. Has that root cause gone away?

oesteban:
It should be. Virtually any `getcwd()` call from a worker that was initiated in a nonexistent folder would crash. But that problem is addressed by the init hook included in that old PR. The chdir being removed is the result of my desperation at that moment.

effigies:
Okay. I'll defer to your memory on that one.

effigies:
@oesteban I was assuming you were going to merge this. Is there a hold-up, or were you waiting for a second approval?

oesteban:
Sorry for the delay

PR #3047, opened 2019-09-24
fix: remove switching directory to load results

satra:
## Summary

Fixes #3014 . @stilley2 - quick change to test directory changes.

## List of changes proposed in this PR (pull-request)
- remove indirectory

## Acknowledgment

- [X] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

PR #3061, opened 2019-10-02
FIX: Minimize scope for directory changes while loading results file

effigies:
## Summary
This PR attempts to handle the results file failures by separating concerns.

1) Load the results file without changing directories. We are pretty consistently seeing "No file named `results_<whatever>.pklz`, which at the minimum is uninformative. There's no reason not to know the full path of the file when we try to open it.

2) Perform `pickle.loads` on the file contents as a single-line `with indirectory` context block.

3) All other operations are directory independent and file-lock independent.

Alternative to #3047.

Sorry to preempt that process, but we're seeing this a lot, and I don't really know how to help move that one along. I'm going to use this branch to see if we can resolve these issues that are popping up a lot in the fMRIPrep tests.

## Acknowledgment


nipype/utils/filemanip.py
```diff
@@ -699,37 +699,49 @@ def loadpkl(infile):
     infile = Path(infile)
     fmlogger.debug('Loading pkl: %s', infile)
     pklopen = gzip.open if infile.suffix == '.pklz' else open
+
+    with SoftFileLock('%s.lock' % infile.name):
+        with pklopen(str(infile), 'rb') as pkl_file:
+            pkl_contents = pkl_file.read()
```
oesteban:
```suggestion
        pkl_contents = infile.read_bytes()
```
although we can wait until only Python +3.5 is supported.

effigies: Yeah, I feel like we've been bitten by trying to jump ahead of ourselves. I'd rather stick with idioms universal to our support window.

effigies: Oh, and actually that wouldn't work, since we need to use `gzip.open` for `pklz` files anyway.

effigies:
dcm2niix bug cropping up again. Merging master. Please see https://github.com/poldracklab/fmriprep/pull/1800 and https://github.com/poldracklab/fmriprep/pull/1803 for fixed tests.

PR #3067, opened 2019-10-07
RF: Redirect nipype.workflows to niflow.nipype1.workflows

### LINKED ISSUE #2991: Replace nipype.workflows with a shim that imports niflow-nipype1-workflows
I've released https://github.com/niflows/nipype1-workflows/releases/tag/0.0.1. For a minor release, at least two releases in the future, we should replace `nipype.workflows` with a shim that does something like:

```Python
try:
    from niflow.nipype1.workflows import *
except ImportError:
    logger.warning("These workflows have been moved to the niflow-nipype1-workflows package. See https://github.com/nipy/nipype/issues/2991.")
```

effigies:
## Summary
Imports from [niflow-nipype1-workflows](https://github.com/niflows/nipype1-workflows/), adding the available modules to `sys.modules` under their old names, so that code that currently imports via `import nipype.workflows.X` or `from nipype.workflows.Y import Z` should continue to work.

Closes #2991.

## Acknowledgment


effigies:
I feel like this would be a good one to get in for RC1, as it has the chance to break things for people, so finding those before the final release would be good.

satra:
i like it and a question.

how are we dealing with circularity? we are installing nipype and these niflows require nipype.

effigies:
Niflows depend on nipype, nipype will detect them at runtime, so there isn't an installation circularity.

For the purpose of testing on docker, it installs the latest release and then installs the current source on top of that. We could just as easily post-install the niflows.

effigies:
Yeah, that would be circular, though it's possible pip is more relaxed for extras. I also don't want to encourage people to continue using this API, so making it a little difficult is fine with me.

PR #3072, opened 2019-10-08
Allow nipype.cfg in cwd to be read even if ~/.nipype does not exist

stilley2:
Fixes a bug where the nipype.cfg in the current directory would not
be read if the folder ~/.nipype does not exist.

## Acknowledgment


effigies:
Please merge/rebase master to fix the tests.

PR #3073, opened 2019-10-08
ENH: Allow Nipype configuration directory to be specified with NIPYPE_CONFIG_DIR environment variable

stilley2:
## Summary

Allow overwriting default location of nipype.cfg (~/.nipype) with the environment variable NIPYPECONFDIR. This is potentially useful for containerized pipelines which require a default nipype.cfg.

## Acknowledgment


stilley2:
Sorry @oesteban , I'm not sure exactly what you mean. Do you want me to add some tests for this?

effigies:
I'm not sure what a good test would be. I guess you could mock the environment and validate that the expected contents are loaded when the variable is set or not set.

In any event, you should add a note about the environment variable to, once this is merged: https://github.com/miykael/nipype_tutorial/blob/master/notebooks/basic_execution_configuration.ipynb

effigies:
Hi @stilley2, this PR is also affected by our recent merges. Could you re-apply these changes on the latest `master`? You'll want to re-run `make specs` entirely, and we're now using the [Black](https://black.readthedocs.io/en/stable/installation_and_usage.html) formatter. You can run it manually, or use [pre-commit](https://pre-commit.com/) to have it automatically run every time you commit.

Please let me know if you need any help or clarifications.

nipype/utils/config.py
```diff
@@ -91,7 +91,7 @@ def __init__(self, *args, **kwargs):
         self._config = configparser.ConfigParser()
         self._cwd = None
 
-        config_dir = os.path.expanduser("~/.nipype")
+        config_dir = os.path.expanduser(os.getenv("NIPYPECONFDIR", default="~/.nipype"))
```
effigies: How do you feel about:

```suggestion
        config_dir = os.path.expanduser(os.getenv("NIPYPE_CONFIG_DIR", default="~/.nipype"))
```

effigies:
Hi @stilley2, do you have a couple minutes to consider the above proposal?

stilley2:
Sorry last week was pretty busy. Change looks good!

effigies:
Thanks for your patience.

PR #3075, opened 2019-10-11
ENH: Minimize the number of calls to ``_load_results`` when populating inputs

oesteban:
This PR attempts to alleviate #3014 by opening the result file of a source node only once when that node feeds into several inputs of the node collecting inputs. Before these changes, a call to ``_load_results`` was issued for every input field that needed to collect its inputs from a past node. Now, all the inputs coming from the same node are put together and the ``_load_results`` function is called just once. 

The PR also modifies the manner the ``AttributeError``s (#3014) were handled to make it easier to spot whether an error occurred while loading results araises when gathering the inputs of a node-to-be-run or elsewhere.

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -516,15 +516,20 @@ directory. Please ensure no other concurrent workflows are racing""", hashfile_u
         This mechanism can be easily extended/replaced to retrieve data from
         other data sources (e.g., XNAT, HTTP, etc.,.)
         """
-        if self._got_inputs:
+        if self._got_inputs:  # Inputs cached
+            return
+
+        if not self.input_source:  # No previous nodes
+            self._got_inputs = True
             return
 
         prev_results = defaultdict(list)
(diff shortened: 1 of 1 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -510,49 +510,55 @@ def _get_hashval(self):
...
-            output_value = Undefined
-            if isinstance(info[1], tuple):
-                output_name = info[1][0]
-                value = getattr(outputs, output_name)
-                if isdefined(value):
-                    output_value = evaluate_connect_function(
-                        info[1][1], info[1][2], value)
-            else:
-                output_name = info[1]
+Error populating the inpus of node "%s": the results file of the source node \
```
oesteban:
```suggestion
Error populating the inputs of node "%s": the results file of the source node \
```

just caught this typo. will fix with your review suggestions.

oesteban:
I'm going to give it a quick try locally with fMRIPrep before merging in.

PR #3083, opened 2019-10-26
ENH: Always fsync pickle files

### LINKED ISSUE #3014: Error checking MapNode hash with fMRIPrep

### LINKED ISSUE #3076: Strange results file checking behavior that breaks workflows at nested mapnodes in nipype 1.2.3
### Summary
Nested MapNodes hang in some workflows when trying to load the node's result file. It appears this is because results.outputs.get() fails since there is a Bunch representation of outputs, rather than a dictionary representation.

### Manifests as follows:
At the logging debug level during execution:
```
Outputs object of loaded result /private/tmp/work/20191012-164048_df69a011-5615-4010-aa22-d4a0c94c5696_wf_single_subject_fmri_0025427/wf_single_sub_0025427_fmri_20191012-164047/meta_wf_0025427/fmri_connectometry_0025427/thresh_func_node/result_thresh_func_node.pklz is a Bunch.
```

and then  if I attempt to load the file manually:
```
In [32]: res.outputs.get()                                                                                                                                                     
TypeError                                 Traceback (most recent call last)
<ipython-input-32-347df97b2ed0> in <module>
----> 1 res.outputs.get()

/usr/local/anaconda3/lib/python3.7/site-packages/nipype-2.0.0.dev0+g7b9c63932-py3.7.egg/nipype/interfaces/base/support.py in get(self, *args)

TypeError: get expected at least 1 arguments, got 0
```
See attached example results file:
[result_thresh_func_node.pklz.zip](https://github.com/nipy/nipype/files/3721280/result_thresh_func_node.pklz.zip)

Interestingly, it seems that `load_resultfile` from `nipyp/epipeline/engine/utils.py` fails, but only for the result file of the primary node (i.e. and not its subnodes if they get parameterized)

Appears to only occurs as of version nipype 1.2.3. Looks like this might also be related to #3009 #3014 #3075 and/or #1798 of fmriprep.

Any ideas @oesteban ?

@dPys

### LINKED ISSUE #3081: No such file or directory: 'result_bidssrc.pklz'
Hello everyone,

### Summary
I am running fmriprep on a remote server.

### Actual behavior
During running, I have several time the same error:

"
Error while checking node hash, forcing re-run. Although this error may not prevent the workflow from running, it could indicate a major problem. Please report a new issue at https://github.com/nipy/nipype/issues adding the following information:

Node: fmriprep_wf.single_subject_p001_wf.bids_info
	Interface: niworkflows.interfaces.bids.BIDSInfo
	Traceback:

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/base.py", line 337, in _local_hash_check
    cached, updated = self.procs[jobid].is_cached()

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 297, in is_cached
    hashed_inputs, hashvalue = self._get_hashval()

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 494, in _get_hashval
    self._get_inputs()

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 521, in _get_inputs
    outputs = _load_resultfile(results_file)[0].outputs

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/utils.py", line 300, in load_resultfile

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/utils/filemanip.py", line 687, in loadpkl
    pkl_file = pklopen(infile.name, 'rb')

  File "/usr/local/miniconda/lib/python3.7/gzip.py", line 53, in open

  File "/usr/local/miniconda/lib/python3.7/gzip.py", line 163, in __init__
    fileobj = self.myfileobj = builtins.open(filename, mode or 'rb')

FileNotFoundError: [Errno 2] No such file or directory: 'result_bidssrc.pklz'
"

### Script/Workflow details
The main command is: docker run --rm -e DOCKER_VERSION_8395080871=19.03.2 -v /opt/freesurfer/license.txt:/opt/freesurfer/license.txt:ro -v /path_to_data:/data:ro -v /path_to_data_output:/out poldracklab/fmriprep:1.5.0 /data /out participant --participant_label sub-p001 --fs-no-reconall

Thank you very much in advance!

effigies:
## Summary
Fully qualifying results file names does not seem to have resolved the issue with finding the files, making it more likely that this is a filesystem synchronization issue. This PR adds an `fsync` call for Pickle files, specifically.

This required a slight refactor for handling gzip (`pklz`) files.

Hopefully:

* Fixes #3014
* Fixes #3076 
* Fixes #3081

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -236,17 +236,17 @@ def save_resultfile(result, cwd, name, rebase=None):
 
     if result.outputs is None:
         logger.warning('Storing result file without outputs')
-        savepkl(resultsfile, result)
+        savepkl(resultsfile, result, sync=True)
         return
     try:
         output_names = result.outputs.copyable_trait_names()
     except AttributeError:
         logger.debug('Storing non-traited results, skipping rebase of paths')
-        savepkl(resultsfile, result)
+        savepkl(resultsfile, result, sync=True)
         return
 
     if not rebase:
-        savepkl(resultsfile, result)
+        savepkl(resultsfile, result, sync=True)
         return
 
     backup_traits = {}
@@ -262,7 +262,7 @@ def save_resultfile(result, cwd, name, rebase=None):
                     backup_traits[key] = old
                     val = rebase_path_traits(result.outputs.trait(key), old, cwd)
                     setattr(result.outputs, key, val)
-        savepkl(resultsfile, result)
+        savepkl(resultsfile, result, sync=True)
     finally:
         # Restore resolved paths from the outputs dict no matter what
         for key, val in list(backup_traits.items()):
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -790,9 +790,9 @@ def read_stream(stream, logger=None, encoding=None):
 
 
 def savepkl(filename, record, versioning=False):
-    pklopen = gzip.open if filename.endswith('.pklz') else open
     with SoftFileLock('%s.lock' % filename):
-        with pklopen(filename, 'wb') as pkl_file:
+        with open(filename, 'wb') as fobj:
+            pkl_file = gzip.GzipFile(fileobj=fobj) if filename.endswith('.pklz') else fobj
             if versioning:
                 from nipype import __version__ as version
                 metadata = json.dumps({'version': version})
@@ -801,6 +801,9 @@ def savepkl(filename, record, versioning=False):
                 pkl_file.write('\n'.encode('utf-8'))
 
             pickle.dump(record, pkl_file)
+            # Pickle files need to be available immediately, so force a sync
+            fobj.flush()
+            os.fsync(fobj.fileno())
 
 
 rst_levels = ['=', '-', '~', '+']
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -789,19 +789,18 @@ def read_stream(stream, logger=None, encoding=None):
     return out.splitlines()
 
 
-def savepkl(filename, record, versioning=False):
-    with SoftFileLock('%s.lock' % filename):
-        with open(filename, 'wb') as fobj:
-            pkl_file = gzip.GzipFile(fileobj=fobj) if filename.endswith('.pklz') else fobj
-            if versioning:
-                from nipype import __version__ as version
-                metadata = json.dumps({'version': version})
+def savepkl(filename, record, versioning=False, sync=False):
+    with open(filename, 'wb') as fobj:
+        pkl_file = gzip.GzipFile(fileobj=fobj) if filename.endswith('.pklz') else fobj
+        if versioning:
+            from nipype import __version__ as version
+            metadata = json.dumps({'version': version})
 
-                pkl_file.write(metadata.encode('utf-8'))
-                pkl_file.write('\n'.encode('utf-8'))
+            pkl_file.write(metadata.encode('utf-8'))
+            pkl_file.write('\n'.encode('utf-8'))
 
-            pickle.dump(record, pkl_file)
-            # Pickle files need to be available immediately, so force a sync
+        pickle.dump(record, pkl_file)
+        if sync:
             fobj.flush()
             os.fsync(fobj.fileno())
 


satra:
@effigies - in a conversation with @oesteban we discussed that the `SoftFileLock` is not useful at all for save/load pickle. in nipype, two nodes would never write to the same file, and for reading, a lock shouldn't be required. so we can simplify the code here.

@oesteban mentioned that the sync issue appears to arise in networked file systems, so we also discussed copying over the timeout code from SGELikePluginBase to load pickle.

effigies:
I thought the lock was for reading and writing...

satra:
in nipype, reading never happens before writing is done. the only issue that could take place is that the filesystem has not synced yet, but even in that case the reading node should be able to assume that writing has been completed. this is because a job doesn't get released till it's dependent jobs are finished (which means result writing has been done).

oesteban:
I think the `flush` can be useful. And we might want to have a paranoia mode where the os.fsync is executed. With all that above said, I think this PR should be reconsidered (and maybe remove the lock for reading, at the same time).

effigies:
Okay. Reopened.

> And we might want to have a paranoia mode where the os.fsync is executed.

I personally find working with the config file to be a pain, especially inside containers. If we make syncing the default behavior, and people can run with [eatmydata](https://www.flamingspork.com/projects/libeatmydata/) to get whatever performance boosts are available by being less paranoid. (I suspect they'll be minimal, as the time spent in the control thread is pretty small relative to most workflow tasks.)

In any event, should I add a `sync` parameter to the function, then, rather than running it unconditionally? This should make it easier to enable/disable paranoid mode from higher level functions, if we decide to go that route.

oesteban:
having the `sync` defaulting to true looks like a good idea

effigies:
This seems to reliably produce failures to find `result_*.pklz` files, so perhaps `fsync` isn't a good idea, regardless.

satra:
@effigies - that's really weird. why would this change exacerbate the problem. might be good to follow up.

on one of the stackoverflow threads i read about the following:

1. write to a temporary file
2. rename the temporary file to the target file (this is an atomic operation)

oesteban:
@effigies you've got the closest to reliably replicate the problem. We can't be sure this is the same problem, but I agree with @satra this is the best we have to investigate it this far :) -  good catch!.

effigies:
Yeah, I don't know it's the same problem. But I think we can say that if `fsync` causes a delay that means the file can't be found, then the strict ordering of tasks doesn't hold.

effigies:
Okay. Restored the branch if you have thoughts on where to go with this.

satra:
@effigies: can we add the timeout check to load_pkl? a version of this is here:

https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/base.py#L476

PR #3089, opened 2019-11-05
Enh/wait

### LINKED ISSUE #3014: Error checking MapNode hash with fMRIPrep

### LINKED ISSUE #3076: Strange results file checking behavior that breaks workflows at nested mapnodes in nipype 1.2.3

### LINKED ISSUE #3081: No such file or directory: 'result_bidssrc.pklz'

satra:
## Summary
Continuation of #3083 

this adds an explicit timeout in reading the results file that may be missing due to file system delays.

Closes #3076

Hopefully:

Fixes #3014
Fixes #3081

- Add waiting for result in reading pickled files.

## Acknowledgment

### DIFF
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -732,7 +732,7 @@ def loadpkl(infile):
         sleep(2)
     if timed_out:
         error_message = (
-            "Ressult file {0} expected, but "
+            "Result file {0} expected, but "
             "does not exist after ({1}) "
             "seconds.".format(infile, timeout)
         )
--- a/nipype/utils/filemanip.py
+++ b/nipype/utils/filemanip.py
@@ -3,8 +3,7 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """Miscellaneous file manipulation functions
 """
-from __future__ import (print_function, division, unicode_literals,
-                        absolute_import)
+from __future__ import print_function, division, unicode_literals, absolute_import
 
 import sys
 import pickle
@@ -21,35 +20,35 @@ import shutil
 import contextlib
 import posixpath
 import simplejson as json
-from filelock import SoftFileLock
+from time import sleep, time
+from glob import glob
 
 from builtins import str, bytes, open
 
 from .. import logging, config
 from .misc import is_container
 from future import standard_library
+
 standard_library.install_aliases()
 
-fmlogger = logging.getLogger('nipype.utils')
+fmlogger = logging.getLogger("nipype.utils")
 
-related_filetype_sets = [
-    ('.hdr', '.img', '.mat'),
-    ('.nii', '.mat'),
-    ('.BRIK', '.HEAD'),
-]
+related_filetype_sets = [(".hdr", ".img", ".mat"), (".nii", ".mat"), (".BRIK", ".HEAD")]
 
 PY3 = sys.version_info[0] >= 3
 
 try:
     from builtins import FileNotFoundError, FileExistsError
 except ImportError:  # PY27
+
     class FileNotFoundError(OSError):  # noqa
         """Defines the exception for Python 2."""
 
         def __init__(self, path):
             """Initialize the exception."""
             super(FileNotFoundError, self).__init__(
-                2, 'No such file or directory', '%s' % path)
+                2, "No such file or directory", "%s" % path
+            )
 
     class FileExistsError(OSError):  # noqa
         """Defines the exception for Python 2."""
@@ -57,7 +56,8 @@ except ImportError:  # PY27
         def __init__(self, path):
             """Initialize the exception."""
             super(FileExistsError, self).__init__(
-                17, 'File or directory exists', '%s' % path)
+                17, "File or directory exists", "%s" % path
+            )
 
 
 USING_PATHLIB2 = False
@@ -65,6 +65,7 @@ try:
     from pathlib import Path
 except ImportError:
     from pathlib2 import Path
+
     USING_PATHLIB2 = True
 
 
@@ -104,11 +105,12 @@ def path_mkdir(path, mode=0o777, parents=False, exist_ok=False):
             return os.mkdir(str(path), mode=mode)
 
 
-if not hasattr(Path, 'write_text'):
+if not hasattr(Path, "write_text"):
     # PY34 - Path does not have write_text
     def _write_text(self, text):
-        with open(str(self), 'w') as f:
+        with open(str(self), "w") as f:
             f.write(text)
+
     Path.write_text = _write_text
 
 
@@ -152,8 +154,7 @@ def split_filename(fname):
     ext = None
     for special_ext in special_extensions:
         ext_len = len(special_ext)
-        if (len(fname) > ext_len) and \
-                (fname[-ext_len:].lower() == special_ext.lower()):
+        if (len(fname) > ext_len) and (fname[-ext_len:].lower() == special_ext.lower()):
             ext = fname[-ext_len:]
             fname = fname[:-ext_len]
             break
@@ -181,11 +182,11 @@ def to_str_py27(value):
     """
 
     if isinstance(value, dict):
-        entry = '{}: {}'.format
-        retval = '{'
+        entry = "{}: {}".format
+        retval = "{"
         for key, val in list(value.items()):
             if len(retval) > 1:
-                retval += ', '
+                retval += ", "
             kenc = repr(key)
             if kenc.startswith(("u'", 'u"')):
                 kenc = kenc[1:]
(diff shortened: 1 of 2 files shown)


satra:
@oesteban @dpys - could you try this branch? this only handles any results file issue, not resource management.

nipype/utils/filemanip.py
```diff
@@ -690,32 +701,50 @@ def load_json(filename):
...
+    t = time()
+    timeout = float(config.get("execution", "job_finished_timeout"))
+    timed_out = True
+    while (time() - t) < timeout:
+        try:
+            glob(str(infile)).pop()
+            timed_out = False
+            break
+        except Exception as e:
+            fmlogger.debug(e)
```
effigies: Is there something special about `glob`, as opposed to, e.g.,

```Python
if infile.exists():
    timed_out = False
    break
fmlogger.debug("'{}' missing; waiting 2s".format(infile))
```

This feels more straightforward, so if `glob` is doing something useful beyond an `exists()` check, a comment would be useful.

nipype/utils/filemanip.py
```diff
@@ -690,32 +701,50 @@ def load_json(filename):
...
+        sleep(2)
+    if timed_out:
+        error_message = (
+            "Ressult file {0} expected, but "
```
effigies:
```suggestion
            "Result file {0} expected, but "
```

nipype/utils/filemanip.py
satra: updated the code - that code is old, and i don't remember why we used glob.

dPys:
Thanks @satra -- this patch appears to fix #3076!

nipype/utils/filemanip.py
```diff
@@ -779,50 +813,53 @@ def read_stream(stream, logger=None, encoding=None):
...
+
+            metadata = json.dumps({"version": version})
 
-                pkl_file.write(metadata.encode('utf-8'))
-                pkl_file.write('\n'.encode('utf-8'))
+            pkl_file.write(metadata.encode("utf-8"))
+            pkl_file.write("\n".encode("utf-8"))
 
-            pickle.dump(record, pkl_file)
+        pickle.dump(record, pkl_file)
```
oesteban: Could we do this to a bytes IO stream outside the context and then open the context and just write it?

satra: it is a bytes io stream and is being done with the context.

satra: see 827-828

oesteban: I meant something like this:

```Python
from io import BytesIO

    with BytesIO() as f:
        if versioning:
            from nipype import __version__ as version
            metadata = json.dumps({"version": version})
            f.write(metadata.encode("utf-8"))
            f.write("\n".encode("utf-8"))
        pickle.dump(record, f)
        content = f.get_value()

    pkl_open = gzip.open if filename.endswith(".pklz") else open
    with pkl_open(filename, "wb") as pkl_file:
        pkl_file.write(content)
```

effigies: Why?

oesteban: minimize being inside the context with an open file?

effigies: Maybe I'm missing something. Are we worried about some kind of race here? If so, why drop the lock?

oesteban: Not really. Separating steps just seemed a good idea to me in this case. We haven't tested versioning thoroughly, and doing the pickle.dump to memory seemed more efficient than all at once.

Also if we keep seeing issues, separating both will make it easier to make more targeted interventions (let's imagine we discover that gzip.open does not get along with NFS)

oesteban: (that said, I agree this is not strictly necessary for this PR, and at the risk of introducing some other bug I am happy dismissing).

satra:
i refactored `savepkl`. and yes there are more stylistic changes, since i have pre-commit turned on.

satra:
here is the direct pointer to the function: https://github.com/nipy/nipype/pull/3089/files#diff-7d84fc0957a520f25bb02479beb87c51R825

satra:
@effigies @oesteban - can we merge this and cut a new release/release candidate?

effigies:
No objections.

satra:
@oesteban - can i merge this?

oesteban:
Sorry for holding this up

PR #3094, opened 2019-11-12
fix: restore checking traits or bunch

satra:
@Shotgunosine - can you see if this works?

@oesteban - this switches back to checking different kinds of results


Shotgunosine:
Seems like it works now.

effigies:
Thanks @Shotgunosine.

PR #3106, opened 2019-11-25
FIX: Mark strings containing regex escapes as raw

effigies:
## Summary

Ran into some deprecation errors working on #3099. The escape characters were easy to handle.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/dipy/base.py
+++ b/nipype/interfaces/dipy/base.py
@@ -158,7 +158,7 @@ def create_interface_specs(class_name, params=None, BaseClass=TraitedSpec):
                     attr[name] = traits_type(desc=desc[-1], mandatory=is_mandatory)
             else:
                 attr[name] = traits_type(
-                    p[3], desc=desc[-1], exists=True, usedefault=True,
+                    p[3], desc=desc[-1], exists=True, usedefault=True
                 )
 
     newclass = type(str(class_name), (BaseClass,), attr)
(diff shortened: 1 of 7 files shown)


nipype/interfaces/spm/model.py
```diff
@@ -474,15 +474,16 @@ def _make_matlab_command(self, _):
             script += "condnames=names;\n"
         else:
             if self.inputs.use_derivs:
-                script += "pat = 'Sn\([0-9]*\) (.*)';\n"
+                script += r"pat = 'Sn\([0-9]*\) (.*)';" "\n"
```
effigies: Hopefully we test these. I'm less comfortable with these, since they need to work when written to MATLAB scripts.

nipype/interfaces/dipy/base.py
```diff
@@ -189,7 +189,7 @@ def dipy_to_nipype_interface(cls_name, dipy_flow, BaseClass=DipyBaseInterface):
     parser = IntrospectiveArgumentParser()
     flow = dipy_flow()
     parser.add_workflow(flow)
-    default_values = inspect.getargspec(flow.run).defaults
+    default_values = inspect.getfullargspec(flow.run).defaults
```
effigies: Guess I fixed another deprecation warning. This works for all Py3.

effigies:
Anybody up for a review?

nipype/interfaces/spm/model.py
satra: as long as we are still running the spm workflows.

satra:
in general looks fine to me.

nipype/interfaces/spm/model.py
effigies: Codecov doesn't seem to hit them.

effigies: I've tested:

```Python
>>> a = ""
>>> a += r"\s+" "\nnextline"                                                                   
>>> print(a)                                                                                   
\s+
nextline
```

So rawstring/string concatenation works fine, and the `+=` is lower precedence, so I'm reasonably confident in this change.

Merging.

PR #3121, opened 2019-12-21
[REF] removed all uses of numpy_mmap

steve19922:
## Summary

Removed all use cases of numpy_mmap as it will always be true.

## Acknowledgment

### DIFF
--- a/nipype/utils/config.py
+++ b/nipype/utils/config.py
@@ -28,7 +28,6 @@ CONFIG_DEPRECATIONS = {
     "filemanip_level": ("logging.utils_level", "1.0"),
 }
 
-NUMPY_MMAP = LooseVersion(np.__version__) >= LooseVersion("1.12.0")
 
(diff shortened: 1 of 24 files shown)


nipype/utils/config.py
```diff
@@ -28,7 +28,6 @@
     "filemanip_level": ("logging.utils_level", "1.0"),
 }
 
-NUMPY_MMAP = LooseVersion(np.__version__) >= LooseVersion("1.12.0")
```
effigies: This leaves the `numpy` import unused.

satra:
> One potential problem is removing the NUMPY_MMAP constant altogether, if any downstream projects use it.

i think we should just remove it. while not explicit, this was not intended to be an external api. hopefully developers will catch this via release candidates if they were re-using this constant.

effigies:
Works for me.

PR #3130, opened 2020-01-02
FIX: DataSink to S3 buckets

### LINKED ISSUE #3105: DataSink S3 support does not work
### Summary
The DataSink interface does not work.

### Actual behavior
Only the local copy works. Oddly, when used in a workflow, DataSink creates a local directory named `s3:/` instead.

### Expected behavior
DataSink creates objects in S3 buckets when base_directory starts with `s3://`.

```
ds = DataSink()

ds.inputs.base_directory = 's3://kimsin-test'
ds.inputs.creds_path = os.path.abspath('aws.csv')
ds.inputs.encrypt_bucket_keys = True
ds.inputs.local_copy = os.path.abspath('log')
setattr(ds.inputs, 'out_s3', 'DELETE.me')

ds.run()
```
This basic S3 test code does not create the desired output in the S3 bucket.

### Platform details:
```
{'commit_hash': '%h',
 'commit_source': 'archive substitution',
 'networkx_version': '2.4',
 'nibabel_version': '2.5.1',
 'nipype_version': '1.3.0-rc1',
 'numpy_version': '1.17.3',
 'pkg_path': '/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/nipype',
 'scipy_version': '1.3.1',
 'sys_executable': '/opt/miniconda-latest/envs/neuro/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.6.7 | packaged by conda-forge | (default, Jul  2 2019, '
                '02:18:42) \n'
                '[GCC 7.3.0]',
 'traits_version': '5.1.2'}
```

### Execution environment
- My python environment inside container

AKSoo:
Due to how `pathlib.Path` handles path strings, S3 bucket paths are missing a `/`.

## Summary
Fixes #3105.

* Changes DataSink input specs for output paths to be `Str` rather than `Directory`.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -440,8 +440,9 @@ class DataSink(IOBase):
             is not a valid S3 path, defaults to '<N/A>'
         """
 
-        # Init variables
-        s3_str = "s3://"
+        # NOTE: due to pathlib.Path in traits.Directory, S3 bucket paths
+        #       follow the format 's3:/bucket_name/...'
+        s3_str = "s3:/"
         bucket_name = "<N/A>"
         base_directory = self.inputs.base_directory
 
@@ -449,18 +450,10 @@ class DataSink(IOBase):
             s3_flag = False
             return s3_flag, bucket_name
 
-        # Explicitly lower-case the "s3"
+        # Check for 's3:/' in base dir
         if base_directory.lower().startswith(s3_str):
-            base_dir_sp = base_directory.split("/")
-            base_dir_sp[0] = base_dir_sp[0].lower()
-            base_directory = "/".join(base_dir_sp)
-
-        # Check if 's3://' in base dir
-        if base_directory.startswith(s3_str):
-            # Expects bucket name to be 's3://bucket_name/base_dir/..'
-            bucket_name = base_directory.split(s3_str)[1].split("/")[0]
+            bucket_name = base_directory[len(s3_str):].split("/")[0]
             s3_flag = True
-        # Otherwise it's just a normal datasink
         else:
             s3_flag = False
 
@@ -618,8 +611,9 @@ class DataSink(IOBase):
 
         from botocore.exceptions import ClientError
 
-        # Init variables
-        s3_str = "s3://"
+        # NOTE: due to pathlib.Path in traits.Directory, S3 bucket paths
+        #       follow the format 's3:/bucket_name/...'
+        s3_str = "s3:/"
         s3_prefix = s3_str + bucket.name
 
         # Explicitly lower-case the "s3"
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -212,12 +212,12 @@ class DataSinkInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
     """
 
     # Init inputspec data attributes
-    base_directory = Directory(desc="Path to the base directory for storing data.")
+    base_directory = Str(desc="Path to the base directory for storing data.")
     container = Str(desc="Folder within base directory in which to store output")
     parameterization = traits.Bool(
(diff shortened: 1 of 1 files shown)


effigies:
I agree with @oesteban that I think converting `base_directory` to `Str` is probably more correct in terms of keeping the inputs consistent with what is passed by the user. On the other hand, this also doesn't seem egregious to me, as `s3:/` is not really going to happen when `s3://` is not intended. So I'm fine with either approach, myself.

In any event, could you rebase onto `maint/1.4.x` and update the base branch? As this is a bug fix, we can aim for the 1.4.1 release.

Screen Shot 2020-01-06 at 13 03 09

nipype/interfaces/io.py
```diff
@@ -440,27 +440,20 @@ def _check_s3_base_dir(self):
...
+        s3_str = "s3:/"
         bucket_name = "<N/A>"
         base_directory = self.inputs.base_directory
 
         if not isdefined(base_directory):
             s3_flag = False
             return s3_flag, bucket_name
 
-        # Explicitly lower-case the "s3"
+        # Check for 's3:/' in base dir
```
satra: how about using: `s3_flag = base_directory.startswith('s3://')` makes it more explicit and potentially avoids other prefixes `ds://` would also say `s3_flag` is true.

AKSoo: I don't think other prefixes would cause problems but cleaned the code a bit with your suggestion.

AKSoo:
As all the reviewers noted, there seems to be little to no advantage of keeping `base_directory` to be a `Directory`. If anything goes wrong, the problem is probably in the user's inputs, and keeping them as `Str` would make the problem more transparent.

Changed the DataSink input spec instead.

nipype/interfaces/io.py
```diff
@@ -212,12 +212,12 @@ class DataSinkInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
...
 
     # Init inputspec data attributes
-    base_directory = Directory(desc="Path to the base directory for storing data.")
+    base_directory = Str(desc="Path to the base directory for storing data.")
     container = Str(desc="Folder within base directory in which to store output")
     parameterization = traits.Bool(
         True, usedefault=True, desc="store output in parametrized structure"
     )
-    strip_dir = Directory(desc="path to strip out of filename")
+    strip_dir = Str(desc="path to strip out of filename")
```
effigies: What is the effect of this change? Was something failing?

nipype/interfaces/io.py
```diff
@@ -449,20 +448,9 @@ def _check_s3_base_dir(self):
...
-        if base_directory.startswith(s3_str):
-            # Expects bucket name to be 's3://bucket_name/base_dir/..'
-            bucket_name = base_directory.split(s3_str)[1].split("/")[0]
-            s3_flag = True
-        # Otherwise it's just a normal datasink
-        else:
-            s3_flag = False
+        s3_flag = base_directory.lower().startswith(s3_str)
+        if s3_flag:
+            bucket_name = base_directory[len(s3_str):].split("/")[0]
```
effigies: No real need to find more than the first `/`.

```suggestion
            bucket_name = base_directory[len(s3_str):].split("/", 1)[0]
```

nipype/interfaces/io.py
AKSoo: No, I don't think this will make any difference in proper use cases.

But since we are setting `base_directory` trait to `Str` to be faithful to the user's inputs, it would not make sense to disable input processing on one path input but not the other.

nipype/interfaces/io.py
AKSoo: apparently `partition` is a faster alternative that only splits once

PR #3133, opened 2020-01-02
FIX: Repair aftermath of docs refactor

oesteban:
- [x] Removes the build_docs step in test_pytest, as we now have a pure-python job for it.
- [x] Fixes matlab code generation within spm.EstimateConstrast. Stylistic edits ended up introducing raw strings with uninterpreted backslashes (e.g., `\n`).
- [x] Run black


oesteban:
After merging this one, only #3134 is left to bring everything back.

PR #3135, opened 2020-01-03
[FIX] immunize shutil.rmtree to node non-existence for remove_node_di…

dPys:
## Summary
- Immunizes shutil.rmtree to node non-existence for remove_node_directories=True in the case that stop_on_first_crash=False. Without ignore_errors=True, workflows fail under these conditions.

## Acknowledgment


effigies:
Do you have a test?

dPys:
I could write a regression test for this, but it might be overkill? I can't really think of a case where ignore_errors=True would cause issues. As you noted, the flag has been around since Python 2. I do like the idea of waiting to rebase this into #3146 though, rather than rush the merge.

effigies:
The regression test would be to make sure future changes don't restore this problem, not whether this fix causes new problems.

effigies:
@dPys #3146 has been merged. Please rebase on `maint/1.4.x` and update the target:

Screen Shot 2020-01-06 at 12 58 41

dPys:
> The regression test would be to make sure future changes don't restore this problem, not whether this fix causes new problems.

Ahh, I see what you mean now.

effigies:
@dPys This history looks a bit complicated. I would suggest the following to clean it up:

```Shell
git fetch upstream
git checkout -b fix/rmtree_ignore_errors upstream/maint/1.4.x
git cherry-pick 1c98d4d
git push -u origin fix/rmtree_ignore_errors
```

Then create a new PR, closing this one.

effigies:
Replaced by #3148.

PR #3137, opened 2020-01-03
FIX fsl CopyGeom caching bug

### LINKED ISSUE #3136: fsl CopyGeom does not change the destination file when caching
### Summary
The geometry is not copied into the given destination file
### Actual behavior
The geometry is not copied into the given destination file but into a file with the same basename located inside the cached folder in `nipype_mem/nipype-interfaces-fsl-utils-CopyGeom/` when caching
### Expected behavior
Geometry of `"/tmp/reference.nii.gz"` should be copied into `"/tmp/anat.nii.gz"` 

```Python
from nipype.caching import Memory
from nipype.interfaces import afni, ants, fsl

memory = Memory('/tmp')
copy_geom = memory.cache(fsl.CopyGeom)
out_copy_geom = copy_geom(dest_file="/tmp/anat.nii.gz", in_file="/tmp/reference.nii.gz")
```
does not change the file `"/tmp/anat.nii.gz"`

### Platform details:

```Shell
200103-23:58:17,498 nipype.utils INFO:
	 Running nipype version 1.5.0-dev+g71ab10b80 (latest: 1.4.0)
{'commit_hash': '71ab10b80',
 'networkx_version': '2.4',
 'nibabel_version': '3.0.0',
 'nipype_version': '1.5.0-dev+g71ab10b80',
 'numpy_version': '1.17.4',
 'pkg_path': '/home/salma/CODE/nipype/nipype',
 'scipy_version': '1.3.2',
 'sys_executable': '/home/salma/miniconda3/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.7.4 (default, Aug 13 2019, 20:35:49) \n[GCC 7.3.0]',
 'traits_version': '5.2.0'}

```

### Execution environment

salma1601:
## Summary

Fixes #3136.

remove ```copyfile=True``` to execute fslcpgeom on the destination file when caching
## Acknowledgment

### DIFF
--- a/nipype/interfaces/fsl/tests/test_auto_CopyGeom.py
+++ b/nipype/interfaces/fsl/tests/test_auto_CopyGeom.py
@@ -7,7 +7,6 @@ def test_CopyGeom_inputs():
         args=dict(argstr="%s",),
         dest_file=dict(
             argstr="%s",
-            copyfile=True,
             extensions=None,
             mandatory=True,
             name_source="dest_file",
--- a/nipype/interfaces/fsl/utils.py
+++ b/nipype/interfaces/fsl/utils.py
@@ -40,11 +40,9 @@ class CopyGeomInputSpec(FSLCommandInputSpec):
         argstr="%s",
         position=1,
         desc="destination image",
-        copyfile=True,
-        output_name="out_file",
-        name_source="dest_file",
-        name_template="%s",
-    )
+        output_name='out_file',
+        name_source='dest_file',
+        name_template='%s')
     ignore_dims = traits.Bool(
         desc="Do not copy image dimensions", argstr="-d", position="-1"
     )


satra:
@salma1601 - this is a weird one and changing copyfile won't work.  one of the principles of nipype for Node based operation is that we never change any inputs. so the only way this would work is to copy the output over using a datasink. 

also if you think of it from a caching perspective, as soon as you overwrite an input the function will run again because the inputs have changed. so if you really want to overwrite, then you use the non-cached mode. if you use a cached mode, you pass on the function outputs to the next step.

salma1601:
@satra thank you for your explanations.
> one of the principles of nipype for Node based operation is that we never change any inputs.

may be we can add to the inputs an `out_file` that would be a copy of the destination file with the new geometry ?

satra:
> may be we can add to the inputs an out_file that would be a copy of the destination file with the new geometry ?

that's what the current output `out_file` is, `a copy of the destination file with the new geometry`. the primary thing that the `Node` version doesn't allow is to locate the output in an external location, which is consistent with a dataflow paradigm.

salma1601:
Ok, shall I close this one ?

satra:
thank you for raising this, but i think this is consistent with nipype at the moment. i'll close it.

PR #3139, opened 2020-01-04
FIX Ants N4BiasFieldCorrection rescale_intensities bug

### LINKED ISSUE #3138: N4BiasFieldCorrection fails for compiled ANTS
### Summary
`N4BiasFieldCorrection` throws an error

### Actual behavior
The `rescale_intensities` trait is set by default to `False` while it must not be specified for ANTS version under 2.1.0
### Expected behavior
`rescale_intensities` should be undefined by default
We can replicate the behavior with compiled ANTs
### Script/Workflow details
My Ants version 
``` console
salma@salma-Latitude-E7450:~/CODE/sammba-mri $ N4BiasFieldCorrection --version
ANTs Version: 0.0.0.0
Compiled: Dec 22 2019 22:03:46
```
Ants version as read by nipype 
```console
salma@salma-Latitude-E7450:~/CODE/sammba-mri $ python -c "from nipype.interfaces import ants; print(ants.base.Info.version())"
200104-19:14:53,203 nipype.utils INFO:
	 Running nipype version 1.5.0-dev+g7f6cc2de2 (latest: 1.4.0)
0.0.0
```
```Python
from nipype.interfaces import ants

bias_correct = ants.N4BiasFieldCorrection().run
bias_correct(input_image='/tmp/func.nii.gz')
```
throws the following error
```
200104-19:20:52,963 nipype.utils INFO:
	 Running nipype version 1.5.0-dev+g7f6cc2de2 (latest: 1.4.0)
Exception                                 Traceback (most recent call last)
<ipython-input-1-9321c82e17d6> in <module>
      2 
      3 bias_correct = ants.N4BiasFieldCorrection().run
----> 4 bias_correct(input_image='/tmp/func.nii.gz')

~/CODE/nipype/nipype/interfaces/base/core.py in run(self, cwd, ignore_exception, **inputs)
    348         self.inputs.trait_set(**inputs)
    349         self._check_mandatory_inputs()
--> 350         self._check_version_requirements(self.inputs)
    351         interface = self.__class__
    352         self._duecredit_cite()

~/CODE/nipype/nipype/interfaces/base/core.py in _check_version_requirements(self, trait_object, raise_exception)
    281                         raise Exception(
    282                             "Trait %s (%s) (version %s < required %s)"
--> 283                             % (name, self.__class__.__name__, version, min_ver)
    284                         )
    285 

Exception: Trait rescale_intensities (N4BiasFieldCorrection) (version 0.0.0 < required 2.1.0)
```
while forcing the trait `rescale_intensities` to be undefined solves the issue
```Python
from nipype.interfaces.base import traits                                                                                                                                                                                                                                 

bias_correct(input_image='/tmp/func.nii.gz', rescale_intensities=traits.Undefined)
```

### Platform details:

<!-- Please run the following code from your shell and place the output between the triple ticks, below.
python -c "import nipype; from pprint import pprint; pprint(nipype.get_info())"
-->
```console
200104-19:19:19,39 nipype.utils INFO:
	 Running nipype version 1.5.0-dev+g7f6cc2de2 (latest: 1.4.0)
{'commit_hash': '7f6cc2de2',
 'networkx_version': '2.4',
 'nibabel_version': '3.0.0',
 'nipype_version': '1.5.0-dev+g7f6cc2de2',
 'numpy_version': '1.17.4',
 'pkg_path': '/home/salma/CODE/nipype/nipype',
 'scipy_version': '1.3.2',
 'sys_executable': '/home/salma/miniconda3/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.7.4 (default, Aug 13 2019, 20:35:49) \n[GCC 7.3.0]',
 'traits_version': '5.2.0'}```

```

```

### Execution environment

### PR DESCRIPTION
**Pull request by salma1601**
## Summary

Addresses #3138 in the short term.

removed `use_default=True` for `rescale_intensities` input
## Acknowledgment

### CODE REVIEW

**Comment by oesteban** [comment]
Hi @salma1601, why is this a bug?

File: nipype/interfaces/ants/segmentation.py [hunk 1]
```diff
@@ -335,7 +335,6 @@ class N4BiasFieldCorrectionInputSpec(ANTSCommandInputSpec):
     )
     rescale_intensities = traits.Bool(
         False,
```
effigies: There was a move a while back to either remove defaults if unused, or set `usedefault=True` if the default was specified. In this case, if we're removing `usedefault`, we should also remove the default.

```suggestion
```

satra:
@effigies and @oesteban - see the discussion in #3138 - there is a more general problem for which this PR is not the answer.

effigies:
@satra I saw that. Is your preference to solve the more general problem instead of this short-term solution, or in addition?

satra:
the short term solution makes an interface inputs change for ants versions greater than `min_ver`. this could potential change any scripts that are relying on the default. in this specific instance it may be fine if that is also the default in ANTs for versions > `min_ver` and that this change would in that case not affect the outcome. it will force things to rerun (since the hash will change). i would be fine with this PR, if we can confirm that the lack of the input on the command line will not make any difference to the outcome for ANTs for versions > `min_ver`

ps. i think we should discuss the more general solution in a more general discussion about how to version inputs and package versions. as the interfaces start evolving a bit, we will be left with what dependencies they support. this is similar in concept to any library api (e.g., numpy), but thus far we have stayed out of it by doing minimal pieces. as we start moving interfaces to their own packages for nipype 2, it may help to discuss the mechanics of this a bit.

oesteban:
> i would be fine with this PR, if we can confirm that the lack of the input on the command line will not make any difference to the outcome for ANTs for versions > `min_ver`

@salma1601 could you confirm this with ANTs devs?



> ps. i think we should discuss the more general solution in a more general discussion about how to version inputs and package versions. as the interfaces start evolving a bit, we will be left with what dependencies they support. this is similar in concept to any library api (e.g., numpy), but thus far we have stayed out of it by doing minimal pieces. as we start moving interfaces to their own packages for nipype 2, it may help to discuss the mechanics of this a bit.

👍

salma1601:
@oesteban @satra Unfortunately the ANTS default (at least for my version compiled 22 december 2019)  is rescale_intensities=True, so actually it will make a difference ...

effigies:
Is there consensus on whether a short-term fix is possible, or only the long-term approach discussed in https://github.com/nipy/nipype/issues/3138#issuecomment-571394934?

effigies:
Just checking back in on this in the run-up to another release. Apologies for the short notice...

PR #3148, opened 2020-01-06
[FIX] immunize shutil.rmtree to node non-existence for remove_node_di…

dPys:
@effigies -- since it seems you're still actively working on the 1.4.x merge, I just went ahead and opened a fresh PR directly onto that branch since rebasing was causing issues...

### DIFF
--- a/nipype/pipeline/plugins/tests/test_base.py
+++ b/nipype/pipeline/plugins/tests/test_base.py
@@ -3,9 +3,10 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """Tests for the engine module
 """
+import pytest
+from pathlib import Path
 import numpy as np
 import scipy.sparse as ssp
-
 
 def test_scipy_sparse():
     foo = ssp.lil_matrix(np.eye(3, k=1))
@@ -39,3 +40,34 @@ wf.base_dir = '/tmp'
 
 wf.run(plugin='MultiProc')
 """
+
+@pytest.mark.regression
+def test_remove_nodes(tmpdir):
+    import os
+    import nipype.interfaces.utility as niu
+    import nipype.pipeline.engine as pe
+
+    wf = pe.Workflow(name='test')
+
+    def func(arg1):
+        try:
+            if arg1 == 2:
+                raise Exception('arg cannot be ' + str(arg1))
+        except:
+            pass
+        return arg1
+
+    funkynode = pe.MapNode(niu.Function(function=func, input_names=['arg1'],
+                                        output_names=['out']),
+                           iterfield=['arg1'],
+                           name = 'functor')
+    funkynode.inputs.arg1 = [1,2]
+
+    wf.add_nodes([funkynode])
+    wf.base_dir = tmpdir
+    wf.config['execution']['remove_node_directories'] = 'true'
+    wf.config['execution']['stop_on_first_crash'] = 'false'
+    wf.config['logging']['crashdump_dir'] = tmpdir
+    res = wf.run(plugin='MultiProc')
+
+    assert os.path.isdir(Path(tmpdir)/'test/functor') is False
--- a/nipype/pipeline/plugins/tests/test_base.py
+++ b/nipype/pipeline/plugins/tests/test_base.py
@@ -42,7 +42,7 @@ wf.run(plugin='MultiProc')
 """
 
 @pytest.mark.regression
-def test_remove_nodes(tmpdir):
+def test_remove_nodes(tmp_path):
     import os
     import nipype.interfaces.utility as niu
     import nipype.pipeline.engine as pe
--- a/nipype/pipeline/plugins/tests/test_base.py
+++ b/nipype/pipeline/plugins/tests/test_base.py
@@ -70,4 +70,4 @@ def test_remove_nodes(tmp_path):
     wf.config['logging']['crashdump_dir'] = str(tmp_path)
     res = wf.run(plugin='MultiProc')
 
-    assert os.path.isdir(Path(tmpdir)/'test/functor') is False
+    assert not Path.is_dir(tmp_path / "test" / "functor")
--- a/nipype/pipeline/plugins/tests/test_base.py
+++ b/nipype/pipeline/plugins/tests/test_base.py
@@ -67,7 +67,7 @@ def test_remove_nodes(tmp_path):
     wf.base_dir = tmpdir
     wf.config['execution']['remove_node_directories'] = 'true'
     wf.config['execution']['stop_on_first_crash'] = 'false'
-    wf.config['logging']['crashdump_dir'] = tmpdir
+    wf.config['logging']['crashdump_dir'] = str(tmp_path)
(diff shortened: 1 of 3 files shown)


dPys:
;-)

effigies:
Thanks. Do you think you could put together a small regression test?

dPys:
@effigies , Not really sure what you have in mind RE: a regression test, but let me know if this is close to what you had in mind:
```
@pytest.mark.regression(issue=3148)
def test_remove_nodes():
    import os

    wf = pe.Workflow(name='test')

    def func(arg1):
        try:
            if arg1 == 2:
                raise Exception('arg cannot be ' + str(arg1))
        except:
            pass
        return arg1

    funkynode = pe.MapNode(niu.Function(function=func, input_names=['arg1'],
                                        output_names=['out']),
                           iterfield=['arg1'],
                           name = 'functor')
    funkynode.inputs.arg1 = [1,2]

    wf.add_nodes([funkynode])
    wf.base_dir = '/tmp'
    wf.config['execution']['remove_node_directories'] = 'true'
    wf.config['execution']['stop_on_first_crash'] = 'false'
    wf.config['logging']['crashdump_dir'] = '/tmp'
    res = wf.run(plugin='MultiProc')
    
    assert os.path.isdir('/tmp/test/functor') is False
```

effigies:
Yup, something of the sort. You'll want to make sure that the test fails before your fix and passes after it.

You can use the `tmp_path` pytest fixture to avoid polluting `/tmp`, but if you're not very familiar with pytest, go ahead and code up a test, and I can make suggestions to clean it up.

dPys:
> Yup, something of the sort. You'll want to make sure that the test fails before your fix and passes after it.
> 
> You can use the `tmp_path` pytest fixture to avoid polluting `/tmp`, but if you're not very familiar with pytest, go ahead and code up a test, and I can make suggestions to clean it up.

With the exception of the regression, I the above update should be a bit closer to what we want. Not sure how you want to approach the regression exactly-- i.e. across previous nipype versions, by overriding the module with a paramaterized fixture for rmtree with the additional flag added?

effigies:
No need to parameterize. Just test on master and on this branch. It should fail and pass, respectively.

dPys:
Removed parameterization. Tested on maint/1.4.x and it passes. Does not yet seem to fail on master as expected so we might still need to tweak the test to add more nodes to workflow (i.e. nodes after the first node that crashes, yet the workflow continues to run and attempt node directory removal for directories that never get created)...

effigies:
Perhaps you could explain more how you were running into this issue? Were you running multiple workflows in parallel?

dPys:
> Perhaps you could explain more how you were running into this issue? Were you running multiple workflows in parallel?

Hmm, it occurred repeatedly when running a single workflow (with nested workflows) using MultiProc, and adding ignore_errors=True completely inoculated the issue. Hard to say what the fail context is precisely, but the solution was surefire.

satra:
could this have something to do with the mapnode fix from yesterday?

effigies:
I tried this test on 1.4.0, and it still passes. So whatever it's testing, it's not the error that was being hit.

Checking the coverage, however, it does at least hit the `shutil.rmtree` line, so that's encouraging. I'm just not sure how the node directory is supposed to get deleted prior to that line. It feels like a race condition, but I also don't see how that's going to happen if jobs aren't being run in parallel.

nipype/pipeline/plugins/tests/test_base.py
```diff
@@ -39,3 +40,34 @@ def func(arg1):
...
+    import nipype.pipeline.engine as pe
+
+    wf = pe.Workflow(name='test')
+
+    def func(arg1):
+        try:
+            if arg1 == 2:
+                raise Exception('arg cannot be ' + str(arg1))
+        except:
+            pass
```
effigies: Is the goal to fail or not? If not, we can skip the entire `try`/`except` block. If so, we should not catch the exception.

nipype/pipeline/plugins/tests/test_base.py
```diff
@@ -39,3 +40,34 @@ def func(arg1):
...
+        return arg1
+
+    funkynode = pe.MapNode(niu.Function(function=func, input_names=['arg1'],
+                                        output_names=['out']),
```
effigies: This is redundant:

```Python
funkynode = pe.MapNode(niu.Function(function=func),
```

Also, is there something intrinsic about the problem and `MapNode`s? If not, then perhaps just make it a `Node`, to reduce the scope as small as possible.

nipype/pipeline/plugins/tests/test_base.py
```diff
@@ -39,3 +40,34 @@ def func(arg1):
 
 wf.run(plugin='MultiProc')
 """
+
+@pytest.mark.regression
+def test_remove_nodes(tmpdir):
```
effigies: You can avoid coercing to `pathlib.Path` if you just use `tmp_path`:

```suggestion
def test_remove_nodes(tmp_path):
```

nipype/pipeline/plugins/tests/test_base.py
```diff
@@ -39,3 +40,34 @@ def func(arg1):
...
+                                        output_names=['out']),
+                           iterfield=['arg1'],
+                           name = 'functor')
+    funkynode.inputs.arg1 = [1,2]
+
+    wf.add_nodes([funkynode])
+    wf.base_dir = tmpdir
+    wf.config['execution']['remove_node_directories'] = 'true'
+    wf.config['execution']['stop_on_first_crash'] = 'false'
+    wf.config['logging']['crashdump_dir'] = tmpdir
```
effigies: We may handle `Path`s cleanly... not sure.

```suggestion
    wf.config['logging']['crashdump_dir'] = str(tmp_path)
```

nipype/pipeline/plugins/tests/test_base.py
```diff
@@ -39,3 +40,34 @@ def func(arg1):
...
+    res = wf.run(plugin='MultiProc')
+
+    assert os.path.isdir(Path(tmpdir)/'test/functor') is False
```
effigies: You can use instance methods as functions on instances:

```suggestion
    assert not Path.is_dir(tmp_path / "test" / "functor")
```

dPys:
> I tried this test on 1.4.0, and it still passes. So whatever it's testing, it's not the error that was being hit.
> 
> Checking the coverage, however, it does at least hit the `shutil.rmtree` line, so that's encouraging. I'm just not sure how the node directory is supposed to get deleted prior to that line. It feels like a race condition, but I also don't see how that's going to happen if jobs aren't being run in parallel.

It does feel very much like a race condition.

dPys:
@effigies -- I'm going to take another look at this on Friday to see if I can't pinpoint exactly what's going on. Things go a little 'nuts' when the iterable expansion starts hitting 10's of thousands of threads...

Stay tuned.

effigies:
Let's rebase and reopen if this becomes an issue again.

Issue #3149, opened 2020-01-07
Potential replacement to most ``_list_outputs`` and/or ``aggregate_outputs``


oesteban:
### Summary
The need for implementing ``_list_outputs`` and/or ``aggregate_outputs`` adds a lot of friction in the creation of new interfaces as these two methods are hard to generate automatically. This design also conflicts with a much-refined design for the inputs.

Since these are hard to automate, interfaces with either one or the other method are difficult to port to a much lighter interface object of Nipype 2.0.

### Proposal
1. Extend metadata / create new traits as follows:

```Python
class _OutputSpec(TearUpTraitedSpec):
    out_file = File('{in_file}_processed', mandatory=True, exists=True)
    out_list_glob = FileList(glob='{in_file}_tissue_*', mandatory=True, min=3, max=5)
    out_list_re = FileList(re=r'{in_file}_tissue_\d{3}', mandatory=True, min=3, max=5)
    out_float = Float(stdout=func_callback, mandatory=True)
```
where the `get_value` of each trait is modified to perform 1) python formatting replacement with the input names as keyword arguments; 2) run pattern matching (`glob` or `re` metadata); and 3) run callbacks to parse standard output or error streams (e.g. that `func_callback`, which is a python function that takes in the std out and returns a parsed value.

2. ``aggregate_outputs`` is then left as an empty hook for backward compatibility, and instead the output trait is instantiated. ``TearUpTraitedSpec`` is a ``TraitedSpec`` that fills in all values for the new special traits in its ``__init__()``. The OutputSpec only lives for the same amount of time it used to live before (basically during ``aggregate_outputs``) and gets cached exactly in the same way - i.e. the API does not really change, is just extended.

3. Most ``_list_inputs`` if not all can be removed.

4. ``SimpleInterface`` can be dropped as it won't be necessary anymore.

## Interesting side-effects
These new traits would be really interesting, for instance, to minimize the boilerplate of ``SelectFiles`` and ``DataSource``s because it would also be possible to use them for input specs in an IdentityInterface, effectively finding the given input patterns. (in particular, I'm considering something like this in the context of poldracklab/smriprep#107)

WDYT?

PR #3150, opened 2020-01-08
ENH: Interfaces with automated population of outputs

oesteban:
Proof of concept for #3149

This is a very rough draft of how such an interface would look like. It only tests automatic templated names filling (glob patterns should be fairly easy to add, and regexes too).

Some wrinkles around the corners need to be ironed out. E.g., implementing a keep_ext metadata for outputs, and raising an error if two inputs are set in the template and extensions are different. Similarly, raise error if not all the arguments were replaced.

But in general, it seems doable.

BTW the pattern matching and replacing part of this PR could also be useful to accept the `name_template` metadata in BaseInterfaces (currently, it is only allowed in command line interfaces).

WDYT @satra, @mgxd, @djarecka, @mgxd ?

EDIT: The new ``AutoOutputInterface`` is basically a copy of the current ``BaseInterface``, but removing the ``_list_outputs`` and ``aggregate_outputs``, and adding this: https://github.com/nipy/nipype/pull/3150/files#diff-1353462138ab420e9b6ce7f8a19ab284R407-R456

### DIFF
--- a/nipype/interfaces/base/tests/test_experimental.py
+++ b/nipype/interfaces/base/tests/test_experimental.py
@@ -2,7 +2,7 @@
 import sys
 from ....utils.filemanip import fname_presuffix
 from ... import base as nib
-from ..experimental import AutoOutputInterface
+from ..experimental import Interface
 
 _TOOL_OUTPUT = r"""\
   _____________
@@ -41,9 +41,9 @@ def test_AutoOutputInterface():
         out_std = nib.Str(stdout=True)
         out_err = nib.traits.Float(stderr=_parse_err_cb)
 
-    class TestInterface(AutoOutputInterface):
-        input_spec = _InputSpec
-        output_spec = _OutputSpec
+    class TestInterface(Interface):
+        _input_spec = _InputSpec
+        _output_spec = _OutputSpec
 
         def _run_interface(self, runtime):
             print(_TOOL_OUTPUT)
--- a/nipype/interfaces/base/tests/test_experimental.py
+++ b/nipype/interfaces/base/tests/test_experimental.py
@@ -1,4 +1,5 @@
 """Proof of concept."""
+import sys
 from ....utils.filemanip import fname_presuffix
 from ... import base as nib
 from ..experimental import AutoOutputInterface
@@ -45,8 +46,8 @@ def test_AutoOutputInterface():
         output_spec = _OutputSpec
 
         def _run_interface(self, runtime):
-            runtime.stdout = _TOOL_OUTPUT
-            runtime.stderr = ' '.join(('a', 'b', '1', '2', '3.0'))
+            print(_TOOL_OUTPUT)
+            print(' '.join(('a', 'b', '1', '2', '3.0')), file=sys.stderr)
             out_fname = fname_presuffix(
                 self.inputs.woo, suffix='_brain')
             open(out_fname, 'w').close()
--- a/nipype/interfaces/base/tests/test_experimental.py
+++ b/nipype/interfaces/base/tests/test_experimental.py
@@ -1,0 +1,60 @@
+"""Proof of concept."""
+from ....utils.filemanip import fname_presuffix
+from ... import base as nib
+from ..experimental import AutoOutputInterface
+
+_TOOL_OUTPUT = r"""\
+  _____________
+< Nipype rocks! >
+  =============
+                  \
+                   \
+                     ^__^
+                     (oo)\_______
+                     (__)\       )\/\
+                         ||----w |
+                         ||     ||
+
+
+"""
+
+
+def _parse_err_cb(in_text):
+    if '3.0' in in_text:
+        return 3.0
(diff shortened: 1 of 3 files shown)


effigies:
This seems reasonable on first glance. I think problems are likely to be found while trying to work through a few examples.

I assume that we'll still support the old style, though? Otherwise a lot of interfaces will need to be rewritten mostly from scratch.

nipype/interfaces/base/tests/test_experimental.py
```diff
@@ -0,0 +1,60 @@
...
+    zoo = nib.File(desc="a file")
+    woo = nib.File(desc="a file")
+
+
+def test_AutoOutputInterface():
+    """Proof of concept."""
+    class _OutputSpec(nib.TraitedSpec):
+        out_woo = nib.File('{woo}_brain', usedefault=True)
+        out_std = nib.Str(stdout=True)
+        out_err = nib.traits.Float(stderr=_parse_err_cb)
```
satra: this functional form is the part that would need to be formalized if we are going to transform the interface into a specification for mapping between software (e.g., nipype1 to pydra spec).

satra:
@oesteban - this looks reasonable.

nipype/interfaces/base/tests/test_experimental.py
oesteban: As I mentioned above/below, the std out/err is actually easier once we convert these into metadata. Parse callables can be reused right away and the semantics of the metadata are clear enough (this is basically one of the most appealing points of this suggestion).

``_list_outputs`` is going to be harder to convert from code to metadata.

oesteban (commented):
> I assume that we'll still support the old style, though?

Yes, that would be the idea. Although, if this turns out easier than we thought we could think of a new config ``compatibility_mode`` that loads the legacy interfaces when on and these new when off.

> Otherwise a lot of interfaces will need to be rewritten mostly from scratch.

For ``aggregate_outputs`` not those many:

```
$ git grep aggregate_outputs
...
nipype/interfaces/afni/preprocess.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/afni/preprocess.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/afni/utils.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/afni/utils.py:        outputs = super(Autobox, self).aggregate_outputs(runtime, needed_outputs)
nipype/interfaces/afni/utils.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/ants/registration.py:    >>> reg4b.aggregate_outputs()  # doctest: +SKIP
nipype/interfaces/ants/registration.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/base/core.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/base/core.py:    This class does not implement aggregate_outputs, input_spec or
nipype/interfaces/base/core.py:            outputs = self.aggregate_outputs(runtime)
nipype/interfaces/base/core.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/base/tests/test_core.py:        nif.aggregate_outputs()
nipype/interfaces/freesurfer/preprocess.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/freesurfer/preprocess.py:        return super(SegmentCC, self).aggregate_outputs(runtime, needed_outputs)
nipype/interfaces/freesurfer/utils.py:        # aggregate_outputs.  Let's try to parse stderr and raise a
nipype/interfaces/freesurfer/utils.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/fsl/model.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/fsl/preprocess.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/fsl/preprocess.py:        outputs = super(FLIRT, self).aggregate_outputs(
nipype/interfaces/fsl/utils.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/niftyseg/label_fusion.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/interfaces/spm/model.py:    def aggregate_outputs(self, runtime=None):
nipype/interfaces/spm/model.py:    def aggregate_outputs(self, runtime=None, needed_outputs=None):
nipype/pipeline/engine/nodes.py:            aggouts = self._interface.aggregate_outputs(
```

Most of these redefine ``aggregate_outputs`` to parse standard out/err.

The use of ``_list_outputs`` is more pervasive and will probably require some kind of high-level parsing (which we would need anyways in the migration to pydra if we want to make interfaces have less weight and boilerplate). I'm worried this part could be a lot of work.

PR #3152, opened 2020-01-11
fix: improve version checking for nodes of workflows

satra:
Reduce number of version checks for nodes of workflows run through multiprocessing .

This is a rather drastic step, and we should think through how to handle this in general. this will prevent this for other libraries which are used as well.

### DIFF
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -124,7 +124,7 @@ def check_latest_version(raise_exception=False):
 if config.getboolean("execution", "check_version"):
     import __main__
 
-    if not hasattr(__main__, "__file__"):
+    if not hasattr(__main__, "__file__") and "NO_NIPYPE_ET" in os.environ:
         from .interfaces.base import BaseInterface
 
         if BaseInterface._etelemetry_version_data is None:
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -14,11 +14,7 @@ Top-level module API
 import os
 from distutils.version import LooseVersion
 
-from .info import (
-    URL as __url__,
-    STATUS as __status__,
-    __version__,
-)
+from .info import URL as __url__, STATUS as __status__, __version__
 from .utils.config import NipypeConfig
 from .utils.logger import Logging
 from .refs import due
@@ -105,6 +101,8 @@ def check_latest_version(raise_exception=False):
                         packname="nipype", version=__version__, latest=latest["version"]
                     )
                 )
+            else:
+                logger.info("No new version available.")
             if latest["bad_versions"] and any(
                 [
                     LooseVersion(__version__) == LooseVersion(ver)
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -124,7 +124,7 @@ def check_latest_version(raise_exception=False):
 if config.getboolean("execution", "check_version"):
     import __main__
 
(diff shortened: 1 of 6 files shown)


effigies:
What if we add a new environment variable like `NIPYPE_SKIP_ET`?

nipype/__init__.py
```diff
@@ -126,7 +124,7 @@ def check_latest_version(raise_exception=False):
 if config.getboolean("execution", "check_version"):
     import __main__
 
-    if not hasattr(__main__, "__file__"):
+    if not hasattr(__main__, "__file__") and "NO_NIPYPE_ET" in os.environ:
```
effigies:
```suggestion
    if not hasattr(__main__, "__file__") and "NO_NIPYPE_ET" not in os.environ:
```

nipype/interfaces/base/core.py
```diff
@@ -168,7 +168,10 @@ class BaseInterface(Interface):
     def __init__(
         self, from_file=None, resource_monitor=None, ignore_exception=False, **inputs
     ):
-        if config.getboolean("execution", "check_version"):
+        if (
+            config.getboolean("execution", "check_version")
+            and "NO_NIPYPE_ET" in os.environ
```
effigies:
```suggestion
            and "NO_NIPYPE_ET" not in os.environ
```

nipype/pipeline/plugins/multiproc.py
```diff
@@ -73,6 +73,12 @@ def run_node(node, updatehash, taskid):
     return result
 
 
+def process_initializer(cwd):
+    """Initializes the environment of the child process"""
+    os.chdir(cwd)
+    os.environ["NO_ET"] = "1"
```
effigies:
```suggestion
    os.environ["NO_NIPYPE_ET"] = "1"
```

nipype/pipeline/plugins/tools.py
```diff
@@ -125,7 +125,14 @@ def create_pyscript(node, updatehash=False, store_exception=True):
     can_import_matplotlib = False
     pass
 
+import os
+value = os.environ.get('NO_NIPYPE_ET', None)
+if value is None:
+    os.environ['NO_NIPYPE_ET'] = "1"
 from nipype import config, logging
+if value is None:
+    del os.environ['NO_NIPYPE_ET']
```
effigies: I don't really understand this section. When is `create_pyscript` used, and why would you reenable etelemetry?

satra: this is used for all HPC execution (SGE, SLURM, etc.,.).

effigies:
This seems to work. See satra#31 for a test and a couple small suggestions.

Also, we currently support two other environment variables: `NIPYPE_NO_MATLAB` and `NIPYPE_CONFIG_DIR`. My inclination would be to stick with the `NIPYPE_*` convention. Unless you have a strong preference for `NO_NIPYPE_ET`?

effigies:
Looking at the [test failure](https://travis-ci.org/nipy/nipype/jobs/637261849), I'm unable to replicate it easily. It appears to be related to https://github.com/matplotlib/matplotlib/issues/15832  and https://discourse.matplotlib.org/t/issues-with-my-installation-of-matplotlib/20746.

I can fiddle more later, but I wouldn't hold up merge for it.

Only remaining comment is that, as a bug fix, we should target `maint/1.4.x` and rebase on it.

PR #3153, opened 2020-01-15
MNT: Update requirements.txt post-1.4

effigies:
## Summary
While looking at the Travis failure, I realized we were still installing things intended for Python 2 and 3.4.

This adds a `tools/update_requirements.py` file to make keeping the requirements file in sync with the setup configuration easy.

## Acknowledgment


effigies:
This also fixes the Travis failures (somewhat unsatisfyingly) by using `--upgrade`. This isn't great because it doesn't help catch cases where upstream pre-releases are breaking things for older versions of *other* libraries, but that's also not our responsibility, and we do want to check the latest releases of all libraries.

PR #3154, opened 2020-01-15
CI: Test Python 3.8

### LINKED ISSUE #3098: Test on Python 3.8
### Summary

Python 3.8 has been out for a month or so, and upstream libraries are mostly providing wheels. We should start 3.8 tests when it's feasible.

Necessary steps:

* [ ] Update `.travis.yml` to use Python 3.8
* [ ] Note any failures, identify whether they exist on our end or in a dependency
* [ ] Find blocking upstream issues and link to them so we can track
* [ ] When tests pass and urgent `FutureWarning`/`DeprecationWarning`s are dealt with, update the classifiers in `nipype/info.py` to indicate we support Python 3.8.

I marked this [good-first-issue](https://github.com/nipy/nipype/labels/good-first-issue) but it's less for a new developer than someone looking to get more involved in project maintenance.

effigies:
Closes #3098.

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -56,6 +56,7 @@ CLASSIFIERS = [
     "Operating System :: POSIX :: Linux",
     "Programming Language :: Python :: 3.6",
     "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
     "Topic :: Scientific/Engineering",
 ]
 PYTHON_REQUIRES = ">= 3.6"
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -110,6 +110,7 @@ DATEUTIL_MIN_VERSION = "2.2"
 FUTURE_MIN_VERSION = "0.16.0"
 SIMPLEJSON_MIN_VERSION = "3.8.0"
 PROV_VERSION = "1.5.2"
+RDFLIB_MIN_VERSION = "5.0.0"
 CLICK_MIN_VERSION = "6.6.0"
 PYDOT_MIN_VERSION = "1.2.3"
 
@@ -144,16 +145,13 @@ REQUIRES = [
     "pydot>=%s" % PYDOT_MIN_VERSION,
     "pydotplus",
     "python-dateutil>=%s" % DATEUTIL_MIN_VERSION,
+    "rdflib>=%s" % RDFLIB_MIN_VERSION,
     "scipy>=%s" % SCIPY_MIN_VERSION,
     "simplejson>=%s" % SIMPLEJSON_MIN_VERSION,
     "traits>=%s,!=5.0" % TRAITS_MIN_VERSION,
     "filelock>=3.0.0",
     "etelemetry>=0.2.0",
 ]
-
-# neurdflib has to come after prov
-# https://github.com/nipy/nipype/pull/2961#issuecomment-512035484
-REQUIRES += ["neurdflib"]
 
 TESTS_REQUIRES = [
     "codecov",
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ language: python
 python:
 - 3.6
 - 3.7
+- 3.8
 
 env:
   global:
--- a/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py
+++ b/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py
@@ -134,6 +134,7 @@ def run_multiproc_nondaemon_with_flag(nondaemon_flag):
     return result
 
 
+@pytest.mark.skipif(sys.version_info >= (3, 8), reason="multiprocessing issues in Python 3.8")
 def test_run_multiproc_nondaemon_false():
     """
     This is the entry point for the test. Two times a pipe of several
@@ -152,6 +153,7 @@ def test_run_multiproc_nondaemon_false():
     assert shouldHaveFailed
 
 
+@pytest.mark.skipif(sys.version_info >= (3, 8), reason="multiprocessing issues in Python 3.8")
 def test_run_multiproc_nondaemon_true():
     # with nondaemon_flag = True, the execution should succeed
     result = run_multiproc_nondaemon_with_flag(True)
--- a/nipype/pipeline/plugins/tests/test_multiproc.py
+++ b/nipype/pipeline/plugins/tests/test_multiproc.py
@@ -4,6 +4,7 @@
 """
 Test the resource management of MultiProc
 """
+import sys
 import os
 import pytest
 from nipype.pipeline import engine as pe
@@ -33,6 +34,7 @@ class MultiprocTestInterface(nib.BaseInterface):
         return outputs
 
 
+@pytest.mark.skipif(sys.version_info >= (3, 8), reason="multiprocessing issues in Python 3.8")
 def test_run_multiproc(tmpdir):
     tmpdir.chdir()
 
@@ -114,6 +116,7 @@ def test_no_more_threads_than_specified(tmpdir):
         pipe.run(plugin="MultiProc", plugin_args={"n_procs": max_threads})
 
 
+@pytest.mark.skipif(sys.version_info >= (3, 8), reason="multiprocessing issues in Python 3.8")
 def test_hold_job_until_procs_available(tmpdir):
     tmpdir.chdir()
 
--- a/nipype/pipeline/plugins/tests/test_oar.py
+++ b/nipype/pipeline/plugins/tests/test_oar.py
@@ -29,8 +29,9 @@ class OarTestInterface(nib.BaseInterface):
         return outputs
 
 
-@pytest.mark.skipif(which('oarsub') is None, reason="OAR not installed")
-def test_run_pbsgraph(tmp_path):
+@pytest.mark.skipif(which("oarsub") is None, reason="OAR not installed")
+@pytest.mark.timeout(60)
+def test_run_oargraph(tmp_path):
     pipe = pe.Workflow(name="pipe", base_dir=str(tmp_path))
     mod1 = pe.Node(interface=OarTestInterface(), name="mod1")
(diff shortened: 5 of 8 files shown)


effigies:
Tests stall on [`nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py::test_run_multiproc_nondaemon_false`](https://github.com/nipy/nipype/blob/1a869991adc024577536689d557fc748c764f15d/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py#L137-L152).

effigies:
It seems that that test actually breaks the `multiprocessing` pool. Locally it will pass, but the next test will fail to start up jobs. Might also be causing problems with `pytest-xdist`.

@oesteban Just in case you have ideas.

oesteban:
Sorry, I was away for the weekend. I'm marking this unread and will come back to it later today.

yarikoptic:
observing the stall of test_run_multiproc_nondaemon_false while building 1.4.2 on Debian (unstable/sid) as of yesterday Mar 22.

yarikoptic:
FWIW test_run_multiproc_nondaemon_true hangs too ;-/

yarikoptic:
and test_run_multiproc

effigies:
To clarify: All for Python 3.8? Or is this a Python 3.7 issue?

@oesteban I'm considering marking these as `skip` if Python >= 3.8.

yarikoptic:
I believe it is 3.8 according to process trace:
```
                               /bin/sh -c export PYTHONPATH=/build/nipype-1.4.2/.pybuild/cpython3_3.8_nipype/build; \ cd build && \          MPLCONFIGDIR=/build/nipype-1.4.2/build LC_ALL=C.UTF-8 py.test-3 -v --doctest-modules ../nipype
                                 /usr/bin/python3 /usr/bin/py.test-3 -v --doctest-modules ../nipype
                                   /usr/bin/python3 /usr/bin/py.test-3 -v --doctest-modules ../nipype
                                   /usr/bin/python3 /usr/bin/py.test-3 -v --doctest-modules ../nipype
```

also `test_hold_job_until_procs_available` stalls. 

<details>
<summary>My current diff with skips now</summary> 

```shell
# quilt diff
Index: nipype-1.4.2/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py
===================================================================
--- nipype-1.4.2.orig/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py
+++ nipype-1.4.2/nipype/pipeline/plugins/tests/test_legacymultiproc_nondaemon.py
@@ -134,6 +134,7 @@ def run_multiproc_nondaemon_with_flag(no
     return result
 
 
+@pytest.mark.skipif(True, reason="hangs, known upstream")
 def test_run_multiproc_nondaemon_false():
     """
     This is the entry point for the test. Two times a pipe of several
@@ -152,6 +153,7 @@ def test_run_multiproc_nondaemon_false()
     assert shouldHaveFailed
 
 
+@pytest.mark.skipif(True, reason="hangs, known upstream")
 def test_run_multiproc_nondaemon_true():
     # with nondaemon_flag = True, the execution should succeed
     result = run_multiproc_nondaemon_with_flag(True)
Index: nipype-1.4.2/nipype/pipeline/plugins/tests/test_multiproc.py
===================================================================
--- nipype-1.4.2.orig/nipype/pipeline/plugins/tests/test_multiproc.py
+++ nipype-1.4.2/nipype/pipeline/plugins/tests/test_multiproc.py
@@ -33,6 +33,7 @@ class MultiprocTestInterface(nib.BaseInt
         return outputs
 
 
+@pytest.mark.skipif(True, reason="hangs, known upstream")
 def test_run_multiproc(tmpdir):
     tmpdir.chdir()
 
@@ -114,6 +115,7 @@ def test_no_more_threads_than_specified(
         pipe.run(plugin="MultiProc", plugin_args={"n_procs": max_threads})
 
 
+@pytest.mark.skipif(True, reason="hangs, known upstream")
 def test_hold_job_until_procs_available(tmpdir):
     tmpdir.chdir()

```
</details>

effigies:
This seems concerning, as I think this means the multiprocessing plugins can't be tested with Python 3.8. The best case scenario is that this is an interaction with pytest, but that would be very bad...

yarikoptic:
eh, I am just an unlucky in general, even `nipype/tests/test_nipype.py::test_no_et` seems to stall (no etelemetry client is installed if that is relevant), although I believe that on previous run it went through.

yarikoptic:
and now they hang on `test_provenance`... and if I skip that one `test_provenance_exists` hangs.
given that it is at 99% I wonder if that is just not a side effect of some previous test ...

effigies:
Do you have `qsub` installed on your test rig?

yarikoptic:
> qsub

I hope not - there is no any pbs/slurm in the build env.  Shouldn't tests requiring it skip if no `qsub`?

effigies:
It's currently being marked `xfail`. I'll propose a PR to make it `skipif`.

yarikoptic:
ok, if I take that patch too, I still get stalling on `no_et` - if I skip it -- on `test_provenance`... didn't try yet again to extend skips.  I will check first if it is etelemetry altogether disabling/removing it.

effigies:
I can't replicate the provenance one. What versions of `prov` and `rdflib` are you using? Are you aware of the weird hack of needing to install `neurdflib` after `prov`?

satra:
hopefully that hack will be resolved after rdflib 5.0.0 is released.

effigies:
Agreed. Does installing from `rdflib@master` have any issues?

satra:
yes, at this point the neurdflib PRs are now in master. so rdflib@master can be used instead of neurdflib.

yarikoptic:
oh, I was not aware of any hacks like `neurdflib` which is not in Debian, and I am not sure if it should be if it is a temporary band-aid.  I cannot "install from master" while building debian package -- all external dependencies should come as debian packages.  Re which versions:

```shell
# python3 -c 'import prov,rdflib; print(prov.__version__)'
INFO:rdflib:RDFLib Version: 4.2.2
1.5.2
```

Could nipype detect deficient rdflib (absent `neurdflib`) and skip tests accordingly?

anyways -- skipping those I get to the same state of a stall upon exit from testing.
I wonder if may be additional warnings like 

```
testing/tests/test_utils.py::test_tempfatfs
  /build/nipype-1.4.2/nipype/testing/tests/test_utils.py:18: UserWarning: Cannot mount FAT filesystems with FUSE
    warnings.warn("Cannot mount FAT filesystems with FUSE")
```
are of relevance here... heh heh

effigies:
I've replaced those warnings with the more appropriate `raise SkipTest` in 858d2dd, and conditioned testing prov on a good version of rdflib in 8a5b619. (neurdflib claims version 5.0, and the next release of rdflib will be 5.0, so this should be fine.)

effigies:
Rebased on #3195 to consolidate. @yarikoptic How are things now?

effigies:
I guess we should go ahead with this, as some tests are better than none, but it's probably not a good thing that we can't test (`Legacy`)`MultiProc` on Python 3.8+. Do we know if it works at all? Has anybody tried using nipype with Python 3.8 yet?

effigies:
The prov issues seem to be stochastic... I'm going to try dropping the `rdflib/neurdflib` dependency and switch to master. That will postpone any inclusion for after the 5.0.0 release, but at this point it seems worth it.

effigies:
Not resolved by `rdflib@master`.

effigies:
Now that I've put a timeout that might give us some debugging output if it does hang again, it won't hang... Rerunning again...

nipype/info.py
```diff
@@ -143,17 +143,14 @@ def get_nipype_gitversion():
     "pydot>=%s" % PYDOT_MIN_VERSION,
     "pydotplus",
     "python-dateutil>=%s" % DATEUTIL_MIN_VERSION,
+    "rdflib @ git+https://github.com/rdflib/rdflib.git@master",
```
yarikoptic, line 146: this would go into `setup.py`... AFAIK ideally such restrictions are to be stated only in `requirements.txt` and here only minimal or problematic versions to be listed.

yarikoptic:
FWIW, I thought that I might just avoid "stall" by simply using older nipype -- nope, elderly nipype 1.1.9 also gets stuck there, so should add skips.

nipype/info.py
effigies: Yeah, this is a temporary measure to ensure that we're testing against latest. rdflib 5.0 is due to be released this weekend, and I'll switch it then.

yarikoptic:
re `test_run_pbsgraph`:

> It's currently being marked `xfail`. I'll propose a PR to make it `skipif`.

could/should it also get a timeout then, since xfail is no good if it just hangs there.

effigies:
@yarikoptic Timeouts added to OAR/PBS tests.

doc/requirements.txt
```diff
@@ -3,6 +3,6 @@ ipython
 matplotlib
 nbsphinx
 sphinx-argparse
-sphinx>=2.1.2
+sphinx>=2.1.2,<3
```
satra, line 6: @effigies - i fixed a sphinx related bug on pydra a couple of days back. if its the same issue, perhaps we can apply that here as well. it had to do with a `:noindex:` warning becoming an error.

effigies: I think the issue is with some custom Sphinx extensions using a deprecated API. In any event, I'd prefer to fix that separately. If we do before this is merged, I can rebase this commit away.

yarikoptic:
FWIW, I have not tried this specific version, but I think I have all the skips - but then good old 1.1.9 still hangs at some later tests (at about 98% completness of the tests).  If someone gets down to the underlying issue which causes the stalls -- would be very appreciated and would open possibility to upload recent nipype to debian proper.

effigies:
This gets things going, so I'm going to merge. I'll create an issue for Python 3.8 + multiprocessing, and feel free to open more outstanding issues.

PR #3170, opened 2020-02-14
FIX: Trait 6 import

mgxd:
Changed import in traits 6


effigies:
Not sure this is needed. I've restarted Travis on master: https://travis-ci.org/nipy/nipype/builds/649672306?utm_medium=notification&utm_source=github_status

If that passes, we should just be able to push a bug-fix release.

mgxd:
Yeah looks like all we need is a new release

we should probably still just import `TraitType` by its [proper file](https://github.com/enthought/traits/blob/ba0c0caf6c58f2632019b8cd7b9985e394812223/traits/trait_type.py#L88).

Issue #3178, opened 2020-02-26
interfaces.ants.segmentation does not have options for PriorLabelImage arguments


eilidhmacnicol:
Summary: PriorLabelImage argument for Atroposis accepted as an initialization, but there is no option for PriorLabelImage declaration and prior weighting isn't included in the command. So:
command-line argument currently reads `--initialization PriorLabelImage[k]`
instead of: `--initialization PriorLabelImage[k, priorlabelimage.nii, weighting]`

effigies:
What are the types of the inputs? Looks like the second is a file, and I assume weighting is a float (does it have a limited range?). Is `k` an integer?

eilidhmacnicol:
That's it exactly: k is the number of segments to be produced, so an integer, then a discrete segmentation image to use as a prior, and a weighting score for the prior (relative to the input image) between 0 and 1.

PR #3180, opened 2020-02-27
FIX: ANTs' tools maintenance overhaul

oesteban:
This PR:
- [x] Revises the inputs of `Atropos` (#3178)
- [x] Upstreams interfaces from https://github.com/nipreps/niworkflows/blob/d6f6d3956c46d634ef7c4b0219d0c6ea94b0412b/niworkflows/interfaces/ants.py
- [x] Removes the `JointFusion` interface, for which there hasn't been any corresponding command line binary for very long time now.


oesteban:
How much testing are we expecting for this @satra, @effigies? Any reasons not to go ahead when all lights are green and merge?

nipype/interfaces/ants/segmentation.py
```diff
@@ -1496,75 +1394,91 @@ class AntsJointFusionInputSpec(ANTSCommandInputSpec):
...
-    out_intensity_fusion_name_format = traits.Str()
-    out_label_post_prob_name_format = traits.Str()
-    out_atlas_voting_weight_name_format = traits.Str()
+    out_intensity_fusion = OutputMultiPath(File(exists=True))
+    out_label_post_prob = OutputMultiPath(File(exists=True))
+    out_atlas_voting_weight = OutputMultiPath(File(exists=True))
 
 
-class AntsJointFusion(ANTSCommand):
+class JointFusion(ANTSCommand):
```
satra: why did this get renamed? i'm pretty positive mindboggle uses this interface.

nipype/interfaces/ants/segmentation.py
```diff
@@ -1577,90 +1491,88 @@ def _format_arg(self, opt, spec, val):
...
+                os.path.abspath(
+                    self.inputs.out_atlas_voting_weight_name_format.replace("%d", "*")
+                )
             )
-
         return outputs
 
 
+# For backwards compatibility
+AntsJointFusion = JointFusion
```
satra: i see this now.

effigies:
I was waiting on CI to pass to read through. I'll look tonight or tomorrow morning.

nipype/interfaces/ants/segmentation.py
oesteban: Yup, I guess the Ants prefix was added not to clobber the old JointFusion. Now that the legacy binary is not supported, we can normalize the naming w.r.t. to the rest of the ants module.

As you found below, I added an alias for backward compatibility.

nipype/interfaces/ants/segmentation.py
```diff
@@ -31,7 +31,12 @@ class AtroposInputSpec(ANTSCommandInputSpec):
         requires=["number_of_tissue_classes"],
         mandatory=True,
     )
-    prior_probability_images = InputMultiPath(File(exists=True))
+    kmeans_init_centers = traits.List(traits.Either(traits.Int, traits.Float), minlen=1)
+    prior_image = traits.Either(
+        File(exists=True),
+        traits.Str,
+        desc="either a string pattern (e.g., 'prior%%02d.nii') or an existing vector-image file.",
```
effigies: Is this string being interpolated somewhere?

```suggestion
        desc="either a string pattern (e.g., 'prior%02d.nii') or an existing vector-image file.",
```

nipype/interfaces/ants/utils.py
```diff
@@ -1,11 +1,456 @@
...
+
+    _cmd = "ImageMath"
+    input_spec = _ImageMathInputSpec
+    output_spec = _ImageMathOuputSpec
+
+    def _list_outputs(self):
+        outputs = super(ImageMath, self)._list_outputs()
+        if self.inputs.copy_header:  # Fix headers
+            _copy_header(self.inputs.op1, outputs["output_image"], keep_dtype=True)
+        return outputs
```
effigies: @satra Isn't `_list_outputs()` supposed to be usable pre-`run()`? This seems like it fail in that case. I believe `aggregate_outputs()` is always post-`run`?

nipype/interfaces/ants/utils.py
```diff
@@ -1,11 +1,456 @@
-# -*- coding: utf-8 -*-
-"""ANTS Apply Transforms interface
-"""
-
+"""ANTs' utilities."""
 import os
+from ...utils.imagemanip import copy_header as _copy_header
+from ..base import traits, isdefined, TraitedSpec, File, Str, InputMultiPath
```
effigies: `InputMultiPath` is now `InputMultiObject`.

```suggestion
from ..base import traits, isdefined, TraitedSpec, File, Str, InputMultiObject
```

nipype/interfaces/ants/utils.py
```diff
@@ -1,11 +1,456 @@
...
+from .base import ANTSCommandInputSpec, ANTSCommand
+
+
+class _ImageMathInputSpec(ANTSCommandInputSpec):
```
effigies: This convention was started in niworkflows but isn't generally used in nipype. Conventions that differ from file to file seems disorienting for contributors.

```suggestion
class ImageMathInputSpec(ANTSCommandInputSpec):
```

nipype/interfaces/ants/utils.py
oesteban: ha, good point. I'll change.

nipype/utils/tests/test_imagemanip.py
```diff
@@ -0,0 +1,33 @@
...
+def test_copy_header(tmp_path, keep_dtype):
+    """Cover copy_header."""
+    fname1 = tmp_path / 'reference.nii.gz'
+    fname2 = tmp_path / 'target.nii.gz'
+
+    nii = nb.Nifti1Image(
+        np.zeros((10, 10, 10), dtype='uint8'), None, None)
+    nii.set_qform(np.diag((1., 2., 3., 1.)), code=2)
+    nii.set_sform(np.diag((1., 2., 3., 1.)), code=1)
+    nii.to_filename(fname1)
```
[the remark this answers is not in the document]
oesteban: will change, thx

nipype/interfaces/ants/utils.py
oesteban: I'd like to see NiPype slowly migrating to this convention. Would you be okay if the whole module included the change (adding aliases so that we don't break the work directories anytime soon)?

effigies: Maybe let's have that conversation in a separate issue, rather than wedging it into this PR?

oesteban: Okay, sounds reasonable

oesteban: BTW - this is going to break fMRIPrep working directory.

effigies:
Suggestions proposed on oesteban#10.

oesteban:
@satra - thumbs up after Chris' suggestions?

oesteban:
Having issues with `N4BiasFieldCorrection`:

```Python
  File "/home/oesteban/workspace/nipype/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/home/oesteban/workspace/nipype/nipype/pipeline/engine/nodes.py", line 516, in run
  File "/home/oesteban/workspace/nipype/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
  File "/home/oesteban/workspace/nipype/nipype/pipeline/engine/nodes.py", line 741, in _run_command
  File "/home/oesteban/workspace/nipype/nipype/interfaces/base/core.py", line 398, in run
    runtime = self._post_run_hook(runtime)
  File "/home/oesteban/workspace/nipype/nipype/interfaces/mixins/fixheader.py", line 132, in _post_run_hook
    _copy_header(inputs[inp], outputs[out], keep_dtype=keep_dtype)
KeyError: 'bias_image'
```
it seems the mixin is not general enough?

effigies:
Oh, I hadn't considered the output file might not exist. Yeah, I think a simple existence check makes sense. I think silent pass-through is fine for a mixin, as it's not responsible to make sure the file exists, only that if the file exists it has the right header.

PR #3182, opened 2020-03-04
FIX: load_resultfile crashes if open resultsfile from crashed job

daniel-ge:
## Problem
If a job crashes on computing nodes a result-file is written containing a dictionary (with traceback, etc). Then the 'controller' job reads the dictionary in order to produce the pickeled crash file. During reading of the results-file an exception arises in `load_resultfile` with `AttributeError: 'dict' object has no attribute 'outputs'`.

## Details
In #2985 `loadpkl` was replaced by `load_resultfile`. But `result_data` could be a dictionary as assumed in the following source snippet:
https://github.com/nipy/nipype/blob/b75a7cebc8dbefa796ed2504ca40197e65a9549e/nipype/pipeline/plugins/base.py#L531-L536
But `result_data` will never be a dictionary, because in `load_resultfile`  it has to have an attribute `outputs`:
https://github.com/nipy/nipype/blob/b75a7cebc8dbefa796ed2504ca40197e65a9549e/nipype/pipeline/engine/utils.py#L293-L294

## Solution
Just check if `result` has an attribute `outputs` before accessing it.

### DIFF
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -291,7 +291,7 @@ def load_resultfile(results_file, resolve=True):
         raise FileNotFoundError(results_file)
 
     result = loadpkl(results_file)
-    if resolve and result.outputs:
+    if resolve and hasattr(result,"outputs") and result.outputs:
         try:
             outputs = result.outputs.get()
         except TypeError:  # This is a Bunch
--- a/nipype/pipeline/engine/utils.py
+++ b/nipype/pipeline/engine/utils.py
@@ -291,7 +291,7 @@ def load_resultfile(results_file, resolve=True):
         raise FileNotFoundError(results_file)
 
     result = loadpkl(results_file)
-    if resolve and hasattr(result,"outputs") and result.outputs:
+    if resolve and getattr(result, "outputs", None):
         try:
             outputs = result.outputs.get()
         except TypeError:  # This is a Bunch
--- a/nipype/pipeline/plugins/tests/test_sgelike.py
+++ b/nipype/pipeline/plugins/tests/test_sgelike.py
@@ -27,8 +27,6 @@ def is_pending(self, taskid):
 @patch.object(SGELikeBatchManagerBase, '_submit_batchtask', new=submit_batchtask)
 @patch.object(SGELikeBatchManagerBase, '_is_pending', new=is_pending)
 def test_crashfile_creation(tmp_path):
-    cur_dir = os.getcwd()
-    with TemporaryDirectory(prefix="test_engine_", dir=cur_dir) as tmpdirname:
         pipe = pe.Workflow(name="pipe", base_dir=tmpdirname)
         pipe.config["execution"]["crashdump_dir"] = tmpdirname
         pipe.add_nodes([pe.Node(interface=Function(function=crasher), 
--- a/nipype/pipeline/plugins/tests/test_sgelike.py
+++ b/nipype/pipeline/plugins/tests/test_sgelike.py
@@ -37,5 +37,5 @@ def test_crashfile_creation(tmp_path):
         assert (str(e.value) == 
                 "Workflow did not execute cleanly. Check log for details")
         
-        crashfiles = glob(join(tmpdirname,"crash*crasher*.pklz"))        
+        crashfiles = tmp_path.glob("crash*crasher*.pklz")
         assert len(crashfiles) == 1
--- a/nipype/pipeline/plugins/tests/test_sgelike.py
+++ b/nipype/pipeline/plugins/tests/test_sgelike.py
@@ -13,8 +13,8 @@ import subprocess
 def crasher():
     raise ValueError()
 
-    
-def submit_batchtask(self, scriptfile, node):    
+
+def submit_batchtask(self, scriptfile, node):
     self._pending[1] = node.output_dir()
     subprocess.call(["bash", scriptfile])
     return 1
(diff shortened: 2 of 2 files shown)


nipype/pipeline/engine/utils.py
```diff
@@ -291,7 +291,7 @@ def load_resultfile(results_file, resolve=True):
         raise FileNotFoundError(results_file)
 
     result = loadpkl(results_file)
-    if resolve and result.outputs:
+    if resolve and hasattr(result,"outputs") and result.outputs:
```
effigies: Simpler and has a space after the commas:

```suggestion
    if resolve and getattr(result, "outputs", None):
```

effigies:
Hi @daniel-ge, any chance of a regression test? Let us know if you need help.

daniel-ge:
That's a first attempt for a regression test. I am pretty sure there is much room for improvements. I tried to make it independent of existing/available installations of SLUM, SGE etc by running the batch script directly.

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
+from nipype.pipeline.plugins.base import SGELikeBatchManagerBase
+from nipype.interfaces.utility import Function
+import nipype.pipeline.engine as pe
+from os.path import join
+import os
+from glob import glob
+import pytest
+from mock import patch
```
effigies: We should be able to use the builtin `unittest.mock`.

```suggestion
from unittest.mock import patch
```

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
...
+    return 1
+
+
+def is_pending(self, taskid):
+    return False
+
+
+@patch.object(SGELikeBatchManagerBase, '_submit_batchtask', new=submit_batchtask)
+@patch.object(SGELikeBatchManagerBase, '_is_pending', new=is_pending)
+def test_crashfile_creation():
```
effigies: Let's use the `tmp_path` fixture. This will replace the `TemporaryDirectory` construct.

```suggestion
def test_crashfile_creation(tmp_path):
```

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
...
+    cur_dir = os.getcwd()
+    with TemporaryDirectory(prefix="test_engine_", dir=cur_dir) as tmpdirname:
```
effigies:
```suggestion
```

And all of the internal block needs to be de-indented. Can you run `black` on this file as well?

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
...
+                            name="crasher")])    
+        sgelike_plugin = SGELikeBatchManagerBase("")
+        with pytest.raises(RuntimeError) as e:
+            assert pipe.run(plugin=sgelike_plugin)
+        assert (str(e.value) == 
+                "Workflow did not execute cleanly. Check log for details")
+        
+        crashfiles = glob(join(tmpdirname,"crash*crasher*.pklz"))        
+        assert len(crashfiles) == 1
+    os.chdir(cur_dir)
```
effigies:
```suggestion
```

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
...
+def is_pending(self, taskid):
+    return False
+
+
+@patch.object(SGELikeBatchManagerBase, '_submit_batchtask', new=submit_batchtask)
+@patch.object(SGELikeBatchManagerBase, '_is_pending', new=is_pending)
+def test_crashfile_creation():
+    cur_dir = os.getcwd()
+    with TemporaryDirectory(prefix="test_engine_", dir=cur_dir) as tmpdirname:
+        pipe = pe.Workflow(name="pipe", base_dir=tmpdirname)
```
effigies:
```suggestion
        pipe = pe.Workflow(name="pipe", base_dir=str(tmp_path))
```

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
...
+        pipe.config["execution"]["crashdump_dir"] = tmpdirname
```
effigies:
```suggestion
        pipe.config["execution"]["crashdump_dir"] = str(tmp_path)
```

nipype/pipeline/plugins/tests/test_sgelike.py
```diff
@@ -0,0 +1,45 @@
...
+        pipe.add_nodes([pe.Node(interface=Function(function=crasher), 
+                            name="crasher")])    
+        sgelike_plugin = SGELikeBatchManagerBase("")
+        with pytest.raises(RuntimeError) as e:
+            assert pipe.run(plugin=sgelike_plugin)
+        assert (str(e.value) == 
+                "Workflow did not execute cleanly. Check log for details")
+        
+        crashfiles = glob(join(tmpdirname,"crash*crasher*.pklz"))
```
effigies:
```suggestion
        crashfiles = tmp_path.glob("crash*crasher*.pklz")
```

daniel-ge:
Thanks for your improvements. I have used `black` for better code style.

effigies:
@daniel-ge I removed some unused imports. Apologies for pushing to your `master` branch.

oesteban:
I suspect this PR might have messed up with nodes parameterized by iterables. Will open an issue if confirmed.

oesteban:
Okay, I just replicated with 1.4.2 so the problem is elsewhere. Sorry for the noise.

PR #3184, opened 2020-03-09
[ENH] Workflow connect performance

HippocampusGirl:
## Summary
Hi :-) I'm using nipype with fmriprep to process hundreds of subjects. When doing that, I noticed that creating the nipype workflows can be quite slow. This was mostly noticeable when there were nested workflows. To find out why, I ran a quick python cProfile.

Screenshot 2020-03-06 21 09 22

It turns out that a prominent performance bottleneck is the `workflow.connect` function, or rather the `_has_attr` function that is called by it. This function checks if a nested workflow has a specific input or output. If the input/output is not found, an error message is printed. 
To check for inputs/outputs, `_has_attr` uses the workflow's `inputs` or `outputs` properties, which return a dictionary of the inputs/outputs of all nested nodes. `_has_attr` now checks whether the target input or output is listed in the dictionary and returns. 
However, accessing these properties calls a getter function that traverses all of the workflow's child nodes to generate the dictionary. As this entire procedure is repeated over again for each new connection, it can be quite slow. 

Instead of generating the full dictionary, I propose that the `_has_attr` function should traverse the individual workflow graphs until it finds the target node (or not). I have created a quick implementation that leads to a ~6x speedup in creating my nipype/fmriprep workflows. The code passes integration tests. Maybe this code will be useful for the larger nipype community. 

Screenshot 2020-03-07 12 51 34

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -798,10 +798,10 @@ connected.
             return False
 
         if subtype == "in":
-            if not hasattr(node.inputs, attrname):
+            if not hasattr(targetnode.inputs, attrname):
                 return False
         else:
-            if not hasattr(node.outputs, attrname):
+            if not hasattr(targetnode.outputs, attrname):
                 return False
 
         if subtype == "in":
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -771,11 +771,25 @@ connected.
         """Checks if a parameter is available as an input or output
         """
         hierarchy = parameter.split(".")
+
+        # Connecting to a workflow needs at least two values,
+        # the name of the child node and the name of the input/output
+        if len(hierarchy) < 2:
+            return False
+
         attrname = hierarchy.pop()
         nodename = hierarchy.pop()
 
+        def _check_is_already_connected(workflow, node, attrname):
+            for _, _, d in workflow._graph.in_edges(nbunch=node, data=True):
+                for cd in d["connect"]:
+                    if attrname == cd[1]:
+                        return False
+            return True
+
         targetworkflow = self
-        for workflowname in hierarchy:
+        while hierarchy:
+            workflowname = hierarchy.pop(0)
             workflow = None
             for node in targetworkflow._graph.nodes():
                 if node.name == workflowname:
@@ -784,6 +798,13 @@ connected.
                         break
             if workflow is None:
                 return False
+            # Verify input does not already have an incoming connection
+            # in the hierarchy of workflows
+            if subtype == "in":
+                hierattrname = ".".join(hierarchy + [nodename, attrname])
+                if not _check_is_already_connected(
+                        targetworkflow, workflow, hierattrname):
+                    return False
             targetworkflow = workflow
 
         targetnode = None
@@ -804,11 +825,12 @@ connected.
             if not hasattr(targetnode.outputs, attrname):
                 return False
 
+        # Verify input does not already have an incoming connection
+        # in the target workflow
         if subtype == "in":
-            for _, _, d in targetworkflow._graph.in_edges(nbunch=targetnode, data=True):
-                for cd in d["connect"]:
-                    if attrname == cd[1]:
-                        return False
+            if not _check_is_already_connected(
+                    targetworkflow, targetnode, attrname):
+                return False
 
         return True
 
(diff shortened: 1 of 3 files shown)


nipype/pipeline/engine/workflows.py
```diff
@@ -770,16 +770,46 @@ def _check_nodes(self, nodes):
...
+                    return False
+                else:
+                    targetnode = node
+                    break
+        if targetnode is None:
+            return False
+
         if subtype == "in":
-            subobject = self.inputs
+            if not hasattr(node.inputs, attrname):
```
effigies: targetnode?

effigies:
This looks great! We'll need to check the failing test...

nipype/pipeline/engine/workflows.py
HippocampusGirl: Yes, that should be targetnode. 
```suggestion
            if not hasattr(targetnode.inputs, attrname):
```

nipype/pipeline/engine/workflows.py
```diff
@@ -770,16 +770,46 @@ def _check_nodes(self, nodes):
...
+                return False
         else:
-            subobject = self.outputs
-        attrlist = parameter.split(".")
-        cur_out = subobject
-        for attr in attrlist:
-            if not hasattr(cur_out, attr):
+            if not hasattr(node.outputs, attrname):
```
HippocampusGirl: This as well
```suggestion
            if not hasattr(targetnode.outputs, attrname):
```

nipype/pipeline/engine/workflows.py
```diff
@@ -770,16 +770,46 @@ def _check_nodes(self, nodes):
...
-            if not hasattr(cur_out, attr):
+            if not hasattr(targetnode.outputs, attrname):
                 return False
-            cur_out = getattr(cur_out, attr)
+
+        if subtype == "in":
+            for _, _, d in targetworkflow._graph.in_edges(nbunch=targetnode, data=True):
+                for cd in d["connect"]:
+                    if attrname == cd[1]:
+                        return False
```
effigies: This block is not very clear to me. A comment would help a lot.

effigies:
```suggestion
        # Verify input does not already have an incoming connection
        if subtype == "in":
            for _, _, d in targetworkflow._graph.in_edges(nbunch=targetnode, data=True):
                for cd in d["connect"]:
                    if attrname == cd[1]:
                        return False
```

effigies:
@hippocampusgirl Do you have a minute to review my suggested comment?

HippocampusGirl:
> @HippocampusGirl Do you have a minute to review my suggested comment?

The comment is great, thank you for the suggestion! I added it. 
Sorry about the slow reply, I'm applying for a PhD fellowship, and the deadline is this week, so I'm a bit swamped.

While I was looking at the comment I found that there was a case where you can connect an input twice that wasn't covered by the check and that I hadn't though of. If you have multiply nested workflows, the previous code did not check if an input was already connected in another  workflow further up in the hierarchy. 
So for example if you have three nested workflow, `flow1` (inner) <- `flow2` (middle) <- `flow3` (outer), then it was possible run first `flow2.connect(node1, "attr", flow1, "inputnode.something")` and then run `flow3.connect(node2, "attr", flow2, "flow1.inputnode.something")` without raising an error. However, it should raise an error, since they're both referring to the same node and the same input in the end. 
This is of course because running connect in a parent workflow doesn't propagate the connection information to the graph of the child workflow. The connection stays in the parent workflow graph, until the workflow is flattened for running. If we want to detect this case, it is necessary to check the inputs to the workflows at every level of the hierarchy. 
As such, I took pulled out the check to an inner function, and call it at every level of the hierarchy. I think this change does not have a big impact, but I think it may be useful to detect these cases, as processing pipelines are constantly getting more complex and deeply nested. I also don't know if the original author of the `_has_attr` and `_get_inputs` functions thought of this possibility and tests for it in another place.
I also wrote a test case in test_workflows.py called `test_nested_workflow_doubleconnect`. The code may be a better explanation of the possible issue.

effigies:
LGTM. Thanks!

oesteban:
I'm going to bookmark this PR as the paradigm of how things should work out in open source/science. Outstanding work @HippocampusGirl, thanks very much!

satra:
thank you @HippocampusGirl

PR #3192, opened 2020-03-21
ENH: Add reverse-ordered transform lists to ants.Registration outputs

sulantha2006:
… ANTS Registration outputs.

## Summary

Fixes # .

## Acknowledgment


sulantha2006:
Any reason why the doctests are failing here. ?

SulanthaM:
Do you think we can incorporate this into pipy soon? 
This output ```reverse_forward_transforms``` is mentioned in the documentation of NIPYPE antsRegistration but was not implemented to be used. So, if anyone tries to access this output based on the documentation, it fails.

effigies:
Can you link to the documentation? It's difficult to review a pull request without context.

SulanthaM:
Here is the documentation. 
https://github.com/nipy/nipype/blob/c37fcfb2ce46c7918ae73c2af391f64f2e123616/nipype/interfaces/ants/registration.py#L646

effigies:
Thanks. Judging by the discussion in https://github.com/nipy/nipype/pull/2187, it looks like `reverse_transforms` was deemed to be what was needed, but the docs didn't change...

@Gilles86 Can you confirm?

SulanthaM:
reverse_transform is actually a collection of inverse transformations. This corresponds to the reverse_invert_flags as well.
The suggested modification is only a list reverse (reverse the elements in the list) of the forward_transforms list and forward_invert_flags list such that you can feed the output directly to a resampling node without a list reverse function node.

sulantha2006:
Any way we can integrate this sooner to pypi?

AlanKuurstra:
I also would appreciate `reverse_forward_transforms` which is mentioned in the documentation but not implemented. I understand `reverse_forward_transforms` to be different from `reverse_transform`.

satra:
@effigies - seems reasonable from a conceptual standpoint and provides a handy helper to ease inverting lists. i think the docstring still says to use composite transforms when possible.

PR #3195, opened 2020-03-24
TEST: Clean up tests

effigies:
Discovered while working through #3154. 

## Acknowledgment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
+os: linux
 dist: xenial
-sudo: true
 
 language: python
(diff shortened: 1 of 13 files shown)


satra:
@effigies - do you want to update using the new etelemetry >= 0.2.0?

see here for an example. https://github.com/dandi/dandi-cli/blob/master/dandi/__init__.py#L44

effigies:
Took a stab at it. To avoid replication in two places, just swapped out the guts in our old implementation.

PR #3213, opened 2020-05-15
volterra_expansion_order documentation error

servoz:
For fMRI model specification, in SPM, matlabbatch{n}.spm.stats.fmri_spec.volt = 2 corresponds to 'Model Interactions' and matlabbatch{n}.spm.stats.fmri_spec.volt =1 corresponds to 'Do not model Interactions'.
Currently, for the corresponding process in nipype, the Level1Design, we see volterra_expansion_order = traits.Enum(1, 2, field="volt", desc=("Model interactions - yes:1, no:2")).
I guess the correct description should rather be volterra_expansion_order = traits.Enum(1, 2, field="volt", desc=("Model interactions - no:1, yes:2"))

### DIFF
--- a/nipype/interfaces/spm/model.py
+++ b/nipype/interfaces/spm/model.py
@@ -81,7 +81,7 @@ Dictionary names of the basis function to parameters:
         mandatory=True,
     )
     volterra_expansion_order = traits.Enum(
-        1, 2, field="volt", desc=("Model interactions - yes:1, no:2")
+        1, 2, field="volt", desc=("Model interactions - no:1, yes:2")
     )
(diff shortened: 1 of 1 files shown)


effigies:
Seems reasonable. Is this consistent across versions?

servoz:
I just checked for spm8 and it's the same thing. Without checking the previous versions (SPM5, SPM2, etc.) I think it's consistent across SPM versions.

effigies:
Sounds good. As it only changes the docs, it won't break scripts that work now, so I'm :+1: for merge.

Apologies for the delay. Very busy time of year.

servoz:
thanks.

PR #3214, opened 2020-05-16
[FIX] BET incorrect output paths

### LINKED ISSUE #2916: Tiny BET configuration and output specification bugs
Hi all,

Just wanted to point out a couple of tiny bugs:
- [BETInputSpec][1]'s default frac value is 0, whereas [BET][2]'s actual default value is 0.5. That should be easy enough to update.
- BETInputSpec includes a "skull" option which does, in fact, result in another .nii file with a __skull_ suffix, however, it is not included in BETOutputSpec's definition or the __list_output_ method. 
- When setting BET's out_file to some location other than the pwd, everything is created properly, but the interface's BETOutputSpec instance (output_spec) seems to list all of the output files' locations (other than the out_file) as if they were still created in the pwd. It seems to have something to do with [FSLCommand][3]'s _gen_fname method and its call to fname_presuffix with _newpath=cwd_, but I haven't yet really tried to understand what might be the repercussions of changing anything there. 

I hope I fork the repo and propose fixes sometime soon, but it's a pretty busy time right now. Just thought it wouldn't hurt to put it here, for future reference if nothing else. 

Thank you for all your amazing work! 
Zvi

[1]: https://github.com/nipy/nipype/blob/master/nipype/interfaces/fsl/preprocess.py
[2]: https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/BET/UserGuide
[3]: https://github.com/nipy/nipype/blob/27490e67a48428699466c0a4cfa36ce25cf7222d/nipype/interfaces/fsl/base.py

ZviBaratz:
Added `cwd` keyword argument to `_gen_fname()` calls so that the returned output will reference the correct paths. This fixes #2916.

- [V] I acknowledge that this contribution will be available under the Apache 2 license.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ __pycache__/
 .ipynb_checkpoints/
 .ruby-version
 .pytest_cache
+.vscode/
+venv/
--- a/nipype/interfaces/fsl/preprocess.py
+++ b/nipype/interfaces/fsl/preprocess.py
@@ -173,45 +173,50 @@ class BET(FSLCommand):
     def _list_outputs(self):
         outputs = self.output_spec().get()
         outputs["out_file"] = self._gen_outfilename()
+
+        basename = os.path.basename(outputs['out_file'])
+        cwd = os.path.dirname(outputs['out_file'])
+        kwargs = {'basename': basename, 'cwd': cwd}
+
         if (isdefined(self.inputs.mesh) and self.inputs.mesh) or (
             isdefined(self.inputs.surfaces) and self.inputs.surfaces
         ):
             outputs["meshfile"] = self._gen_fname(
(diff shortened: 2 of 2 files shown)


ZviBaratz:
I forgot to mention - _pytest_ fails 1 test unless the `bids` package is manually installed.

satra:
thanks @ZviBaratz - this looks reasonable to me. the failures appear to be unrelated to this PR (linked to nilearn tests). @oesteban - do you want to take this opportunity to fix the other issue about lengths of paths?

effigies:
With no activity regarding the other issue, I'm inclined to merge this. Any objections?

satra:
none from my side.

PR #3222, opened 2020-06-02
ENH: Restore ants.legacy interfaces

### LINKED ISSUE #3197: antslegacy module
I notice that in the newest version the `nipype.interfaces.ants.legacy` module is missing, and I cannot find a class for `antsIntroduction()` anywhere in the package.

I know that using antsIntroduction is probably not the best idea, in fact I need it for an article recommending against it for preclinical imaging, but I need it for the benchmark. Is ti not coming back, and should I thus depend on older nipype for the article source code?

effigies:
## Summary

Fixes #3197.

I restored all three interfaces. I can limit it to `antsIntroduction` if preferred.

## Acknowledgment


satra:
we could keep all three as long as ants distributes them. if they don't i think someone would have to maintain some form of deprecation/version-based interface. now that ants is versioning, this does become a little easier.

Issue #3232, opened 2020-08-05
TypeError: '<' not supported between instances of 'int' and 'str'


TheChymera:
### Summary
I am getting the following error from all my ANTs bias field correction nodes:

```
chymera@ihost ~/docsrc/opfvta/prepare $ nipypecli crash /home/chymera/.scratch/opfvta/crashdump/crash-20200805-014215-chymera-f_biascorrect.a00-9c5faaa4-c317-4a54-abef-658b91035d5d.pklz


/mnt/data/scratch/opfvta/crashdump/crash-20200805-014215-chymera-f_biascorrect.a00-9c5faaa4-c317-4a54-abef-658b91035d5d.pklz
Node: preprocess_work.f_biascorrect
Working directory: /mnt/data/scratch/opfvta/preprocess_work/_ind_type_6/f_biascorrect


Node inputs:

args = <undefined>
bias_image = <undefined>
bspline_fitting_distance = 10.0
bspline_order = 4
convergence_threshold = 1e-11
copy_header = False
dimension = 3
environ = {'NSLOTS': '1'}
histogram_sharpening = <undefined>
input_image = <undefined>
mask_image = <undefined>
n_iterations = [150, 100, 50, 30]
num_threads = 1
output_image = <undefined>
rescale_intensities = False
save_bias = False
shrink_factor = 2
weight_image = <undefined>



Traceback:
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/usr/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
    result = self._run_interface(execute=True)
  File "/usr/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
    return self._run_command(execute)
  File "/usr/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 353, in run
    self._check_version_requirements(self.inputs)
  File "/usr/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 279, in _check_version_requirements
    if min_ver > version:
  File "/usr/lib/python3.7/distutils/version.py", line 64, in __gt__
    c = self._cmp(other)
[...]

PR #3234, opened 2020-08-07
ENH: no more auto-failing on misparsed versions

TheChymera:
Fixes https://github.com/nipy/nipype/issues/3232

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -276,6 +276,15 @@ class BaseInterface(Interface):
             version = LooseVersion(str(self.version))
             for name in names:
                 min_ver = LooseVersion(str(trait_object.traits()[name].min_ver))
+                try:
+                    min_ver > version
+                except TypeError:
+                    iflogger.warning(
+                        'Nipype is having issues parsing the package version '
+                        f'for Trait {name} ({self.__class__.__name__})'
+                        f'You may want to check whether {version} is larger than {min_ver}'
+                        )
+                    continue
                 if min_ver > version:
                     unavailable_traits.append(name)
                     if not isdefined(getattr(trait_object, name)):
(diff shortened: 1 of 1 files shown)


satra:
thanks @TheChymera - can we add a test for this by explicitly setting the version info of dummy interface to raise a type error and not to raise one.

nipype/interfaces/base/core.py
```diff
@@ -276,6 +276,15 @@ def _check_version_requirements(self, trait_object, raise_exception=True):
...
                 min_ver = LooseVersion(str(trait_object.traits()[name].min_ver))
+                try:
+                    min_ver > version
+                except TypeError:
+                    iflogger.warning(
+                        'Nipype is having issues parsing the package version '
+                        'for Trait {} ({})'
+                        'You may want to check whehter {} is larger than {}'
+                        .format(name, self.__class__.__name__, version, min_ver)
+                        )
```
mgxd, line 287: We can use f-strings now that nipype is 3.6+. Also, you'll want to escape here, otherwise you'll run into the same problem on the following line.
```suggestion
                        'Nipype is having issues parsing the package version '
                        f'for Trait {name} ({self.__class__.__name__})'
                        f'You may want to check whether {version} is larger than {min_ver}'
                        )
                    return
```

effigies: Just to add: We should use `continue`, not `return`. And we should do this for `max_ver`, as well, ~15 lines down.

TheChymera:
@effigies @mgxd looking good? Tests seem to fail with some sort of permissions error :-/

TheChymera:
@satra can you help me out here? As for the tests, I'm really unsure how to test this one :-/

satra:
@djarecka - could you please help @TheChymera over here. the test should simply focus on evaluating the min/max version check of a dummy interface. it shouldn't be specific to ants or any tool at all.

effigies:
Proposed a test and small refactor at TheChymera#2.

effigies:
If we want this in 1.5.1, please rebase onto `maint/1.5.x`.

PR #3236, opened 2020-08-12
FIX: ANTs' utilities revision - bug fixes and add more operations to ``ants.ImageMath``

oesteban:
These revisions are necessary for the antsBrainExtraction workflow we have in niworkflows (and hence niflows too):

- Missing `ImageMath` operations, overwriting the behavior of `copy_header` if the operation is `PadImage` (see comment below).
- Disable the `copy_header` of `ResampleImageBySpacing` as the original affine is not valid anymore in the output


oesteban:
For the `PadImage` operation we need to work around the copy of the header - the center of the image should not be copied from the input header.

nipype/interfaces/ants/utils.py
```diff
@@ -99,13 +104,46 @@ class ImageMath(ANTSCommand, CopyHeaderInterface):
     ...     op2='0.005 0.999 256').cmdline
     'ImageMath 3 structural_maths.nii TruncateImageIntensity structural.nii 0.005 0.999 256'
 
+    >>> pad = ImageMath(
+    ...     op1='structural.nii',
+    ...     operation='PadImage',
+    ...     op2='0.005 0.999 256')
+    >>> pad.inputs.copy_header
+    False
+
```
effigies, line 113: Can you add some text explaining why you're showing these?

effigies:
Looks reasonable.

nipype/interfaces/ants/utils.py
oesteban: Added, please let me know if this looks good now.

nipype/interfaces/ants/utils.py
```diff
@@ -104,17 +104,21 @@ class ImageMath(ANTSCommand, CopyHeaderInterface):
     ...     op2='0.005 0.999 256').cmdline
     'ImageMath 3 structural_maths.nii TruncateImageIntensity structural.nii 0.005 0.999 256'
 
+    >>> # Test the special case of PadImage. In this case, copy_header is disabled
+    >>> # because the header cannot be copied from inputs to ouptuts.
+    >>> # When operation is PadImage, the copy_header is set to False automatically:
```
effigies, line 109:
```suggestion
    By default, Nipype copies headers from the first input image (``op1``)
    to the output image.
    For the ``PadImage`` operation, the header cannot be copied from inputs to
    outputs, and so ``copy_header`` option is automatically set to ``False``.

```

nipype/interfaces/ants/utils.py
```diff
@@ -104,17 +104,21 @@ class ImageMath(ANTSCommand, CopyHeaderInterface):
...
     >>> pad = ImageMath(
     ...     op1='structural.nii',
-    ...     operation='PadImage',
-    ...     op2='0.005 0.999 256')
+    ...     operation='PadImage')
     >>> pad.inputs.copy_header
     False
 
+    >>> # While operation remains as PadImage, copy_header will not accept True
```
effigies, line 116:
```suggestion
    While the operation is set to ``PadImage``, setting ``copy_header = True`` will fail.

```

nipype/interfaces/ants/utils.py
```diff
@@ -104,17 +104,21 @@ class ImageMath(ANTSCommand, CopyHeaderInterface):
...
     >>> pad.inputs.copy_header = True
     >>> pad.inputs.copy_header
     False
 
+    >>> # For any other operation, copy_header gets its default behavior back.
```
effigies, line 121:
```suggestion
    For any other operation, ``copy_header`` is able to be set:

```

nipype/interfaces/ants/utils.py
oesteban: If I remember correctly, we need Python comments so that numpydoc/napoleon conversions work well (this is partly because we are using a modified version of napoleon). I'm not 100% sure about this, but I don't think it's worth finding out.

I'll apply your suggestions, but as Python comments, if you are okay with that.

oesteban: Scratch that, looking at antsRegistration it seems we correctly handle it. I just force pushed my last commit.

PR #3237, opened 2020-08-14
FIX: Permit identity transforms in list of transforms given to ants.ApplyTransforms

effigies:
## Summary

antsApplyTransforms currently permits `identity` or a chain of transformations, but it can be useful to insert an `identity` into a chain.

* Changes transforms from `Either(InputMultiObject(File), "identity")` to `InputMultiObject(Either(File, "identity"))`.
* Change `InputMultiPath` to `InputMultiObject` throughout the file, in passing.
* Simplifies the logic for constructing `--transform` arguments.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/ants/resampling.py
+++ b/nipype/interfaces/ants/resampling.py
@@ -502,8 +502,6 @@ class ApplyTransforms(ANTSCommand):
         if opt == "output_image":
             return self._get_output_warped_filename()
         elif opt == "transforms":
-            if val == "identity":
-                return "-t identity"
             return self._get_transform_filenames()
         elif opt == "interpolation":
             if self.inputs.interpolation in [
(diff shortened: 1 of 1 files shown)


nipype/interfaces/ants/resampling.py
```diff
@@ -369,15 +369,14 @@ class ApplyTransformsInputSpec(ANTSCommandInputSpec):
             traits.Float(), traits.Float()  # Gaussian/MultiLabel (sigma, alpha)
         ),
     )
-    transforms = traits.Either(
-        InputMultiPath(File(exists=True)),
-        "identity",
+    transforms = InputMultiObject(
+        traits.Either(File(exists=True), "identity"),
```
oesteban, line 373: In the latest releases of traits they are recommending Union instead of Either. I don't know when Union started to be available or when they started to recommend it though. I'm just writing this down and don't think it requires any particular action in this PR (cc @satra)

effigies:
Okay, tests passing. Thanks for having a look.

PR #3239, opened 2020-08-15
FIX: test_no_et hanging on Python 3.8

effigies:
## Summary

Travis is consistently hanging on Python 3.8, apparently at test_no_et. This adds verbosity, and splits up the test to hopefully make it easier to track down.

Cannot reproduce this locally by running the specific file, but there may be state changes in the test workers...

## Acknowledgment


effigies:
Unclear why this fixed the hanging, but I guess at least we can get through tests.

PR #3241, opened 2020-08-16
FIX: Warn for min/max_ver traits when tool version can't be parsed

effigies:
## Summary

@TheChymera Apologies for taking over #3234, but I would like to get a bug-fix release out quickly.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -277,15 +277,15 @@ class BaseInterface(Interface):
             for name in names:
                 min_ver = LooseVersion(str(trait_object.traits()[name].min_ver))
                 try:
-                    min_ver > version
+                    too_old = min_ver > version
                 except TypeError:
                     iflogger.warning(
-                        'Nipype is having issues parsing the package version '
-                        f'for Trait {name} ({self.__class__.__name__})'
-                        f'You may want to check whether {version} is larger than {min_ver}'
-                        )
+                        f"Nipype cannot validate the package version {version!r} for "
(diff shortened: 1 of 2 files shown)


nipype/interfaces/base/core.py
```diff
@@ -276,7 +276,16 @@ def _check_version_requirements(self, trait_object, raise_exception=True):
...
                 min_ver = LooseVersion(str(trait_object.traits()[name].min_ver))
-                if min_ver > version:
+                try:
+                    too_old = min_ver > version
+                except TypeError:
+                    iflogger.warning(
+                        f"Nipype cannot validate the package version {version!r} for "
+                        f"{self.__class__.__name__}. Trait {name} requires version "
+                        f">={min_ver}. Please verify validity."
+                    )
```
satra, line 286: should this and the next one be a warning or an exception if you can't parse? i feel warnings are one of those things that people, especially users never see in complex workflows. and since the underlying tool could be installed separately, could cause major headaches for debugging.

this almost sounds like we should add a config option that allows a specific interface or all interfaces to work when there is a version mismatch or not. we already have this execution variable `stop_on_unknown_version` which we could use here i think.

effigies:
@satra Updated.

nipype/interfaces/base/core.py
satra: thanks @effigies. LGTM.

TheChymera:
@effigies no worries, in fact, thank you!

PR #3248, opened 2020-09-10
FIX: Handle changes in CLI structure of mrtrix3.DWIBiasCorrect

### LINKED ISSUE #3247: Mrtrix3's dwibiascorrect interface crashes due to "-"
Interface crashes because of "-" before algorithm specification.

```python
RuntimeError: Command:
```
```bash
dwibiascorrect -mask <mask_file> -ants <in_file> <out_file>

Standard output:

Standard error:

Error: argument algorithm: invalid choice: <in_file> (choose from 'ants', 'fsl')
Usage: dwibiascorrect algorithm [ options ] ...
       (Run dwibiascorrect -help for more information)

Return code: 1
```
The expected behavior would be to execute the following command:
```bash
dwibiascorrect ants [options] <in_file> <out_file>
```
To replicate the behavior, simply execute:
```python
import nipype.interface.mrtrix3 as mrt
bc = mrt.DWIBiasCorrect()
bc.inputs.in_file = <in_file>
bc.inputs.out_file = <out_file>
bc.inputs.use_ants = True
bc.run()
```
Running on a 64 bit Linux machine (Ubuntu)

```
{'commit_hash': '72945ef34',
 'networkx_version': '2.4',
 'nibabel_version': '3.0.0',
 'nipype_version': '1.5.0',
 'numpy_version': '1.19.0',
 'pkg_path': '/home/groot/Projects/labbing/venv/lib/python3.7/site-packages/nipype',
 'scipy_version': '1.5.0',
 'sys_executable': '/home/groot/Projects/labbing/venv/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.7.8 (default, Jun 29 2020, 04:26:04) \n[GCC 9.3.0]',
 'traits_version': '6.1.1'}
```

GalBenZvi:
Resolves #3247 .


effigies:
I think this may be a change in CLI...

[This line](https://github.com/MRtrix3/mrtrix3/blob/97946e44b77f14b1cbc857659a5b6f2b8ba2461b/lib/mrtrix3/dwibiascorrect/fsl.py#L53-L55) shows:

```Python
  app.warn('Use of -fsl option in dwibiascorrect script is discouraged due to its strong dependence ' + \
           'on brain masking (specifically its inability to correct voxels outside of this mask).' + \
           'Use of the -ants option is recommended for quantitative DWI analyses.')
```

If this has changed, then we need to adjust our behavior based on the version of MRtrix3 installed.

effigies:
If I'm reading it correctly, https://github.com/MRtrix3/mrtrix3/pull/1449 changed from `-ants` and `-fsl` flags to `ants` and `fsl` subcommands, which means this behavior is new as of MRtrix3 3.0. @lestropie is that correct?

Given that MRtrix3 is Python, I wonder if there's a cleaner way to invoke these through a function API, rather than subprocesses. Though that also may have changed.

Hershkovitz-hub:
@effigies I think you're right, but I wonder whether to create a version-specific inputspec, or simply change the relevant trait in accordance with the user's Mrtrix3's version?
I've also posted a message in Mrtrix3's forum to try and verify the exact version in which the change occurred.

Lestropie:
>  MRtrix3/mrtrix3#1449 changed from `-ants` and `-fsl` flags to ants and fsl subcommands, which means this behavior is new as of MRtrix3 3.0.

Correct; they changed from command-line options to using the *MRtrix3* `algorithm` submodule (essentially subcommands), as this is more consistent with other *MRtrix3* Python scripts. This also means that the algorithm name needs to be the second command-line argument, whereas previously the command-line option was not positional.

> Given that at least some MRtrix3 commands are Python, I wonder if there's a cleaner way to invoke these through a function API, rather than subprocesses. Though that also may have changed.

Given the degree to which the *MRtrix3* Python API is custom and very much nonstandard, I suspect not; at least not without essentially duplicating the functionality of the `dwibiascorrect` wrapper script. The majority of *MRtrix3* is C++ terminal commands, so when Python scripts started to get added they were designed around the same philosophy.

> I've also posted a message in Mrtrix3's forum to try and verify the exact version in which the change occurred.

The change is from `3.0_RC3` to `3.0.0`. There's a couple of other tags (`3.0_RC3_latest`, `3.0_RC3_last`, `3.0_RC4`) that were not actually intended for public use. The relevant change is present in `3.0_RC4`; but I'd advise avoiding that tag anyway, it was somewhat erroneous to include that tag at all on the main repo.

Lestropie:
Also MRtrix3/mrtrix3#2171 may have contributed to the confusion; sorry about that.

nipype/interfaces/mrtrix3/preprocess.py
```diff
@@ -181,13 +181,13 @@ class DWIBiasCorrectInputSpec(MRTrix3BaseInputSpec):
     )
     in_mask = File(argstr="-mask %s", desc="input mask image for bias field estimation")
     use_ants = traits.Bool(
-        argstr="-ants",
+        argstr="ants",
         mandatory=True,
         desc="use ANTS N4 to estimate the inhomogeneity field",
```
effigies, line 186:
```suggestion
        desc="use ANTS N4 to estimate the inhomogeneity field",
        position=0,
```

nipype/interfaces/mrtrix3/preprocess.py
```diff
@@ -181,13 +181,13 @@ class DWIBiasCorrectInputSpec(MRTrix3BaseInputSpec):
...
         xor=["use_fsl"],
     )
     use_fsl = traits.Bool(
-        argstr="-fsl",
+        argstr="fsl",
         mandatory=True,
         desc="use FSL FAST to estimate the inhomogeneity field",
```
effigies, line 192:
```suggestion
        desc="use FSL FAST to estimate the inhomogeneity field",
        position=0,
```

.zenodo.json
```diff
@@ -697,6 +697,10 @@
       "affiliation": "Sagol School of Neuroscience, Tel Aviv University",
       "name": "Baratz, Zvi"
     },
+    {
+      "affiliation": "Sagol School of Neuroscience, Tel Aviv University",
+      "name": "Ben-Zvi,Gal"
```
effigies, line 702:
```suggestion
      "name": "Ben-Zvi, Gal"
```

effigies (commented):
Okay, then I think we want to add:

```Python
class DWIBiasCorrect(MRTrix3Base):
    ...
        if name in ('use_ants', 'use_fsl'):
            ver = self.version
            # Changed in version 3.0, after release candidates
            if ver is not None and (ver[0] < '3' or ver.startswith("3.0_RC")):
                return f"-{trait_spec.argstr}"
        super()._format_arg(name, trait_spec, value)
```

And we need to update the doctest to say `dwibiascorrect ants dwi.mif dwi_biascorr.mif`.

effigies:
@GalBenZvi Do you have time to finish this one off? I think we're probably due for a release soon, and it would be nice to have this in.

GalBenZvi:
@effigies I'm sorry, I was under the impression that it was finished, could you maybe explain to me what do you want me to do? Obviously I would love to contribute.

effigies:
Sure, my previous review had a number of suggestions. If that doesn't seem actionable, I can send a PR against your branch with my suggested changes.

GalBenZvi:
Sure, I'll do my best to get to it as soon as possible. Sorry for the delay.

nipype/interfaces/mrtrix3/preprocess.py
```diff
@@ -224,14 +226,20 @@ class DWIBiasCorrect(MRTrix3Base):
...
     input_spec = DWIBiasCorrectInputSpec
     output_spec = DWIBiasCorrectOutputSpec
-
+    def _format_arg(self, name, trait_spec, value):
+        if name in ("use_ants", "use_fsl"):
+            ver = self.version
+            # Changed in version 3.0, after release candidates
+            if ver is not None and (ver[0] < "3" or ver.startswith("3.0_RC")):
+                return f"-{trait_spec.argstr}"
+        super()._format_arg(name, trait_spec, value)
```
effigies, line 242:
```suggestion
        return super()._format_arg(name, trait_spec, value)
```

effigies:
Okay, this looks all set. Thanks for this addition!

Issue #3254, opened 2020-09-21
Currently it is not possible to launch SPM SliceTiming if the input parameters slice_order and ref_slice are set in slice times (ms).


servoz:
In fact it is only possible to use SPM SliceTiming if slice_order and ref_slice are set to slice indices.

Indeed, the SliceTimingInputSpec class of the module nipype.interfaces.spm.preprocess defines ref_slice = traits.Int(.....).
So if we set ref_slice in (ms) an execption is raised because refslice is a float and not an integer:
`traits.trait_errors.TraitError: The 'ref_slice' trait of a SliceTimingInputSpec instance must be an integer, but a value of 1471.6990943396227 <class 'float'> was specified.`

 I propose a PR with a small correction for ref_slice (and also slice_order) so that both cases (index, or times (ms)) will work properly.

servoz:
PR #3255 is now merged, this issue is fixed.

PR #3255, opened 2020-09-21
SPM SliceTiming must accept either Int or float for ref_slice and sli…

servoz:
#3254

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -276,12 +276,14 @@ class SliceTimingInputSpec(SPMCommandInputSpec):
         mandatory=True,
     )
     slice_order = traits.List(
-        traits.Float(),
+        traits.Either(traits.Int(),traits.Float()),
         field="so",
         desc=("1-based order or onset (in ms) in which slices are acquired"),
         mandatory=True,
     )
-    ref_slice = traits.Int(
+    ref_slice = traits.Either(
+        traits.Int(),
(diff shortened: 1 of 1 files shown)


effigies:
This doesn't make much sense to me on its face, but if SPM accepts these values, that seems fine. Do you have a reference? I don't see indications of this behavior in https://en.wikibooks.org/wiki/SPM/Slice_Timing.

servoz:
[In the spm12 manual, it is stated that slice_order and ref_slice can be provided in slice time instead of slice indices](https://www.fil.ion.ucl.ac.uk/spm/doc/manual.pdf#page=21). In this case TA is set to zero. It is true that we (I) generally use slice indices, but in some cases (multi-band accelerated EPI data) slice times are appropriate.

servoz:
If the spm documentation is not enough, there is another way to convince yourself:
In the Slice Timing module of spm (Batch Editor > SPM > Temporal> Slice Timing) use for example :
`nslices` = 10 ;
`tr`= 2.95 ;
`ta`= 0 ;
`so`= [0:1:9]*2.95/10 ;
`refslice` = 11.8 ;
all these parameters are accepted by SPM. This is a direct proof that `refslice` can be a float and `so` can be a list of floats.

effigies:
Ah, no sorry, I've just been very busy. I'll look again today.

effigies:
Yup, this looks good. Looks like the trait documentation actually already included these options, but the types didn't permit. Thanks for this!

PR #3256, opened 2020-09-23
FIX: Raise version error when using ``-g`` with ``antsAI`` < 2.3.0

oesteban:
## Summary

Fixes # .

## Acknowledgment


effigies:
make specs?

effigies:
May need to upgrade `black`. I think they changed some of their rendering rules, and it looks like you're reverting some changes in 8c4a8b09669e076f128318f40006a3de1bb46611.

effigies (approved):

PR #3260, opened 2020-10-18
[ENH] Optimize workflow.run performance

HippocampusGirl:
## Summary
A few months back, I submitted pull request #3184 to improve the performance of using `connect` when creating large workflows. Specifically, I had discovered that the use of the `inputs` or `outputs` properties of workflows can create a performance bottleneck if there are many child nodes or nested workflows. 

Recently, I noticed that the same bottleneck can cause a delay between calling `workflow.run()` and the start of the actual execution, meaning when nodes and interfaces start to run. 

Running cProfile suggests that the delay occurs in `_create_flat_graph`. Note that the profile does not include the full workflow execution, but was cancelled immediately when the first node started to run.
Screen Shot 2020-08-07 at 10 47 49 AM

As far as I can tell, before execution starts, nested workflows are merged into one overall workflow using `_create_flat_graph`. To resolve the final connections between nodes in this merged workflow, `_create_flat_graph` calls `_get_parameter_node` for each input from or output to a nested workflow, and then modifies the connection information accordingly.
https://github.com/nipy/nipype/blob/e9217c299149be829a985d8f822c3a91e29b2f4b/nipype/pipeline/engine/workflows.py#L975-L979

As a result, for each connection to/from a nested workflow, `_get_parameter_node` constructs the entire `inputs` or `outputs` data structure of the nested workflow, and then uses it to resolve the correct connection information. Just as for #3184, constructing this entire data structure over and over again for each connection can reduce performance.

Instead of generating the full `inputs` or `outputs` data structure, I propose that the `_get_parameter_node` function should traverse the individual workflow graphs until it finds the target node (or not). 

I have created a quick implementation that leads to a significant speedup. This implementation is a slightly modified copy of the code from #3184. 
Screen Shot 2020-08-07 at 11 47 01 AM

I hope that this code will be useful for the nipype community.

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -838,15 +838,28 @@ connected.
         """Returns the underlying node corresponding to an input or
         output parameter
         """
-        if subtype == "in":
-            subobject = self.inputs
-        else:
-            subobject = self.outputs
         attrlist = parameter.split(".")
-        cur_out = subobject
-        for attr in attrlist[:-1]:
-            cur_out = getattr(cur_out, attr)
-        return cur_out.traits()[attrlist[-1]].node
+        _ = attrlist.pop()  # take attribute name
+        nodename = attrlist.pop()
+
+        targetworkflow = self
+        while attrlist:
+            workflowname = attrlist.pop(0)
+            workflow = None
+            for node in targetworkflow._graph.nodes():
+                if node.name == workflowname and isinstance(node, Workflow):
+                    workflow = node
+                    break
+            if workflow is None:
+                return
+            targetworkflow = workflow
+
+        for node in targetworkflow._graph.nodes():
+            if node.name == nodename:
+                if isinstance(node, Workflow):
+                    return
+                else:
+                    return node
 
     def _check_outputs(self, parameter):
         return self._has_attr(parameter, subtype="out")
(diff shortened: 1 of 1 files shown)


effigies:
Thanks for this, that looks like a great improvement!

I've read your solution and my only suggestions would be aesthetic ones, but I realized that we probably already have a function to fetch a particular node, and found `get_node()`:

https://github.com/nipy/nipype/blob/e9217c299149be829a985d8f822c3a91e29b2f4b/nipype/pipeline/engine/workflows.py#L369-L383

I think we might be able to replace all calls to `wf._get_parameter_node(parameter)` with `wf.get_node(parameter.rsplit(".", 1)[0])`, and probably not lose much efficiency over your solution. Would you mind giving that a shot?

HippocampusGirl:
That's a really good idea :-)

effigies:
Test failures appear to be #3261.

How does the profiling look? Do we need to clean up `get_node()` at all? The `outnode = [...]` comprehension seems like it could be wasteful on large graphs, but if there's no discernible difference with your solution, then I think we should just get this in.

effigies (approved):
Using fMRIPrep with the latest commit, this gets a ~100x speedup for `_generate_flatgraph()`.

LGTM, though let me know if there's anything else you want to include before merge.

satra:
thank you @HippocampusGirl - much appreciated.

HippocampusGirl:
Thank you for benchmarking that @effigies! No, I don't have anything to add

Issue #3261, opened 2020-10-20
Travis builds broken


effigies:
Working build on October 18 ~4pm EDT: https://travis-ci.com/github/nipy/nipype/builds/190839373

Failed build on October 19 ~8pm EDT: https://travis-ci.com/github/nipy/nipype/builds/191039325

The test that's timing out might be `nipype/interfaces/fsl/tests/test_maths.py::test_smooth[NIFTI_GZ]`, which is a bit odd, since I don't think FSL has been updated. So it's probably some Python dependency that was released in the last 48 hours or so.

effigies:
Restarted and am just watching one of these. It is moving, but at an absolute snail's pace. I don't know if maybe Travis dropped the resources available to each job?

effigies:
Reported on the Travis discourse: https://travis-ci.community/t/builds-timing-out-in-last-several-days/10234

Not really sure what to do here.

effigies:
Seems resolved. IDK.

PR #3262, opened 2020-10-21
FIX: CI issues

effigies:
Trying to figure out what's going on with #3261. Possibly xenial is being relegated to slow machines?

Also found a small bug while looking at the Travis outputs.


effigies:
I guess this worked? Merging to see if master starts acting right.

PR #3272, opened 2020-11-21
FIX: Purge deprecated exception content accesses

### LINKED ISSUE #3270: Consider alternative to using Exception.message to get message text
### Summary
Apparently, some `Exception` classes may not have a `.message` attribute. This can be more problematic when catching base `Exception` and then trying to log/warn using a string incorporating the original Exception's `.message` attribute. Maybe using `str(ex)` is more compatible, or more specific exception classes?

Example (from `master`):
https://github.com/nipy/nipype/blob/b356ad16828afd4c63f862dfc72fa7e841863391/nipype/utils/filemanip.py#L866-L877

will cause a throw exception if an `OSError` is caught, since OSError does not have a `.message` attribute (at least in python 3.8.3)
 
### Actual behavior
```
Python 3.8.3 (default, May 19 2020, 18:47:26) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
>>> a = OSError('darn')
>>> a.message
  File "<stdin>", line 1, in <module>
AttributeError: 'OSError' object has no attribute 'message'
```
### Expected behavior
```
>>> str(a)
'darn'
```

effigies:
## Summary
Apparently this was deprecated in Python 2.6 and we missed it. `BaseException.__str__` is the generic way to access messages.

Fixes #3270.

* Remove uses of `BaseException.message; Update to `BaseException.__str__`
* Remove uses of `OSError.__getitem__`; Update to `OSError.errno`

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/tests/test_specs.py
+++ b/nipype/interfaces/base/tests/test_specs.py
@@ -183,7 +183,7 @@ def test_deprecation():
         except nib.TraitError:
             not_raised = False
         assert not_raised
-        assert len(w) == 1, "deprecated warning 1 %s" % [w1.message for w1 in w]
+        assert len(w) == 1, f"deprecated warning 1 {[str(w1) for w1 in w]}"
 
     with warnings.catch_warnings(record=True) as w:
         warnings.filterwarnings("always", "", UserWarning)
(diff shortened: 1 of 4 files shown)


effigies:
Ready for review.

PR #3284, opened 2020-12-20
MNT: Drop support for numpy < 1.15.3

effigies:
## Summary

We forgot to drop support for numpy < 1.14 for the last minor release, but the next [NEP29+1y](https://github.com/nipy/nibabel/issues/803) milestone is Jan 7 2021, and I doubt we'll have a minor release before then. This drops support for <1.15.3, which means we can drop our split between 3.6 and 3.7+ dependencies.

## Acknowledgment

### DIFF
--- a/nipype/info.py
+++ b/nipype/info.py
@@ -100,10 +100,9 @@ existing pipeline systems.
 # versions
 NIBABEL_MIN_VERSION = "2.1.0"
 NETWORKX_MIN_VERSION = "2.0"
-NUMPY_MIN_VERSION = "1.13"
 # Numpy bug in python 3.7:
 # https://www.opensourceanswers.com/blog/you-shouldnt-use-python-37-for-data-science-right-now.html
-NUMPY_MIN_VERSION_37 = "1.15.3"
+NUMPY_MIN_VERSION = "1.15.3"
 SCIPY_MIN_VERSION = "0.14"
 TRAITS_MIN_VERSION = "4.6"
 DATEUTIL_MIN_VERSION = "2.2"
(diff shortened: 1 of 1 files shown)


oesteban (approved):

nipype/info.py
```diff
@@ -100,10 +100,9 @@ def get_nipype_gitversion():
 # versions
 NIBABEL_MIN_VERSION = "2.1.0"
 NETWORKX_MIN_VERSION = "2.0"
-NUMPY_MIN_VERSION = "1.13"
 # Numpy bug in python 3.7:
 # https://www.opensourceanswers.com/blog/you-shouldnt-use-python-37-for-data-science-right-now.html
-NUMPY_MIN_VERSION_37 = "1.15.3"
+NUMPY_MIN_VERSION = "1.15.3"
```
oesteban, line 105:
```suggestion
NUMPY_MAX_VERSION_36 = "1.20"
NUMPY_MIN_VERSION = "1.15.3"
```

nipype/info.py
```diff
@@ -138,8 +137,7 @@ def get_nipype_gitversion():
     "click>=%s" % CLICK_MIN_VERSION,
     "networkx>=%s" % NETWORKX_MIN_VERSION,
     "nibabel>=%s" % NIBABEL_MIN_VERSION,
-    'numpy>=%s ; python_version < "3.7"' % NUMPY_MIN_VERSION,
-    'numpy>=%s ; python_version >= "3.7"' % NUMPY_MIN_VERSION_37,
+    'numpy>=%s' % NUMPY_MIN_VERSION,
```
oesteban, line 140:
```suggestion
    f'numpy>={NUMPY_MIN_VERSION},<{NUMPY_MAX_VERSION_36} ; python_version < "3.7"',
    f'numpy>={NUMPY_MIN_VERSION} ; python_version >= "3.7"',
```

oesteban (commented):
I think we need to keep special pinnings for PY36, although if we are to follow Numpy's schedule we can just drop PY36

effigies:
While it's not *hard* to support the edge case of people using Python 3.6 and setup.py and not already having numpy installed, it might be better to let deprecated use cases fail in order to prod people to update their practices. New users should not experience this, since the docs have instructed people to use `pip install` instead of `easy_install` or `setup.py` since 0.13.0 (#1977).

PR #3287, opened 2021-01-05
CI: Build docker images with Python 3.8

effigies:
Circle tests are failing because h5py no longer builds wheels for Python 3.6, and our docker images are not equipped to build the wheels. Python 3.6 is on its way out, so upgrading to Python 3.8.

Or should I target 3.9?


effigies:
In the absence of response, I'm just going to do it.

PR #3290, opened 2021-01-08
FIX: Restore generate_gantt_chart functionality

### LINKED ISSUE #2982: generate_gantt_chart fails on logfile
### Summary
generate_gantt_chart fails on generating charts based on profiler log file
### Actual behavior

Exception occurs:

 > Traceback (most recent call last):
  File "./profiler_chart.py", line 2, in <module>
    generate_gantt_chart('/mnt/connectomics/profiler.log', cores=4)
  File "/home/siegfriedwagner/.local/lib/python3.6/site-packages/nipype-1.2.0-py3.6.egg/nipype/utils/draw_gantt_chart.py", line 507, in generate_gantt_chart
    duration = (last_node['finish'] - start_node['start']).total_seconds()
TypeError: unsupported operand type(s) for -: 'str' and 'str'

### Expected behavior

Graph generates flawlessly 

Run script presented bellow on log from https://gist.github.com/SiegfriedWagner/891bb05bff3cb8224f23915634672a83
### Script/Workflow details

> from nipype.utils.draw_gantt_chart import generate_gantt_chart
> generate_gantt_chart('profiler.log', cores=4)

### Platform details:

```
{'commit_hash': 'a0f756d1c',
 'networkx_version': '2.3',
 'nibabel_version': '2.4.1',
 'nipype_version': '1.2.0',
 'numpy_version': '1.17.0',
 'pkg_path': '/home/siegfriedwagner/.local/lib/python3.6/site-packages/nipype-1.2.0-py3.6.egg/nipype',
 'scipy_version': '1.3.0',
 'sys_executable': '/usr/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.6.8 (default, Jan 14 2019, 11:02:34) \n'
                '[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]',
 'traits_version': '5.1.2'}

```

### Execution environment

Choose one

shnizzedy:
## Summary

Fixes #2982.

All tests pass locally. <sup>8</sup>/<sub>13</sub> jobs pass [on Travis](https://travis-ci.com/github/shnizzedy/nipype/builds/212518190). The Travis failures seem unrelated to the changes in this PR.

- [x] Updates https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/nipype/utils/draw_gantt_chart.py#L391 to handle changes to profiler
   - [x] filter log file to [include only logged nodes with timing information](https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/nipype/utils/draw_gantt_chart.py#L528)
   - [x] [convert datetime strings to datetime objects](https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/nipype/utils/draw_gantt_chart.py#L522) before doing datetime math
   - [x] for nodes with timing information but no `name`, [use `id` or an empty string for `name`](https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/nipype/utils/draw_gantt_chart.py#L297) instead of crashing
   - [x] [warn if a node is suspected of being included twice instead of raising an exception](https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/nipype/utils/draw_gantt_chart.py#L74-L75)
   - [x] skip nodes with non-timestamp `start` and `finish` values like `"N/A"` and `"Unknown"`
       https://github.com/nipy/nipype/blob/a08ee57aa54868a6ce85f0799dfe4fd7cc00df3e/nipype/utils/draw_gantt_chart.py#L144-L147
       https://github.com/nipy/nipype/blob/a08ee57aa54868a6ce85f0799dfe4fd7cc00df3e/nipype/utils/draw_gantt_chart.py#L151-L154
- [x] Adds a test https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/nipype/pipeline/plugins/tests/test_callback.py#L66-L98 to make sure the Gantt chart HTML page generates without error
- [x] [Adds myself as a contributor to .zenodo.json](https://github.com/nipy/nipype/blob/fa65cafa83e937538827d8ba6ac6f32296711b82/.zenodo.json#L791-L795)

## Acknowledgment

### DIFF
--- a/nipype/pipeline/plugins/tests/test_callback.py
+++ b/nipype/pipeline/plugins/tests/test_callback.py
@@ -61,3 +61,36 @@ def test_callback_exception(tmpdir, plugin, stop_on_first_crash):
 
     sleep(0.5)  # Wait for callback to be called (python 2.7)
     assert so.statuses == [("f_node", "start"), ("f_node", "exception")]
+
+
+@pytest.mark.parametrize("plugin", ["Linear", "MultiProc", "LegacyMultiProc"])
+def test_callback_gantt(tmpdir, plugin):
+    import logging
+    import logging.handlers
+
+    from os import path
+
+    from nipype.utils.profiler import log_nodes_cb
+    from nipype.utils.draw_gantt_chart import generate_gantt_chart
+
+    log_filename = 'callback.log'
+    logger = logging.getLogger('callback')
+    logger.setLevel(logging.DEBUG)
+    handler = logging.FileHandler(log_filename)
+    logger.addHandler(handler)
+
+    #create workflow
+    wf = pe.Workflow(name="test", base_dir=tmpdir.strpath)
+    f_node = pe.Node(
+        niu.Function(function=func, input_names=[], output_names=[]), name="f_node"
+    )
+    wf.add_nodes([f_node])
+    wf.config["execution"] = {"crashdump_dir": wf.base_dir, "poll_sleep_duration": 2}
+
+    plugin_args = {"status_callback": log_nodes_cb}
+    if plugin != "Linear":
+        plugin_args['n_procs'] = 8
+    wf.run(plugin=plugin, plugin_args=plugin_args)
+
+    generate_gantt_chart('callback.log', 1 if plugin == "Linear" else 8)
+    assert path.exists('callback.log.html')
--- a/nipype/pipeline/plugins/tests/test_callback.py
+++ b/nipype/pipeline/plugins/tests/test_callback.py
@@ -73,13 +73,13 @@ def test_callback_gantt(tmpdir, plugin):
     from nipype.utils.profiler import log_nodes_cb
     from nipype.utils.draw_gantt_chart import generate_gantt_chart
 
-    log_filename = 'callback.log'
-    logger = logging.getLogger('callback')
+    log_filename = "callback.log"
+    logger = logging.getLogger("callback")
     logger.setLevel(logging.DEBUG)
     handler = logging.FileHandler(log_filename)
     logger.addHandler(handler)
 
-    #create workflow
+    # create workflow
     wf = pe.Workflow(name="test", base_dir=tmpdir.strpath)
     f_node = pe.Node(
         niu.Function(function=func, input_names=[], output_names=[]), name="f_node"
@@ -89,8 +89,8 @@ def test_callback_gantt(tmpdir, plugin):
 
     plugin_args = {"status_callback": log_nodes_cb}
     if plugin != "Linear":
-        plugin_args['n_procs'] = 8
+        plugin_args["n_procs"] = 8
     wf.run(plugin=plugin, plugin_args=plugin_args)
 
-    generate_gantt_chart('callback.log', 1 if plugin == "Linear" else 8)
-    assert path.exists('callback.log.html')
+    generate_gantt_chart("callback.log", 1 if plugin == "Linear" else 8)
+    assert path.exists("callback.log.html")
--- a/nipype/pipeline/plugins/tests/test_callback.py
+++ b/nipype/pipeline/plugins/tests/test_callback.py
@@ -73,7 +73,7 @@ def test_callback_gantt(tmpdir, plugin):
     from nipype.utils.profiler import log_nodes_cb
     from nipype.utils.draw_gantt_chart import generate_gantt_chart
 
-    log_filename = "callback.log"
+    log_filename = path.join(tmpdir, "callback.log")
     logger = logging.getLogger("callback")
     logger.setLevel(logging.DEBUG)
     handler = logging.FileHandler(log_filename)
@@ -92,5 +92,7 @@ def test_callback_gantt(tmpdir, plugin):
         plugin_args["n_procs"] = 8
     wf.run(plugin=plugin, plugin_args=plugin_args)
 
-    generate_gantt_chart("callback.log", 1 if plugin == "Linear" else 8)
-    assert path.exists("callback.log.html")
+    generate_gantt_chart(
+        path.join(tmpdir, "callback.log"), 1 if plugin == "Linear" else 8
+    )
+    assert path.exists(path.join(tmpdir, "callback.log.html"))
--- a/nipype/utils/draw_gantt_chart.py
+++ b/nipype/utils/draw_gantt_chart.py
@@ -292,7 +292,7 @@ def draw_nodes(start, nodes_list, cores, minute_scale, space_between_minutes, co
             "offset": offset,
             "scale_duration": scale_duration,
             "color": color,
-            "node_name": node["name"],
+            "node_name": node.get("name", node.get("id", "")),
             "node_dur": node["duration"] / 60.0,
             "node_start": node_start.strftime("%Y-%m-%d %H:%M:%S"),
             "node_finish": node_finish.strftime("%Y-%m-%d %H:%M:%S"),
@@ -512,6 +512,20 @@ def generate_gantt_chart(
 
     # Read in json-log to get list of node dicts
     nodes_list = log_to_dict(logfile)
+
+    # Only include nodes with timing information, and covert timestamps
+    # from strings to datetimes
+    nodes_list = [{
+        k: datetime.datetime.strptime(
+            i[k], "%Y-%m-%dT%H:%M:%S.%f"
+        ) if k in {"start", "finish"} else i[k] for k in i
+    } for i in nodes_list if "start" in i and "finish" in i]
+
+    for node in nodes_list:
+        if "duration" not in node:
+            node["duration"] = (
+                node["finish"] - node["start"]
+            ).total_seconds()
 
     # Create the header of the report with useful information
     start_node = nodes_list[0]
--- a/nipype/utils/draw_gantt_chart.py
+++ b/nipype/utils/draw_gantt_chart.py
@@ -138,13 +138,14 @@ def calculate_resource_timeseries(events, resource):
     all_res = 0.0
 
     # Iterate through the events
+    nan = {"Unknown", "N/A"}
     for _, event in sorted(events.items()):
         if event["event"] == "start":
-            if resource in event and event[resource] != "Unknown":
+            if resource in event and event[resource] not in nan:
                 all_res += float(event[resource])
             current_time = event["start"]
         elif event["event"] == "finish":
-            if resource in event and event[resource] != "Unknown":
+            if resource in event and event[resource] not in nan:
                 all_res -= float(event[resource])
             current_time = event["finish"]
         res[current_time] = all_res
--- a/nipype/utils/draw_gantt_chart.py
+++ b/nipype/utils/draw_gantt_chart.py
@@ -138,15 +138,20 @@ def calculate_resource_timeseries(events, resource):
     all_res = 0.0
 
     # Iterate through the events
-    nan = {"Unknown", "N/A"}
     for _, event in sorted(events.items()):
         if event["event"] == "start":
-            if resource in event and event[resource] not in nan:
-                all_res += float(event[resource])
+            if resource in event:
+                try:
+                    all_res += float(event[resource])
+                except ValueError:
+                    next
             current_time = event["start"]
         elif event["event"] == "finish":
-            if resource in event and event[resource] not in nan:
-                all_res -= float(event[resource])
+            if resource in event:
+                try:
+                    all_res -= float(event[resource])
+                except ValueError:
+                    next
             current_time = event["finish"]
         res[current_time] = all_res
 
--- a/nipype/utils/draw_gantt_chart.py
+++ b/nipype/utils/draw_gantt_chart.py
@@ -12,6 +12,7 @@ import datetime
 import simplejson as json
 
 from collections import OrderedDict
+from warnings import warn
 
 # Pandas
 try:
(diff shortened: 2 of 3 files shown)


shnizzedy:
As noted 

> [T]here is an issue with the number of threads being estimated by the callback, or the gantt chart creation script is pulling in the wrong numbers. Some of the nodes are reporting using 210 threads! 

_Originally posted by @ccraddock in https://github.com/FCP-INDI/C-PAC/issues/1404#issuecomment-756389510_

> I thought maybe `runtime_threads` was counting something different than I expected.

I see the profile uses [`cpu_percent`](https://github.com/giampaolo/psutil/blob/ede5e46dbee0c0f6c33f80629f75aecc1ad4143c/psutil/__init__.py#L942-L963) for `runtime_threads` which returns a percentage of a CPU, so I think something like `math.ceil(cpu_percent)/100` would be an estimate of the number of threads, but there's [some disconnected code](https://github.com/nipy/nipype/blob/e9217c299149be829a985d8f822c3a91e29b2f4b/nipype/utils/profiler.py#L184-L282) that looks like it collects the actual number of threads used (as opposed to percentage of 1 CPU).

_Originally posted by @shnizzedy in https://github.com/FCP-INDI/C-PAC/issues/1404#issuecomment-756407584_

> I think estimating the number of threads (by dividing by `cpu_percent` 100 and rounding up) is good enough for what I'm trying to do.

[callback.log.html screenshot](https://github.com/FCP-INDI/C-PAC/files/5788868/callback.log.html.zip)

_Originally posted by @shnizzedy in https://github.com/FCP-INDI/C-PAC/issues/1404#issuecomment-756934321_

I think the issues of 
1. what `runtime_threads` is logging and
2. whether the number of threads used by a node is recorded

are related to this PR and issue, but beyond the scope of these changes. C-PAC has [its own callback function](https://github.com/FCP-INDI/C-PAC/blob/c55377fd369bfa7e9c5df2cde27ae1c806285ea3/CPAC/utils/monitoring.py#L33) in which I'm [dividing and rounding](https://github.com/FCP-INDI/C-PAC/blob/c55377fd369bfa7e9c5df2cde27ae1c806285ea3/CPAC/utils/monitoring.py#L62-L65), so I made no changes regarding `runtime_threads` in Nipype.

shnizzedy:
My only hesitance is the potentially misleading `runtime_threads` ― maybe that should be fixed before restoring this functionality?

nipype/utils/draw_gantt_chart.py
```diff
@@ -139,12 +140,18 @@ def calculate_resource_timeseries(events, resource):
     # Iterate through the events
     for _, event in sorted(events.items()):
         if event["event"] == "start":
-            if resource in event and event[resource] != "Unknown":
-                all_res += float(event[resource])
+            if resource in event:
+                try:
+                    all_res += float(event[resource])
+                except ValueError:
+                    next
```
mgxd, line 147:
```suggestion
                    pass
```

nipype/utils/draw_gantt_chart.py
```diff
@@ -139,12 +140,18 @@ def calculate_resource_timeseries(events, resource):
...
             current_time = event["start"]
         elif event["event"] == "finish":
-            if resource in event and event[resource] != "Unknown":
-                all_res -= float(event[resource])
+            if resource in event:
+                try:
+                    all_res -= float(event[resource])
+                except ValueError:
+                    next
```
mgxd, line 154:
```suggestion
                    pass
```

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -61,3 +61,38 @@ def test_callback_exception(tmpdir, plugin, stop_on_first_crash):
 
     sleep(0.5)  # Wait for callback to be called (python 2.7)
     assert so.statuses == [("f_node", "start"), ("f_node", "exception")]
+
+
+@pytest.mark.parametrize("plugin", ["Linear", "MultiProc", "LegacyMultiProc"])
+def test_callback_gantt(tmpdir, plugin):
+    import logging
+    import logging.handlers
```
mgxd, line 69: not needed?
```suggestion
```

nipype/utils/draw_gantt_chart.py
```diff
@@ -69,9 +70,9 @@ def create_event_dict(start_time, nodes_list):
         finish_delta = (node["finish"] - start_time).total_seconds()
 
         # Populate dictionary
-        if events.get(start_delta) or events.get(finish_delta):
+        if events.get(start_delta):
             err_msg = "Event logged twice or events started at exact same time!"
-            raise KeyError(err_msg)
+            warn(str(KeyError(err_msg)), category=Warning)
```
mgxd, line 75:
```suggestion
            warn(err_msg, category=Warning)
```

mgxd (commented):
Looks good, just some minor nits.

> My only hesitance is the potentially misleading runtime_threads ― maybe that should be fixed before restoring this functionality?

I agree 👍

nipype/utils/draw_gantt_chart.py
shnizzedy: whoops, good catch! I actually meant https://github.com/nipy/nipype/blob/25dd1fc9d0e29fe382f7aad35bb91598fd01f0cc/nipype/utils/draw_gantt_chart.py#L147

https://github.com/nipy/nipype/pull/3290/commits/25dd1fc9d0e29fe382f7aad35bb91598fd01f0cc

nipype/utils/draw_gantt_chart.py
shnizzedy: whoops, good catch! I actually meant https://github.com/nipy/nipype/blob/25dd1fc9d0e29fe382f7aad35bb91598fd01f0cc/nipype/utils/draw_gantt_chart.py#L154

https://github.com/nipy/nipype/pull/3290/commits/25dd1fc9d0e29fe382f7aad35bb91598fd01f0cc

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -61,3 +61,37 @@ def test_callback_exception(tmpdir, plugin, stop_on_first_crash):
 
     sleep(0.5)  # Wait for callback to be called (python 2.7)
     assert so.statuses == [("f_node", "start"), ("f_node", "exception")]
+
+
+@pytest.mark.parametrize("plugin", ["Linear", "MultiProc", "LegacyMultiProc"])
```
effigies, line 66: This test fails if `pandas` is not installed. Since it's not a hard requirement, we need something like:

```suggestion
@pytest.mark.parametrize("plugin", ["Linear", "MultiProc", "LegacyMultiProc"])
@pytest.mark.skipif(not has_pandas, "Test requires pandas")
```

And at the top:

```Python
try:
    import pandas
    has_pandas = True
except ImportError:
     has_pandas = False
```

effigies:
> > My only hesitance is the potentially misleading runtime_threads ― maybe that should be fixed before restoring this functionality?
> 
> I agree

Was this fixed? What needs doing?

shnizzedy:
> Was this fixed? What needs doing?

I haven't fixed it (yet at least). The issue is that the chart uses `runtime_threads` from the callback log as a count of threads observed being used at runtime, but the value actually stored there is `cpu_percent`,

https://github.com/nipy/nipype/blob/e9217c299149be829a985d8f822c3a91e29b2f4b/nipype/utils/profiler.py#L143

> [a float representing the current process CPU utilization as a percentage](https://github.com/giampaolo/psutil/blob/ede5e46dbee0c0f6c33f80629f75aecc1ad4143c/psutil/__init__.py#L942-L963)

This leads to thread counts in the hundreds when they're expected to be in the ones, like Gantt chart screenshot with CPU percent in "Threads"

So I think the "threads" part of these charts should be changed before the chart functionality is restored, either
* by updating the log to include an integer count of threads and use this value in the chart
* change the column from threads to CPU percentage
* something else?

effigies:
Yeah, seems like we want something like:

```
if status_dict['runtime_threads'] != "N/A":
    status_dict['runtime_threads'] //= 100
```

shnizzedy:
An existing unit test does https://github.com/nipy/nipype/blob/6c060304f380c46b2f05c5afdc7171dbbdfadc58/nipype/interfaces/base/tests/test_resource_monitor.py#L76-L78

which is similar to [what we're doing for now in C-PAC](https://github.com/FCP-INDI/C-PAC/blob/e63d16c61c217a5df8a8ef620af32e315e7369f5/CPAC/utils/monitoring/monitoring.py#L63-L64):

```Python
if runtime_threads != 'N/A':
    runtime_threads = math.ceil(runtime_threads/100)
```

My concern is that, as I read

> **Note:** the returned value can be > 100.0 in case of a process running multiple threads on different CPU cores.
> **Note:** the returned value is explicitly _not_ split evenly between all available CPUs (differently from `psutil.cpu_percent()`). This means that a busy loop process running on a system with 2 logical CPUs will be reported as having 100% CPU utilization instead of 50%. This was done in order to be consistent with `top` UNIX utility and also to make it easier to identify processes hogging CPU resources independently from the number of CPUs. It must be noted that `taskmgr.exe` on Windows does not behave like this (it would report 50% usage instead). To emulate Windows `taskmgr.exe` behavior you can do: `p.cpu_percent() / psutil.cpu_count()`.

― [psutil documentation: Process.cpu_percent](https://psutil.readthedocs.io/en/latest/#psutil.Process.cpu_percent)

this number can be a misleading estimate. For example, if a process is using 25% of each of 4 CPUs, I believe this would report 100%, which would reduce to 1 or 2 threads depending on how we're rounding up or not. I'd be happy to learn that either I'm misunderstanding the number or that the number is good enough.

effigies:
@shnizzedy Can you rebase/merge master to resolve conflicts? I think we let this go too long and should just merge and let people find bugs and fix them.

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -1,12 +1,20 @@
...
 import pytest
 import nipype.interfaces.utility as niu
 import nipype.pipeline.engine as pe
 
+try:
+    import pandas
+
+    has_pandas = True
+except ImportError:
+    has_pandas = False
```
effigies, line 16: Can you just add `pandas` to the test dependencies?

https://github.com/nipy/nipype/blob/1ccee342f6455af20e3c641fdc36caca32ef0b00/nipype/info.py#L154-L163

An explicit minimum version would be appreciated. Pandas 1.2 (first version with Python 3.9 wheels) or 2.0 ([SPEC 0](https://scientific-python.org/specs/spec-0000/)) would be reasonable choices, or anything in between. Shouldn't go later than 2.0 until SPEC0 does.

shnizzedy: Sure, good idea!

effigies: LOL, looking back, I suggested that you do this. I'm sorry to jerk you around. Though at a three year lag, it may not qualify as a "jerk".

nipype/info.py
```diff
@@ -153,6 +153,7 @@ def get_nipype_gitversion():
 
 TESTS_REQUIRES = [
     "coverage >= 5.2.1",
+    "pandas > 1.5.0",
```
effigies, line 156:
```suggestion
    "pandas >= 1.5.0",
```

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -60,3 +61,51 @@ def test_callback_exception(tmpdir, plugin, stop_on_first_crash):
...
+        loglines = _f.readlines()
+
+        # test missing duration
+        first_line = json.loads(loglines[0])
+        if "duration" in first_line:
+            del first_line["duration"]
+        loglines[0] = f"{json.dumps(first_line)}\n"
+
+        # test duplicate timestamp warning
+        loglines.append(loglines[-1])
```
effigies, line 104:
```suggestion
    # test missing duration
    first_line = json.loads(loglines[0])
    if "duration" in first_line:
        del first_line["duration"]
    loglines[0] = f"{json.dumps(first_line)}\n"

    # test duplicate timestamp warning
    loglines.append(loglines[-1])
```

nipype/pipeline/plugins/tests/test_callback.py
```diff
@@ -60,3 +61,51 @@ def test_callback_exception(tmpdir, plugin, stop_on_first_crash):
...
+    )
+    wf.add_nodes([f_node])
+    wf.config["execution"] = {"crashdump_dir": wf.base_dir, "poll_sleep_duration": 2}
+
+    plugin_args = {"status_callback": log_nodes_cb}
+    if plugin != "Linear":
+        plugin_args["n_procs"] = 8
+    wf.run(plugin=plugin, plugin_args=plugin_args)
+
+    with open(log_filename, "r") as _f:
```
effigies, line 94: Out of curiosity, why the `_` prefix?

```suggestion
    with open(log_filename, "r") as _f:
```

effigies (approved):
Passing tests! Small comments.

nipype/pipeline/plugins/tests/test_callback.py
shnizzedy: Habit ― I like to indicate temporary scope things like open files are internal variables, but it's definitely not required to be that way

effigies: Got it.

shnizzedy:
Do we want to do [this](https://github.com/nipy/nipype/pull/3290#issuecomment-833141167)

> ```Python
> if status_dict['runtime_threads'] != "N/A":
>     status_dict['runtime_threads'] //= 100
> ```

in this PR or kick the can on it?

effigies:
Let's kick the can. If you want to open another PR in 5 minutes, that's fine with me. Last time we had that question, it delayed things 3 years.

PR #3291, opened 2021-01-11
ENH: Interface for R

Terf:
## Summary

Like matlab, R has many useful neuroimaging tools but isn't as nice of a scripting language as python. This PR creates an R interface akin to the matlab interface, so R code may be more easily wrapped by nipype.

- Added R interface

## Acknowledgment


effigies (commented):
Overall looks good. I started reviewing this a while back, but I guess got lost in my sea of tabs. Here are some suggestions, and could you make a test (in `nipype/interfaces/tests/test_r.py`) that skips if R is not found? e.g.:

```Python
@pytest.mark.skipif(no_r, "No R command found")
def test_basic_Rscript():
    ...
```

effigies:
@Terf I'm going to try to make a release in the next couple days. If you want to try to finish this off, I'd love to include it!

Terf:
@effigies awesome! I just committed some tests, let me know if there any problems

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,100 @@
...
+        ' -e "1 + 1"'
+    )
+
+    assert ri.inputs.script == "1 + 1"
+    assert ri.inputs.script_file == "testscript"
+    assert not os.path.exists(ri.inputs.script_file), "scriptfile should not exist"
+    assert not os.path.exists(
+        default_script_file
+    ), "default scriptfile should not exist."
+
```
effigies, line 41: Style fixup.

```suggestion


```

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.mark.skipif(no_r, reason="R is not available")
+def test_r_init():
+    default_script_file = clean_workspace_and_get_default_script_file()
+
+    assert r.RCommand._cmd == "R"
```
effigies, line 46: This won't be true if `$RCMD` is defined as something other than `R`.

```suggestion
    assert r.RCommand._cmd == r.get_r_command()
```

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,100 @@
...
+    assert r.RCommand.input_spec == r.RInputSpec
```
effigies, line 47: This feels like an odd assertion. I suspect you can just drop it, unless there's a case where this might fail?

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,100 @@
...
+
+    assert r.RCommand().cmd == r_cmd
+    rc = r.RCommand(r_cmd="foo_m")
+    assert rc.cmd == "foo_m"
+
+
+@pytest.mark.skipif(no_r, reason="R is not available")
+def test_run_interface(tmpdir):
```
effigies, line 55: This should all be done in `tmpdir`.

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,100 @@
...
+    assert not os.path.exists(default_script_file), "scriptfile should not exist 3."
+    with pytest.raises(IOError):
+        rc.run()  # foo_m is not an executable
+    assert os.path.exists(default_script_file), "scriptfile should exist 3."
+    if os.path.exists(default_script_file):  # cleanup
+        os.remove(default_script_file)
+
+    cwd = tmpdir.chdir()
+
+    # bypasses ubuntu dash issue
```
effigies, line 76: The dash issue is referenced here with no explanation.

effigies (commented):
You're getting some test failures that should be fixed in `master`, if you want to merge.

nipype/interfaces/r.py
```diff
@@ -0,0 +1,127 @@
...
+    """Interface that runs R code
+
+    >>> import nipype.interfaces.r as r
+    >>> r = r.RCommand(rfile=False)  # don't write script file
+    >>> r.inputs.script = "Sys.getenv('USER')"
+    >>> out = r.run()  # doctest: +SKIP
+    """
+
+    _cmd = get_r_command()
+    _default_r_cmd = None
```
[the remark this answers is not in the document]
Terf: Admittedly, I wrote the entire R interface and tests by copying the matlab implementation and find+replacing matlab with R. So it's done that way because that's the way the matlab interface works, but I agree it's extraneous and removed it (would we want to remove it from the matlab interface as well for consistency's sake?)

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def clean_workspace_and_get_default_script_file():
+    # Make sure things are clean.
+    default_script_file = r.RInputSpec().script_file
+    if os.path.exists(default_script_file):
+        os.remove(
+            default_script_file
+        )  # raise Exception('Default script file needed for tests; please remove %s!' % default_script_file)
+    return default_script_file
```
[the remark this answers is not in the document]
Terf: Agreed; fixed

nipype/interfaces/tests/test_r.py
Terf: It's an odd assertion indeed, a relic from the matlab implementation so I dropped it

nipype/interfaces/tests/test_r.py
Terf: Another relic from the matlab implementation, I'm not sure what the dash issue is about either; removed.

effigies:
Can you merge latest `master` and run `make specs`? That should fix the failing tests.

nipype/interfaces/r.py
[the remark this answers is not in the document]
effigies: I'm inclined to leave it alone in MATLAB, since it's been there so long. Somebody may use it for something, and we'd be breaking things for them...

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,62 @@
...
+def test_set_rcmd(tmpdir):
+    os.chdir(tmpdir)
+    default_script_file = r.RInputSpec().script_file
+
+    ri = r.RCommand()
+    _default_r_cmd = ri._cmd
+    ri.set_default_r_cmd("foo")
+    assert not os.path.exists(default_script_file), "scriptfile should not exist."
+    assert ri._cmd == "foo"
+    ri.set_default_r_cmd(_default_r_cmd)
```
effigies, line 62: We probably don't do it everywhere, but I would recommend `chdir`ing in and out. IIRC there were some versions of pytest and coverage that interacted badly when tests `chdir`ed into temporary directories without going back before they got deleted. Anyway, burned once, twice shy.

```suggestion
    cwd = tmpdir.chdir()
    default_script_file = r.RInputSpec().script_file

    ri = r.RCommand()
    _default_r_cmd = ri._cmd
    ri.set_default_r_cmd("foo")
    assert not os.path.exists(default_script_file), "scriptfile should not exist."
    assert ri._cmd == "foo"
    ri.set_default_r_cmd(_default_r_cmd)
    cwd.chdir()
```

nipype/interfaces/tests/test_r.py
```diff
@@ -0,0 +1,63 @@
...
+    assert ri.inputs.script_file == default_script_file
+    assert not os.path.exists(ri.inputs.script_file), "scriptfile should not exist"
+    assert not os.path.exists(
+        default_script_file
+    ), "default scriptfile should not exist."
+
+
+@pytest.mark.skipif(no_r, reason="R is not available")
+def test_run_interface(tmpdir):
+    os.chdir(tmpdir)
```
effigies, line 31: Can you un-chdir in this one, too?

effigies (approved):
I'm good with this, at this point. I notice that the outputs are pretty verbose, so I suspect that at some point we'll want to add inputs to allow for control of that. If you want to adjust the behavior at all before merge, I'm happy to wait. Otherwise we can merge now and improve later.

effigies:
Awesome, thanks. If you're not already in the Zenodo file and would like to be included as an author on releases, please post on #3336 with your name, affiliation and ORCID.

Terf:
Since this PR’s already merged I’ll submit a new PR with inputs to control the verbosity of outputs. Right now, calling `run()` on an `RCommand` will return a “Bunch” object with fields like the command line executed, command output, env var dump, timestamps, etc. Perhaps the less verbose version should just return the command output as the command would?

I wonder if these changes should just apply to the R interface, or we should abstract the R and matlab interfaces into an “AbstractLanguage” class that could be extended by other language interfaces to reduce code reuse. After all, the R interface copied much of the code from the matlab interface, and I think it could be useful to have interfaces for Java and Julia which would worsen the code duplication problem between them.

I would be honored to be included as an author on releases! Thanks, and I hope to make more significant contributions in the future.

PR #3297, opened 2021-02-04
ENH: Skip newline before Python call in batch submission to facilitate containerized runs

d-vogel:
## Summary

- make possible the integration of the call to the python scipt generated from nodes into a larger call
- one example this is the use of singularity with the slurm template file.

## Fixes
- Remove the newline at the end of template to avoid having an empty line after joining with the node command. This allows to have the node command integrated in more complex call, such as a `singularity exec ...` call.

### Before:
```bash
#/!bin/bash
singularity exec $SINGULARITY_IMG.sif\

python node.py
```

### After:
```bash
#/!bin/bash
singularity exec $SINGULARITY_IMG.sif\
python node.py
```


effigies:
Hey, thanks for this. I'm trying to look into the test failure when I have time... Hope to get back to this soon.

nipype/pipeline/plugins/base.py
```diff
@@ -543,7 +543,7 @@ def _submit_job(self, node, updatehash=False):
         pyscript = create_pyscript(node, updatehash=updatehash)
         batch_dir, name = os.path.split(pyscript)
         name = ".".join(name.split(".")[:-1])
-        batchscript = "\n".join((self._template, "%s %s" % (sys.executable, pyscript)))
+        batchscript = "\n".join((self._template[:-1], "%s %s" % (sys.executable, pyscript)))
```
effigies, line 546: Looking at this, I think we want to be a little more discriminating, as it's not guaranteed that a template file will end in a newline. What if we did :

```suggestion
        batchscript = "\n".join(
            (self._template.rstrip("\n"), "%s %s" % (sys.executable, pyscript))
        )
```

d-vogel: Indeed, my approach was naïve.

effigies:
Thanks!

PR #3316, opened 2021-03-30
FIX: Update SmoothEstimateOutputSpec resels description

HippocampusGirl:
## Summary

As mentioned previously (https://github.com/poldracklab/ds003-post-fMRIPrep-analysis/issues/7), the `resels` output of FSL's smoothest refers to the size of a resel, not the *number of resels*. 

I propose to update the output spec to avoid misunderstandings.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/fsl/model.py
+++ b/nipype/interfaces/fsl/model.py
@@ -1886,7 +1886,7 @@ class SmoothEstimateInputSpec(FSLCommandInputSpec):
 class SmoothEstimateOutputSpec(TraitedSpec):
     dlh = traits.Float(desc="smoothness estimate sqrt(det(Lambda))")
     volume = traits.Int(desc="number of voxels in mask")
-    resels = traits.Float(desc="number of resels")
+    resels = traits.Float(desc="number of voxels per resel")
 
 
 class SmoothEstimate(FSLCommand):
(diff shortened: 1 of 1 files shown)


nipype/interfaces/fsl/model.py
```diff
@@ -1886,7 +1886,7 @@ class SmoothEstimateInputSpec(FSLCommandInputSpec):
 class SmoothEstimateOutputSpec(TraitedSpec):
     dlh = traits.Float(desc="smoothness estimate sqrt(det(Lambda))")
     volume = traits.Int(desc="number of voxels in mask")
-    resels = traits.Float(desc="number of resels")
+    resels = traits.Float(desc="number of voxels per resel")
```
effigies, line 1889: I've had difficulty tracking down documentation from FSL on this, but found the following footnote in a @nicholst [white paper](https://www.fmrib.ox.ac.uk/datasets/techrep/tr08tn1/tr08tn1.pdf):

> Beware that FSL's `smoothest` program reports `RESELS`, which is actually the the size of one Resel, i.e. volume of space with dimensions FWHM_x, FWHM_y, FWHM_z, i.e. from `smoothest`, `RESELS` = FWHMx FWHMy FWHMz. In SPM, “RESEL” generically refers to the Resel count, R_3. And, for completeness, note that `smoothest`'s DHL=|Λ|^1/2, “Determinant Half-Power Lambda.”

Further, if I'm reading it correctly, it seems that these FWHM are in voxels for each dimension, so I think "number of voxels per resel" is a correct phrasing, but it might be clearer to write:

```suggestion
    resels = traits.Float(desc="volume of resel, in voxels, defined as FWHM_x * FWHM_y * FWHM_z")
```

effigies (approved):
LGTM. Did some digging to verify and propose an alternative phrasing, but okay with it going in as-is.

HippocampusGirl:
Sounds good!

nicholst:
Belatedly giving this a thumbs up. Thanks!

PR #3331, opened 2021-04-30
[REF] Cache nodes in workflow to speed up construction, other optimizations

HippocampusGirl:
## Summary

I'm working with fMRIPrep and large data sets (via the ENIGMA consortium). In my testing, Nipype workflows with more than a few hundred thousand nodes can be prohibitively slow, so that it can take days before processing starts. This comes specifically from the function `_generate_flatgraph` and `generate_expanded_graph`. While my previous pull requests #3184 and #3260 fixed some bottlenecks, I recently had the time to do some more profiling and try to fix some more. 
I understand that Pydra is just around the corner, and that switching to it may solve my performance troubles then. However, the small changes outlined here can reduce the time before processing starts by approximately half (in my testing with a workflow of 75k nodes). 
I have tried to divide up the changes into independent commits, so that you can decide which ones may be interesting for inclusion.

Here is my profile before starting out. Please not that the absolute durations are not interpretable, because the profiler reduces performance by a lot.

Here is the profile with all proposed changes. Note that `_create_flat_graph` has become the dark gray box without caption on the bottom right.

- The `workflow._has_node` function is called for every new connection, and currently iterates through all nodes in the workflow and nested workflows. It then compares each node to the `wanted_node`. We can speed this up, because `networkx` graphs are dictionaries and have support the `in` statement. This way we can reduce the scaling behavior from linear to almost constant time, but only in the best-case scenario that the `wanted_node` is in the workflow. 
Otherwise, we still need to iterate to find all nested workflows. To overcome this last limitations, we can cache the nested workflows in a set, and update it whenever the graph is changes.
https://github.com/nipy/nipype/blob/4e8f54d8e304dd940cfdc5d42f37288651bc8a03/nipype/pipeline/engine/workflows.py#L906-L913

- The `_merge_graphs` function uses `np.unique` for a list of strings. We can use the built-in `set` to avoid casting to Numpy array. `np.unique` additionally sorts the input array, which we really don't need here
https://github.com/nipy/nipype/blob/4e8f54d8e304dd940cfdc5d42f37288651bc8a03/nipype/pipeline/engine/utils.py#L756

- Use sets wherever order is not important and we just need to check whether a string or a node is contained or not, again reducing linear scaling to a constant time look-up.
https://github.com/nipy/nipype/blob/4e8f54d8e304dd940cfdc5d42f37288651bc8a03/nipype/pipeline/engine/workflows.py#L173-L174

- Remove a topological sort that may not be necessary. Please correct me if I'm wrong :-)
https://github.com/nipy/nipype/blob/4e8f54d8e304dd940cfdc5d42f37288651bc8a03/nipype/pipeline/engine/workflows.py#L942

- Avoid re-compiling regular expressions in a loop by using Python's built-in functions
https://github.com/nipy/nipype/blob/4e8f54d8e304dd940cfdc5d42f37288651bc8a03/nipype/pipeline/engine/utils.py#L1114-L1116

## Acknowledgment

### DIFF
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -333,7 +333,7 @@ connected.
         newnodes = []
         all_nodes = self._get_all_nodes()
         for node in nodes:
-            if self._has_node(node):
+            if node in all_nodes:
                 raise IOError("Node %s already exists in the workflow" % node)
             if isinstance(node, Workflow):
                 for subnode in node._get_all_nodes():
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -970,7 +970,7 @@ connected.
             raise Exception(
                 ("Workflow: %s is not a directed acyclic graph " "(DAG)") % self.name
             )
-        nodes = list(nx.topological_sort(self._graph))
+        nodes = list(self._graph.nodes)
         for node in nodes:
             logger.debug("processing node: %s", node)
             if isinstance(node, Workflow):
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -912,10 +912,12 @@ connected.
         return allnodes
 
     def _has_node(self, wanted_node):
-        for node in self._graph.nodes():
+        if wanted_node in self._graph:
+            return True  # best case scenario
+        for node in self._graph:  # iterate otherwise
             if wanted_node == node:
                 return True
-            if isinstance(node, Workflow):
+            if hasattr(node, "_has_node"):  # hasattr is faster than isinstance
                 if node._has_node(wanted_node):
                     return True
         return False
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -144,7 +144,7 @@ class Workflow(EngineBase):
             self.disconnect(connection_list)
             return
 
-        newnodes = []
+        newnodes = set()
         for srcnode, destnode, _ in connection_list:
             if self in [srcnode, destnode]:
                 msg = (
@@ -154,9 +154,9 @@ class Workflow(EngineBase):
 
                 raise IOError(msg)
             if (srcnode not in newnodes) and not self._has_node(srcnode):
-                newnodes.append(srcnode)
+                newnodes.add(srcnode)
             if (destnode not in newnodes) and not self._has_node(destnode):
-                newnodes.append(destnode)
+                newnodes.add(destnode)
         if newnodes:
             self._check_nodes(newnodes)
             for node in newnodes:
--- a/nipype/pipeline/engine/workflows.py
+++ b/nipype/pipeline/engine/workflows.py
@@ -59,6 +59,9 @@ class Workflow(EngineBase):
         super(Workflow, self).__init__(name, base_dir)
         self._graph = nx.DiGraph()
 
+        self._nodes_cache = set()
+        self._nested_workflows_cache = set()
+
     # PUBLIC API
     def clone(self, name):
         """Clone a workflow
@@ -269,6 +272,8 @@ connected.
                 "(%s, %s): new edge data: %s", srcnode, destnode, str(edge_data)
             )
 
+        self._update_node_cache()
+
     def disconnect(self, *args):
         """Disconnect nodes
         See the docstring for connect for format.
@@ -314,6 +319,8 @@ connected.
             else:
                 self._graph.add_edges_from([(srcnode, dstnode, edge_data)])
 
+        self._update_node_cache()
+
     def add_nodes(self, nodes):
         """ Add nodes to a workflow
 
@@ -346,6 +353,7 @@ connected.
             if node._hierarchy is None:
                 node._hierarchy = self.name
         self._graph.add_nodes_from(newnodes)
+        self._update_node_cache()
 
     def remove_nodes(self, nodes):
         """ Remove nodes from a workflow
@@ -356,6 +364,7 @@ connected.
             A list of EngineBase-based objects
         """
         self._graph.remove_nodes_from(nodes)
+        self._update_node_cache()
 
     # Input-Output access
     @property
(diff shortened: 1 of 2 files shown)


nipype/pipeline/engine/workflows.py
```diff
@@ -903,21 +913,32 @@ def _set_node_input(self, node, param, source, sourceinfo):
...
+        removed_nodes = self._nodes_cache.difference(nodes)
+
+        self._nodes_cache = nodes
+        self._nested_workflows_cache.difference_update(removed_nodes)
+
+        for node in added_nodes:
+            if isinstance(node, Workflow):
+                self._nested_workflows_cache.add(node)
+
     def _has_node(self, wanted_node):
```
effigies, line 936: This method is a depth-first search, and it seems that there should be ways to cut branches early and save effort. Does `wanted_node` have information we can use, such as `wanted_node.name.startswith(subworkflow.name)`? I can't really remember when and where the hierarchical names get attached.

nipype/pipeline/engine/workflows.py
```diff
@@ -903,21 +913,32 @@ def _set_node_input(self, node, param, source, sourceinfo):
...
-            if wanted_node == node:
+        if wanted_node in self._nodes_cache:
+            return True
+        for node in self._nested_workflows_cache:
+            if node._has_node(wanted_node):
                 return True
-            if isinstance(node, Workflow):
-                if node._has_node(wanted_node):
-                    return True
         return False
```
effigies, line 942: Small style thought:

```Python
def _has_node(self, wanted_node):
    return (
        wanted_node in self._nodes_cache or
        any(wf._has_node(wanted_node) for wf in self._nested_workflows_cache)
    )
```

This feels more descriptive than procedural but does the same comparisons in the same order, with short-circuiting. Feel free to reject.

nipype/pipeline/engine/workflows.py
```diff
@@ -949,7 +970,7 @@ def _generate_flatgraph(self):
             raise Exception(
                 ("Workflow: %s is not a directed acyclic graph " "(DAG)") % self.name
             )
-        nodes = list(nx.topological_sort(self._graph))
+        nodes = list(self._graph.nodes)
```
effigies, line 973: I think you're right that we don't need a topological sort here.

effigies (commented):
Awesome. This all looks correct. A few thoughts, but nothing blocking. Looking at the style check, it looks like `black` released a new version that affects a lot of files, so it should not block this PR, either.

nipype/pipeline/engine/workflows.py
HippocampusGirl: I like this. GitHub doesn't let me commit this as a suggestion, so I will need to pull it out and commit it manually :-)

nipype/pipeline/engine/workflows.py
```diff
@@ -269,6 +273,8 @@ def connect(self, *args, **kwargs):
                 "(%s, %s): new edge data: %s", srcnode, destnode, str(edge_data)
             )
 
+        self._update_node_cache()
```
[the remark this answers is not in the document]
HippocampusGirl: We would still need to go though every one of `newnodes` and determine whether it's a `Workflow` or a `Node`, but we could save the set difference calculation. I'm not sure if the performance impact is worth the effort.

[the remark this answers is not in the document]
HippocampusGirl: Looking at the profiler output, it seems we could save around 40 seconds of total runtime, or 17% of the total time spent on `connect`.

nipype/pipeline/engine/workflows.py
effigies: Okay, so we're literally just looking for "is this node object *somewhere* in this potentially hierarchical workflow"? So it's just following a tree of disjoint sets, which isn't the worst. Are we still spending a significant amount of time in `_has_node()` or are the cache updates a more worthwhile target?

The reason I ask is because it's not obvious we really need to recurse. Looking at each call to `_has_node()`, most make sense for the local workflow, and the recursive case seems to be covered by using `_get_all_nodes()`:

https://github.com/nipy/nipype/blob/0b49a70a007d9de7387b63db20dcea43fb263d49/nipype/pipeline/engine/workflows.py#L334-L344

Maybe there's a pathological construction strategy that we're heading off by recursing, so I wouldn't drop it without making sure we have tests for those strategies.

effigies: > I'm not sure I follow what you mean by needing recursion

I think in >95% of use cases, `_has_node(node)` could be replaced with `node in self._node_cache` and not recursing down into sub-workflows. Something like fmriprep's style of doing:

```Python
def init_x_wf():
    wf = Workflow()
    wf.add_nodes([locally, defined, nodes])
    return wf
```

Ensures that nodes that are added to sub- or super-workflows aren't in-scope to be added. If we're doing an object-identity check, then there's no point in me checking sub-workflows for a node that the subworkflow can't know about.

> or pathological construction.

Here I mean a situation where you're making multiple workflows in the same scope and you actually could do:

```Python
wf1 = Workflow()
wf2 = Workflow()
bad_node = Node()
wf2.add_nodes([bad_node])
wf1.add_nodes([wf2])
wf1.add_nodes([bad_node])
```

I'm pretty sure this should fail now, and if we were to stop recursing in `_has_node()`, we'd need to make sure that this case is still causing an error by some other check.

effigies: Also, my thought with the `_get_all_nodes()` check was:

```diff
  all_nodes = self._get_all_nodes()
  for node in nodes:
-     if self._has_node(node):
+     if node in all_nodes:
          raise IOError("Node %s already exists in the workflow" % node)
      if isinstance(node, Workflow):
          for subnode in node._get_all_nodes():
              if subnode in all_nodes:
                  raise IOError(
                      ("Subnode %s of node %s already exists " "in the workflow")
                      % (subnode, node)
                  )
```

This is a case where the `_has_node()` recursion is necessary, and we are already `_get_all_nodes()`ing, so that's why I thought we might be able to drop it.

HippocampusGirl: Thank you for the explanation! That makes sense. Personally, I would not worry too much about the recursion, the code is still quite efficient.

That's a good point, I can add that.

nipype/pipeline/engine/workflows.py
```diff
@@ -314,6 +320,8 @@ def disconnect(self, *args):
             else:
                 self._graph.add_edges_from([(srcnode, dstnode, edge_data)])
 
+        self._update_node_cache()
+
```
effigies, line 324: Looking at your latest profile, you spend a chunk of time in `disconnect()`, but it does not remove nodes, only edges.

```suggestion
```

nipype/pipeline/engine/workflows.py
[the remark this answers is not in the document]
effigies:
```suggestion
        if newnodes:
            self._update_node_cache()
```

effigies (approved):
Two last thoughts, but otherwise this LGTM.

HippocampusGirl:
Thank you for your quick feedback! I always fell like I'm learning a lot from your code reviews

effigies:
Thank you!

PR #3346, opened 2021-06-22
STY: Make private member name consistent with the rest of them

oesteban:
## Summary

Fixes # .

## Acknowledgment


oesteban:
The documentation now renders just a tiny bit nicer. If we want to leverage doi2bib and make it better, it is now as hard/easy as it was before. Since these changes seem pretty innocuous, I'll go ahead and merge.

PR #3347, opened 2021-06-22
RF: Clean-up the BaseInterface ``run()`` function using context

oesteban:
Python contexts seem the most appropriate pattern to follow.

## Summary

Fixes # .

## Acknowledgment

### DIFF
--- a/nipype/interfaces/base/core.py
+++ b/nipype/interfaces/base/core.py
@@ -1,10 +1,8 @@
-# -*- coding: utf-8 -*-
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """
 Nipype interfaces core
 ......................
-
 
 Defines the ``Interface`` API and the body of the
 most basic interfaces.
@@ -12,20 +10,15 @@ The I/O specifications corresponding to these base
 interfaces are found in the ``specs`` module.
 
 """
-from copy import deepcopy
-from datetime import datetime as dt
 import os
-import platform
 import subprocess as sp
 import shlex
-import sys
 import simplejson as json
-from dateutil.parser import parse as parseutc
 from traits.trait_errors import TraitError
 
 from ... import config, logging, LooseVersion
 from ...utils.provenance import write_provenance
-from ...utils.misc import str2bool, rgetcwd
+from ...utils.misc import str2bool
 from ...utils.filemanip import split_filename, which, get_dependencies, canonicalize_env
 from ...utils.subprocess import run_command
 
@@ -39,7 +32,12 @@ from .specs import (
     MpiCommandLineInputSpec,
     get_filecopy_info,
 )
-from .support import Bunch, InterfaceResult, NipypeInterfaceError, format_help
+from .support import (
+    RuntimeContext,
+    InterfaceResult,
+    NipypeInterfaceError,
+    format_help,
+)
 
 iflogger = logging.getLogger("nipype.interface")
 
@@ -63,8 +61,15 @@ class Interface(object):
 
     """
 
-    input_spec = None  # A traited input specification
-    output_spec = None  # A traited output specification
+    input_spec = None
+    """
+    The specification of the input, defined by a :py:class:`~traits.has_traits.HasTraits` class.
+    """
+    output_spec = None
+    """
+    The specification of the output, defined by a :py:class:`~traits.has_traits.HasTraits` class.
+    """
+
     _can_resume = False  # See property below
     _always_run = False  # See property below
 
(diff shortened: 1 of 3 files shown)


oesteban:
I've been noticing that under some (yet unknown) conditions, the report will show definitely set inputs as undefined, and some errors that escape generating a crash file.

This PR, in collaboration with #3349, makes the flow of operations a bit more clear and makes the treatment of errors more reliable.

oesteban:
To note, I have tested this and #3349 locally with fMRIPrep and dMRIPrep and they seem to work as expected.

nipype/interfaces/base/core.py
```diff
@@ -365,131 +370,42 @@ def run(self, cwd=None, ignore_exception=None, **inputs):
...
-                self.__class__.__name__,
-                proc_pid,
-            )
-            mon_sp = ResourceMonitor(proc_pid, freq=mon_freq)
-            mon_sp.start()
 
-        # Grab inputs now, as they should not change during execution
-        inputs = self.inputs.get_traitsfree()
-        outputs = None
+        with rtc(self, cwd=cwd, redirect_x=self._redirect_x) as runtime:
```
satra, line 380: This changes the logic of exceptions i believe. previously the run interface only raised/trapped exceptions related to the underlying interface, while other nipype related exceptions would never be skipped by ignore_exception. in the current context, any exception would be either re-raised or ignored.

for example if mandatory checks failed, the interface would raise an exception. now that would be supressed if ignore_exception is true.

this is a change in design which may be worth discussing a bit.

satra (commented):
some things to discuss about exceptions.

nipype/interfaces/base/core.py
```diff
@@ -63,8 +61,15 @@ class Interface(object):
 
     """
 
-    input_spec = None  # A traited input specification
-    output_spec = None  # A traited output specification
+    input_spec = None
+    """
```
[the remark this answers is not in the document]
oesteban: Sphinx follows this. It was a PEP that finally never passed. But Sphinx kept it.

nipype/interfaces/base/core.py
oesteban: I believe we could revert this particular deviation from the original behavior by moving the next three lines outside the context.

I think I added them to make sure they are run while the current directory is `cwd`, but we can use the `indirectory` context Chris wrote some time ago for that.

WDYT?

satra: as an initial step, we can do that to maintain current compatibility. although cwd should have no role to play in those three statements, so could simply be run by themselves without `indirectory`.

however, i think the key question here is what are the scenarios, if any, in which your current implementation (which i like) would be troublesome.

oesteban: because contexts behave like try/catch blocks, I think this is pretty safe. I also created the mock resource monitor, so that we don't need to branch depending on whether it is active or not.



> as an initial step, we can do that to maintain current compatibility. although cwd should have no role to play in those three statements, so could simply be run by themselves without `indirectory`.

I'm not sure about this for the ``self.inputs.trait_set(**inputs)`` line. If I'm not wrong, this is going to set those paths in ``inputs``, and if they are provided relative then they will likely fail (unless the user is very aware that the paths should be given relative to the current working directory, not the intended working directory of the interface.

What would you like to see for going forward with this PR?

nipype/interfaces/base/core.py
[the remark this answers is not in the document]
oesteban: meaning, if you want a class attribute to be documented with sphinx, you have to add the comment after it is defined.

nipype/interfaces/base/core.py
satra: The question is whether there is any reason to ignore the exception if any of the following three lines fail? In the past a failure of those lines would have always generated an exception. Under the current change, those errors can be ignored. 

Also, while unlikely that many people used cwd, for backwards compatibility purposes, we shouldn't adjust the relative path behind the scenes I think.

So to move forward, perhaps simply pull the three statements out of the context manager. That should make the behavior as is.

oesteban: Cool, please note the `os.chdir(cwd)` that used to live a few lines ahead these three... Meaning that we still need to change the directory. Just to make sure we're on the same page.

satra: i missed that. in that case `inDirectory` is necessary.

oesteban: Yup

oesteban: (actually that lone chdir was one of the reasons to start this)

satra (approved):

effigies (approved):
LGTM too.

PR #3351, opened 2021-06-29
ENH: interface for whitestripe normalization

Terf:
## Summary

A processing pipeline for MS usually follows these preprocessing steps:
1. N4 correction
2. Registration
3. Skull stripping
4. Intensity normalization

nipype provides interfaces for tools for all these steps except intensity normalization. The WhiteStripe method is especially well suited for MS brains as lesions can skew mean/variance.

- Added WhiteStripe interface

## Acknowledgment


Terf:
@effigies It looks like I have some more work to fix the failings CI tests, but I'm wondering if this is something you'd accept into nipype in the first place?

effigies:
Absolutely! Please let me know if you want feedback before working on the tests, or if you want to get things fixed up and then get feedback.

Terf:
@effigies I’d love some feedback before working on the tests as I’m not sure how stringent they should be. Right now the test runs whitestripe on a nifti downloaded from a public dataset, then simply checks the output file got created. A more stringent test might actually compare the intensities of the normalized vs original image, but I’m not sure whether our tests should be for whitestripe itself, or just testing that the interface generates the correct command-line string. I think a robust test would actually run whitestripe (maybe in addition to checking the command-line string?), but that’s complicated by the fact that neither R nor the whitestripe R package are installed in the official nipype docker image. As the whitestripe interface isn’t usable unless whitestripe is installed, perhaps as a first step I should submit a PR to create (perhaps a separate?) container image with R and whitestripe installed?

effigies:
> I’m not sure whether our tests should be for whitestripe itself, or just testing that the interface generates the correct command-line string

Generally nipype concerns itself with the command-line, if we're not implementing the algorithms ourselves. There may be some rare cases where we do exercise the underlying tools, but that's typically going to be when the operations are pretty cheap.

> As the whitestripe interface isn’t usable unless whitestripe is installed, perhaps as a first step I should submit a PR to create (perhaps a separate?) container image with R and whitestripe installed?

If you want to install R and whitestripe, then go ahead and do it in this PR.

Terf:
@effigies It seems one of the checks is failing because the auto test, `nipype/interfaces/tests/test_auto_WhiteStripe.py` doesn't exist. I tried running `make specs` but that didn't seem to generate the file? I also tried running everything with just `make` and got a segmentation fault from sphinx, not sure if that's just a problem with my environment though.

But otherwise I rewrote the test (which passes for me locally) and added R and whitestripe to the nipype docker image. One thing to note is the main image is now `6.98GB` on my machine, I'm not sure how much of a size reduction you get by compressing layers afterwards but R is a hefty dependency and so are the neuroimaging packages required by whitestripe. Maybe it would make sense to offer a "slim" version of the image without R, and a "full" version with everything?

effigies:
I ran `make specs` and pushed the new file to 3a64072779ce6ca077abde3ef80249f0e68a8f65. If you fetch from the main repository, you should be able to merge it into your branch.

Terf:
@effigies looks like all the tests are passing now.

On a somewhat unrelated note, I wanted to get your input / approval on another interface I wanted to add for the [mimosa lesion segmentation method](https://github.com/avalcarcel9/mimosa). Since the training requires manually segmented images I imagine nearly everyone will use the built-in models my lab has trained, but I think including the training step in the interface could still be useful —  but should the training and test parts be bundled into the same interface, or should there be two interfaces for training new model / using that model to test new data?

I also wanted to write a tutorial (like the ones [here](https://nipype.readthedocs.io/en/latest/examples.html)) showing how to segment MS lesions by building a nipype workflow that uses ANTS for bias correction / skull striping, whitestripe for normalization and then finally mimosa. I’m a bit confused where the source files are for the documentation are though — I looked through the [docs](https://github.com/nipy/nipype/tree/master/doc) but don’t see the many articles listed under the User Guide and Examples section of the website?

PR #3357, opened 2021-07-29
[FIX] Also allow `errno.EBUSY` during `emptydirs` on NFS

HippocampusGirl:
## Summary

Running a workflow with nipype 1.6.1, I sometimes get the following exception:
```
│     Traceback (most recent call last):
│       File "/usr/local/miniconda/lib/python3.7/site-packages/halfpipe/plugins/multiproc.py", line 71, in run_node
│         result["result"] = node.run(updatehash=updatehash)
│       File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 493, in run
│         emptydirs(outdir, noexist_ok=True)
│       File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/utils/filemanip.py", line 794, in emptydirs
│         raise ex
│       File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/utils/filemanip.py", line 781, in emptydirs                                                                                           │         shutil.rmtree(path)
│       File "/usr/local/miniconda/lib/python3.7/shutil.py", line 494, in rmtree
│         _rmtree_safe_fd(fd, path, onerror)
│       File "/usr/local/miniconda/lib/python3.7/shutil.py", line 452, in _rmtree_safe_fd
│         onerror(os.unlink, fullname, sys.exc_info())
│       File "/usr/local/miniconda/lib/python3.7/shutil.py", line 450, in _rmtree_safe_fd
│         os.unlink(entry.name, dir_fd=topfd)
│     OSError: [Errno 16] Device or resource busy: '.nfs0000000002273a940000254d'
│     During handling of the above exception, another exception occurred:
│     Traceback (most recent call last):
│       File "/usr/local/miniconda/lib/python3.7/concurrent/futures/process.py", line 239, in _process_worker
│         r = call_item.fn(*call_item.args, **call_item.kwargs)
│       File "/usr/local/miniconda/lib/python3.7/site-packages/halfpipe/plugins/multiproc.py", line 74, in run_node
│         result["result"] = node.result
│       File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 217, in result
│         op.join(self.output_dir(), "result_%s.pklz" % self.name)
│       File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/utils.py", line 291, in load_resultfile
│         raise FileNotFoundError(results_file)
│     FileNotFoundError
```
As far as I can tell, if a file on an NFS mount is still open somewhere, when it is deleted, NFS will rename that file to a hidden file ("placeholder") in the same directory so that the data stays available to the open file handle (according to <https://unix.stackexchange.com/a/348678>). When `shutil` discovers that hidden file, it will try to delete it, and we get an `OSError` with `errno.EBUSY`

- [x] This pull request catches that error number in addition to `errno.ENOTEMPTY`, which is already handled by current code

## Acknowledgment


effigies:
Might need to merge/rebase master to fix the dipy issue.

nipype/utils/filemanip.py
```diff
@@ -780,8 +780,8 @@ def emptydirs(path, noexist_ok=False):
     try:
         shutil.rmtree(path)
     except OSError as ex:
-        elcont = os.listdir(path)
-        if ex.errno == errno.ENOTEMPTY and not elcont:
+        elcont = [p for p in os.listdir(path) if not p.startswith(".nfs")]
+        if ex.errno in [errno.ENOTEMPTY, errno.EBUSY] and not elcont:
```
effigies, line 784: Can we try putting in a test to make sure this works as expected? Something like:

```Python
import os
from pathlib import Path
from unittest import mock

def nfs_unlink(pathlike, *, dir_fd=None):
    if dir_fd is None:
        path = Path(pathlike)
        deleted = path.with_name(".nfs00000000")
        path.rename(deleted)
    else:
        os.rename(pathlike, ".nfs1111111111", src_dir_fd=dir_fd, dst_dir_fd=dir_fd)

def test_emptydirs_dangling_nfs(tmp_path):
    busyfile = tmp_path / "base" / "subdir" / "busyfile"
    busyfile.parent.mkdir(parents=True)
    busyfile.touch()
    with mock.patch("os.unlink") as mocked:
        mocked.side_effect = nfs_unlink
        emptydirs(tmp_path / "base")

    assert Path.exists(tmp_path / "base")
    assert not busyfile.exists()
    assert busyfile.parent.exists()  # Couldn't remove
```

HippocampusGirl: That sounds like a great idea :-) Your code works as is, I just had to update my patch to also condsider the `subdir`. For the `nfs_unlink` mock, I also removed the case for `dir_fd` being `None`, because as far as I can tell, the `os.rename` line also handles that (`src_dir_fd` and `dst_dir_fd` are `None` by default).

HippocampusGirl:
I just noticed that I had this error, because I had another instance of the workflow open in a different terminal, still open in the debugger. Closing that instance of Python solved the issue. As a result, I think the issue I had comes down to user error. I am not sure if it's worth the additional code complexity to handle this kind of problem in `nipype` though, so it may make sense to discard the changes here and keep only the unit test.

nipype/utils/filemanip.py
effigies: The reason for the other case is that, if we don't get the dir_fd, we would get more than just the basename.

HippocampusGirl: That makes sense! Thank you for explaining

nipype/utils/tests/test_filemanip.py
```diff
@@ -670,3 +671,21 @@ def test_write_rst_list(tmp_path, items, expected):
     else:
         with pytest.raises(expected):
             write_rst_list(items)
+
+
+def nfs_unlink(pathlike, *, dir_fd=None):
+    os.rename(pathlike, ".nfs1111111111", src_dir_fd=dir_fd, dst_dir_fd=dir_fd)
```
HippocampusGirl, line 677:
```suggestion
    if dir_fd is None:
        path = Path(pathlike)
        deleted = path.with_name(".nfs00000000")
        path.rename(deleted)
    else:
        os.rename(pathlike, ".nfs1111111111", src_dir_fd=dir_fd, dst_dir_fd=dir_fd)
```

HippocampusGirl (commented):
I wanted to try commiting like this, let's see if it works :-)

nipype/utils/tests/test_filemanip.py
```diff
@@ -670,3 +671,26 @@ def test_write_rst_list(tmp_path, items, expected):
...
             write_rst_list(items)
+
+
+def nfs_unlink(pathlike, *, dir_fd=None):
+    if dir_fd is None:
+        path = Path(pathlike)
+        deleted = path.with_name(".nfs00000000")
+        path.rename(deleted)
+    else:
+        os.rename(pathlike, ".nfs1111111111", src_dir_fd=dir_fd, dst_dir_fd=dir_fd)
```
effigies, line 682: I do like your idea of avoiding branching. What if we just construct the target correctly?

```suggestion
    target = os.path.join(os.path.dirname(pathlike), ".nfs00000000")
    os.rename(pathlike, target, src_dir_fd=dir_fd, dst_dir_fd=dir_fd)
```

Demo:

```Python
>>> [os.path.join(os.path.dirname(pathlike), ".nfs00000000")
...  for pathlike in ["path/subdir/busyfile", "busyfile"]]
['path/subdir/.nfs00000000', '.nfs00000000']
```

effigies (approved):
Green check in any case. Happy to merge as-is, or if you want to make that change.

PR #3369, opened 2021-08-21
[FIX] coord for mrconvert

rmukh:
mrconvert in the newest version of MRtrix3 crashes if -coord is float

## Summary

Fixes # .

## Acknowledgment


nipype/interfaces/mrtrix3/utils.py
```diff
@@ -560,7 +560,7 @@ class MRConvertInputSpec(MRTrix3BaseInputSpec):
         desc="output image",
     )
     coord = traits.List(
-        traits.Float,
+        traits.Int,
```
effigies, line 563: If previous versions allowed floats, then this change could break workflows that were written for those versions. Do you know what the history is here?

PR #3376, opened 2021-09-17
ENH: Stop printing false positive differences when logging cached nodes

oesteban:
The function calculating changed input fields did not preempt the problem of comparing tuples vs. lists in most of the cases. This PR simplifies the logic of the `dict_diff` accessory function, and fixes the issue - resulting in way shorter indices of changed inputs (esp. when an interface has several inputs and some of them contain long lists).

This issue translated in practice into showing long lists of unchanged values as if they had changed, when ofter, the changed input invalidating the cache would be some other field.


oesteban:
I don't think the failing test is related, right?

effigies:
Looks like git annex/datalad is getting confused.

```
[INFO] Author identity unknown
| 
| *** Please tell me who you are.
| 
| Run
| 
|   git config --global user.email "you@example.com"
|   git config --global user.name "Your Name"
| 
| to set your account's default identity.
| Omit --global to set the identity only in this repository.
| 
| fatal: unable to auto-detect email address (got 'neuro@63b92567e55a.(none)') 
[INFO] git-annex: user error (git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","-c","annex.dotfiles=true","commit-tree","e976c1453f70daf99659701b820cabde9fb00d2f","--no-gpg-sign","-p","refs/heads/git-annex"] exited 128) 
```

@yarikoptic I think this is a datalad 0.15 regression. The fix seems to be to set user info, but I assume this is not intended behavior.

nipype/utils/misc.py
```diff
@@ -296,12 +296,16 @@ def dict_diff(dold, dnew, indent=0):
 
     typical use -- log difference for hashed_inputs
     """
-    # First check inputs, since they usually are lists of tuples
-    # and dicts are required.
-    if isinstance(dnew, list):
+    try:
         dnew = dict(dnew)
-    if isinstance(dold, list):
         dold = dict(dold)
```
effigies, line 301: If `dict(dold)` fails while `dict(dnew)` succeeds, then we will get a less comparable error message. What about:

```Python
dnew, dold = dict(dnew), dict(dold)
```

This way, if either fails, the creation of the RHS tuple fails and assignment to the LHS never occurs.

nipype/utils/misc.py
```diff
@@ -321,21 +325,28 @@ def dict_diff(dold, dnew, indent=0):
 
     diffkeys = len(diff)
 
+    def _shorten(value):
+        if isinstance(value, str) and len(value) > 50:
+            return f"{value[:10]}...{value[-10:]}"
+        if isinstance(value, (tuple, list)) and len(value) > 10:
+            return tuple(list(value[:2]) + "..." + list(value[-2:]))
```
effigies, line 332: I'm guessing we're never hitting this because you can't add a list and str.

```suggestion
            return tuple(list(value[:2]) + ["..."] + list(value[-2:]))
```

nipype/utils/misc.py
```diff
@@ -321,21 +325,28 @@ def dict_diff(dold, dnew, indent=0):
...
-            diff += ["  * %s: %r != %r" % (k, dnew[k], dold[k])]
+        new = dnew[k]
+        old = dold[k]
+        # Reading from JSON produces lists, but internally we typically
+        # use tuples. At this point these dictionary values can be
+        # immutable (and therefore the preference for tuple).
+        if isinstance(dnew[k], (list, tuple)):
+            new = tuple([tuple(el) if isinstance(el, list) else el for el in dnew[k]])
+        if isinstance(dnew[k], (list, tuple)):
+            old = tuple([tuple(el) if isinstance(el, list) else el for el in dold[k]])
```
effigies, line 346: What if there are dicts, or are nested beyond two levels? Should we just have a generic value uniformizer?

```Python
def uniformize(val):
    if isinstance(val, dict):
        return {k: uniformize(v) for k, v in val.items()}
    if isintsance(val, (list, tuple)):
        return tuple(uniformize(el) for el in val)
    return val
```

Then this is just:

```Python
new = uniformize(dnew[k])
old = uniformize(dold[k])
```

PR #3377, opened 2021-09-22
ENH: ``verbose`` input should not be hashed in ``ants.Registration``

oesteban:
A tiny correction.


effigies:
Seems to break tests. `make specs` and update the hash value here: https://github.com/nipy/nipype/runs/3672742514?check_suite_focus=true#step:10:6479

PR #3378, opened 2021-09-23
MNT: Add user name and email to Docker to appease git/annex/datalad

effigies:
Fixing an apparent datalad 0.15 regression that causes failures if `user.name` and `user.email` aren't set in Docker containers.


effigies:
Quick fix. Merging.

PR #3380, opened 2021-09-27
FIX: Extension not extensions, after pybids v0.9

adelavega:
## Summary
Updates `BIDSDataGrabber` to reflect that `extensions` is a deprecated keyword, in favor of `extension`

Related issue: https://neurostars.org/t/issue-with-bidsdatagrabber-in-nipype-1-6-1/20243/3

## Acknowledgment

- [ X ] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -2935,12 +2935,12 @@ class BIDSDataGrabber(LibraryBaseInterface, IOBase):
                 "bold": {
                     "datatype": "func",
                     "suffix": "bold",
-                    "extensions": ["nii", ".nii.gz"],
+                    "extension": ["nii", ".nii.gz"],
                 },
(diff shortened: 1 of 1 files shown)


effigies:
In retrospect, I wish we'd put this in PyBIDS rather than Nipype. This will break things for people with old versions of pybids, but it's obviously been broken for recent pybids.

adelavega:
Yeah. We could check the bids version and update based on that. 

At least its easy for the user the over-ride the output query but for many users the error will be confusing

effigies:
Yeah. I think this is fine. It would be good to start allowing interfaces to say "this requires `pybids>=0.9`". Right now they can only say "this requires this set of modules".

PR #3385, opened 2021-09-30
[FIX] Removed exists=True from MathsOutput

### LINKED ISSUE #3384: MathsCommand().inputs.out_file has to exist
When setting `MathsCommand().inputs.out_file` argument, a Trait error is thrown because of the `exists=True`:

https://github.com/nipy/nipype/blob/6a7837c0994367a5f34bb576bb0a97ec70669b8f/nipype/interfaces/fsl/maths.py#L44

I think this doesn't make any sense, because `out_file` is to be created by `MathsCommand()`.
In my case, I just want to get rid of SPM nan's in an .nii file, but I also don't want fsl_maths to append a suffix to the filename - this is why I also provide `out_file` to `MathsCommand().inputs`.
A short example:
```
nan2zero = pe.MapNode(interface=MathsCommand(), name='nan2zero', iterfield=['in_file', 'out_file'])
nan2zero.inputs.nan2zeros = True

```

0nu:
## Summary
When setting MathsCommand().inputs.out_file argument, a Trait error is thrown because of the exists=True, although this file is to be created. This PR fixed this by removing the `exists=True` argument.

Fixes #3384 .

* Removed `exists=True` from `MathsOutput` `File()`.

## Acknowledgment

### DIFF
--- a/nipype/interfaces/fsl/maths.py
+++ b/nipype/interfaces/fsl/maths.py
@@ -41,7 +41,7 @@ class MathsInput(FSLCommandInputSpec):
 
 class MathsOutput(TraitedSpec):
 
-    out_file = File(exists=True, desc="image written after calculations")
+    out_file = File(desc="image written after calculations")
 
 
 class MathsCommand(FSLCommand):


effigies:
I think you're going to need to rerun `make check-before-commit`. (See https://github.com/nipy/nipype/blob/master/CONTRIBUTING.md#making-a-change)

effigies:
Ah, nevermind. It seems that `exists=True` isn't a thing we check for in specs, so no changes would result. Thanks!

0nu:
Thanks!

PR #3407, opened 2021-11-09
RF: Optimize ICC_rep_anova using a global cache

### LINKED ISSUE #3406: [enhancement] ICC_rep_anova could be significantly faster than it is when running on images
### Summary
ICC_rep_anova calculates ICC(3,1) on a table of subjects and repeated measures.  Setting up the design matrix is far more computationally expensive than actually calculating ICC from the input data (by around a factor of 100 for 2000 subjects and 2 repeated measures). For repeated calculations (such as calculating ICC on every voxel of a set of images), this goes much more slowly than it should.

I made a modified version of ICC_rep_anova for a project I'm working on where I needed the speedup for the calculation to be practical; It's simple enough to cache the design matrix calculations and put in some if statements to decide if they need to be calculated/recalculated.  I don't know if this is something anybody else actually ever wants to do, but if so, it's an easy fix.

### Actual behavior
As implemented, the design matrix setup is performed on each call. 

### Expected behavior
The setup only depends on the shape of the input table.  When running the calculation on an image, this shape will be the same for each voxel, so does not need to be redone. If the routine is uninitialized, the design information should be calculated and cached.  If the routine is called again, and the design matrix is unchanged, this information should be retrieved and used.  If the design matrix is changed, the cached information should be discarded and recalculated.

### Script/Workflow details

[A proposed fix is here in lines 89 and 96-123](https://github.com/bbfrederick/nipype/blob/master/nipype/algorithms/icc.py)

This seems to pass all of the tests.

bbfrederick:
## Summary

Fixes #3406 .

* Caches design matrix setup information for speed.
* Adds a nan_to_num to the final ICC calculation to catch bad values.

## Acknowledgment


nipype/algorithms/icc.py
```diff
@@ -109,20 +130,22 @@ def ICC_rep_anova(Y):
...
 
     # create the design matrix for the different levels
-    x = kron(eye(nb_conditions), ones((nb_subjects, 1)))  # sessions
-    x0 = tile(eye(nb_subjects), (nb_conditions, 1))  # subjects
-    X = hstack([x, x0])
+    if not icc_inited:
+        x = kron(eye(nb_conditions), ones((nb_subjects, 1)))  # sessions
+        x0 = tile(eye(nb_subjects), (nb_conditions, 1))  # subjects
+        X = hstack([x, x0])
+        centerbit = dot(dot(X, pinv(dot(X.T, X))), X.T)
```
effigies, line 137:
```suggestion
        centerbit = X @ pinv(X.T @ X) @ X.T
```

nipype/algorithms/icc.py
```diff
@@ -109,20 +130,22 @@ def ICC_rep_anova(Y):
...
 
     # Sum Square Error
-    predicted_Y = dot(dot(dot(X, pinv(dot(X.T, X))), X.T), Y.flatten("F"))
+    predicted_Y = dot(centerbit, Y.flatten("F"))
```
effigies, line 140:
```suggestion
    predicted_Y = centerbit @ Y.flatten("F")
```

effigies (approved):
@bbfrederick Thanks for your patience. I think when I first looked at this, I wanted to try to figure out an alternative to using `global`, but I don't think that should hold this up. Giving up on that dream and revisiting this, the only suggestion I have is that we can use the `@` operator to have more readable matrix multiplication. Happy to merge this in as-is, though.

effigies:
Actually, yes, let's revert the `nan_to_num`, and leave that decision to downstream consumers. Otherwise we'll be changing things on people who've been using the presence of nans to make decisions.

effigies:
@bbfrederick Looking to do another release next week, if you want to try to get this one finished off?

effigies:
Consider #3454 as an alternative.

PR #3421, opened 2021-12-10
[ENH]  Enable BidsDataGrabber to load saved layouts

gaunab:
A new input `load_layout` enables the `BIDSDataGrabber` to read a previously saved BIDSLayout. This makes the use of BidsDatagrabber much faster, since it doesn't need to reinterpret all Bidsfiles every time. 

Layouts can be saved via `layout.save(<path>)`.


effigies (approved):
Can merge on passing tests.

PR #3443, opened 2022-04-01
ENH: ROBEX interface

mfmachado:
Code modified from: https://neurostars.org/t/nipype-interfaces/1225

## Acknowledgment

https://neurostars.org/t/nipype-interfaces/1225


.zenodo.json
```diff
@@ -263,7 +263,7 @@
     },
     {
       "affiliation": "CIBIT, UC",
-      "name": "Machado, F\u00e1tima",
+      "name": "Dias, Maria de Fatima",
```
effigies, line 266: Could you also add an entry to [`.mailmap`](https://github.com/nipy/nipype/blob/master/.mailmap)? The format would be:

```
Maria de Fatima Dias <your.preferred@email.tld>
```

If you commit with multiple email addresses, you can add additional lines to ensure that all your commits are associated to a single `Name <email>`:

```
Maria de Fatima Dias <your.preferred@email.tld> <other.1@email.tld>
Maria de Fatima Dias <your.preferred@email.tld> <other.2@email.tld>
```

nipype/interfaces/robex/preprocess.py
```diff
@@ -0,0 +1,84 @@
...
+        outputs["out_file"] = self.inputs.out_file
+        if isdefined(self.inputs.out_mask):
+            outputs["out_mask"] = self.inputs.out_mask
+        return outputs
+
+    def run(self, cwd=None, ignore_exception=None, **inputs):
+        if not isdefined(self.inputs.out_file):
+            _, base, extension = split_filename(self.inputs.in_file)
+            self.inputs.out_file = base + "_brain_robex" + extension
+        return super(RobexSegment, self).run(cwd, ignore_exception, **inputs)
```
effigies, line 84: I think all of this should be done automatically. Can you try removing it and seeing if it works for you?

```suggestion
```

nipype/interfaces/robex/preprocess.py
```diff
@@ -0,0 +1,84 @@
...
+        hash_files=False,
+        name_template='%s_brainmask',
+        keep_extension=True,
+    )
+    seed = traits.Int(desc="Seed for random number generator", position=3, argstr="%i")
+
+
+class RobexOutputSpec(TraitedSpec):
+    out_file = File(desc="Output volume", exists=True)
+    out_mask = File(desc="Output mask", exists=True)
```
effigies, line 40: We haven't been entirely consistent before, but `exists=True` is not a great check for output specs.

```suggestion
    out_file = File(desc="Output volume")
    out_mask = File(desc="Output mask")
```

nipype/interfaces/robex/preprocess.py
```diff
@@ -0,0 +1,84 @@
...
+    constrained by a shape model, to the probabilistic output of a supervised brain boundary classifier.
+    Because the shape model cannot perfectly accommodate unseen cases, a small free deformation is subsequently allowed.
+    The deformation is optimized using graph cuts.
+    The method ROBEX is based on was published in IEEE Transactions on Medical Imaging;
+    please visit the website http://www.jeiglesias.com to download the paper.
+
+    Examples
+    --------
+    >>> path_mr = 'structural.nii'
+    >>> robex = RobexSegment(in_file=path_mr)
```
effigies, line 57: One useful check is the cmdline:

```suggestion
    >>> robex.cmdline
    'runROBEX.sh structural.nii structural_brain.nii structural_mask.nii'
```

effigies (commented):
Thanks for this! And looks great! I've made a few suggestions that you should definitely check locally before just taking my word for it.

One additional suggestion would be to add in `robex/__init__.py`:

```Python
from .preprocess import RobexSegment
```

nipype/interfaces/robex/preprocess.py
mfmachado: Thanks for your comments :). I did this because:

- **_format_arg**: Without this line if I run this  interface in a Node, the file is not created inside the node's folder.
-  **_list_outputs(self)** without this line the out_file in the outputs is undefined. 
- *** if not isdefined(self.inputs.out_file)** this was because the runRobex.sh requires that the output filename is provided, but I rather prefer not define it myself everything I have to run the code..  :) Do you preferer that I remove it?

effigies: Ah, right. That's because you need `name_source=["in_file"]` in your input traits. I'll make a suggestion...

nipype/interfaces/robex/preprocess.py
```diff
@@ -0,0 +1,89 @@
...
+        hash_files=False,
+        name_template='%s_brain',
+        keep_extension=True,
+    )
+    out_mask = File(
+        desc="Output mask",
+        position=2,
+        argstr="%s",
+        hash_files=False,
+        name_template='%s_brainmask',
```
effigies, line 32:
```suggestion
        name_template='%s_brain',
        name_source=["in_file"],
        keep_extension=True,
    )
    out_mask = File(
        desc="Output mask",
        position=2,
        argstr="%s",
        hash_files=False,
        name_template='%s_brainmask',
        name_source=["in_file"],
```

nipype/interfaces/robex/preprocess.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Examples
+    --------
+    >>> from nipype.interfaces.robex.preprocess import RobexSegment
+    >>> robex = RobexSegment()
+    >>> robex.inputs.in_file = 'structural.nii'
+    >>> robex.inputs.out_file = 'structural_brain.nii'
+    >>> robex.inputs.out_mask = 'structural_mask.nii'
+    >>> robex.cmdline # doctest: +SKIP
+    'runROBEX.sh structural.nii structural_brain.nii structural_mask.nii'
```
effigies, line 64: Let's let the interface create the output names for us. And then we can actually run `robex.cmdline`, rather than skipping it:

```suggestion
    >>> robex.cmdline
    'runROBEX.sh structural.nii structural_brain.nii structural_brainmask.nii'
```

Issue #3450, opened 2022-04-20
ICC_rep_anova speedup suggestion


kristoferm94:
### Summary
ICC_rep_anova can be sped up by rearranging the matrix multiplication order and using the hermitian=True flag in pinv

### Actual behavior
As implemented, pinv is called without the hermitian=True flag.
dot(X.T, X) will always be a hermitian matrix so speedup can be achieved by setting the hermitian flag in pinv to True.

The matrix multiplication in line 117 in icc.py is arranged in a way that fat matrices end up being multiplied together.
The matrix multiplication order can be rearranged in a way that we end up multiplying fat matrices with column vectors most of the time.

### Script/Workflow details
Suggested modification in lines 117: https://github.com/kristoferm94/nipype/blob/linalg-speedup/nipype/algorithms/icc.py

I.e. replace:
```
predicted_Y = dot(dot(dot(X, pinv(dot(X.T, X))), X.T), Y.flatten("F"))
```
with
```
predicted_Y = X @ (pinv(X.T @ X, hermitian=True) @ (X.T @ Y.flatten("F")))
```

effigies:
Seems like a good idea. Would you mind also changing `dot(a, b)` with `a @ b` in these expressions?

kristoferm94:
Sure. I've updated the first comment.

effigies:
Sorry, should have been clearer: Do you want to submit a quick PR?

kristoferm94:
No problem.
Sure, I've submitted a pull request with the proposed change: https://github.com/nipy/nipype/pull/3453

PR #3453, opened 2022-04-21
RF: Rearranging matmul order and using hermitian flag in ICC_rep_anova

kristoferm94:
## Summary
See https://github.com/nipy/nipype/issues/3450

Calculation of predicted_Y refactored such that:
* numpy.dot is replaced with the @ operator 
* The pseudo-inverse of X.T@X is now calculated with the hermitian flag set to True (pinv uses a more efficient algorithm when the matrix input is assumed to be hermitian, which will always be the case with X.T@X in this function).
* The matrix multiplication order is changed in a way that we don't end up multiplying as many fat matrices as before.


effigies (approved):
Nice! I ran some quick tests just to see what this looks like. The reordering seems to dominate, but both changes seem well worth making.

effigies:
I think you are going to want to remove the `dot` import from the top of the file, since it's no longer used. Otherwise looks good to me.

kristoferm94:
I removed the numpy.dot import

effigies:
Thanks!

PR #3454, opened 2022-04-22
RF: Optimize ICC_rep_anova with a memoized helper function

effigies:
@bbfrederick Post #3453, I got motivated to look at your PR again. I grokked the code a little better and realized we could do this more Pythonically.

WDYT? I set the `maxsize=1` to match your effective cache size. I don't really know how big these matrices can get, so I'm not sure whether that was a memory optimization or just an assumption that the number of switches back and forth between shapes is likely to be small.

@kristoferm94 Would be interested in your thoughts. This does break up the improved matrix ordering you provided, but I suspect the caching benefit is going to dominate here.

As some basic tests:

```Python
In [1]: import numpy as np

In [2]: rng = np.random.default_rng()

In [3]: from nipype.algorithms.icc import *

In [4]: %time ICC_rep_anova(rng.normal(size=(50, 10)))
CPU times: user 26.4 ms, sys: 67 ms, total: 93.4 ms
Wall time: 15.4 ms
Out[4]: 
(-0.0010955283641284907,
 -0.0012372157263673023,
 1.130569661036011,
 0.01640374953027528,
 9,
 441)

In [5]: ICC_projection_matrix.cache_clear()

In [6]: %time ICC_rep_anova(rng.normal(size=(50, 10)))
CPU times: user 431 µs, sys: 18.8 ms, total: 19.3 ms
Wall time: 3.14 ms
Out[6]: 
(0.052737606398966394,
 0.049204080066708754,
 0.8837938966422438,
 0.028379415383302194,
 9,
 441)

In [7]: %timeit ICC_rep_anova(rng.normal(size=(50, 10)))
160 µs ± 7.95 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Cells 4-6 is just to distinguish compilation effects. Cell 7 shows the effect of caching.

### DIFF
--- a/nipype/algorithms/icc.py
+++ b/nipype/algorithms/icc.py
@@ -86,20 +86,41 @@ class ICC(BaseInterface):
         return outputs
 
 
-def ICC_rep_anova(Y):
+def ICC_rep_anova(Y, nocache=False):
     """
     the data Y are entered as a 'table' ie subjects are in rows and repeated
     measures in columns
-
     One Sample Repeated measure ANOVA
-
     Y = XB + E with X = [FaTor / Subjects]
-    """
 
-    [nb_subjects, nb_conditions] = Y.shape
-    dfc = nb_conditions - 1
-    dfe = (nb_subjects - 1) * dfc
-    dfr = nb_subjects - 1
+    This is a hacked up (but fully compatible) version of ICC_rep_anova
+    from nipype that caches some very expensive operations that depend
+    only on the input array shape - if you're going to run the routine
+    multiple times (like, on every voxel of an image), this gives you a
+    HUGE speed boost for large input arrays.  If you change the dimensions
+    of Y, it will reinitialize automatially.  Set nocache to True to get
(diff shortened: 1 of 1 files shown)


bbfrederick:
Sorry it took so long to get back to you - kept finding myself on mobile with no screen big enough to look at code.

Yes, that looks like a good switch - I wasn't aware of lru_cache (I need to get out more)!  Maxsize=1 is fine - my use case is calculating it on ~260k voxels in a set of 3D images from 1000 subjects with 4 conditions - exactly the same matrix size each time, so the repeated pseudoinverse was killing me....  I don't imagine you'd need to switch the matrix up very frequently, although I'm sure you could come up with some pathological application that would.

effigies:
Sounds good. I agree that there's probably not a huge advantage in storing more than one precomputed pseudoinverse. I'll go ahead and merge. If someone can come up with a relevant use case for increasing the cache, we can revisit.

PR #3458, opened 2022-04-28
MNT: Drop distutils

### LINKED ISSUE #3439: Drop `distutils`
### Summary

distutils is deprecated. Here is the output of `grep -rI distutils .` in the nipype root:

```
./doc/changelog/0.X.X-changelog.rst:* ENH: Abandon distutils, only use setuptools (https://github.com/nipy/nipype/pull/1627)
./nipype/interfaces/afni/base.py:from distutils import spawn
./nipype/interfaces/dipy/preprocess.py:from distutils.version import LooseVersion
./nipype/interfaces/dipy/reconstruction.py:from distutils.version import LooseVersion
./nipype/interfaces/dipy/registration.py:from distutils.version import LooseVersion
./nipype/interfaces/dipy/setup.py:    from numpy.distutils.misc_util import Configuration
./nipype/interfaces/dipy/setup.py:    from numpy.distutils.core import setup
./nipype/interfaces/dipy/stats.py:from distutils.version import LooseVersion
./nipype/interfaces/dipy/tracks.py:from distutils.version import LooseVersion
./nipype/interfaces/mrtrix3/connectivity.py:            from distutils.spawn import find_executable
./nipype/interfaces/niftyreg/base.py:from distutils.version import StrictVersion
./nipype/interfaces/slicer/generate_classes.py:    from numpy.distutils.misc_util import Configuration
./nipype/interfaces/slicer/generate_classes.py:    from numpy.distutils.core import setup
./nipype/utils/misc.py:from distutils.version import LooseVersion
./nipype/utils/misc.py:        distutils.version.LooseVersion.
./nipype/utils/config.py:from distutils.version import LooseVersion
./nipype/__init__.py:from distutils.version import LooseVersion
./tools/toollib.py:from distutils.dir_util import remove_tree
```

We can break this down into groups:

* `LooseVersion` - We import this directly from distutils in many places and also import it into `nipype.__init__.py` from which it is imported in many other places. The semantics of LooseVersion are not the same as `packaging.version.Version`, although in most places it should generally be close enough. Are we okay with a blanket replace?
* `StrictVersion` - Probably safe to replace with `Version`?
* `distutils.spawn.find_executable` - We can use `shutil.which`
* `numpy.distutils` - Used in `dipy/setup.py` and files generated by `slicer/generate_classes.py`; I don't understand this use case so I'm hesitant to do anything.
* `distutils.dir_util.remove_tree` - I don't think `tools/toollib.py` is part of anybody's workflow. Easier just to remove? Alternately, can use `shutil.rmtree`.

Assigning to 1.8 so that there will be a semver hurdle.

effigies:
## Summary
FreeSurfer interfaces have some fairly deep LooseVersion integration that I didn't want to touch right now, so I just vendored LooseVersion in externals.

Fixes #3439.


nipype/interfaces/base/core.py
```diff
@@ -747,7 +748,7 @@ def _run_interface(self, runtime, correct_return_codes=(0,)):
 
         # which $cmd
         executable_name = shlex.split(self._cmd_prefix + self.cmd)[0]
-        cmd_path = which(executable_name, env=runtime.environ)
+        cmd_path = shutil.which(executable_name, path=runtime.environ.get("PATH"))
```
mgxd, line 751:
```suggestion
        cmd_path = shutil.which(executable_name, path=runtime.environ.get("PATH"), os.defpath)
```

nipype/interfaces/mrtrix3/connectivity.py
```diff
@@ -208,17 +209,15 @@ def _parse_inputs(self, skip=None):
...
+            path = shutil.which(self._cmd)
             if path is None:
                 path = os.getenv(MRTRIX3_HOME, "/opt/mrtrix3")
             else:
                 path = op.dirname(op.dirname(path))
 
             self.inputs.in_config = op.join(
                 path,
-                "src/dwi/tractography/connectomics/" "example_configs/fs_default.txt",
+                "src/dwi/tractography/connectomics/example_configs/fs_default.txt",
```
mgxd, line 220: shouldn't this be path expanded?
```suggestion
                "src",
                "dwi",
                "tractography",
                "connectomics",
                "example_configs",
                "fs_default.txt",
```

nipype/utils/filemanip.py
```diff
@@ -28,6 +28,9 @@
 
 related_filetype_sets = [(".hdr", ".img", ".mat"), (".nii", ".mat"), (".BRIK", ".HEAD")]
 
+# Keep anybody who depended on the old hack from breaking
+which = shutil.which
```
mgxd, line 32: just a note, the keyword args (which can also be positional) are different

mgxd (commented):
Much needed, thanks for pushing on this.

One thing - `packaging.version.LegacyVersion` might be a better fit for a lot of the non-Python tools, since they may not follow PEP440 (especially if using dev versions [this has bitten me before when working with dev FreeSurfer versions]). Or alternatively, we can use `packaging.version.parse` to produce either classes depending on the version string.

effigies:
I tried this for a bit... LegacyVersion has deprecation warnings.

mgxd:
I think that sounds reasonable

nipype/interfaces/mrtrix3/connectivity.py
effigies: It can be, but I don't see much advantage to it. Windows interprets forward slashes as path separators, and this fits on one line.

nipype/utils/filemanip.py
effigies: Yeah. I guess we should keep and deprecate the old one instead.

ghisvail: Can't you wrap calls to `shutil.which` in `n/u/filemanip.py#which` and keep the old function signature alive ?

I'd expect users who want to use `shutil.which` to import it directly instead of using nipype as a proxy for it anyway. However, anyone using nipype's `which` with named arguments would get an error which is sad. It depends how popular you think these utilities are in client code. I know Clinica uses some and we are trying to move away from them to reduce coupling. 

Or a solution in between would be start with a `DeprecationWarning` advising to port to `shutil.which` and later remove or proxy it?

ghisvail: Can't you wrap calls to `shutil.which` in `n/u/filemanip.py#which` and keep the old function signature alive ?

I'd expect users who want to use `shutil.which` to import it directly instead of using nipype as a proxy for it anyway. However, anyone using nipype's `which` with named arguments would get an error which is sad. It depends how popular you think these utilities are in client code. I know Clinica uses some and we are trying to move away from them to reduce coupling. 

Or a solution in between would be start with a `DeprecationWarning` advising to port to `shutil.which` and later remove or proxy it?

ghisvail:
> Maybe we should consider externals.version.LooseVersion a long term solution rather than a bridge. If it's Python, use Version, if it's not, use LooseVersion unless a custom method is needed



> Maybe we should consider externals.version.LooseVersion a long term solution rather than a bridge. If it's Python, use Version, if it's not, use LooseVersion unless a custom method is needed



> Maybe we should consider externals.version.LooseVersion a long term solution rather than a bridge. If it's Python, use Version, if it's not, use LooseVersion unless a custom method is needed

Now that I have understood the needs a bit more, I believe that's the best you can do with what's available to you right now.

effigies:
Thanks for your input @ghisvail. I've decided to defer the deprecation of `filemanip.which` and converting what we sensibly can of `LooseVersion` to `Version` for a future release so we can just get this out the door, as I've held up 1.8.0 for too long. I will make issues of these things.

effigies:
Opened #3464 and #3465. Merging.

PR #3463, opened 2022-05-07
ENH: Add random seed option to ANTs registration

ghisvail:
## Context

Clinica uses the ANTs registration interfaces for some of its pipelines. Since reproducibility is a major focus for us, it would be desirable to expose and use the random seed option recently introduced to the ANTs registration CLI.

## Summary

This PR exposes the corresponding random seed options for `antsRegistration` and `antsRegistrationSynQuick.sh`. Both are named differently but expects its value as an integer. If unspecified, the seed value is computed from the system clock internally according to the documentation.


effigies (approved):
LGTM. Thanks, I thought I'd actually sent you a request to add `min_ver`, but it was sitting here in my pending review.

Could you also run `make specs`?

ghisvail:
> It runs fine on my end.

My bad, I thought they were some sorts of additional checks.

Thanks for updating them.

PR #3470, opened 2022-05-11
ENH: Add inputs to mrtrix3.DWIPreprocInputSpec and remove mandatory annotation for pe_dir

### LINKED ISSUE #3469: Mrtrix3's DWIPreproc force inputs that are not mandatory
### Summary
The interface to Mrtrix3's dwifslpreproc forces some inputs that are not really necessary according to [the original CLI on which it relies](https://mrtrix.readthedocs.io/en/dev/reference/commands/dwifslpreproc.html).

### Actual behavior
The interface requires inputs to be set for out_grad_mrtrix, pe_dir, etc.

### Expected behavior
Ideally, it would change the necessary inputs (that are different and depends on the way you use this CLI).
A simpler solution would be to simply reduce the mandatory inputs to those that truly are.

python -c "import nipype; from pprint import pprint; pprint(nipype.get_info())"
-->

```
{'commit_hash': 'd77fc3c',
 'networkx_version': '2.6.3',
 'nibabel_version': '3.2.1',
 'nipype_version': '1.8.0',
 'numpy_version': '1.20.0',
 'pkg_path': '/home/groot/Projects/PhD/venv/lib/python3.9/site-packages/nipype',
 'scipy_version': '1.7.3',
 'sys_executable': '/home/groot/Projects/PhD/venv/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.9.12 (main, Mar 24 2022, 16:20:11) \n[GCC 9.4.0]',
 'traits_version': '6.3.2'}
```

### Execution environment

GalBenZvi:
## Summary
Simply removed the "mandatory" tag from pe_dir keyword argument as it isn't really mandatory, for example when using rpe_header (in which dwifslpreproc will read the pe_dir from the associated metadata)

Fixes #3469 .


nipype/interfaces/mrtrix3/preprocess.py
```diff
@@ -267,7 +284,7 @@ class DWIPreprocInputSpec(MRTrix3BaseInputSpec):
     )
     pe_dir = traits.Str(
         argstr="-pe_dir %s",
-        mandatory=True,
+        mandatory=False,
```
effigies, line 287:
```suggestion
```

nipype/interfaces/mrtrix3/tests/test_auto_DWIPreproc.py
```diff
@@ -71,18 +83,15 @@ def test_DWIPreproc_inputs():
...
-            argstr="%s",
+            argstr="-export_grad_mrtrix %s",
             extensions=None,
-            requires=["export_grad_mrtrix"],
-            usedefault=True,
         ),
         pe_dir=dict(
             argstr="-pe_dir %s",
-            mandatory=True,
+            mandatory=False,
```
effigies, line 94:
```suggestion
```

effigies (commented):
Conventionally, we only use `mandatory=True`, as `mandatory=False` is equivalent to not providing it at all. I will commit these changes. @GalBenZvi can you merge master afterwards?

GalBenZvi:
> Conventionally, we only use `mandatory=True`, as `mandatory=False` is equivalent to not providing it at all. I will commit these changes. @GalBenZvi can you merge master afterwards?

Sue, no problem. (:

PR #3472, opened 2022-05-13
ENH: Enable compression with Gunzip interface

### LINKED ISSUE #3471: Add compression to Gunzip interface
In Clinica, we use the [Gunzip](https://github.com/nipy/nipype/blob/6c060304f380c46b2f05c5afdc7171dbbdfadc58/nipype/algorithms/misc.py#L282) interface quite a bit, especially for SPM-based pipelines since SPM does not handle compressed NIfTI files. Afaik, this interface only supports decompression.

Would there be an intereset to enhance this interface to also support compression, so that results from SPM can be persisted in compressed form? Unless there is an alternative option I missed.

ghisvail:
See #3471 for context.

Closes #3471


ghisvail:
> but it feels weird to add compress to Gunzip, rather than a Gzip interface

I guess the goal would be to be consistent with the behaviour of `gzip` and `gunzip` on the shell.

I'll merge your changes in.

ghisvail:
Looks like it's ready now. Thanks @effigies for your help.

effigies:
Looks great! Thanks.

PR #3489, opened 2022-07-12
FIX: Add tolerance parameter to ComputeDVARS

### LINKED ISSUE #3487: Voxelwise dvars explodes with voxels with variance ~1e-19
### Summary

In `compute_dvars` we remove zero-variance voxels:

https://github.com/nipy/nipype/blob/3e7e61339ee7d0e67c917477fa989759fa69a577/nipype/algorithms/confounds.py#L1045-L1054

However, if `func_sd` is small but nonzero, this will propagate to `diff_sdhat`:

https://github.com/nipy/nipype/blob/3e7e61339ee7d0e67c917477fa989759fa69a577/nipype/algorithms/confounds.py#L1061-L1063

Which then can explode the values in 

https://github.com/nipy/nipype/blob/3e7e61339ee7d0e67c917477fa989759fa69a577/nipype/algorithms/confounds.py#L1078-L1080

If sufficiently small, we will overflow `float32`.

### Proposal

We should set a default threshold for "zerovariance". I think 1e-10 is more than small enough. In the example dataset I have, only one additional voxel is caught by this less stringent threshold, and would avoid the failure.

Will propose a fix soon, but wanted to give people a chance to comment.

effigies:
Closes #3487.

Also made some changes to avoid numpy deprecation warnings, and used nibabel 3+ idioms for coercing data types.

### DIFF
--- a/nipype/algorithms/confounds.py
+++ b/nipype/algorithms/confounds.py
@@ -50,6 +50,11 @@ class ComputeDVARSInputSpec(BaseInterfaceInputSpec):
     remove_zerovariance = traits.Bool(
         True, usedefault=True, desc="remove voxels with zero variance"
     )
+    variance_tol = traits.Float(
+        1e-7,
+        usedefault=True,
+        desc="maximum variance to consider \"close to\" zero for the purposes of removal",
+    )
     save_std = traits.Bool(True, usedefault=True, desc="save standardized DVARS")
     save_nstd = traits.Bool(False, usedefault=True, desc="save non-standardized DVARS")
     save_vxstd = traits.Bool(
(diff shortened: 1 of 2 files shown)


mgxd (approved):

nipype/algorithms/confounds.py
```diff
@@ -994,8 +1000,19 @@ def _list_outputs(self):
...
+    return AR_est_YW(x, order, rxx=rxx)[0]
+
+
 def compute_dvars(
-    in_file, in_mask, remove_zerovariance=False, intensity_normalization=1000
+    in_file,
+    in_mask,
+    remove_zerovariance=False,
+    variance_tol=0.0,
+    intensity_normalization=1000,
```
effigies, line 1015:
```suggestion
    intensity_normalization=1000,
    variance_tol=0.0,
```

PR #3501, opened 2022-08-22
MNT: Pin traits < 6.4

effigies:
This basically sets a lifetime on nipype 1.x. Refactoring to work with breaking changes in traits seems more work than it's worth (and would presumably break with older traits).

## Summary

Fixes # .

### DIFF
--- a/.github/workflows/contrib.yml
+++ b/.github/workflows/contrib.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-18.04']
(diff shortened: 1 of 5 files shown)


effigies:
This fixes the breaking tests in master.

mgxd (commented):
+1 on capping traits over a massive refactor. We should get this in for the next batch of nipreps releases.

re: the failing tests - it seems not all output compcor lists are numpy arrays. weird that this hasn't shown up in other PRs, but the fix should be a simple coercion

effigies:
@mgxd Please merge if you're happy with this.

PR #3506, opened 2022-09-14
FIX: Use interpolation/method in numpy.percentile as available

effigies:
Bug introduced in #3489. Numpy 1.22 introduced method and deprecated interpolation in a single shot. Typically, they deprecate a few versions after they introduce an alternative, so I missed this and now things that have pinned, e.g., numpy 1.21 are going to break on nipype.


effigies:
@mgxd Would you mind giving this a quick glance?

mgxd (approved):
lgtm - is this something we should track and remove once 1.22 becomes the minimum?

effigies:
I generally try to grep for `PY` strings when bumping minimum Python and `NP` strings when bumping minimum numpy, so the comment there is my tracking plan.

PR #3509, opened 2022-09-20
FIX: Deployed Matlab applications must not call addpath()

### LINKED ISSUE #3507: MATLAB:mpath:PathAlterationNotSupported / Modifying the search path is not supported by MATLAB Compiler using MAtlab 2019b
[It seems modifying the matlab path using addpath() is not likely to be supported at any point after MATLAB MCR_R2018b](https://github.com/populse/populse_mia/issues/238).
However, [nipype use addpath()](https://github.com/nipy/nipype/blob/master/nipype/interfaces/matlab.py#L194) command in the matlab script generated .
So [the easy fix is to stick with R2018b.](https://fr.mathworks.com/matlabcentral/answers/554821-compiling-failing-using-matlab-2019b)
But it's a real problem not to be able to use newer MCRs in the future.
Can you remove the addpath() command when generating the script ? Or is there another way to use MCR after R2018b?

DimitriPapadopoulos:
## Summary

https://mathworks.com/help/compiler/isdeployed.html

> The path of a deployed application is fixed at compile time and cannot change. Use `isdeployed` to ensure that the application uses path modifying functions, such as `addpath`, before deployment.

As suggested by @servoz, this fixes the case where nipype calls SPM12 Standalone, associated to Matlab Compiler Runtime (MCR) R2019a (9.6) or later. Upcoming versions of SPM12 Standalone will most probably be compiled with Matlab R2021b, and hence will be subject to this bug.

Fixes #3507.

* Do not call `addpath` in Matlab scripts called from a deployed Matlab application.

### DIFF
--- a/nipype/interfaces/matlab.py
+++ b/nipype/interfaces/matlab.py
@@ -191,7 +191,7 @@ class MatlabCommand(CommandLine):
         else:
             prescript.insert(0, "fprintf(1,'Executing code at %s:\\n',datestr(now));")
         for path in paths:
-            prescript.append("addpath('%s');\n" % path)
+            prescript.append("if ~(ismcc || isdeployed), addpath('%s'); end;\n" % path)
 
         if not mfile:
             # clean up the code of comments and replace newlines with commas
--- a/nipype/interfaces/matlab.py
+++ b/nipype/interfaces/matlab.py
@@ -191,6 +191,9 @@ class MatlabCommand(CommandLine):
         else:
             prescript.insert(0, "fprintf(1,'Executing code at %s:\\n',datestr(now));")
         for path in paths:
+            # addpath() is not available after compliation
+            # https://www.mathworks.com/help/compiler/ismcc.html
+            # https://www.mathworks.com/help/compiler/isdeployed.html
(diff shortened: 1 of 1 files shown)


nipype/interfaces/matlab.py
```diff
@@ -191,7 +191,7 @@ def _gen_matlab_command(self, argstr, script_lines):
         else:
             prescript.insert(0, "fprintf(1,'Executing code at %s:\\n',datestr(now));")
         for path in paths:
-            prescript.append("addpath('%s');\n" % path)
+            prescript.append("if ~(ismcc || isdeployed), addpath('%s'); end;\n" % path)
```
effigies, line 194: Maybe a comment to explain what's going on here?

```suggestion
            # addpath() is not available after compliation
            # https://www.mathworks.com/help/compiler/ismcc.html
            # https://www.mathworks.com/help/compiler/isdeployed.html
            prescript.append("if ~(ismcc || isdeployed), addpath('%s'); end;\n" % path)
```

effigies (approved):
LGTM. Small suggestion.

PR #3512, opened 2022-09-21
DOC: Fix typos found by codespell

DimitriPapadopoulos:
## Summary

Fix typos.

* Fix the rest of the typos found by [codespell](https://github.com/codespell-project/codespell).
* Remove spaces at end of line.
* Remove empty lines at end of file.
* Run black on files that require it to pass tests.


DimitriPapadopoulos:
Do I need to run [`black`](https://github.com/psf/black) on modified files?

nipype/interfaces/fsl/model_templates/feat_contrastmask_element.tcl
```diff
@@ -1,3 +1,2 @@
 # Mask real contrast/F-test $c1 with real contrast/F-test $c2?
 set fmri(conmask${c1}_${c2}) 0
-
```
effigies, line 3: I think for template files, we should leave them alone. These are meant to be snippets to compile as inputs to tools, and I don't have an immediate plan to generate before/after files that would use all of these and decide whether the whitespace improved readability or possibly was significant.

effigies (changes requested):
Overall looks good. I think we should revert the `nipype/interfaces/fsl/model_templates/` directory.

nipype/interfaces/fsl/model_templates/feat_contrastmask_element.tcl
DimitriPapadopoulos: I have reverted `nipype/interfaces/fsl/model_templates/*`.

effigies (approved):
LGTM!

PR #3514, opened 2022-09-21
CI: Allow tutorial test cancellation

mgxd:
Currently, after the tutorial workflow is dispatched, it runs to completion, regardless of our workflow ending early. This adds a step to cancel the workflow upon failure or cancellation, to avoid overbearing the external runners.


effigies (approved):
Ah, this is good, as combined with concurrency it will ensure that multiple commits to a release branch don't just flood the the tutorial CI.

PR #3522, opened 2022-09-27
[CI] python-version should be a string, not a float

DimitriPapadopoulos:
## Summary

The underlying issue shows clearly with `3.10`, which is interpreted as `3.1`.

This does not fix anything that wouldn't currently work. However, for documentation purposes, I feel it is better to use explicit strings. Otherwise, other maintainers may for example trigger this error when updating a version from `3.8` to `3.10`.


effigies:
Done in #3589

PR #3523, opened 2022-09-28
FIX: Check for non-mandatory output in DWIBiasCorrect

MatthieuJoulot:
The function `_list_outputs` in `DWIBiasCorrect` doesn't take into account the fact that `output_file` is not a mandatory input, thus making any pipeline using this interface without specifying the output crash. It crashes with the following error:

Screenshot 2022-09-27 at 14 55 19

### DIFF
--- a/nipype/interfaces/mrtrix3/preprocess.py
+++ b/nipype/interfaces/mrtrix3/preprocess.py
@@ -256,11 +256,12 @@ class DWIBiasCorrect(MRTrix3Base):
         return super()._format_arg(name, trait_spec, value)
 
     def _list_outputs(self):
-        outputs = self.output_spec().get()
-        outputs["out_file"] = op.abspath(self.inputs.out_file)
-        if self.inputs.bias:
-            outputs["bias"] = op.abspath(self.inputs.bias)
-        return outputs
+        if self.inputs.out_file:
+            outputs = self.output_spec().get()
+            outputs["out_file"] = op.abspath(self.inputs.out_file)
+            if self.inputs.bias:
+                outputs["bias"] = op.abspath(self.inputs.bias)
+            return outputs
 
 
 class DWIPreprocInputSpec(MRTrix3BaseInputSpec):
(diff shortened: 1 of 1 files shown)


nipype/interfaces/mrtrix3/preprocess.py
```diff
@@ -256,8 +256,9 @@ def _format_arg(self, name, trait_spec, value):
         return super()._format_arg(name, trait_spec, value)
 
     def _list_outputs(self):
-        outputs = self.output_spec().get()
-        outputs["out_file"] = op.abspath(self.inputs.out_file)
+        if self.inputs.out_file:
+            outputs = self.output_spec().get()
```
effigies, line 260: L263 will fail if `self.inputs.bias`.

```suggestion
        outputs = self.output_spec().get()
        if self.inputs.out_file:
```

effigies (approved):

effigies:
I'm actually a little confused by this. I thought `name_template` was supposed to populate the field automatically if no value was passed. With this fix, are you getting properly populated outputs?

ghisvail:
What we know is that our DWI pipeline used to work before the changes introduced in version 1.8.3. Our CI had version 1.8.2 locked in the test dependencies so we were only able to catch the regression later, when we deployed a brand new environment which included the latest version of `nipype`.

Based on the error message, we suspected the issue would come from the missing `if` guard. @MatthieuJoulot could you try validating the fix by updating your environment and running the pipeline again? :pray:

MatthieuJoulot:
Hey @effigies ! Sorry for the slow response. I have checked and it doesn't work unless I specify the `out_file` in the parameters of the node. It does run and DWIBiasCorrect does it's thing, populates an output, but this output is not found by the next node, causing the pipeline to crash right after.

effigies:
Hmm. That seems like a problem, but one that I'm not prepared to debug right now. I think it's sensible to get this fix in in the short term.

Issue #3530, opened 2022-12-08
networkx >= 3.0 unsupported


thomas-vincent:
to_scipy_sparse_matrix will be deprecated, use to_scipy_sparse_array instead

As stated in networkx:
```
warnings.warn(
        (
            "\n\nThe scipy.sparse array containers will be used instead of matrices\n"
            "in Networkx 3.0. Use `to_scipy_sparse_array` instead."
        ),
        DeprecationWarning,
        stacklevel=2,
    )
```

Either nipype code is adapted to handle network 2 and 3 or setup.py should specify that networkx must be < 3 in the dependencies.

effigies:
Good catch. Would you mind submitting a PR?

lamaresh:
this may be userful for you https://github.com/scikit-image/scikit-image/pull/6564

TheChymera:
I am getting the same issue, among others, over [here](https://ppb.chymera.eu/b4e05b.log). Tired to ~fix : https://github.com/nipy/nipype/pull/3535

PR #3535, opened 2023-01-26
Tentative `=>networkx-3.0` support

TheChymera:
Closes: https://github.com/nipy/nipype/issues/3530

Well, except it doesn't really fix the issue.

Without the patch: https://ppb.chymera.eu/b4e05b.log
With the patch: https://ppb.chymera.eu/9e1476.log

Basically instead of `AttributeError: module 'networkx' has no attribute 'to_scipy_sparse_matrix'` we now get `IndexError: tuple index out of range`... which, dunno, I guess solves the issue in the most narrow and unhelpful sense.

It seems our parameter usage is still correct, at least per the documentation of the new function...  Any ideas?

### DIFF
--- a/nipype/interfaces/cmtk/cmtk.py
+++ b/nipype/interfaces/cmtk/cmtk.py
@@ -226,7 +226,8 @@ def cmat(
     # Add node information from specified parcellation scheme
     path, name, ext = split_filename(resolution_network_file)
     if ext == ".pck":
-        gp = nx.read_gpickle(resolution_network_file)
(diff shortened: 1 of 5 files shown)


TheChymera:
Caught a few more live ones, but there are still 20 test failures, at least they appear to be nipype-internal now:

https://ppb.chymera.eu/fdae94.log

TheChymera:
Superseded by https://github.com/nipy/nipype/pull/3538#pullrequestreview-1274191472

PR #3541, opened 2023-01-31
BF: Update dcmstack interface for Py3 / newer pydicom

moloney:
## Summary

Fixes issues with dcmstack interface on Python3 with newer pydicom.


effigies:
Thanks, Brendan. It looks like this doesn't get exercised by any tests. Is it feasible to add a test that would catch a future breakage?

PR #3550, opened 2023-03-09
Add github action to shellcheck master on push and PRs

yarikoptic:
Currently it will fail with good number of errors etc to fix up. Let's see if workflow works and shows them.


effigies:
@yarikoptic Reran so you can check the output.

Issue #3534, opened 2023-01-19
ProbTrackXBaseInputSpec random_seed type error


lamaresh:
in ProbTrackXBaseInputSpec the random_seed input is set as a boolean, instead it should be a number used to initialise the random number generator and it's useful in the setting of splitting tractography over multiple threads (https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=ind1603&L=FSL&D=0&P=34677)

So i think that in ProbTrackXBaseInputSpec should be something like this:

`random_seed = traits.Int(argstr="--rseed=%s", desc="a number to initialize the random number generator")`

PR #3556, opened 2023-03-31
[FIX] update sample_random_points and random_seed type of ProbTrackXBaseInputSpec

mauriliogenovese:
updated random_seed to accept an integer value and sample_random_points to accept a float value

https://github.com/nipy/nipype/issues/3534

### DIFF
--- a/nipype/interfaces/fsl/dti.py
+++ b/nipype/interfaces/fsl/dti.py
@@ -674,8 +674,8 @@ class ProbTrackXBaseInputSpec(FSLCommandInputSpec):
     c_thresh = traits.Float(
         argstr="--cthr=%.3f", desc="curvature threshold - default=0.2"
     )
-    sample_random_points = traits.Bool(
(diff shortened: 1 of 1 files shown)


effigies:
Does this stem from a change in probtrackx or has this always been a deficiency of the interface?

effigies:
Sorry, I don't know what you're saying "yes" to.

mauriliogenovese:
I think probtrackx has always been like that, sry I misread your question and I was updating my reply

mauriliogenovese:
I've found some other similar little inconsistencies. It's ok if I propose other pull requests or do you prefer that I just open issues? I'm not too practiced in git

effigies:
Due to the low maintenance level in nipype (nobody has much time to spend on it), it's probably most efficient to open PRs unless it would be more costly to attempt to fix things before discussing them.

It may also be most efficient to copy/fix interfaces in the packages that use them, so that you are not dependent on our release cycle, which is basically unspecified at this point. Anyway, thanks for this.

PR #3558, opened 2023-04-03
[ENH] Enable SPM voxel-based FDR

mmbannert:
Hi,

The SPM interface currently does not support voxel-based FDR. The MATLAB version of SPM readily provides this functionality, however, if you change your SPM defaults accordingly. There is no easy way to achieve this in nipype. I believe that FDR is a very useful way of correcting for multiple comparisons when analyzing fMRI data. Other users share this view because a similar PR has already been made by Lorenzo [here](https://github.com/nipy/nipype/pull/3333). The changes proposed in my PR are minimally invasive and thus are very unlikely to cause any problems with respect to backward compatibility. 

* Added a single new binary (Boolean) input for whether or not the user wants to use voxel-based FDR.
* Default is False.
* ThresholdInputSpecs checks if the input for this parameter contradicts the input made for FWE because you cannot have both FWE and voxel-based FDR correction.
* If conflicting inputs are found, it raises a ValueError.

I hope you will find these changes useful as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,13 +2,13 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.4.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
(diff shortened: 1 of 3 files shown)


effigies (approved):
Thanks! This looks backwards-compatible, so if it works for you, I have no objections.

PR #3561, opened 2023-04-06
[REF] Stop using deprecated import internally

asmacdo:
## Summary

Remove internal imports of deprecated `nipype.LooseVersion` for `looseversion.LooseVersion`

### DIFF
--- a/nipype/__init__.py
+++ b/nipype/__init__.py
@@ -13,7 +13,7 @@ Top-level module API
 """
 import os
 
-# XXX Deprecate this import
+# XXX Deprecated import. Not used internally but could be used externally.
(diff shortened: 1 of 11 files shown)


nipype/__init__.py
```diff
@@ -13,7 +13,7 @@
 """
 import os
 
-# XXX Deprecate this import
+# XXX Deprecated import. Not used internally but could be used externally.
```
effigies, line 16: Nipype is in maintenance mode. No sense marking things as deprecated or to-deprecate at this point.
```suggestion
# No longer used internally but could be used externally.
```

PR #3568, opened 2023-05-10
STY: Remove duplicated trait in EddyOutputSpec

yibeichan:
`out_shell_pe_translation_parameters` in `fsl.epi.eddy` appeared twice, removed the second one to avoid redundancy


effigies:
Surprised this isn't something flake8 catches.

PR #3569, opened 2023-05-15
Normalize and Normalize12 use "estimate" and not "est" as the jobtype

servoz:
Normalize and Normalize12 of SPM use in the _list_outputs method "estimate" as jobtype and not "est". In addition, SPM uses "estimate" for the jobtype.

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -1110,7 +1110,7 @@ class NormalizeInputSpec(SPMCommandInputSpec):
         copyfile=True,
     )
     jobtype = traits.Enum(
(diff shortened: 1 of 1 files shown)


servoz:
Actually, there are some real problems with `estimate' job_type in Normalise12 with nipype ... Nipype's SPM Normalize12 process does not work at all with `estimate`. Bad output files, bad pyscript.m, generated, etc. I am investigating. I think the quick and dirty fix would be to keep `est` in the input traits and fix the _list_outputs ...

servoz:
I am closing this PR. The issue requires a bigger change I think and I don't have the time right now. I will open a new and more complete PR about spm.Normalize12 later

Issue #3576, opened 2023-06-08
Missing contrast outputs for SPM EstimateModel


manuegrx:
Hi, 
It seems that if factorial design is used for fMRI model specification (parameter "factor_info" in Level1design), spm automatically creates contrasts (con*, ess*, spmF* and spmT* images) during model estimation (EstimateModel). 

Those outputs are not yet in EstimateModel in Nipype, it could be useful to have them. 
I will create a PR in order to add them if it is okay 

Best,

effigies:
Closed by #3577.

PR #3577, opened 2023-06-08
add contrast outputs for EstimatedModel 

manuegrx:
If factorial design is used for fMRI model specification (parameter "factor_info" in Level1design), spm automatically creates contrasts (con*, ess*, spmF* and spmT* images) during model estimation (EstimateModel). 

Those outputs are not yet in EstimateModel in Nipype, the aim of this PR is to add them. 

Issue : https://github.com/nipy/nipype/issues/3576


effigies (approved):
This seems reasonable to me. Please note that nipype is not very actively maintained and does not have a specific release schedule, and you may find it easier to maintain your own interfaces in the packages that need them.

If you are interested in helping with the maintenance, let us know.

manuegrx:
Thank you very much !

PR #3599, opened 2023-09-01
Fix typo in spm.Normalize12 process

servoz:
Please consider this PR that will be fast to review:

1-  `deformation_file` and `image_to_align` are mutually exclusive. So we can't use the `mandatory=True` option for these two traits, as one of them will always be Undefined.
2- the `self.inputs.jobtype == "estimate"` statement is useless because the `jobtype` trait can only take values in [["est", "write", "estwrite"]](https://github.com/nipy/nipype/blob/dac6aee9a0f331c4f498f14a7ffb84abc989594a/nipype/interfaces/spm/preprocess.py#L1315).

For point 2- it's not very important, it doesn't stop the run, it's just useless.
For point 1-, it's critical because the condition for one or the other can never be met, and this blocks the use of this process in our application, which uses nipype.

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -1290,6 +1290,7 @@ class Normalize12InputSpec(SPMCommandInputSpec):
         field="subj.vol",
         desc=("file to estimate normalization parameters with"),
         xor=["deformation_file"],
+        mandatory=True,
         copyfile=True,
     )
     apply_to_files = InputMultiPath(
@@ -1302,6 +1303,7 @@ class Normalize12InputSpec(SPMCommandInputSpec):
     )
     deformation_file = ImageFileSPM(
         field="subj.def",
+        mandatory=True,
         xor=["image_to_align", "tpm"],
         copyfile=False,
         desc=(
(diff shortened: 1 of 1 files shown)


effigies:
`mandatory=True` is fine for mutually exclusive options. It means there will be an error if neither is defined. Have you ever had a problem?

But the other fix looks good.

servoz:
We never use nipype directly, we use it in the [populse_mia](https://github.com/populse/populse_mia) application.

In this application, there's a method that checks whether an input trait with the `mandatory=True` option has a defined value. If it doesn't, this indicates that a mandatory parameter has not been filled in, which blocks the calculation...

Ok, I've just run nipype directly, with deformation_file as Udefined and I can't see any issue ...

```
$ python
Python 3.9.9 (main, Nov 19 2021, 00:00:00) 
[GCC 10.3.1 20210422 (Red Hat 10.3.1-1)] on linux
>>> import nipype.interfaces.spm as spm
>>> norm12 = spm.Normalize12()
>>> norm12.inputs.jobtype = 'est'
>>> norm12.inputs.image_to_align = '/home/SPM_Matlab_tests_normalize/directlyInnipype/anat.nii'
>>> norm12.run()
<nipype.interfaces.base.support.InterfaceResult object at 0x7f1d771570a0>

```
Well, it's on our side that we need to change the way we check if there's a missing mandatory parameter!

So I'm going to cancel the modification for the mandatory trait option, and keep only the unnecessary part.

servoz:
Done!

effigies (approved):

PR #3600, opened 2023-09-04
Use out_prefix instead of "w" in spm.Normalize12

servoz:
Please consider this PR that will be fast to review:

I think there are two solutions.
1- Remove the `out_prefix` parameter and the prefix remains always at "w" 
or
2- use the `out_prefix` parameter instead of the hard-coded "w" value.

I think the second solution is the right one. 
I propose a slight modification to this effect.

### DIFF
--- a/nipype/interfaces/spm/preprocess.py
+++ b/nipype/interfaces/spm/preprocess.py
@@ -1491,13 +1491,13 @@ class Normalize12(SPMCommand):
                 filelist = ensure_list(self.inputs.apply_to_files)
                 for f in filelist:
                     if isinstance(f, list):
-                        run = [fname_presuffix(in_f, prefix="w") for in_f in f]
+                        run = [fname_presuffix(in_f, prefix=self.inputs.out_prefix) for in_f in f]
                     else:
-                        run = [fname_presuffix(f, prefix="w")]
+                        run = [fname_presuffix(f, prefix=self.inputs.out_prefix)]
                     outputs["normalized_files"].extend(run)
             if isdefined(self.inputs.image_to_align):
                 outputs["normalized_image"] = fname_presuffix(
-                    self.inputs.image_to_align, prefix="w"
+                    self.inputs.image_to_align, prefix=self.inputs.out_prefix
                 )
 
         return outputs


effigies:
This seems reasonable, but does it work? I don't see where the updated file names would be passed to SPM to tell it to write to those files.

servoz:
Yes it is working.
Using this little change in nipype.interfaces.spm.Normalize12:

```
$ python
Python 3.11.4 (main, Jun  7 2023, 00:00:00) [GCC 12.3.1 20230508 (Red Hat 12.3.1-1)] on linux
>>> import nipype.interfaces.spm as spm
>>> import os
>>> matlab_cmd = '/data/softs/SPM12standalone/spm12_r7771_Linux_R2019b/spm12/run_spm12.sh /data/softs/MATLAB_Runtime/v97 script'
>>> spm.SPMCommand.set_mlab_paths(matlab_cmd=matlab_cmd, use_mcr=True)
>>> norm12 = spm.Normalize12()
>>> norm12.inputs.jobtype = 'estwrite'
>>> norm12.inputs.image_to_align = '/home/Desktop/testNipypeNormalize12/anat.nii'
>>> norm12.inputs.out_prefix = 'test'
>>> os.listdir('/home/Desktop/testNipypeNormalize12')
['anat.nii', 'ref']
>>> norm12.run()
<nipype.interfaces.base.support.InterfaceResult object at 0x7f537d58fc10>
>>> os.listdir('/home/Desktop/testNipypeNormalize12')
['y_anat.nii', 'anat.nii', 'ref', 'testanat.nii']

```

>  I don't see where the updated file names would be passed to SPM to tell it to write to those files.

I confess I didn't look into the nipype machinery to see exactly where the updated filenames would be passed to SPM to tell it to write to those files. ... However, concretely, with the small modification I propose, it works directly with nipype or in our environment (populse).

effigies:
Sounds good. Could you go ahead and run `black` on this file? My guess is we have a line length issue.

servoz:
> Could you go ahead and run black on this file?

Done!

effigies:
Thanks!

Issue #3604, opened 2023-09-15
Doesn't quote paths in the invocation of dcm2niix - directory with a space makes execution fail


yarikoptic:
### Summary

Originally reported by a fella co-developer against heudiconv that it crashed when path contained a space in it.

In heudiconv we depend on nipype to just run `dcm2niix` interface: https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py#L803

### Actual behavior

```
*(Pdb) p convertnode.inputs

anon_bids = <undefined>
args = <undefined>
bids_format = True
comment = <undefined>
compress = y
compression = <undefined>
crop = False
environ = {}
has_private = False
ignore_deriv = <undefined>
merge_imgs = False
out_filename = sub-b0dwiForFmap_acq-b0dwi_epi_heudiconv133
output_dir = /tmp/tr icky/pytest-of-yoh/pytest-1/test_b0dwi_for_fmap0/sub-b0dwiForFmap/fmap
philips_float = <undefined>
series_numbers = <undefined>
single_file = False
source_dir = <undefined>
source_names = ['/home/yoh/proj/heudiconv/heudiconv-master/heudiconv/tests/data/b0dwiForFmap/b0dwi_for_fmap+00001.dcm', '/home/yoh/proj/heudiconv/heudiconv-master/heudiconv/tests/data/b0dwiForFmap/b0dwi_for_fmap+00002.dcm', '/home/yoh/proj/heudiconv/heudiconv-master/heudiconv/tests/data/b0dwiForFmap/b0dwi_for_fmap+00003.dcm']
to_nrrd = <undefined>
verbose = False
```

observe that `output_dir` has space in `/tmp/tr icky` and then running that node kabooms:

```
(Pdb) convertnode.run()
230915-17:50:37,584 nipype.workflow INFO:
	 [Node] Setting-up "convert" in "/tmp/tr icky/dcm2niixiftfysju/convert".
230915-17:50:37,588 nipype.workflow INFO:
	 [Node] Executing "convert" <nipype.interfaces.dcm2nii.Dcm2niix>
230915-17:50:37,633 nipype.interface INFO:
	 stdout 2023-09-15T17:50:37.633592:Chris Rorden's dcm2niiX version v1.0.20220720  (JP2:OpenJPEG) GCC13.2.0 x86-64 (64-bit Linux)
230915-17:50:37,633 nipype.interface INFO:
[...]

PR #3605, opened 2023-09-15
Run tests whenever TMPDIR points to some folder with spaces in the name

yarikoptic:
See https://github.com/nipy/nipype/issues/3604 for the motivation


yarikoptic:
oh well, I guess since no actual dcm2niix is invoked -- all is good here besides a single test `test_crashfile_creation` leading to `/opt/hostedtoolcache/Python/3.11.5/x64/bin/python: can't open file '/tmp/t': [Errno 2] No such file or directory`

Issue #3579, opened 2023-06-12
dcm2niix converted_files duplicate entry


mauriliogenovese:
### Summary
Converting mosaic dti series with dmc2niix results in a duplicated output

### Actual behavior
converted_files output contains twice the generated nifti file

### Expected behavior
converted_files should contain a single entry

###Possible solution
 _parse_files in Dcm2niix should check if the file is already in lists before appending

effigies:
Patches are welcome.

PR #3610, opened 2023-10-09
[FIX]Deduplicate dcm2niix output

mauriliogenovese:
fix for #3579

### DIFF
--- a/nipype/interfaces/dcm2nii.py
+++ b/nipype/interfaces/dcm2nii.py
@@ -485,6 +485,11 @@ class Dcm2niix(CommandLine):
                     mvecs.append(fl)
                 elif fl.endswith(".json") or fl.endswith(".txt"):
                     bids.append(fl)
+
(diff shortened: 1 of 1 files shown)


effigies:
Apologies for the delay. Is it possible to write a test that reproduces the issue?

nipype/interfaces/dcm2nii.py
```diff
@@ -485,6 +485,11 @@ def _parse_files(self, filenames):
                     mvecs.append(fl)
                 elif fl.endswith(".json") or fl.endswith(".txt"):
                     bids.append(fl)
+
+        # in siemens mosaic conversion nipype misread dcm2niix output and generate a duplicate list of results
+        # next line remove duplicates from output files array
+        outfiles = [*set(outfiles)]
```
effigies, line 491: If no test is feasible, fair enough. This will at least preserve the ordering, in case anything depends on it:

```suggestion
        outfiles = list(dict.fromkeys(outfiles))
```

PR #3424, opened 2021-12-15
ENH: Add "GenericLabel" to antsRegistration interpolation option.

axiezai:
## Summary
Fixes #3423, i did `make check-before-commit` but I need some help editing the docs so that the new option is reflected in the HTML. 

## List of changes proposed in this PR (pull-request)
 - Add `GenericLabel` to antsRegistration interface's interpolation option

## Acknowledgment

- [x] \(Mandatory\) I acknowledge that this contribution will be available under the Apache 2 license.

Issue #3423, opened 2021-12-15
[ENH] Add `GenericLabel` to antsRegistration interface as MultiLabel is outdated. 


axiezai:
From `antsRegistration` manual page:

```
 -n, --interpolation Linear
                         NearestNeighbor
                         MultiLabel[<sigma=imageSpacing>,<alpha=4.0>]
                         Gaussian[<sigma=imageSpacing>,<alpha=1.0>]
                         BSpline[<order=3>]
                         CosineWindowedSinc
                         WelchWindowedSinc
                         HammingWindowedSinc
                         LanczosWindowedSinc
                         GenericLabel[<interpolator=Linear>]
```

The `GenericLabel` option is currently missing from the input spec:
https://github.com/nipy/nipype/blob/5ac2f18ac116832d81a13f0c83e0a6aeda1457cf/nipype/interfaces/ants/registration.py#L411-L423

Additionally, `MultiLabel` is no longer recommended and `GenericLabel` is the preferred option.

Will make a PR.

PR #3617, opened 2023-12-05
[ENH] Add GenericLabel to ANTS ApplyTransforms.

tsalo:
## Summary

Fixes none, but is related to #3423 and #3424.

- Add "GenericLabel" to list of valid interpolation types in ANTS's ApplyTransforms interface.


nipype/interfaces/ants/resampling.py
```diff
@@ -349,6 +349,7 @@ class ApplyTransformsInputSpec(ANTSCommandInputSpec):
         "MultiLabel",
         "Gaussian",
         "BSpline",
+        "GenericLabel",
```
effigies, line 352: I think you need to add `traits.Tuple(traits.Str())` to `interpolation_parameters` options, as well.

PR #3630, opened 2024-02-20
ENH: added 'hsvs' option to 5ttgen + mrtransform fix

teanijarv:
- added Hybrid Surface and Volume Segmentation ('hsvs') option to the 5TT image generation class
- removed the in_file existing requirement (exists=True -> exists=False) for 5ttgen because hsvs requires a directory not a file as input
- changed '-inverse' parameter location from 1 to 2 in mrtransform class (to make it possible to insert it in the command together with other parameters)

### DIFF
--- a/nipype/interfaces/mrtrix3/utils.py
+++ b/nipype/interfaces/mrtrix3/utils.py
@@ -11,6 +11,7 @@ from ..base import (
     traits,
     TraitedSpec,
     File,
+    Directory,
     InputMultiPath,
     isdefined,
 )
@@ -230,8 +231,10 @@ class Generate5ttInputSpec(MRTrix3BaseInputSpec):
         mandatory=True,
         desc="tissue segmentation algorithm",
     )
-    in_file = File(
-        exists=False, argstr="%s", mandatory=True, position=-2, desc="input image"
+    in_file = traits.Either(
+        File(exists=True),
+        Directory(exists=True),
+        argstr="%s", mandatory=True, position=-2, desc="input image / directory"
     )
     out_file = File(argstr="%s", mandatory=True, position=-1, desc="output image")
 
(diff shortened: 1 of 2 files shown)


nipype/interfaces/mrtrix3/utils.py
```diff
@@ -224,13 +224,14 @@ class Generate5ttInputSpec(MRTrix3BaseInputSpec):
...
         "freesurfer",
+        "hsvs",
         argstr="%s",
         position=-3,
         mandatory=True,
         desc="tissue segmentation algorithm",
     )
     in_file = File(
-        exists=True, argstr="%s", mandatory=True, position=-2, desc="input image"
+        exists=False, argstr="%s", mandatory=True, position=-2, desc="input image"
```
effigies, line 234: So it seems like you need either a file or a directory, not a missing file.

```
in_file = traits.Either(
    File(exists=True),
    Directory(exists=True),
    argstr="%s", mandatory=True, position=-2, desc="input image / directory"
)

File: nipype/interfaces/mrtrix3/utils.py [hunk 2]
```diff
@@ -822,7 +823,7 @@ class MRTransformInputSpec(MRTrix3BaseInputSpec):
     )
     invert = traits.Bool(
         argstr="-inverse",
-        position=1,
+        position=2,
```
effigies, line 826: This means `-invert` will appear after all of the things with `position=1`. Does it need to come last, before in_files/out_file?

nipype/interfaces/mrtrix3/utils.py
teanijarv: Thanks! I added this solution now. :)

nipype/interfaces/mrtrix3/utils.py
teanijarv: I removed the position change for now, but I just added this previously because if I wanted to use -inverse with some other parameter and they both had 'position=1' then only one would show up on the command. But it can also be fixed by just using inputs.args string.

effigies (approved):

PR #2298, opened 2017-11-22
Nipype multiproc plugin modification to use GPU(s) as resources.

schahid:
Fixes # .

Changes proposed in this pull request
- Workflow nodes that run on GPUs can be scheduled similar to that of CPU resources. The plugin args will now require additionally "n_gpus" for total no. of GPUs to use and "ngpuproc" for no. of threads to start on a single GPU.

PR #3642, opened 2024-03-22
support for gpu queue

mauriliogenovese:
I wrote a simpler implementation of [this](https://github.com/nipy/nipype/pull/2298) old pull request to handle a queue of threads to be executed on GPU.
The user can specify the maximum number of parallel threads with the plugin option n_gpu_procs
The multiprocplugin will raise exception if a node require more threads than allowed in a similar way as classic CPU threads.
Note that in this implementation any GPU node will also allocate a CPU slot (is that necessary? We can change that behavior ).
Moreover the plugin doesn't check that the system actually has a cuda capable GPU (we can add such check if you think we need it)


effigies:
Just to check my understanding: in this model, a GPU-enabled job gets exclusive access to one full GPU, so the GPU queue is simply the number of available GPUs and the number of GPU-enabled jobs? There's no notion of a job acquiring multiple GPUs or partial GPUs?

From some quick searching, it's at least possible (though I don't know how common) to write programs that utilize multiple GPUs, so I think we should allow nodes to be tagged with multiple GPU threads.

If the CPU usage of a process is negligible, I think it would be reasonable to say:

```Python
myproc = pe.Node(ProcessInterface(), n_threads=0, n_gpus=2)
```

mauriliogenovese:
In the current implementation the user specify how many n_gpu_procs the plugin should manage and reserve them for nodes based on the node.n_threads property. If you think it's useful we can allow the user to specify different values for "gpu_procs" and "cpu_procs" for each node.
What should be the behaviour if the user does not specify the n_gpus property? n_gpus=n_threads?

nipype/pipeline/engine/nodes.py
```diff
@@ -821,6 +821,11 @@ def update(self, **opts):
         """Update inputs"""
         self.inputs.update(**opts)
 
+    def is_gpu_node(self):
+        return (hasattr(self.inputs, 'use_cuda') and self.inputs.use_cuda) or (
+            hasattr(self.inputs, 'use_gpu') and self.inputs.use_gpu
+        )
```
effigies, line 827:
```suggestion
        return bool(getattr(self.inputs, 'use_cuda', False)) or bool(
            getattr(self.inputs, 'use_gpu', False))
```

nipype/pipeline/plugins/multiproc.py
```diff
@@ -130,10 +131,23 @@ def __init__(self, plugin_args=None):
...
+        # GPU found on system
+        self.n_gpus_visible = MultiProcPlugin.gpu_count()
+        # proc per GPU set by user
+        self.n_gpu_procs = self.plugin_args.get('n_gpu_procs', self.n_gpus_visible)
+
+        # total no. of processes allowed on all gpus
+        if self.n_gpu_procs > self.n_gpus_visible:
+            logger.info(
+                'Total number of GPUs proc requested (%d) exceeds the available number of GPUs (%d) on the system. Using requested GPU slots at your own risk!'
+                % (self.n_gpu_procs, self.n_gpus_visible)
```
effigies, line 143: Loggers accept format strings and their arguments and only actually interpolate them if the logging event is emitted:

```suggestion
                'Total number of GPUs proc requested (%d) exceeds the available number of GPUs (%d) on the system. Using requested GPU slots at your own risk!',
                self.n_gpu_procs, self.n_gpus_visible)
```

nipype/pipeline/plugins/multiproc.py
```diff
@@ -352,6 +393,8 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
                 self._remove_node_dirs()
                 free_memory_gb += next_job_gb
                 free_processors += next_job_th
+                if is_gpu_node:
+                    free_gpu_slots -= next_job_gpu_th
```
effigies, line 397: Note that this is releasing resource claims that were made around line 356 so the next time through the loop sees available resources.

```suggestion
                if is_gpu_node:
                    free_gpu_slots += next_job_gpu_th
```

nipype/pipeline/plugins/multiproc.py
```diff
@@ -304,28 +334,39 @@ def _send_procs_to_workers(self, updatehash=False, graph=None):
...
                     next_job_gb,
                     next_job_th,
+                    next_job_gpu_th,
                 )
                 continue
 
             free_memory_gb -= next_job_gb
             free_processors -= next_job_th
+            if is_gpu_node:
+                free_gpu_slots -= next_job_gpu_th
```
effigies, line 359: I would expect this to be hit by your test, but coverage shows it's not. Can you look into this?

effigies (dismissed):
I'm extremely sorry about how long it took me to get back to this. If you're still around and up to work on this, here's the review I started last May and just finished.

nipype/pipeline/plugins/multiproc.py
mauriliogenovese: Maybe I missed that because I never used updatedhash=True, but it seems that no test includes that. Should we add a test with that option?

mauriliogenovese: Moreover that error does not impact "common" use (I have a project including this gpu support code)

mauriliogenovese: While I was looking into this I found two error about updatehash functionality. I sent a pull request https://github.com/nipy/nipype/pull/3709 to fix the biggest.
The second is that in multiproc plugin EVERY node will be executed in main thread if updatehash=True, so no multi process is enabled. I will try to send a pull request for that too (maybe after this gpu support is merged to avoid to handle merge conflicts)

PR #3648, opened 2024-05-05
STY: Apply ruff/refurb rules

DimitriPapadopoulos:
## Summary

Partially apply `FURB` rules, where they make sense, except `FURB101` (26 errors), `FURB103` (21 errors), `FURB113` (2 errors), `FURB118` (3 errors), `FURB140` (2 errors), `FURB171` (25 errors).

* STY: Apply ruff/refurb rule FURB113
     FURB113 Use `list.extend(...)` instead of repeatedly calling `list.append()`
* STY: Apply ruff/refurb rule FURB129
     FURB129 Instead of calling `readlines()`, iterate over file object directly
* STY: Apply ruff/refurb rule FURB136
    FURB136 Replace `20 if len(nodiffidx) >= 20 else len(nodiffidx)` with `min(20, len(nodiffidx))`
* STY: Apply ruff/refurb rule FURB148
    FURB148 `enumerate` index is unused, use `for x in y` instead
* STY: Apply ruff/refurb rule FURB167
    FURB167 Use of regular expression alias `re.X`
    FURB167 Use of regular expression alias `re.M`

### DIFF
--- a/nipype/algorithms/misc.py
+++ b/nipype/algorithms/misc.py
@@ -531,7 +531,7 @@ def remove_identical_paths(in_files):
         commonprefix = op.commonprefix(in_files)
         lastslash = commonprefix.rfind("/")
         commonpath = commonprefix[0 : (lastslash + 1)]
-        for fileidx, in_file in enumerate(in_files):
+        for in_file in in_files:
             path, name, ext = split_filename(in_file)
             in_file = op.join(path, name)
             name = in_file.replace(commonpath, "")
--- a/nipype/interfaces/ants/registration.py
+++ b/nipype/interfaces/ants/registration.py
@@ -1127,14 +1127,10 @@ class Registration(ANTSCommand):
         return retval
 
     def _format_transform(self, index):
-        retval = []
-        retval.append("%s[ " % self.inputs.transforms[index])
         parameters = ", ".join(
             [str(element) for element in self.inputs.transform_parameters[index]]
         )
-        retval.append("%s" % parameters)
-        retval.append(" ]")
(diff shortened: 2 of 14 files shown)


nipype/pipeline/engine/tests/test_join.py
```diff
@@ -640,14 +641,11 @@ def sq(x):
...
 
-    def exponent(x, p):
-        return x**p
-
     wf = pe.Workflow("wf", base_dir=tmpdir.strpath)
 
     xs = pe.Node(IdentityInterface(["x"]), iterables=[("x", [1, 2])], name="xs")
     ps = pe.Node(IdentityInterface(["p"]), iterables=[("p", [3, 4])], name="ps")
-    exp = pe.Node(Function(function=exponent), name="exp")
+    exp = pe.Node(Function(function=operator.pow), name="exp")
```
effigies, line 648: I don't think the `Function` interface can handle builtin functions.

DimitriPapadopoulos: I see, that would explain this error message, right?
```
TypeError: module, class, method, function, traceback, frame, or code object was expected, got builtin_function_or_method
```
The problem is that [`operator.pow()`](https://docs.python.org/3/library/operator.html#operator.pow) is part of the Python library, it is not documented to be a built-in function. I guess it's an undocumented alias for the built-in [`pow()`](https://docs.python.org/3/library/functions.html#pow).

effigies (approved):
Looks good. A couple small suggestions, but this could go in as-is.

PR #3651, opened 2024-05-30
RF: Simplify high-pass filtering in algorithms.confounds

effigies:
Legendre and cosine detrending are implemented almost identically, although with several minor variations. Here I separate regressor creation from detrending to unify the implementations.

This now uses `np.linalg.pinv(X)` to estimate the betas in both cases, rather than using `np.linalg.lstsq` in the cosine filter. lstsq uses SVD and can thus fail to converge in rare cases. Under no circumstances should (X.T @ X) be singular, so the pseudoinverse is unique and precisely what we want.

Issue raised in https://neurostars.org/t/fmriprep-numpy-linalg-linalg-linalgerror-svd-did-not-converge/29525.


effigies:
@jhlegarreta I wonder if I could bug you for a review. I suspect this would be a quick one for you, but let me know if it's not.

jhlegarreta (approved):
Take my review with some care: I am not familiar with fMRI data processing.

Changes look sensible; the methods being changed are not tested, though :grimacing:. Documenting the methods would help :book:.

Thanks for addressing the reported issue so fast.

PR #3652, opened 2024-06-01
ENH: FILMGLS gifti output support in surface mode

bogpetre:
## Summary

Fixes # .

FILMGLS is an interface for the FSL utility (film_gls) that fits GLM models to fMRI data. Most of FSL is only compatible with NIFTI files and NIFTI variants (e.g. NIFTI_GZ), but modern versions of film_gls in particular support gifti processing and output .func.gii files if a surface mode argument is included when invoking film_gls. Surface mode selection is supported by nipype but cannot be used in nipype workflows due to automated substitution of supported extensions for FSL outputs. With most FSL utilities the extension used to specify output files on the command line automatically selects the output format FSL uses, but film_gls in surface mode produces *.func.gii file in spite of this substitution. When workflow nodes downstream of FILMGLS then attempt to use the outputs of a FILMGLS node they're given the wrong filenames, namely files with NIFTI extensions rather than GIFTI extensions. This produces an error because such files do not exist. This enhancement addresses this issue in two ways.

* addition of GIFTI output file type to class Info of interfaces.fsl.base
* automatic assignment of GIFTI to output_type in FILMGLS constructor if (1) mode is set to 'surface' (2) no output_type is specified and (3) FSL version > 5.0.6

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -929,7 +929,7 @@
       "affiliation": "Department of Psychological and Brain Sciences, Dartmouth College",
       "name": "Petre, Bogdan",
       "orcid": "0000-0002-8437-168X"
-    },
+    }
   ],
   "keywords": [
     "neuroimaging",
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -924,7 +924,12 @@
       "affiliation": "MIT, HMS",
       "name": "Ghosh, Satrajit",
       "orcid": "0000-0002-5312-6729"
-    }
+    },
+    {
+      "affiliation": "Department of Psychological and Brain Sciences, Dartmouth College",
+      "name": "Petre, Bogdan",
+      "orcid": "0000-0002-8437-168X"
+    },
   ],
   "keywords": [
     "neuroimaging",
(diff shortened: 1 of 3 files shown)


effigies (approved):
LGTM. Thanks!

nipype/interfaces/fsl/base.py
```diff
@@ -57,6 +57,7 @@ class Info(PackageInfo):
         "NIFTI_PAIR": ".img",
         "NIFTI_GZ": ".nii.gz",
         "NIFTI_PAIR_GZ": ".img.gz",
+        "GIFTI": ".func.gii"
```
effigies, line 60: Small style fixup

```suggestion
        "GIFTI": ".func.gii",
```

.zenodo.json
```diff
@@ -924,7 +924,12 @@
       "affiliation": "MIT, HMS",
       "name": "Ghosh, Satrajit",
       "orcid": "0000-0002-5312-6729"
-    }
+    },
+    {
+      "affiliation": "Department of Psychological and Brain Sciences, Dartmouth College",
+      "name": "Petre, Bogdan",
+      "orcid": "0000-0002-8437-168X"
+    },
```
effigies, line 932: Invalid JSON

```suggestion
    }
```

PR #3681, opened 2024-09-22
STY: Apply ruff/Pyflakes rules (F)

DimitriPapadopoulos:
## Summary

Apply [Pyflakes (F)](https://docs.astral.sh/ruff/rules/#pyflakes-f) rules, except:
* [F401](https://docs.astral.sh/ruff/rules/unused-import/) (831 errors!)
* [F403](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star/)
* [F811](https://docs.astral.sh/ruff/rules/redefined-while-unused/) (left for later)
* [F821](https://docs.astral.sh/ruff/rules/undefined-name/) (fixed some occurrences and left a few for later)
* [F841](https://docs.astral.sh/ruff/rules/unused-variable/) (fixed some occurrences and left a few for later)


nipype/algorithms/tests/test_misc.py
```diff
@@ -8,7 +8,6 @@
 
 from nipype.algorithms import misc
 from nipype.utils.filemanip import fname_presuffix
-from nipype.testing.fixtures import create_analyze_pair_file_in_directory
```
effigies, line 11: You've removed a lot of fixtures from the module namespaces. This could be generally rectified by either moving or importing fixtures from `nipype.testing.fixtures` into `nipype.conftest`.

nipype/utils/filemanip.py
```diff
@@ -702,8 +702,8 @@ def savepkl(filename, record, versioning=False):
         except FileNotFoundError as e:
             fmlogger.debug(str(e))
             sleep(2)
-    else:
-        raise e
+        else:
+            raise
```
effigies, line 706: This was meant to be a `for`/`else`, not a `try`/`except`/`else`. What we need to do is instead:

```python
for ...:
    try:
        ...
        break
    except FileNotFoundError as e:
        err = e
        ...
else:
    raise err
```

DimitriPapadopoulos: By the way, `sleep(2)` should be skipped in the last iteration.

nipype/algorithms/tests/test_misc.py
DimitriPapadopoulos: I've put them back for now.

Issue #3654, opened 2024-06-06
nipype.interfaces.cat12.preprocess.CAT12Segment FileNotFoundError: No such file or directory '/[....]/catROIs_T1w.xml' for output 'label_rois' of a CAT12Segment interface


JohannesWiesner:
### Summary

I am running the `Cat12Segment` on a single T1w image. I don't want to run the surface reconstruction workflow so I set `surface_and_thickness_estimation=0`, and `surface_measures=0`.

### Actual behavior

This is the error I get:

```
NodeExecutionError: Exception raised while executing Node cat12segment.

Traceback:
	Traceback (most recent call last):
	  File "/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/nipype/interfaces/base/core.py", line 453, in aggregate_outputs
	    setattr(outputs, key, val)
	  File "/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/nipype/interfaces/base/traits_extension.py", line 330, in validate
	    value = super(File, self).validate(objekt, name, value, return_pathlike=True)
	  File "/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/nipype/interfaces/base/traits_extension.py", line 135, in validate
	    self.error(objekt, name, str(value))
	  File "/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/traits/base_trait_handler.py", line 74, in error
	    raise TraitError(
	traits.trait_errors.TraitError: The 'label_rois' trait of a CAT12SegmentOutputSpec instance must be a pathlike object or string representing an existing file, but a value of '/pandora/home/johannes.wiesner/work/testing/debug_cat12/cat12segment/label/catROIs_T1w.xml' <class 'str'> was specified.

	During handling of the above exception, another exception occurred:

	Traceback (most recent call last):
	  File "/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/nipype/interfaces/base/core.py", line 400, in run
	    outputs = self.aggregate_outputs(runtime)
	  File "/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/nipype/interfaces/base/core.py", line 460, in aggregate_outputs
	    raise FileNotFoundError(msg)
[...]

PR #3683, opened 2024-09-25
Fix nipy#3653 and nipy#3654 in the CAT12 Interface

### LINKED ISSUE #3653: ipype.interfaces.cat12.preprocess.CAT12Segment does not accept compressed `.nii.gz` files
### Summary

The `Cat12Segment` interface does not accept compressed `.nii.gz` files. Everything works fine when one inputs an `.nii` file.

### Actual behavior

This is the error I get:

```
TraitError: Each element of the 'in_files' trait of a CAT12SegmentInputSpec instance must be a pathlike object or string representing an existing file, but a value of '/pandora/home/johannes.wiesner/work/testing/debug_cat12/T1w.nii.gz' <class 'str'> was specified.
```

which by the way is a little bit misleading because the file exists but the file extension if the problem (but this is another topic I guess). 

### Expected behavior

Run Segmentation on a `.nii.gz` file should work because newer versions of CAT12 are able to deal with `.nii.gz` files. My CAT12 version is CAT12.9 (r2560) inside an spm12/9.14 environment module.

See also: https://www.neuro.uni-jena.de/cat12/CAT12-Manual-old.pdf

>  The segmentation module now supports the input of nii.gz files (not
possible for the longitudinal pipeline).

Here's an example script that should reproduce the error:

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""
Debug CAT12Segment
"""

import subprocess
from nipype.interfaces.cat12.preprocess import CAT12Segment
from nipype import Node
import os 

###############################################################################
# Preparation: Download an anatomical files from a public available AWS-bucket 
###############################################################################

# download T1w image
command = """
curl \
--create-dirs https://s3.amazonaws.com/openneuro.org/ds004302/sub-02/anat/sub-02_T1w.nii.gz?versionId=93eQ.AwPcMeccJAT3sO9otYv4A_WH3Bj  \
-o ./T1w.nii.gz
"""
subprocess.run(command,shell=True)

# gunzip image to get from .nii.gz to .nii
subprocess.run("gunzip -c ./T1w.nii.gz > T1w.nii",shell=True)

# get the full path to the directory where this script is being executed
script_dir = os.path.dirname(os.path.realpath(__file__))

# get full path to both images (nipype always needs absolut paths to inputs)
image_nii = os.path.join(script_dir,'T1w.nii')
image_nii_gz = os.path.join(script_dir,'T1w.nii.gz')

###############################################################################
# Run CAT12Segment inside a node on one image
###############################################################################

# create node
cat12segment = Node(CAT12Segment(surface_and_thickness_estimation=0,
                                 surface_measures=0),
                    name='cat12segment')

# input image
cat12segment.inputs.in_files = image_nii_gz

# set working directory of node to script directory
cat12segment.base_dir = script_dir

# run the node
cat12segment.run()

```

### Platform details:

```
{'commit_hash': '%h',
 'commit_source': 'archive substitution',
 'networkx_version': '3.2.1',
 'nibabel_version': '5.2.0',
 'nipype_version': '1.8.6',
 'numpy_version': '1.26.3',
 'pkg_path': '/csp-tools/anaconda3/envs/csp_wiesner_johannes/lib/python3.9/site-packages/nipype',
 'scipy_version': '1.12.0',
 'sys_executable': '/csp-tools/anaconda3/envs/csp_wiesner_johannes/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.9.18 | packaged by conda-forge | (main, Dec 23 2023, '
                '16:33:10) \n'
                '[GCC 12.3.0]',
 'traits_version': '6.3.2'}

```

### Execution environment

Choose one

JailanOweda:
## Summary

Fixes [nipy#3653](https://github.com/nipy/nipype/issues/3653) and [nipy#3654](https://github.com/nipy/nipype/issues/3654).

* Added input trait for the option of no surface measures (noROI) in the CAT12 Segment function.
* Added missing input traits (that were added to the Matlab GUI but not updated in Nipype's CAT12 Interface) for the templates used for surface measures (e.g. thalamus, thalamic_nuclei).
* Added option to process compressed nifti files (.nii.gz) like the latest GUI version allows.
* Corrected if statement for saving ROI and ROIs (surface measure) files.

### DIFF
--- a/nipype/interfaces/cat12/preprocess.py
+++ b/nipype/interfaces/cat12/preprocess.py
@@ -113,9 +113,9 @@ class CAT12SegmentInputSpec(SPMCommandInputSpec):
         'rhe "Optimized Shooting - superlarge ventricles" option for "Spatial registration" is ! '
         "required Values: \nnone: 0;\nlight: 1;\nfull: 2;\ndefault: 1070."
     )
-    initial_segmentation = traits.Int(
-        0, field="extopts.spm_kamap", desc=_help_initial_seg, usedefault=True
-    )
+    # initial_segmentation = traits.Int(
+    #     0, field="extopts.spm_kamap", desc=_help_initial_seg, usedefault=True
+    # )
 
     _help_las = (
         "Additionally to WM-inhomogeneities, GM intensity can vary across different regions such as the motor"
@@ -232,10 +232,8 @@ class CAT12SegmentInputSpec(SPMCommandInputSpec):
     surface_measures = traits.Int(
         1,
         field="output.surf_measures",
-        # usedefault=True,
+        usedefault=True,
         desc="Extract surface measures",
-        # requires=["neuromorphometrics", "lpba40", "cobra", "hammers", "thalamus", "thalamic_nuclei", "suit", "ibsr"],
-        # xor=["noROI"],
     )
 
     # Templates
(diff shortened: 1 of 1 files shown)


nipype/interfaces/cat12/preprocess.py
```diff
@@ -239,33 +240,71 @@ class CAT12SegmentInputSpec(SPMCommandInputSpec):
     neuromorphometrics = traits.Bool(
         True,
         field="output.ROImenu.atlases.neuromorphometrics",
-        usedefault=True,
+        # usedefault=True,
```
effigies, line 243: We can't change the interface like this. The reason to have defaults expliciltly added is so that when tools change their defaults, old workflows that depended on those defaults would continue using the old settings.

If the changes are this extensive, new interfaces will be needed.

effigies (changes requested):
As noted before, any changes must be backwards compatible. If this cannot be done in the current interface, a new interface must be written.

PR #3685, opened 2024-10-03
MNT: Drop traits upper bound

### LINKED ISSUE #3661: Please consider making nipype compatible with traits >= 6.4
I'm aware that traits 6.4.0 introduced breaking changes and that for that reason the version is pinned to `<6.4` (#3501). I'm also aware that there are [currently no plans](https://github.com/nipy/nipype/issues/3536#issuecomment-1406789730) to overhaul nipype.

However, something will break eventually, e.g with a newer Python release, requiring adaption to a newer traits version. Since current supported Fedora releases have all updated traits to `>6.4`, it's preventing me from packaging nipype for Fedora.

I'm not sure it's within my capabilities to help refactoring (probably not). But is there some documentation, ongoing work or similar, which would allow me to get an idea what needs changing where and what possible culprits may be?

So far, I've seen a single test out of the 2000+ tests failing due to the newer traits version. That's running tests without optional dependencies, which may hide a few additional failures.

effigies:
This PR:

* Pins numpy<2. Until dipy releases a version compatible with numpy 2, it's not worth it. Numpy 1.26 is the default in present in major distros for now.
* Subclasses `traits.BaseTuple` to recover the previously lax behavior that allowed `Tuple()` to check the length and types of arguments without insisting on the container. This fixes the problem that traits. 6.4 introduced, which particularly broke JSON load/save of input specs.
* Small compatibility fix for dipy and traits.
* Drops Python 3.8 support, which is scheduled to be EOL next week.
* Adapts to change in networkx graphviz outputs, which removed a space. It doesn't seem worth making the logic even more complex to test a cosmetic difference on a wide range of networkx versions.

Closes #3661.

### DIFF
--- a/nipype/interfaces/afni/preprocess.py
+++ b/nipype/interfaces/afni/preprocess.py
@@ -11,6 +11,7 @@ from ..base import (
     CommandLine,
     TraitedSpec,
     traits,
+    Tuple,
     isdefined,
     File,
     InputMultiPath,
@@ -2439,7 +2440,7 @@ class TCorrMapInputSpec(AFNICommandInputSpec):
     mask = File(exists=True, argstr="-mask %s")
     automask = traits.Bool(argstr="-automask")
     polort = traits.Int(argstr="-polort %d")
-    bandpass = traits.Tuple((traits.Float(), traits.Float()), argstr="-bpass %f %f")
+    bandpass = Tuple((traits.Float(), traits.Float()), argstr="-bpass %f %f")
     regress_out_timeseries = File(exists=True, argstr="-ort %s")
     blur_fwhm = traits.Float(argstr="-Gblur %f")
     seeds_width = traits.Float(argstr="-Mseed %f", xor=("seeds"))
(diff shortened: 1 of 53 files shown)


nipype/interfaces/afni/preprocess.py
```diff
@@ -3821,8 +3822,8 @@ class QwarpInputSpec(AFNICommandInputSpec):
         maxlen=5,
         xor=["wmask"],
     )
-    traits.Tuple((traits.Float(), traits.Float()), argstr="-bpass %f %f")
-    wmask = traits.Tuple(
+    bandpass = Tuple((traits.Float(), traits.Float()), argstr="-bpass %f %f")
```
effigies, line 3825: @Shotgunosine I assume this was the original intent?

nipype/interfaces/dipy/base.py
```diff
@@ -162,9 +162,17 @@ def create_interface_specs(class_name, params=None, BaseClass=TraitedSpec):
...
+                    if isinstance(traits_type, traits.List) and not isinstance(
+                        default_value, list
+                    ):
+                        default_value = [default_value]
                     attr[name] = traits_type(
-                        p[3], desc=desc[-1], usedefault=True, mandatory=is_mandatory
+                        default_value,
+                        desc=desc[-1],
+                        usedefault=True,
+                        mandatory=is_mandatory,
```
effigies, line 175: @skoudoro Could you check my work here? There were a couple failures without this, where a list with default values of `None` or `2` would fail.

effigies (commented):
@mgxd Could I bother you for a review? It hits a lot of files, but the changes aren't complicated.

mgxd (approved):
🚀 nice!

PR #3699, opened 2024-11-18
FIX: Address numpy and traits deprecations

effigies:
To unlock projects that want to support numpy 2, we need to disable our pin and fix some bugs that were probably generated in the 2to3 transition, namely using `np.savetxt(..., fmt=b'...')`.

Started to test on Python 3.13, but we're blocked on a release of traits. Ran with latest master and addressed deprecations.

@mgxd Would you be up for a quick review?


effigies:
Okay, moving to tox because this CI is a pain to work with.

mgxd (approved):

PR #3704, opened 2024-12-17
FIX: Missed np.savetxt bstring

mgxd:
Popped up in tests - remaining from https://github.com/nipy/nipype/commit/b41fb1cbe582fe659fd0914f0b3abe44788d15db

PR #3705, opened 2024-12-17
REL: 1.9.2

effigies:
Need to release #3704.


moravveji:
Hi. Is this (1.9.2) the latest stable release for building on HPC clusters?
On the main github repo page, I see the latest release being version 1.8.4. Is there a reason for the difference between versions?

effigies:
Oh, it looks like we've pushed tags but not created releases. 1.9.2 is the latest.

PR #3709, opened 2025-01-19
fix: Resolve crashes when running workflows with updatehash=True

mauriliogenovese:
Any workflow running with updatehash=True will crash

That happens because in nodes.py line 457 we try to update the hashfile even if the node was never executed before, so no hashfile is found and _update_hash function raises an exception.

### DIFF
--- a/nipype/pipeline/engine/nodes.py
+++ b/nipype/pipeline/engine/nodes.py
@@ -452,7 +452,7 @@ class Node(EngineBase):
         cached, updated = self.is_cached()
 
         # If the node is cached, check on pklz files and finish
-        if cached and not force_run and (updated or (not updated and updatehash)):
+        if cached and not force_run and (updated or updatehash):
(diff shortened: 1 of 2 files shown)


nipype/pipeline/engine/nodes.py
```diff
@@ -452,7 +452,7 @@ def run(self, updatehash=False):
         cached, updated = self.is_cached()
 
         # If the node is cached, check on pklz files and finish
-        if not force_run and (updated or (not updated and updatehash)):
+        if cached and not force_run and (updated or (not updated and updatehash)):
```
effigies, line 455:
```suggestion
        if cached and not force_run and (updated or updatehash):
```

effigies (approved):
Lgtm. Small logic simplification.

mauriliogenovese:
I add another commit to keep multi process functionality with updatehash=True.
Maybe there is a more elegant way to address it, but it seems to work in my tests

PR #3713, opened 2025-02-11
[FIX] Set length to 1 when np.squeeze returns a 0D array

### LINKED ISSUE #3712: [BUG] Merging rois outputs from bedpostx_parallel fails
I initially posted on neurostars about this: https://neurostars.org/t/nipype-merging-outputs-from-bedpostx-parallel-fails/31733/1. But I think I have a quick fix, so reporting it here. Will also do PR.

As reported in the neurostars post, I am trying to run [bedpostx_parallel workflow in old (?) nipype](https://github.com/niflows/nipype1-workflows/blob/b82a820f9e82bf92fed34add8006f4c5202b3f23/package/niflow/nipype1/workflows/dmri/fsl/dti.py#L137). It works fine during all of my test runs on downsampled data (and even some previous runs on full res data). 

However, when I recently tried to run it on the full res data, it failed at the step where it tried to merge all the outputs from parallel nodes into single files.

I looked into the issue, and it seems that on this line:
https://github.com/nipy/nipype/blob/bc456dd985b785783fb94094be685b2d6759ffaf/nipype/algorithms/misc.py#L1465

the output of `np.squeeze` is a 0D array, and getting the length of 0D array in a subsequent line isn’t possible.
https://github.com/nipy/nipype/blob/bc456dd985b785783fb94094be685b2d6759ffaf/nipype/algorithms/misc.py#L1467

I could just be an anomaly due to the heuristic nature of the algorithm, and it would probably not be an issue on a subsequent run. But I still think that it could be handled within nipype.

Here's the error traceback:

man-shu:
## Summary

Fixes #3712 .

- try getting the len of `idxs`
- if it fails with TypeError (which happens when `idxs` is a 0D array)
- set `nels` to 1

### DIFF
--- a/nipype/algorithms/misc.py
+++ b/nipype/algorithms/misc.py
@@ -1495,7 +1495,11 @@ def merge_rois(in_files, in_idxs, in_ref, dtype=None, out_file=None):
             for d, fname in enumerate(nii):
                 data = np.asanyarray(nb.load(fname).dataobj).reshape(-1)
                 cdata = nb.load(cname).dataobj[..., d].reshape(-1)
-                nels = len(idxs)
+                try:
+                    nels = len(idxs)
+                except TypeError:
+                    nels = 1
+
                 idata = (idxs,)
                 data[idata] = cdata[0:nels]
                 nb.Nifti1Image(data.reshape(rsh[:3]), aff, hdr).to_filename(fname)
--- a/nipype/algorithms/misc.py
+++ b/nipype/algorithms/misc.py
@@ -1490,18 +1490,13 @@ def merge_rois(in_files, in_idxs, in_ref, dtype=None, out_file=None):
 
         for cname, iname in zip(in_files, in_idxs):
             f = np.load(iname)
-            idxs = np.squeeze(f["arr_0"])
+            idxs = np.atleast_1d(np.squeeze(f["arr_0"]))
+            nels = len(idxs)
 
             for d, fname in enumerate(nii):
                 data = np.asanyarray(nb.load(fname).dataobj).reshape(-1)
                 cdata = nb.load(cname).dataobj[..., d].reshape(-1)
-                try:
-                    nels = len(idxs)
-                except TypeError:
-                    nels = 1
-
-                idata = (idxs,)
-                data[idata] = cdata[0:nels]
+                data[idxs] = cdata[:nels]
                 nb.Nifti1Image(data.reshape(rsh[:3]), aff, hdr).to_filename(fname)
 
         imgs = [nb.load(im) for im in nii]


nipype/algorithms/misc.py
```diff
@@ -1495,7 +1495,11 @@ def merge_rois(in_files, in_idxs, in_ref, dtype=None, out_file=None):
...
                 cdata = nb.load(cname).dataobj[..., d].reshape(-1)
-                nels = len(idxs)
+                try:
+                    nels = len(idxs)
+                except TypeError:
+                    nels = 1
+
                 idata = (idxs,)
                 data[idata] = cdata[0:nels]
                 nb.Nifti1Image(data.reshape(rsh[:3]), aff, hdr).to_filename(fname)
```
effigies, line 1505: A few things:

1) `np.atleast_1d()` allows us to avoid branching logic.
2) The `idata` tuple does not change the meaning of `idxs`, even with fancy indexing.
3) `nels` does not need to be recalculated inside the inner loop.

I would instead do:

```diff
         for cname, iname in zip(in_files, in_idxs):
             f = np.load(iname)
-            idxs = np.squeeze(f["arr_0"])
+            idxs = np.atleast_1d(np.squeeze(f["arr_0"]))
+            nels = len(idxs)
 
             for d, fname in enumerate(nii):
                 data = np.asanyarray(nb.load(fname).dataobj).reshape(-1)
                 cdata = nb.load(cname).dataobj[..., d].reshape(-1)
-                nels = len(idxs)

-                idata = (idxs,)
-                data[idata] = cdata[0:nels]
+                data[idxs] = cdata[:nels]
```

man-shu: Yep, makes sense. Thanks!

effigies:
@man-shu I'm not sure if this code actually gets tested by our tests. If you're testing on your end, please make sure that it works in cases where it was previously failing.

man-shu:
> @man-shu I'm not sure if this code actually gets tested by our tests. If you're testing on your end, please make sure that it works in cases where it was previously failing.

Ok will update you on that.

effigies:
@man-shu Have you had a chance to test this?

man-shu:
Sorry not yet! Our HPC has been down since the day I opened this PR. Our IT dept is telling us that it should be ok by next week, so I hope to get back to you then.

(I tried copying the original nipype cache locally and then run it, but then it just started a fresh run because the checksum didn't match.)

man-shu:
Hello @effigies ! I finally ran the full pipeline with and without this fix and both of them worked. 
So I don't think either of them were hitting this case. 

I do not have access to the cache for the failing run either. 

Do you have any suggestions on how to proceed from here?

effigies:
Just need to eyeball it, then.

It still looks right to me after a few weeks, so let's merge.

PR #3718, opened 2025-03-04
reimplementation of gpu_count

### LINKED ISSUE #3717: [ENH] Multiproc always looks for GPUs making it impossible to run pipelines on CPU-only machines
Hello,

I noticed that in #3642 GPU support has been added, which is indeed much appreciated.

However, the current implementation always checks for GPUs availability and there's no way to turn this off. Users (like me) might need to do so, for example, when they are trying to run a pipeline on a CPU-only HPC (where no NVIDIA-SMI is installed). 

Could we maybe have another parameter that would skip checking for GPUs? I would be up for making a PR if you think this makes sense. Please let me know.

Thanks!

mauriliogenovese:
fix #3717 

I tried to write a simpler implementation of `getGPUs` from gputil package. 
I used a separate file to include their license.

This need to be tested under windows with Python 3.12+ and I don't have such system, so it's **untested** for now.

### DIFF
--- a/nipype/utils/gpu_count.py
+++ b/nipype/utils/gpu_count.py
@@ -1,0 +1,55 @@
+# -*- DISCLAIMER: this file contains code derived from gputil (https://github.com/anderskm/gputil)
+# and therefore is distributed under to the following license:
+#
+# MIT License
+#
+# Copyright (c) 2017 anderskm
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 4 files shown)


effigies:
We need to remove gputil from the dependencies.

I'm not terribly worried about Windows, as there are many years of development since last time anybody tried to get nipype working on Windows.

mauriliogenovese:
> We need to remove gputil from the dependencies.
Done

> I'm not terribly worried about Windows, as there are many years of development since last time anybody tried to get nipype working on Windows.
I meant that I don't have a computer without a nvidia GPU to test if this code fix the error reported, I assumed it was on Windows. So even a test on other OS may be useful.

nipype/utils/gpu_count.py
```diff
@@ -0,0 +1,55 @@
...
+            stdout=PIPE,
+        )
+        stdout, stderror = p.communicate()
+
+        output = stdout.decode('UTF-8')
+        lines = output.split(os.linesep)
+        num_devices = len(lines) - 1
+        return num_devices
+    except:
+        return 0
```
effigies, line 55: If the goal is to keep it as close to the source as possible, that's fine. Here are some cleanups for your consideration:

1) Use `shutil.which()` unconditionally. We can detect a `FileNotFoundError` before calling the process and simply return 0.
2) Use `subprocess.run`, which is the recommended API. By using text mode, we don't have to handle decoding or newline normalization.
3) Use targeted errors. `FileNotFoundError` and `PermissionError` (both `OSError`s) should catch failures to run. `UnicodeDecodeError` will catch bad output.
4) Use `p.output.splitlines()` to get content-full lines. A trailing newline does not produce an empty string at the end of the list.

```suggestion
import subprocess
import os


def gpu_count():
    nvidia_smi = shutil.which('nvidia-smi')
    if nvidia_smi is None and platform.system() == "Windows":
        nvidia_smi = f'{os.environ["systemdrive"]}\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe'

    if nvidia_smi is None:
        return 0

    try:
        p = subprocess.run(
            [nvidia_smi, "--query-gpu=name", "--format=csv,noheader,nounits"],
            stdout=subprocess.PIPE,
            text=True,
        )
    except (OSError, UnicodeDecodeError):
        return 0

    return len(output.splitlines())
```

PR #3721, opened 2025-03-21
[FIX]eddy command and multithreading

mauriliogenovese:
I will send some PRs to improve/fix fsl cuda-capable tools with the new gpu support.
This one is about eddy:
eddy_openmp is an old command, new fls uses eddy_cpu
eddy_cuda is missing in some version, in those cases if user requests cuda, just use eddy

eddy_cpu does not enable multithread with OMP_NUM_THREADS variable but with --nthr argument


mauriliogenovese:
the second commit fix this: 
`EddyInputError:  The version compiled for GPU can only use 1 CPU thread (i.e. --nthr=1)
`

PR #3724, opened 2025-04-25
Slicer _cmd revision and quotes around %s

### LINKED ISSUE #2918: BRAINSResample interface changes
### Summary
I was trying to use the Slicer BRAINSResample interface and needed to make two main changes to the interface class definition:

1) The first change was to the command line string that is stored in the interface - right now it is just "BRAINSResample ".  This does not work for me, even if my PATH environment variable points to the CLI module folder in Slicer (a library .so file is missing).  I have to change it to "Slicer --launch BRAINSResample ", which seems to be in line with what the Slicer folks consider as a standard way to call these modules.  Perhaps in older versions of Slicer, you can call the command line modules directly in the terminal without the "Slicer --launch"? So therefore:

`_cmd = "BRAINSResample "`  becomes `_cmd = "Slicer --launch BRAINSResample "`

Even though this worked for me, I don't know if the current nipype version works for most people, but I am inadvertently doing something else wrong, which is solved by this hack.  Would all other command line modules in Slicer be affected in the same way?  

2) My image filenames all have spaces in them, which mess up the argument parsing when the function is executed on the command line.  On the command line outside nipype, the problem is avoided when there are quotation marks enclosing the pathname.  Therefore I changed the argstr value whenever a File input is defined in BRAINSResampleInputSpec so that it will enclose the pathname with double quotes.  For example, if the original code is:

`class BRAINSResampleInputSpec(CommandLineInputSpec):
    inputVolume = File(
        desc="Image To Warp", exists=True, argstr="--inputVolume %s")`

Then I change it to:

`class BRAINSResampleInputSpec(CommandLineInputSpec):
    inputVolume = File(
        desc="Image To Warp", exists=True, argstr="--inputVolume \"%s\"")`

andrewcyung:
## Summary

Fixes #2918 

- changed the command line string used to call the BRAINSresample utility that is included in 3D Slicer.  The master branch version will not properly call the BRAINSresample utility if the user's PATH does not specifically include the full path to the command line utility.  Using "Slicer --launch <name_of_Slicer_command_line_utility>" is the correct way to call it.

- add quotation marks (with escape sequence \") to the format specifier argument when parsing file path inputs, in order to allow filenames with spaces in them.

### DIFF
--- a/nipype/interfaces/slicer/registration/brainsresample.py
+++ b/nipype/interfaces/slicer/registration/brainsresample.py
@@ -12,18 +12,18 @@ from nipype.interfaces.base import (
 
 
 class BRAINSResampleInputSpec(CommandLineInputSpec):
-    inputVolume = File(desc="Image To Warp", exists=True, argstr="--inputVolume %s")
+    inputVolume = File(desc="Image To Warp", exists=True, argstr="--inputVolume \"%s\"")
     referenceVolume = File(
         desc="Reference image used only to define the output space. If not specified, the warping is done in the same space as the image to warp.",
         exists=True,
-        argstr="--referenceVolume %s",
+        argstr="--referenceVolume \"%s\"",
     )
     outputVolume = traits.Either(
         traits.Bool,
         File(),
         hash_files=False,
         desc="Resulting deformed image",
-        argstr="--outputVolume %s",
+        argstr="--outputVolume \"%s\"",
     )
     pixelType = traits.Enum(
         "float",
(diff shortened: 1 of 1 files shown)


effigies:
This updates the generator but doesn't regenerate. What happens when you run:

```
uv run nipype/interfaces/slicer/generate_classes.py
```

I'm concerned by the comment:

https://github.com/nipy/nipype/blob/745c1ba738db21f8dc0281fcb3e5eb48bce31fd0/nipype/interfaces/slicer/generate_classes.py#L596

This suggests that people who have slicer installed one way versus another will need different behavior. It would be bad to break workflows for one group of users to fix them for another. A larger change might be needed to allow for detecting whether the launcher is needed at runtime.

But can you go ahead and run this and commit the results so we can see the change?

PR #3745, opened 2026-03-02
FIX: Propagate errors from `run_without_submitting`

mgxd:
Came up in https://github.com/nipreps/fmriprep/issues/3611 , https://github.com/nipreps/nibabies/issues/392

This adds a  `_run_errors` attribute to `DistributedPluginBase` which is only utilized during the `run()` lifecycle, to ensure any encountered errors are properly reported.

Each distributed plugin may need to be updated, this only touches on `MultiProc` and `LegacyMultiProc`.


effigies (approved):
Looks good. I can't readily think of something cleaner.

PR #3753, opened 2026-05-18
[ENH] Add T1Prep/CatSurf interfaces

ChristianGaser:
## Summary
Add interfaces for T1Prep – T1-weighted MRI preprocessing pipeline (PyCAT). T1Prep performs skull-stripping, tissue segmentation, and cortical surface reconstruction using DeepMriPrep and the CAT-Surface library (via cat-surf).

* Add new interface at nipype/interfaces/t1prep
* Add test functions at nipype/interfaces/t1prep/tests
* Updates doc/interfaces.rst 

## Sub-module interfaces
:class:`T1Prep`
    Full pipeline: skull-strip → segment → surface estimate.  Wraps ``python -m t1prep.t1prep``.

:class:`T1PrepSegment`
    Segmentation stage only.  Wraps ``python -m t1prep.segment``.

:class:`T1PrepSurfaceEstimation`
    Surface-estimation stage for one hemisphere.  Wraps ``python -m t1prep.surface_estimation``.

:class:`T1PrepRealignLongitudinal`
    Rigid realignment of longitudinal T1w time-points.  Wraps ``python -m t1prep.realign_longitudinal``.

## Base classes
:class:`Info`
    T1Prep package version detection.

:class:`T1PrepCommand`
    Base ``CommandLine`` for ``python -m t1prep.<module>`` invocations.

##  Interfaces for the full ``cat_surf`` / ``t1prep.cat_surf`` Python API.

Every public function in ``cat_surf`` that T1Prep exposes is wrapped as a separate :class:`~nipype.interfaces.base.BaseInterface` sub-class so that each operation can be used as an independent node inside a Nipype workflow.

# References
* https://github.com/ChristianGaser/T1Prep
* https://github.com/ChristianGaser/CAT-Surface


nipype/interfaces/t1prep/cat_surf.py
effigies, line 1: Have you written a workflow with these interfaces, where you read the values in one node, use them in another, and write them in yet another? I'm a bit skeptical of how well that will work with nipype's data transfer model, but it's possible it would work. It would ultimately be using Python's pickling as its I/O interface, and seems unlikely to be more efficient than just loading files as needed. So a more typical `CatSurfGetArea` would be:

```py
class CatSurfGetAreaInputSpec(TraitedSpec):
    in_file = File(
        exists=True,
        mandatory=True,
        desc="Surface file to read (.gii, FreeSurfer, or .obj).",
    )
    out_file = File(desc="Output per-vertex area file path.")


class CatSurfGetAreaOutputSpec(TraitedSpec):
    out_file =  File(desc="Output per-vertex area file path.")
    total_area = traits.Float(desc="Total surface area (mm²).")


class CatSurfGetArea(SimpleInterface):
    """Compute per-vertex surface area.

    Wraps ``cat_surf.get_area(vertices, faces) → (area, total_area)``.

    Each vertex receives the sum of one-third of the areas of its adjacent
    triangles (the standard Voronoi tessellation-based area).

    Examples
    --------
    >>> node = CatSurfGetArea()
    >>> node.inputs.in_file = 'lh.pial'
    >>> res = node.run()            # doctest: +SKIP
    >>> per_vertex_area = res.outputs.area  # doctest: +SKIP
    """

    input_spec = CatSurfGetAreaInputSpec
    output_spec = CatSurfGetAreaOutputSpec

        # Set default output name
        if not (out_file := self.inputs.out_file):
            out_file = fname_presuffix(self.inputs.in_file, suffix='_area', newpath=runtime.cwd)
        # Ensure absolute path, but respect a user-entered absolute path
        out_file = Path(runtime.cwd) / self.inputs.out_file
        cs = _import_cat_surf()

        v, fcs = cs.read_surface(self.inputs.in_file)
        area, total = cs.get_area(v, fcs)
        cs.write_values(out_file, area)

        self._results.update(total_area=float(total), out_file=out_file)
        return runtime
```

Additional notes:

1) There is no longer any difference between `TraitedSpec` and `BaseInterfaceInputSpec`, so just use `TraitedSpec`.
2) `nipype.interfaces.base.SimpleInterface` allows you to set the outputs in the `self._results` dict in `_run_interface`, and avoid writing a custom `_list_outputs` method.
3) Given that these interfaces will be imported from `nipype.interfaces.t1prep.cat_surf`, it's not strictly necessary to prefix them all with `CatSurf`. But interface names are up to the contributor, so whatever works for you.

effigies (approved):

PR #3755, opened 2026-06-18
fix: DataGrabber._list_outputs TypeError when single file matches template (Fixes #3732)

### LINKED ISSUE #3732: DataGrabber.run() fails with TypeError when only one file matches template
## Summary
When using `DataGrabber` from Nipype with only **one** file matching the template path, `DataGrabber.run()` raises:

```
TypeError: 'in <string>' requires string as left operand, not NoneType
```

If two or more files match the template path, it works as intended.

## Actual behavior
With a single match to the template path, `dg.run()` fails with the above `TypeError`.

## Expected behavior
If there is only one file matching the template, it should return the single file path in the outputs without error, consistent with the behavior when multiple files match.

## How to replicate the behavior
A minimal dataset is provided here: [NipypeTest.zip](https://github.com/user-attachments/files/21749273/NipypeTest.zip).

**Failing case: one file matches**
```python
from nipype.interfaces.io import DataGrabber

dg = DataGrabber()
dg.inputs.sort_filelist = True
dg.inputs.template = "/path/to/NipypeTest/*/MyFile-A.nii.gz"
dg.run()
```

**Error output:**
```
  File "<python-input-20>", line 1, in <module>
    dg.run()
  File ".../nipype/interfaces/base/core.py", line 404, in run
    outputs = self.aggregate_outputs(runtime)
  File ".../nipype/interfaces/base/core.py", line 433, in aggregate_outputs
    predicted_outputs = self._list_outputs()
  File ".../nipype/interfaces/io.py", line 1300, in _list_outputs
    if None in outputs[key]:
TypeError: 'in <string>' requires string as left operand, not NoneType
```

**Working case: two files match**
```python
from nipype.interfaces.io import DataGrabber

dg = DataGrabber()
dg.inputs.sort_filelist = True
dg.inputs.template = "/path/to/NipypeTest/*/MyFile-*.nii.gz"
dg.run()
# <nipype.interfaces.base.support.InterfaceResult object ...>
```

## Execution environment
- **Python:** 3.12.5  
- **Nipype:** 1.10.0

rtmalikian:
Fixes #3732

## Problem

`DataGrabber._list_outputs()` raises `TypeError: 'in <string>' requires string as left operand, not NoneType` when exactly one file matches the template and `template_args` is empty.

**Root cause:** When `glob.glob()` returns exactly one file and there are no template args, `simplify_list(filelist)` returns a string (not a list). The code at line 1251 assigns this string directly to `outputs[key]`:

```python
outputs[key] = simplify_list(filelist)
```

Later, the check `if None in outputs[key]:` fails because `None in "string"` raises `TypeError` — Python's `in` operator for strings requires a string as the left operand.

The same bug exists in `S3DataGrabber._list_outputs()` at line 976.

## Solution

Changed both `DataGrabber` and `S3DataGrabber` to use `outputs[key].append(simplify_list(filelist))` instead of `outputs[key] = simplify_list(filelist)`, consistent with the pattern already used in the template-args code path (line 1297). This keeps `outputs[key]` as a list throughout, so the later `None in outputs[key]` check works correctly.

## Verification

All existing DataGrabber tests pass, plus two new regression tests:

```
nipype/interfaces/tests/test_io.py::test_datagrabber PASSED
nipype/interfaces/tests/test_io.py::test_datagrabber_order PASSED
nipype/interfaces/tests/test_io.py::test_datagrabber_single_file PASSED
nipype/interfaces/tests/test_io.py::test_datagrabber_single_glob_match PASSED
```

---

**About the Author:** Raphael Malikian — Clinical AI Solutions Architect. I specialise in building and fixing AI/ML systems for healthcare, including vector databases, RAG pipelines, and clinical NLP. If you need help with your project or think I can add value to your organisation, feel free to reach out — I'd love to connect.

📧 rtmalikian@gmail.com
🔗 GitHub: https://github.com/rtmalikian
🔗 LinkedIn: http://www.linkedin.com/in/raphael-t-malikian-mbbs-bsc-hons-71075436a

---

**Disclosure:** This code was developed with assistance from **mimo-2.5-pro** (Xiaomi) via **Hermes Agent** (Nous Research). All changes were reviewed, tested against the actual codebase, and verified for correctness.

### DIFF
--- a/nipype/interfaces/io.py
+++ b/nipype/interfaces/io.py
@@ -1023,12 +1023,13 @@ class S3DataGrabber(LibraryBaseInterface, IOBase):
                         if self.inputs.sort_filelist:
                             outfiles = human_order_sorted(outfiles)
                         outputs[key].append(simplify_list(outfiles))
-            if None in outputs[key]:
-                outputs[key] = []
-            if len(outputs[key]) == 0:
-                outputs[key] = None
-            elif len(outputs[key]) == 1:
-                outputs[key] = outputs[key][0]
+            if isinstance(outputs[key], list):
+                if None in outputs[key]:
+                    outputs[key] = []
+                if len(outputs[key]) == 0:
+                    outputs[key] = None
+                elif len(outputs[key]) == 1:
+                    outputs[key] = outputs[key][0]
         # Outputs are currently stored as locations on S3.
         # We must convert to the local location specified
         # and download the files.
@@ -1295,15 +1296,16 @@ class DataGrabber(IOBase):
                         if self.inputs.sort_filelist:
                             outfiles = human_order_sorted(outfiles)
                         outputs[key].append(simplify_list(outfiles))
-            if self.inputs.drop_blank_outputs:
-                outputs[key] = [x for x in outputs[key] if x is not None]
-            else:
-                if None in outputs[key]:
-                    outputs[key] = []
-            if len(outputs[key]) == 0:
-                outputs[key] = None
-            elif len(outputs[key]) == 1:
-                outputs[key] = outputs[key][0]
+            if isinstance(outputs[key], list):
+                if self.inputs.drop_blank_outputs:
+                    outputs[key] = [x for x in outputs[key] if x is not None]
+                else:
+                    if None in outputs[key]:
+                        outputs[key] = []
+                if len(outputs[key]) == 0:
+                    outputs[key] = None
+                elif len(outputs[key]) == 1:
+                    outputs[key] = outputs[key][0]
         return outputs
 
 
--- a/nipype/interfaces/tests/test_io.py
+++ b/nipype/interfaces/tests/test_io.py
@@ -299,6 +299,43 @@ def test_datagrabber_order(tmpdir):
     assert "sub002_L3_R10" in outfiles[2][1]
 
 
+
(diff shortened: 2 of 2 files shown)


effigies:
Instead of changing the type of the output and potentially breaking workflows, the simpler thing would be to check if the type is list before searching for None.

rtmalikian:
Hi @effigies, thanks for the feedback! You're right — changing the output type would be a breaking change.

I've updated the fix to guard the `None in outputs[key]` check with `isinstance(outputs[key], list)` instead. This preserves the original output type (string for single-file matches) while preventing the TypeError. Both `S3DataGrabber` and `DataGrabber` are fixed.

The changes:
- `DataGrabber._list_outputs`: wrap the `drop_blank_outputs`/None-in/len check block with `isinstance(outputs[key], list)`
- `S3DataGrabber._list_outputs`: same guard on the None-in/len check block
- Regression tests confirm single-file matches still return a string

Ready for re-review.

effigies (approved):
Just looking at the coverage, we don't really have tests that explain what's supposed to be happening here.

I'm merging because it does fix a bug, but the behavior here is so complex that it's hard to be sure that we're not changing behavior someone is relying on.

PR #3759, opened 2026-08-02
[ENH] Support renamed binary (fsl-cluster) in FSL Cluster interface

mauriliogenovese:
### Summary

Update the `fsl.Cluster` interface to use the `fsl-cluster` executable when available, while preserving compatibility with old FSL installations where the command were only exposed as `cluster`.

### Motivation

Recent FSL conda installations do not create the command wrappers provided by standard FSL installations. In particular, installing `fsl-cluster` through conda provides the `fsl-cluster` executable, but the `cluster` alias will be missing.
This causes the Nipype `Cluster` interface to fail.
I wrote to the FSL developers and @pauldmccarthy confirmed that this is expected behaviour: wrapper scripts such as `cluster -> fsl-cluster` are only generated by the standard FSL installation workflow. For custom/conda installations, users may need to create the wrappers manually or use the original command names directly.

### Changes
The `Cluster` interface now dynamically selects the executable:
```python
_cmd = "fsl-cluster" if which("fsl-cluster") else "cluster"

### CODE REVIEW

File: nipype/interfaces/fsl/model.py [hunk 1]
```diff
@@ -2055,7 +2055,7 @@ class Cluster(FSLCommand):
 
     input_spec = ClusterInputSpec
     output_spec = ClusterOutputSpec
-    _cmd = "cluster"
+    _cmd = "fsl-cluster" if which("fsl-cluster") else "cluster"
```
effigies, line 2059: Will fsl-cluster always exist (including in old installations) or do we definitely need the conditional?

Cc @pauldmccarthy

pauldmccarthy: @effigies The `$FSLDIR/bin/cluster` binary was renamed to `$FSLDIR/bin/fsl-cluster` in FSL 6.0.6 (November 2022), and I don't envisage it changing in future versions.

effigies, line 2059:
```suggestion
    # Renamed in FSL 6.0.6 (Nov 2022)
    _cmd = "fsl-cluster" if which("fsl-cluster") else "cluster"
```

effigies (approved):

